PR #44, opened 2025-12-12
Making the BMM use fp4 weights

omuhamma:
## Motivation

Porting over fp4 bmm to ATOM
## Test Plan
Checking the trace to see if kernel runtime is lower + if the correct kernel call is there 


k50112113 (commented):
I'd suggest we move all the functions (except for the un-used ones) in `atom/quantization/quark/utils.py` to the end of `atom/model_ops/utils.py`

PR #48, opened 2025-12-12
Enable RMSNorm + Quant Fusion on DSR1 FP8

farlukas:
## Motivation

Port fused RMSNorm + quant in vLLM to ATOM.

## Technical Details

When enabled, it replaces `input_layernorm` with a call to the correct fused RMSNorm + quant kernel depending on the quant dtype and skips the quant prior to the following GEMM operation.

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -92,7 +92,7 @@ ENABLE_ALLREDUCE_RMSNORM_FUSION = envs.ATOM_ENABLE_ALLREDUCE_RMSNORM_FUSION
 ENABLE_RMSNORM_QUANT_FUSION = envs.ATOM_ENABLE_RMSNORM_QUANT_FUSION
 
 
-def _fuse_rmsnorm_quant_fake(
+def _fused_rms_fp8_group_quant_fake(
     x1: torch.Tensor,
     x1_weight: torch.Tensor,
     x1_epsilon: float,
@@ -101,8 +101,6 @@ def _fuse_rmsnorm_quant_fake(
     x2_epsilon: Optional[float] = None,
     res1: Optional[torch.Tensor] = None,
     dtype_quant: torch.dtype = dtypes.fp8,
-    shuffle: bool = False,
-    scale_shuffle_padding: bool = False,
     group_size: int = 128,
     output_unquantized_inp1: bool = False,
     transpose_scale: bool = False,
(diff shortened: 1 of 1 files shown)


atom/models/deepseek_v2.py
```diff
@@ -788,12 +796,38 @@ def forward(
         residual: Optional[torch.Tensor],
     ) -> torch.Tensor:
         # Self Attention
-        if residual is None:
-            residual = hidden_states
-            hidden_states = self.input_layernorm(hidden_states)
+        if self.fuse_rmsnorm_quant:
+            weight = self.input_layernorm.weight
+            eps = self.input_layernorm.eps
+            if residual is None:
```
ChuanLi1101, line 879: if residual is None may not be true since when reach inside the above if self.fuse_rmsnorm_quant: meaning residual is not none?

atom/models/deepseek_v2.py
```diff
@@ -788,12 +796,38 @@ def forward(
...
+            hidden_states = (hidden_states_quant, hidden_states_quant_scale)
         else:
-            hidden_states, residual = self.input_layernorm(
-                hidden_states, residual)
+            if residual is None:
+                residual = hidden_states
+                hidden_states = self.input_layernorm(hidden_states)
+            else:
+                hidden_states, residual = self.input_layernorm(
+                    hidden_states, residual)
```
ChuanLi1101, line 909: This results in double RMSNorm application in some cases as self.input_layernorm is called once unconditionally and again inside both branches?

atom/models/deepseek_v2.py
```diff
@@ -788,12 +796,38 @@ def forward(
...
+            eps = self.input_layernorm.eps
+            if residual is None:
+                residual = hidden_states
+                (hidden_states_quant, hidden_states_quant_scale), _,  _, _ = _fuse_rmsnorm_quant(
+                    x1=hidden_states,
+                    x1_weight=weight,
+                    x1_epsilon=eps,
+                    res1=None,
+                    dtype_quant=self.quant_dtype,
+                    group_size=128,
```
ChuanLi1101, line 888: why hardcode this group size?

farlukas: so dsr1 fp8 does a 1x128 quant hence the `group_size=128`, though this is just for the meantime and will need to be 32 for fp4, which Omar is working on. I am coordinating with him to make sure that this condition is correctly set depending on the quant dtype

atom/models/deepseek_v2.py
farlukas: The unconditional call to `self.input_layernorm` is the old logic and is moved inside the most outer else block, so if fused RMSNorm + quant is disabled, it will go through this outer else block and call `self.input_layernorm` once depending on the path it chooses (whether `residual` is or is not `None`). I hope this clears it up. Let me know if you have any other questions or concerns.

atom/models/deepseek_v2.py
farlukas: The `if self.fuse_rmsnorm_quant` condition is the first condition to be checked in the `forward` method. This essentially just checks if the environment variable that enables fused RMSNorm + quant is set or not. If it's set, it will do the same thing that `self.input_layernorm` has been doing prior to my change (including the `residual` checks), but use the fused RMSNorm + quant function instead of the normal RMSNorm.

atom/models/deepseek_v2.py
ChuanLi1101: Maybe we can simplify the control flow by fully isolating the fused and non-fused RMSNorm paths, so that input_layernorm is only visible in one place.

atom/models/deepseek_v2.py
ChuanLi1101: Thanks for the explanation. I think my original concern is specifically about reachability: since residual is already initialized before the if self.fuse_rmsnorm_quant block, the inner if residual is None in the fused path seems unreachable in the current structure.

If the intention is to fully mirror the old input_layernorm behavior, maybe we can simplify or refactor the control flow so that residual initialization is fully owned by either the fused or non-fused path.

PR #49, opened 2025-12-14
feat: add skip_tokenizer option for pre-tokenized input

ChuanLi1101:
Add skip_tokenizer option to LLMEngine for pre-tokenized input. See commit message for details.

### DIFF
--- a/atom/model_engine/llm_engine.py
+++ b/atom/model_engine/llm_engine.py
@@ -23,14 +23,26 @@ logger = logging.getLogger("atom")
 
 class LLMEngine:
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, skip_tokenizer=False, **kwargs):
         config_fields = {field.name for field in fields(Config)}
         config_kwargs = {k: v for k, v in kwargs.items() if k in config_fields}
         data_parallel_size = kwargs.get('data_parallel_size', 1)
         config = Config(model, **config_kwargs)
-        self.tokenizer = AutoTokenizer.from_pretrained(config.model, use_fast=True)
-        config.bos_token_id = self.tokenizer.bos_token_id
-        config.eos_token_id = self.tokenizer.eos_token_id
+        
+        if skip_tokenizer:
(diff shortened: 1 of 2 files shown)


atom/model_engine/llm_engine.py
```diff
@@ -125,16 +137,29 @@ def preprocess(
         self, prompt_or_tokens: str | list[int], sampling_params: SamplingParams, stream_callback=None
     ):
         """responsible for:
-        1) Tokenize
+        1) Tokenize (if tokenizer is available and input is string)
         2) Create Sequence object"""
-        tokens = (
-            self.tokenizer.encode(prompt_or_tokens)
-            if isinstance(prompt_or_tokens, str)
-            else prompt_or_tokens
```
valarLip, line 133: this else is for "pre-tokenized token", in another word we support "pre-tokenized token" already...

ChuanLi1101: my thought was the tokenizer is still loaded during LLMEngine.__init__ even when using pre-tokenized input:

The skip_tokenizer=True option allows users to completely skip tokenizer loading, which is useful when:
> - Users have their own tokenizer and only need to pass token IDs
> - Running in offline/air-gapped environments without HuggingFace access
> - Reducing engine startup time
>
If this use case isn't common enough, I'm happy to close the PR.

valarLip: i would like to keep code clean, as normally each model will force to use their own tokenizer...  like this one for deepseek https://huggingface.co/deepseek-ai/DeepSeek-R1/blob/main/tokenizer.json

ChuanLi1101: Fair point. What if we just make the tokenizer lazy-loaded instead?
> - Only load tokenizer when preprocess() receives a string
> - No new API, no breaking changes
> - Users passing list[int] automatically skip tokenizer loading
>
This keeps the code clean while still allowing the optimization.
BTW, my main motivation for this PR is to accommodate Meena team's MLPerf request.

PR #56, opened 2025-12-15
fuse rmsnorm + quant for llama fp8

scxiao:
## Motivation

Fuse two scenarios in the llama attn decode layer:
 - rmsnorm and quant
 - act_mul and quan

This PR depends on the aiter PR: https://github.com/ROCm/aiter/pull/1658

## Technical Details

port the same changes from the vllm here. 

## Test Plan

Try the llama 70B model and wrapped up a unit test to verify corretness.

## Submission Checklist

### DIFF
--- a/atom/models/llama.py
+++ b/atom/models/llama.py
@@ -107,7 +107,7 @@ class LlamaMLP(nn.Module):
         x = self.gate_up_proj(x, x_scale=x_scale)
         scale = getattr(self.down_proj, "input_scale", None)
         x = self.act_fn(x, scale)
-        if self.fused_act_quant:
+        if scale is not None and self.fused_act_quant:
             x, scale = x
         else:
             scale = None
@@ -321,7 +321,7 @@ class LlamaDecoderLayer(nn.Module):
             hidden_states, residual = self.input_layernorm(
                 hidden_states, residual, x_scale=scale
             )
-        if self.use_fused_rmsnorm_quant:
+        if scale is not None and self.use_fused_rmsnorm_quant:
             hidden_states, scale = hidden_states
         else:
             scale = None
@@ -335,7 +335,7 @@ class LlamaDecoderLayer(nn.Module):
         hidden_states, residual = self.post_attention_layernorm(
             hidden_states, residual, scale
         )
-        if self.use_fused_rmsnorm_quant:
+        if scale is not None and self.use_fused_rmsnorm_quant:
             hidden_states, scale = hidden_states
         else:
             scale = None


atom/models/llama.py
```diff
@@ -96,10 +121,18 @@ def __init__(
...
         self.act_fn = SiluAndMul()
 
-    def forward(self, x):
-        x = self.gate_up_proj(x)
-        x = self.act_fn(x)
-        x = self.down_proj(x)
+    def forward(self, x, x_scale: Optional[torch.Tensor] = None):
+        x = self.gate_up_proj(x, x_scale=x_scale)
+        scale = getattr(self.down_proj, "input_scale", None)
+        if scale is not None and ATOM_USE_AITER_TRITON_FUSED_SILU_MUL_FP8_QUANT:
```
valarLip: move this logic into "from atom.model_ops.activation import SiluAndMul"

atom/models/llama.py
```diff
@@ -292,16 +326,75 @@ def forward(
...
     ) -> tuple[torch.Tensor, torch.Tensor]:
         # Self Attention
-        if residual is None:
-            residual = hidden_states
-            hidden_states = self.input_layernorm(hidden_states)
+
+        scale = getattr(self.self_attn.qkv_proj, "input_scale", None)
+        # scale = self.self_attn.qkv_proj.input_scale
+
+        if scale is not None and ATOM_USE_AITER_TRITON_FUSED_RMSNORM_FP8_QUANT:
```
valarLip: let's keep no hack in model itself

atom/models/llama.py
scxiao: I am wondering How I can get the `scale` inside `SimuAndMul`? do you have a similar example? Thanks

valarLip: something like this op, https://github.com/ROCm/ATOM/blob/57d426a6f3c5f300685dc31cb088fcffdd75de49/atom/model_ops/linear.py#L236

valarLip (dismissed):
LGTM, but please resolve the conflicts..

atom/model_ops/activation.py
```diff
@@ -2,22 +2,52 @@
...
-
+    def forward(
+        self, x: torch.Tensor, x_scale: Optional[torch.Tensor] = None
+    ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
+        if x_scale is not None and ATOM_USE_AITER_TRITON_FUSED_SILU_MUL_FP8_QUANT:
+            x = fused_silu_mul_fp8_per_tensor_static_quant(
+                x, x_scale, dtype_quant=rocm_aiter_fp8_dtype
+            )
+            return x, x_scale
+        else:
```
valarLip, line 45: also add x_scale support for branch "else"?

scxiao: The `scale` in the else side is None. Also, the caller side checks whether it is a tuple.

PR #73, opened 2025-12-18
CI: Fix output issues and add gsm8k accuracy tests in CI

gyohuangxin:
Related to https://github.com/orgs/ROCm/projects/48/views/1?filterQuery=sprint%3A%40current+assignee%3A%22gyohuangxin%22&pane=issue&itemId=143941128&issue=ROCm%7Cframeworks-internal%7C14818

- Fix the output issues in simple inference
- Add gsm8k accuracy tests

### DIFF
--- a/.github/scripts/atom_test.sh
+++ b/.github/scripts/atom_test.sh
@@ -42,19 +42,8 @@ if [ "$TYPE" == "accuracy" ]; then
 
   echo ""
   echo "========== Running accuracy test =========="
-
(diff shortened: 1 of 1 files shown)


gyohuangxin:
> if the scripts are for CI, I recommend using hidden folders like `ATOM/.precheckin`.

Thanks for your comments, but currently all the CI workflow files and scripts are located in `.github` folder which is already hidden.

PR #77, opened 2025-12-19
[ds3.2](fix): add one param for top_k_per_row_prefill ops

PerryZhang01:
after fix the param error, it can run normally.
image


valarLip (approved):
lgtm

PR #80, opened 2025-12-19
[Perf](bench): save perfermance info in beautiful format

PerryZhang01:
## Motivation

this PR aims to automatically print all performance information and save it into excel file in clean format.

## Technical Details

step 1: please pip install pandas openpyxl  
step 2: git clone https://github.com/kimbochen/bench_serving.git in current path  
step 3: change the variables in scripts 
step 4: bash benchmark.sh  

## Test Result

the screen will print as below:
image
and the excel file format as:
image

### DIFF
--- a/scripts/performance/benchmark.sh
+++ b/scripts/performance/benchmark.sh
@@ -19,8 +19,12 @@ if [ -f "$LOG_FILE" ]; then
     rm "$LOG_FILE"
 fi
 
-SCRIPT_DIR=$(dirname "$(realpath "${BASH_SOURCE[0]}")")
-RUN_PAYH="$SCRIPT_DIR/../../atom/benchmarks/benchmark_serving.py"
+# health check
+curl -sf "http://localhost:$PORT/health" > /dev/null || {
+    echo "ERROR: Server not running on port $PORT at vllm backend"
+    exit 1
+}
+
 for ISL in "${ISL_LIST[@]}"; do
 
     for CONC in "${CONC_LIST[@]}"; do
@@ -28,7 +32,7 @@ for ISL in "${ISL_LIST[@]}"; do
     echo "Start Test ISL=$ISL, OSL=$OSL, CONC=$CONC" | tee -a "$LOG_FILE"
     echo "=========================================" | tee -a "$LOG_FILE"
         
-        python $RUN_PAYH \
+        python -m atom.benchmarks.benchmark_serving \
             --model="$MODEL_PATH" \
             --backend=vllm \
             --base-url="http://localhost:$PORT" \


scripts/performance/benchmark.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+# step 1: please pip install pandas openpyxl
+# step 2: git clone https://github.com/kimbochen/bench_serving.git in current path
```
valarLip: haonan had adapt this one into https://github.com/ROCm/ATOM/blob/main/atom/benchmarks/benchmark_serving.py we can use this one now

PerryZhang01: okay, I modified it to use already existing benchmark serving and I think the `benchmarks` and `example` folders should be taken out of atom and placed outside.

scripts/performance/benchmark.sh
```diff
@@ -0,0 +1,51 @@
...
+for ISL in "${ISL_LIST[@]}"; do
+
+    for CONC in "${CONC_LIST[@]}"; do
+    echo "=========================================" | tee -a "$LOG_FILE"
+    echo "Start Test ISL=$ISL, OSL=$OSL, CONC=$CONC" | tee -a "$LOG_FILE"
+    echo "=========================================" | tee -a "$LOG_FILE"
+        
+        python $RUN_PAYH \
+            --model="$MODEL_PATH" \
+            --backend=vllm \
```
ChuanLi1101, line 37: Consider sth like to avoide hardcode the server mode?
curl -sf "http://localhost:$PORT/health" >/dev/null || {
  echo "ERROR: vLLM server not running on port $PORT"
  exit 1

scripts/performance/save_csv.py
```diff
@@ -0,0 +1,248 @@
...
+        r'Start Test ISL=(\d+), OSL=(\d+), CONC=(\d+)(.*?)==================================================',
+        content,
+        re.DOTALL
+    )
+    
+    for isl, osl, conc, block in test_blocks:
+        mean_ttft = re.search(r'Mean TTFT \(ms\):\s+([\d.]+)', block)
+        mean_tpot = re.search(r'Mean TPOT \(ms\):\s+([\d.]+)', block)
+        total_throughput = re.search(r'Total Token throughput \(tok/s\):\s+([\d.]+)', block)
+
```
ChuanLi1101, line 26: Function returns after first block, All other test cases are silently dropped, consider sth like:
for isl, osl, conc, block in test_blocks:
    ...
    if mean_ttft and mean_tpot and total_throughput:
        results.append(...)

return results

ChuanLi1101 (dismissed):
Minor issues, leave my comment and approved for unblock check in.

scripts/performance/save_csv.py
PerryZhang01: `test_blocks` contains all test cases, if `if mean_ttft and mean_tpot and total_throughput:` is outside the `for` loop, it seems the `results` only appends the last case. now function can include all the cases as shown in the figure.

scripts/performance/benchmark.sh
```diff
@@ -0,0 +1,51 @@
...
+##########################################################
+
+LOG_FILE="result.txt"
+# clean old file
+if [ -f "$LOG_FILE" ]; then
+    rm "$LOG_FILE"
+fi
+
+SCRIPT_DIR=$(dirname "$(realpath "${BASH_SOURCE[0]}")")
+RUN_PAYH="$SCRIPT_DIR/../../atom/benchmarks/benchmark_serving.py"
```
valarLip: 1. typo: RUN_PAYH ->RUN_PATH 
2. "$SCRIPT_DIR/../../atom/benchmarks/benchmark_serving.py" -> “python -m atom.benchmarks.benchmark_serving”

PerryZhang01: I tried to use `python -m atom.benchmarks.benchmark_serving` and it failed as shown:
image
I think when using the method of `python -m`, the sys path does not include the path of  `benchmark_serving.py`, so it import failed, it should use `from .backend_request_func import (ASYNC_REQUEST_FUNCS, RequestFuncInput,`.
but haonan runs it well, I am confused now.

valarLip: rebase main should help

scripts/performance/benchmark.sh
PerryZhang01: good suggestion！I added a health check as yours.

scripts/performance/benchmark.sh
PerryZhang01: that`s right!

PR #105, opened 2025-12-30
[PA](fix): add ATOM_GPT_OSS_MODEL env for prefill attention

PerryZhang01:


atom/model_ops/attention_mha.py
```diff
@@ -54,6 +54,7 @@ def __init__(
             else 1.0
         )
         self.kv_scale = torch.tensor(self.kv_scale_float, dtype=torch.float32)
+        self.one_scale = torch.tensor(1.0, dtype=torch.float32)
```
valarLip: let's keep using kv_scale  instead of 1.0

PerryZhang01: yes sir!

PR #113, opened 2026-01-06
[fix] disable gluon pa for llama

gbyu-amd:
## Motivation
Gluon pa is not fully verified with llama, thus disable this path for llama for now.

## Submission Checklist

### DIFF
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -405,4 +405,6 @@ class Attention(nn.Module):
                 return self.paged_attention_triton
             else:
                 # Qwen only uses gluon pa decode when bs=64
-                return self.paged_attention_triton if ctx.batch_size == 64 else self.paged_attention_asm
+                if ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION:
+                    return self.paged_attention_triton if ctx.batch_size == 64 else self.paged_attention_asm
+                return self.paged_attention_asm


atom/model_ops/attention_mha.py
```diff
@@ -405,4 +405,6 @@ def dispatch_backend(self, fwd_args: ForwardContext):
                 return self.paged_attention_triton
             else:
                 # Qwen only uses gluon pa decode when bs=64
-                return self.paged_attention_triton if ctx.batch_size == 64 else self.paged_attention_asm
+                if ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION:
```
scxiao, line 408: should we also make a change for the `paged_attention_triton` for line 405?

gbyu-amd: llama and qwen will not trigger the if condition in line 404, check line 132. One problem is that some if conditions here maybe not orthogonal enough for different models. 
The accuracy of llama is back to normal with this pr:
image

scxiao: But the Qwen3 model failed for the gluon attn here, https://github.com/ROCm/ATOM/actions/runs/20798013041/job/59736412001?pr=56. Do you know why it failed here?

gbyu-amd: image
the gluon pa api has been changed in aiter side, but the integration remains unchanged in ATOM. bernard_ps_pa_upstream gave a fix and waiting for merge.

PR #126, opened 2026-01-12
[1/N][feat] Make ATOM work with vLLM and SGLang

zejunchen-zejun:
This PR is used to make ATOM work with vLLM and SGLang, which keeps the OOB of popular frameworks and provides the optimizations from ATOM.

For vLLM, this PR uses the vLLM official out-of-tree mechanism and make ATOM provide platform, model and attention to vLLM. Here is the design diagram and performance snapshot. Compared to vLLM, vLLM+ATOM has 6-20% performance uplift.
image
image

Here is the RFC:
- vLLM RFC: https://github.com/ROCm/ATOM/issues/201
- vLLM PR: **No PRs are needed. The design is transparent to vLLM.**

For SGLang, this PR uses the official model impl backend mechanism. Here is the design diagram.
image

For attention, this PR constructs the BaseAttention and makes paged attention/radix attention inherits from this base class. The implementation details of ATOM server mode and plugin mode have been moved into the PagedAttentionImpl
image

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -17,6 +17,10 @@ from atom.utils.distributed.utils import stateless_init_torch_distributed_proces
 from torch.distributed import ProcessGroup, ReduceOp
 from transformers import AutoConfig, GenerationConfig, PretrainedConfig
 
+# plugin-related utilities
+from atom.plugin import is_plugin_mode
+from atom.plugin.config import PluginConfig
+
 logger = logging.getLogger("atom")
 
 
@@ -598,6 +602,9 @@ class Config:
     torch_dtype: torch.dtype = field(init=False)
     speculative_config: Optional[SpeculativeConfig] = None
 
+    # only use for plugin mode
+    plugin_config: Optional[PluginConfig] = None
+
     def _set_cudagraph_sizes(self):
         if self.compilation_config.cudagraph_capture_sizes:
             self.graph_bs = self.compilation_config.cudagraph_capture_sizes
@@ -621,6 +628,7 @@ class Config:
             if rope_params is None:
                 rope_params = {}
             rope_params["rope_theta"] = getattr(self.hf_config, "rope_theta", None)
+            rope_params["rope_type"] = getattr(self.hf_config, "rope_type", "default")
             self.hf_config.rope_parameters = rope_params
 
         self.generation_config = get_generation_config(self.model)
@@ -640,16 +648,25 @@ class Config:
                 self.max_model_len, hf_config_max_position_embeddings
             )
         # assert self.max_num_batched_tokens >= self.max_model_len
-        if self.torch_profiler_dir is not None:
-            os.makedirs(self.torch_profiler_dir, exist_ok=True)
-        assert self.torch_profiler_dir is None or os.path.isdir(
-            self.torch_profiler_dir
-        ), f"torch_profiler_dir {self.torch_profiler_dir} is not a valid directory"
-        if self.compilation_config.level == CompilationLevel.PIECEWISE:
-            self.compilation_config.set_splitting_ops_for_v1()
-            self._set_cudagraph_sizes()
-            self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
-            self.compilation_config.init_with_cudagraph_sizes()
+        if not is_plugin_mode():
+            if self.torch_profiler_dir is not None:
+                os.makedirs(self.torch_profiler_dir, exist_ok=True)
+            assert self.torch_profiler_dir is None or os.path.isdir(
+                self.torch_profiler_dir
+            ), f"torch_profiler_dir {self.torch_profiler_dir} is not a valid directory"
+
+        # only for server mode or plugin mode(vllm)
+        # for torch compile policy, plugin mode(vllm) uses the ATOM compile policy
+        # for cuda graph capture, plugin mode(vllm) uses the vLLM's cuda graph capture policy
+        if not is_plugin_mode() or (
+            self.plugin_config is not None and self.plugin_config.is_vllm
+        ):
+            if self.compilation_config.level == CompilationLevel.PIECEWISE:
+                self.compilation_config.set_splitting_ops_for_v1()
+                self._set_cudagraph_sizes()
+                self.compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
+                self.compilation_config.init_with_cudagraph_sizes()
+
         self.torch_dtype = (
             self.hf_config.dtype
             if getattr(self.hf_config, "dtype", None) is not None
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -15,8 +15,15 @@ from torch import nn
 
 from .attention_mla import MLAModules
 
+from atom.plugin.prepare import is_plugin_mode, is_vllm
+from atom.plugin.attention_mha import PagedAttentionImplDecoratorForPluginMode
 
-class Attention(nn.Module):
+
+@PagedAttentionImplDecoratorForPluginMode
+class PagedAttentionImpl(nn.Module):
+    """
+    Attention paged implementation
+    """
 
     def __init__(
         self,
@@ -24,11 +31,15 @@ class Attention(nn.Module):
         head_dim,
         scale,
         num_kv_heads,
+        alibi_slopes: list[float] | None,
+        sliding_window: Optional[int] = None,
         kv_cache_dtype="bf16",
+        logits_soft_cap: float | None = None,
+        attn_type=None,
+        kv_sharing_target_layer_name: int | None = None,
         layer_num=0,
         mla_modules: Optional[MLAModules] = None,
         sinks: Optional[nn.Parameter] = None,
-        sliding_window: Optional[int] = None,
         rotary_emb: Optional[torch.nn.Module] = None,
         q_norm: Optional[torch.nn.Module] = None,
         k_norm: Optional[torch.nn.Module] = None,
@@ -37,12 +48,16 @@ class Attention(nn.Module):
         super().__init__()
         self.num_heads = num_heads
         self.head_dim = head_dim
+        # for upper framework, it uses head_size in built-in methods
+        self.head_size = head_dim
         self.scale = scale
         self.num_kv_heads = num_kv_heads
+        self.alibi_slopes = alibi_slopes
         self.k_cache = self.v_cache = torch.tensor([])
         self.kv_cache_dtype = kv_cache_dtype
         self.max_model_len = 0
         self.k_scale = self.v_scale = None
+        self.device = "cuda:" + str(torch.cuda.current_device())
         self.layer_num = layer_num
         self.kv_scale_float = (
             torch.finfo(torch.float8_e4m3fn).max / torch.finfo(aiter.dtypes.fp8).max
@@ -56,7 +71,11 @@ class Attention(nn.Module):
         self.q_norm = q_norm
         self.k_norm = k_norm
 
-    def forward(
+        # for plugin mode(vllm), the query quant is disabled for now
+        if is_vllm():
+            self.supports_quant_query_input = False
+
+    def forward_impl_server_mode(
         self,
         q: torch.Tensor,
         k: torch.Tensor,
@@ -416,3 +435,39 @@ class Attention(nn.Module):
                 if atom_config.kv_cache_block_size == 1024:
                     return self.paged_attention_persistent_asm
                 return self.paged_attention_asm
+
+    def forward(
+        self,
+        layer: torch.nn.Module,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        kv_cache: torch.Tensor = None,
+        attn_metadata=None,
+        position: torch.Tensor = None,
+        q_scale: Optional[torch.Tensor] = None,
+        qkv: torch.Tensor = None,
+        output: torch.Tensor = None,
+        **kwargs,
+    ):
+        if is_plugin_mode():
+            # forward impl method are added by the decorator
+            # PagedAttentionImplDecoratorForPluginMode
+            return self.forward_impl_plugin_mode(
+                layer=layer,
+                query=query,
+                key=key,
+                value=value,
+                kv_cache=kv_cache,
+                attn_metadata=attn_metadata,
+                position=position,
+                q_scale=q_scale,
+                qkv=qkv,
+            )
+        else:
+            # only for server mode, keep the original method
+            o = self.forward_impl_server_mode(
+                q=query, k=key, v=value, position=position, q_scale=q_scale, qkv=qkv
+            )
+
+            return o
--- a/atom/model_ops/paged_attention.py
+++ b/atom/model_ops/paged_attention.py
@@ -1,0 +1,191 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+
+# from flash_attn import flash_attn_with_kvcache
+from typing import Optional
+
+import torch
+from torch import nn
+
+from .attention_mla import MLAModules
+from .base_attention import BaseAttention
+from atom.config import get_current_atom_config
+from atom.utils.selector import get_attn_backend
+from atom.plugin.prepare import is_sglang, is_vllm
+from atom.plugin.attention import unified_attention_with_output_base_for_plugin_mode
+
+
+class PagedAttention(BaseAttention):
+    """
+    Attention paged implementation
+    """
+
+    def __init__(
+        self,
+        num_heads,
+        head_dim,
+        scale,
+        num_kv_heads,
+        alibi_slopes: list[float] = None,
+        kv_cache_dtype="bf16",
+        layer_num=0,
+        use_mla: bool = False,
+        mla_modules: Optional[MLAModules] = None,
+        sinks: Optional[nn.Parameter] = None,
+        per_layer_sliding_window: Optional[int] = None,
+        rotary_emb: Optional[torch.nn.Module] = None,
+        prefix: Optional[str] = None,
+        q_norm: Optional[torch.nn.Module] = None,
+        k_norm: Optional[torch.nn.Module] = None,
+        **kwargs,
+    ):
+        # plugin mode(sglang) is not support paged attention
+        # for now, only support plugin mode(vllm) and atom server mode
+        assert (
+            not is_sglang()
+        ), "PagedAttention is not supported for plugin mode(sglang) for now"
+        super().__init__(
+            num_heads=num_heads,
+            head_dim=head_dim,
+            scale=scale,
+            num_kv_heads=num_kv_heads,
+            kv_cache_dtype=kv_cache_dtype,
+            layer_num=layer_num,
+            use_mla=use_mla,
+            mla_modules=mla_modules,
+            sinks=sinks,
+            per_layer_sliding_window=per_layer_sliding_window,
+            rotary_emb=rotary_emb,
+            prefix=prefix,
+            **kwargs,
+        )
+
+        # for plugin mode
+        if is_vllm():
+            self.use_mla = use_mla
+            self.rotary_emb = rotary_emb
+
+            try:
+                from vllm.attention.layer import Attention, AttentionType
+            except ImportError:
+                from vllm.model_executor.layers.attention import Attention
+                from vllm.v1.attention.backend import AttentionType
+
+            atom_config = get_current_atom_config()
+            assert (
+                atom_config is not None
+            ), "atom_config is required for plugin mode to vllm"
+
+            # use vllm cache config and quant config to follow the convention of vllm
+            cache_config = atom_config.plugin_config.vllm_cache_config
+            quant_config = atom_config.plugin_config.vllm_quant_config
+
+            # add extra impl args, which are needed to be passed to the impl class
+            # while it only works for custom attention backend for vllm
+            extra_impl_args = {}
+            if atom_config.plugin_config.vllm_use_atom_attention:
+                extra_impl_args["sinks"] = sinks
+                extra_impl_args["rotary_emb"] = rotary_emb
+                extra_impl_args["q_norm"] = q_norm
+                extra_impl_args["k_norm"] = k_norm
+
+            self.attn = Attention(
+                num_heads=num_heads,
+                head_size=head_dim,
+                scale=scale,
+                num_kv_heads=num_kv_heads,
+                alibi_slopes=alibi_slopes,
+                cache_config=cache_config,
+                quant_config=quant_config,
+                logits_soft_cap=None,
+                per_layer_sliding_window=per_layer_sliding_window,
+                prefix=f"{prefix}",
+                attn_type=AttentionType.DECODER,
+                kv_sharing_target_layer_name=None,
+                **extra_impl_args,
+            )
+
+            compilation_config = atom_config.compilation_config
+            self.layer_name = prefix
+            if self.layer_name in compilation_config.static_forward_context:
+                raise ValueError("Duplicate layer: {}".format(self.layer_name))
+            compilation_config.static_forward_context[self.layer_name] = self
+            return
+
+        self.num_heads = num_heads
+        self.head_dim = head_dim
+        self.scale = scale
+        self.num_kv_heads = num_kv_heads
+        self.k_cache = self.v_cache = torch.tensor([])
+        self.kv_cache_dtype = kv_cache_dtype
+        self.max_model_len = 0
+        self.k_scale = self.v_scale = None
+        self.layer_num = layer_num
+        self.mla_modules = mla_modules
+        self.use_mla = use_mla
+        self.base_attention = None
+        self.kv_cache = torch.tensor([])
+        self.indexer = mla_modules.indexer if mla_modules is not None else None
+        self.sinks = sinks
+
+        atom_config = get_current_atom_config()
+        dtype = atom_config.torch_dtype
+        block_size = atom_config.kv_cache_block_size
+        self.attn_backend = get_attn_backend(
+            block_size,
+            use_mla=self.use_mla,
+        )
+        impl_cls = self.attn_backend.get_impl_cls()
+        self.impl = impl_cls(
+            num_heads=num_heads,
+            head_dim=head_dim,
+            scale=scale,
+            num_kv_heads=num_kv_heads,
+            alibi_slopes=alibi_slopes,
+            kv_cache_dtype=kv_cache_dtype,
+            layer_num=layer_num,
+            mla_modules=mla_modules,
+            sinks=sinks,
+            sliding_window=per_layer_sliding_window,
+            rotary_emb=rotary_emb,
+            dtype=dtype,
+            q_norm=q_norm,
+            k_norm=k_norm,
+            **kwargs,
+        )
+
+        compilation_config = atom_config.compilation_config
+        default_name = f"MLA_{layer_num}" if self.use_mla else f"MHA_{layer_num}"
+        self.layer_name = prefix if prefix is not None else default_name
+        if self.layer_name in compilation_config.static_forward_context:
+            raise ValueError("Duplicate layer: {}".format(self.layer_name))
+        compilation_config.static_forward_context[self.layer_name] = self
+
+    def forward(
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        positions: torch.Tensor = None,
+        q_scale: Optional[torch.Tensor] = None,
+        qkv: torch.Tensor = None,
+        **kwargs,
+    ):
+        if is_vllm():
+            output = unified_attention_with_output_base_for_plugin_mode(
+                query,
+                q_scale,
+                key,
+                value,
+                positions,
+                layer_name=self.layer_name,
+                use_mla=self.use_mla,
+                qkv=qkv,
+            )
+            return output
+
+        # for atom server mode
+        output = torch.ops.aiter.unified_attention_with_output_base(
+            query, q_scale, key, value, positions, self.layer_name, self.use_mla, qkv
+        )
+        return output
--- a/atom/models/qwen3.py
+++ b/atom/models/qwen3.py
@@ -24,14 +24,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional
+from typing import Any, Iterable
 
 import torch
 
 # import torch.distributed as dist
 from aiter.dist.parallel_state import get_tp_group
 from aiter.rotary_embedding import get_rope
-from atom.config import Config, QuantizationConfig
+from atom.config import Config
 from atom.model_ops.activation import SiluAndMul
 
 # from atom.model_ops.attention import Attention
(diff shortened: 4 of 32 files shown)


atom/model_ops/paged_attention.py
```diff
@@ -0,0 +1,176 @@
...
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        positions: torch.Tensor = None,
+        q_scale: Optional[torch.Tensor]=None,
+        qkv: torch.Tensor = None,
+        **kwargs,
+    ):
+        if is_vllm():
+            output = unified_attention_with_output_base_for_plugin_mode(
```
wuhuikx, line 175: why we need this two paths here? vLLM also supports quantized q tensor.

atom/plugin/attention.py
```diff
@@ -0,0 +1,620 @@
...
+        while len(sliding_window_sizes) > 0:
+            sliding_window_config = sliding_window_sizes.pop()
+            if sliding_window_config is not None and sliding_window_config[0] != -1:
+                assert self.aot_sliding_window is None, (
+                    "Aiter Backend only support one valid sliding window"
+                )
+                self.aot_sliding_window = sliding_window_config
+
+        # for extend path to store the fetched key and value
+        self.extend_workspace = torch.empty(
```
wuhuikx, line 250: This buffer is out of control by vLLM or SGLang Cache manger. It's risky to meet OOM issue. Can you add some comments here for issue reminder?

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,746 @@
...
+
+        attn_metadata = attention_metadata
+
+        use_triton_attn = (
+            self.sliding_window != -1 or self.head_dim != 128
+        )
+        self.use_triton_attn = use_triton_attn
+
+        if ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION:
+            fused_qk_norm_rope_cache_quant_shuffle(
```
wuhuikx, line 110: Recently, vllm has separate the reshape_and_cache from attention backend. We can refer to their design and make the attention backend cleaner. 

only for Qwen, we have qk_norm_rope_cache_quant fusion, for the other models, we have not qk_norm fusion. 
Please let Qwen modeling to handle this specific logics instead of the common attention module.

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,746 @@
...
+                ],
+                slot_mapping=attn_metadata.plugin_metadata.slot_mapping[
+                    num_decodes : num_decodes + num_extends
+                ],
+                k_scale=k_scale,
+                v_scale=v_scale,
+            )
+
+        # calculate for decodes
+        if num_decodes > 0:
```
wuhuikx, line 661: We recently have a unified pa API in AITER, please take a look at that one and simplify the dispatch logic here.

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,746 @@
...
+            block_tables_stride0=attn_metadata.plugin_metadata.block_table[:num_decodes].stride(0),
+            K_QScale=k_scale,
+            V_QScale=v_scale,
+            out_=out[:num_decode_tokens],
+            high_precision=0,
+        )
+
+        return
+
+    def prefill_attention_asm_plugin_mode(self,
```
wuhuikx: why we need a function wrapper here? We can directly call the flash_attn_varlen_func in the forward function.

atom/config.py
```diff
@@ -18,6 +18,10 @@
 from aiter import QuantType
 from aiter.utility.dtypes import d_dtypes
 
+# only for plugin mode
```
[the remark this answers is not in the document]
zejunchen-zejun: make sense

atom/plugin/attention_mha.py
zejunchen-zejun: Move to next PR because it is related to the attention refine. For now let's follow the existed design and change it in future.

recipes/SGLang-ATOM-Model-Impl-Backend.md
wuhuikx, line 1: I suggest we remove the SGL RFC in this PR while focus on vLLM OOT in this PR. We can file another PR for SGL with minor revisions.

recipes/vLLM-ATOM-OOT-Plugin-Backend.md
```diff
@@ -0,0 +1,89 @@
+# vLLM out-of-tree ATOM Plugin Backend
+ATOM can work as the OOT plugin backend of vLLM. The OOT register mechanism is quite mature and most of accelerators have leveraged this design to register their devices into vLLM without any code changes in upper framework. ATOM follows this design and provide the layer/op and model implementations to vLLM. The frontend users can launch vLLM server like before and there is no need to specify any arguments. Meanwhile the ATOM platform can leverage most of the vLLM features and focus more on model- and kernel-level optimizations. For the overall design, here is a RFC to enable ATOM work as the OOT plugin platform of vLLM: https://github.com/ROCm/ATOM/issues/201
+
+## Preparing environment for vLLM with ATOM model backend
+Pull the vLLM official docker for ROCm. If you are using the vLLM nightly docker, there could be incompatible error because vLLM is changing its code and may break the class/module import in ATOM
+```bash
+docker pull rocm/vllm-dev:nightly_main_20260118
```
wuhuikx, line 7: Please try the latest vllm docker when you have time. It's not urgent.
Latest Docker: Use this one: rocm/vllm-private:rocm7.2.1_kineto_amd_dev_20260225

recipes/vLLM-ATOM-OOT-Plugin-Backend.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+Then the ATOM should be installed. When the following PR merged, you can use ATOM main branch
+```bash
+git clone https://github.com/zejunchen-zejun/ATOM.git
+cd ATOM
+git checkout origin/zejun/plugin_for_atom_1223
+pip install -e . 2>&1 | tee build.log
+```
+For AITER, there is no specific requirements, however, if you find any latest fusion kernels are missing, you may need to upgrade the AITER
```
wuhuikx: Actually, we depend on the AITER version. We should use the same AITER version as ATOM (now ATOM use the latest main)

recipes/SGLang-ATOM-Model-Impl-Backend.md
zejunchen-zejun: ok, remove SGLang RFC

recipes/vLLM-ATOM-OOT-Plugin-Backend.md
zejunchen-zejun: ok, will change it in recipe.
For now, the plugin can work with aiter latest main branch:
```
commit 31ba9bae995e52acdd27b2054612c1689d4c4a67 (HEAD -> main, origin/main, origin/HEAD)
Author: Xin Huang <Xin.Huang@amd.com>
Date:   Fri Feb 27 14:50:27 2026 +0800

    Add runner-config.yml for runner->GPU mapping used by frameworks-devops-dashboard to display GPU architecture (#2126)
```

recipes/vLLM-ATOM-OOT-Plugin-Backend.md
zejunchen-zejun: For now, the plugin can work compatible with docker rocm/vllm-private:rocm7.2.1_kineto_amd_dev_20260225.
The detailed usage is:
1. git pull rocm/vllm-private:rocm7.2.1_kineto_amd_dev_20260225
2. install this PR
3. install latest aiter main
4. **upgrade the triton to 3.6.0**
5. run your workloads

@ChuanLi1101 @wuhuikx 
For the step 4, is it ok to upgrade the triton installed in the docker?
If we do not upgrade the triton version, there will be following error:
```
(EngineCore_DP0 pid=38454)   File "/usr/local/lib/python3.12/dist-packages/vllm/v1/executor/multiproc_executor.py", line 358, in get_response
(EngineCore_DP0 pid=38454)     raise RuntimeError(
(EngineCore_DP0 pid=38454) RuntimeError: Worker failed with error 'module 'triton.experimental.gluon.language.amd' has no attribute 'cdna3'', 
please check the stack trace above for the root cause
```

atom/utils/backends.py
```diff
@@ -255,6 +255,24 @@ class SplitItem:
...
+    if node.op == "call_function" and (
+        hasattr(node.target, "spliting_op") and (node.target.spliting_op)
+    ):
+        return True
+
+    # When plugin mode(vLLM), the attention impl op is registered
+    # as unified_attention
+    from atom.plugin import is_vllm
+
+    if is_vllm() and "unified_attention" in node.name:
```
ZhangLirong-amd, line 270: I remember attention is always spliting op by @mark_spliting_op decorator, why we need this additional condition?

zejunchen-zejun: Hi, @ZhangLirong-amd 
I move the original attention forward into a class named PagedAttention. Here is its forward code with this PR:
```
    def forward(
        self,
        query: torch.Tensor,
        key: torch.Tensor,
        value: torch.Tensor,
        positions: torch.Tensor = None,
        q_scale: Optional[torch.Tensor] = None,
        qkv: torch.Tensor = None,
        **kwargs,
    ):
        if is_vllm():
            output = unified_attention_with_output_base_for_plugin_mode(
                query,
                q_scale,
                key,
                value,
                positions,
                layer_name=self.layer_name,
                use_mla=self.use_mla,
                qkv=qkv,
            )
            return output

        # for atom server mode
        output = torch.ops.aiter.unified_attention_with_output_base(
            query, q_scale, key, value, positions, self.layer_name, self.use_mla, qkv
        )
        return output
```
For ATOM plugin mode(ATOM work as out-of-tree platform of vLLM), the `unified_attention_with_output_base_for_plugin_mode` will be called for this path and it has not been decorated by the `mark_spliting_op`. It is not a registered op of torch.compile because it will call vLLM official Attention interface. The vLLM official Attention forward will call a registered op `torch.ops.vllm.unified_attention`, so the compiled graph here will already have a standalone node named `unified_attention`. We just need to pick out this node and split it from the root graph on ATOM side.

For ATOM server mode(ATOM work as standalone engine), the logic has no change and the unified_attention_with_output_base will be called, meanwhile it is a registered op because it is decorated by `mark_spliting_op`.

According to above, I developed a heuristic split judging function `_split_judge_func` to handle different scenarios.

ZhangLirong-amd: Thanks, zejun. Got it. And I notice vllm mark these ops as splitting_ops https://github.com/vllm-project/vllm/blob/main/vllm/config/compilation.py#L665-L679 , in this version, we only splitting unified_attention right?

zejunchen-zejun: `vllm mark these ops as splitting_ops`
Yes, vLLM has marked these attention related ops as splitting ops, but it doesn't affect the ATOM side because when ATOM work as out-of-tree platform, the compilation is owned on ATOM side. ATOM calls torch.compile to compile the model and decide how to split the graph. The only thing vLLM did is call the torch.ops.vllm.unified_attention.

`we only splitting unified_attention right?`
Yes, for now we only split unified_attention. For future usage, we can extend the split heuristic here.

atom/model_ops/paged_attention.py
zejunchen-zejun: We need 2 paths here for different modes. It is not related to quantized q_tensor.
`unified_attention_with_output_base_for_plugin_mode` is for OOT because in this function, the vLLM official attention should be called.
`unified_attention_with_output_base` is for server mode and it keeps the original ATOM code logic.

atom/plugin/attention.py
zejunchen-zejun: ok, added some comments here

atom/plugin/attention_mha.py
zejunchen-zejun: Move to next PR

atom/plugin/attention_mha.py
zejunchen-zejun: yes, it makes sense, fixed

atom/model_ops/attention_mha.py
```diff
@@ -56,7 +71,16 @@ def __init__(
...
 
-    def forward(
+        # for plugin mode(vllm), the query quant is disabled for now
+        if is_vllm():
+            self.supports_quant_query_input = False
+
+    # this method will just be called by vLLM and there is no logic in this method
+    # as ATOM handles the process after loading weights for all ops by itself
+    def process_weights_after_loading(self, act_dtype: torch.dtype = torch.bfloat16):
+        pass
```
valarLip: maybe raise not implement here?

atom/model_ops/paged_attention.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # for plugin mode
+        if is_vllm():
+            self.use_mla = use_mla
+            self.rotary_emb = rotary_emb
+
+            try:
+                from vllm.attention.layer import Attention, AttentionType
+            except ImportError:
+                from vllm.model_executor.layers.attention import Attention
```
valarLip, line 71: what's the reason for using vllm/sglang attn class?

atom/models/deepseek_v2.py
```diff
@@ -59,7 +59,7 @@
 )
 from atom.model_ops.activation import SiluAndMul
 from atom.model_ops.attention_mla import MLAModules, is_rocm_aiter_fp4bmm_enabled
-from atom.model_ops.base_attention import Attention
```
valarLip, line 62: can we find a better way... instead of modify all the model files

atom/model_ops/paged_attention.py
```diff
@@ -0,0 +1,191 @@
...
+                alibi_slopes=alibi_slopes,
+                cache_config=cache_config,
+                quant_config=quant_config,
+                logits_soft_cap=None,
+                per_layer_sliding_window=per_layer_sliding_window,
+                prefix=f"{prefix}",
+                attn_type=AttentionType.DECODER,
+                kv_sharing_target_layer_name=None,
+                **extra_impl_args,
+            )
```
valarLip, line 106: maybe we can override process_weights_after_loading here instead if..else in loader.py for following logic
if isinstance(module, Attention):
        module.process_weights_after_loading(act_dtype=act_dtype)

atom/model_ops/paged_attention.py
zejunchen-zejun: Attention op is a little bit different from other ops(Linear/FusedMoE). We need to register the attention backend to vLLM and call official vLLM attention op in ATOM because vLLM has the ownership for 2 things when ATOM running the plugin mode.
1. trigger build metadata for attention op in vLLM model runner. If we do not register the attention backend to vLLM, the metadata cannot be built before running the model forward
2. trigger `get_impl_cls` to find the attention implementation class. This method is called in official attention op https://github.com/vllm-project/vllm/blob/f5d1281c9d1b96cb4f046f1ec2c53a525f319098/vllm/model_executor/layers/attention/attention.py#L315. If we do not call official attention op, the impl class will not be called by vLLM

atom/model_ops/attention_mha.py
zejunchen-zejun: The method `process_weights_after_loading` of `PagedAttentionImpl` has no logic code because indeed we don't have any post processing logic after weight finish loading. However even if we have no logic here, the method still will be called by vLLM here: https://github.com/vllm-project/vllm/blob/f5d1281c9d1b96cb4f046f1ec2c53a525f319098/vllm/model_executor/layers/attention/attention.py#L503, so we add this method to `PagedAttentionImpl` to bypass the calling.

It is only used for OOT path, so we removed this method here and added into the class through the decorator.

atom/model_ops/paged_attention.py
zejunchen-zejun: It makes perfect sense. We have override the `process_weights_after_loading` method, as the result, the if-else code logic in loader.py is removed in this PR.

atom/models/deepseek_v2.py
zejunchen-zejun: Yes. It makes sense. Fixed. I change the code here and hide the attention selection logic into the base_attention file, as a result, there is no code changes to all model files. The developers can construct the model and call `self.attn = Attention(...)` like before.

atom/plugin/attention.py
```diff
@@ -0,0 +1,625 @@
...
+            raise NotImplementedError(
+                "AttentionMetadataBuilder for sglang is not implemented yet"
+            )
+
+        # create the new class
+        new_class = type(cls.__name__, (base_class,), class_dict)
+
+        # replace the inherit base class for plugin mode, meanwhile support generic base class
+        if needs_generic and generic_base is not None:
+            new_class.__orig_bases__ = (generic_base[new_class],)
```
ChuanLi1101: generic_base refers to the AttentionMetadataBuilder class (line 270). In Python, a regular class does not support [] subscript syntax unless it implements __class_getitem__. This will directly raise a TypeError.

How about this?
 new_class.__orig_bases__ = (generic_base,)

Or from typing import Generic

atom/plugin/prepare.py
```diff
@@ -0,0 +1,89 @@
...
+    # import here to avoid partial initialization
+    from .register import (
+        _ATOM_SUPPORTED_MODELS,
+        # register_ops_to_vllm,
+        register_ops_to_sglang,
+        init_aiter_dist,
+        set_attn_cls,
+    )
+
+    if model_arch not in _ATOM_SUPPORTED_MODELS:
```
ChuanLi1101, line 63: If `config` is neither vLLM nor SGLang (theoretically unlikely, but for defensive programming), then `model_arch` would be undefined and trigger a `NameError`.

**Suggested fix:**
If vLLM is not supported in this path, `prepare_model` should either return early for the vLLM branch or explicitly clean up the logic to avoid ambiguity.

More importantly, in vLLM the `model_wrapper.py` bypasses `prepare_model` entirely (it directly calls `generate_atom_config_for_plugin_mode` + `_prepare_env`). So currently this function is effectively only used by SGLang.

It would be safer to add an `else` branch like:

```python
else:
    raise ValueError(f"Unsupported config type: {type(config)}")
```

This avoids silent failure and makes the control flow explicit.

atom/plugin/vllm/platform.py
```diff
@@ -0,0 +1,38 @@
...
+class ATOMPlatform(RocmPlatform):
+    # For multi-modality models, to make AiterBackend supported by ViT,
+    # get_supported_vit_attn_backends may need to be overridden.
+    @classmethod
+    def get_attn_backend_cls(cls, selected_backend, attn_selector_config) -> str:
+        # Fall back to original behavior of vLLM mainline.
+        if disable_vllm_plugin_attention:
+            logger.info("Fallback to original vLLM attention backend")
+            return super().get_attn_backend_cls(selected_backend, attn_selector_config)
+
```
ChuanLi1101: When `disable_vllm_plugin=True`, `RocmPlatform` falls back to `object`. In that case, `ATOMPlatform(object)` does not implement `get_attn_backend_cls`, so calling `super()` will raise an `AttributeError`.

**Suggested fix:**
Instead of relying on a fallback base class, define the class conditionally:

```python
if not disable_vllm_plugin:
    from vllm.platforms.rocm import RocmPlatform

    class ATOMPlatform(RocmPlatform):
        @classmethod
        def get_attn_backend_cls(cls, selected_backend, attn_selector_config) -> str:
            if disable_vllm_plugin_attention:
                return super().get_attn_backend_cls(
                    selected_backend, attn_selector_config
                )
            return "atom.model_ops.attentions.aiter_attention.AiterBackend"
else:
    ATOMPlatform = None
```

This avoids inheriting from `object` unintentionally and makes the control flow explicit, preventing unexpected `AttributeError`s at runtime.

atom/models/qwen3.py
```diff
@@ -283,3 +325,10 @@ def compute_logits(
     ) -> torch.Tensor:
         logits = self.lm_head(hidden_states)
         return logits
+
+    def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
+        # load weights in plugin mode and discard passed weights generator
+        loaded_weights_record = load_model_in_plugin_mode(
+            model=self, config=self.atom_config, prefix="model."
+        )
+        return loaded_weights_record
```
ChuanLi1101, line 338: `Qwen3ForCausalLM` is structured as:

```python
self.model = Qwen3Model(...)
self.lm_head = ...
```

The checkpoint parameter names are:

```
model.layers.0.xxx
lm_head.weight
```

If `prefix="model."` is passed into `load_model_in_plugin_mode`, the loader will prepend the prefix again, producing keys like:

```
model.model.layers.0.xxx
```

This does **not** match the actual checkpoint keys and will cause weight loading mismatches.

The same issue exists in `atom/models/qwen3_moe.py`.

### Suggested Fix

Change the prefix to an empty string:

```python
loaded_weights_record = load_model_in_plugin_mode(
    model=self,
    config=self.atom_config,
    prefix=""
)
```

This ensures the parameter names align exactly with the checkpoint structure and avoids accidental double-prefixing.

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,896 @@
...
+                if num_decodes == _QWEN_GLUON_PA_DECODE_BS:
+                    self.paged_attention_triton_plugin_mode(
+                        q=query[:num_decode_tokens],
+                        k_cache=new_key_cache,
+                        v_cache=new_value_cache,
+                        k_scale=k_scale,
+                        v_scale=v_scale,
+                        out=output_actual_tokens[:num_decode_tokens],
+                        attn_metadata=attn_metadata,
+                    )
```
ChuanLi1101, line 700: `paged_attention_triton_plugin_mode` has the following signature:

```python
def paged_attention_triton_plugin_mode(
    self,
    q: torch.Tensor,
    k_cache: torch.Tensor,
    v_cache: torch.Tensor,
    ...
)
```

So in the call at line 839–840, passing `k=` and `v=` is incorrect. Those names do not exist in the function signature and will either:

* Be swallowed into `**kwargs` (if present), or
* Directly raise a `TypeError` for unexpected keyword arguments.

### Suggested Fix

Update the call site to match the actual parameter names:

```python
self.paged_attention_triton_plugin_mode(
    q=query[:num_decode_tokens],
    k_cache=new_key_cache,
    v_cache=new_value_cache,
    ...
)
```

This keeps the interface consistent with the function definition and avoids subtle runtime failures.

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,890 @@
...
+        attn_metadata: "AttentionMetaData",
+        query: torch.Tensor,
+        key_cache,
+        value_cache,
+        output: torch.Tensor,
+        cu_seqlens_q: torch.Tensor,
+        max_seqlen_q: int,
+        block_table: torch.Tensor,
+        k_scale: float,
+        v_scale: float,
```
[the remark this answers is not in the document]
ChuanLi1101: echo:
The function signature defines:

```python
k_scale: float
v_scale: float
```

(lines 462–463)

However, these are later passed to `cp_mha_gather_cache` as `k_scales` and `v_scales`, and `cp_mha_gather_cache` expects `torch.Tensor` (lines 138–139).

This creates a type inconsistency and can lead to runtime errors or silent misuse.

### Suggested Fix

Update the function signature to use `torch.Tensor` instead of `float`:

```python
def extend_for_sliding_window(
    ...
    k_scale: torch.Tensor,
    v_scale: torch.Tensor,
):
```

This aligns the declared type with actual downstream usage and avoids subtle type bugs during cache gather operations.

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,896 @@
...
+                k_scale=k_scale,
+                v_scale=v_scale,
+            )
+
+            qkv = qkv.view(qkv.shape[0], -1, self.head_dim)
+            q, k, v = qkv.split(
+                [self.num_heads, self.num_kv_heads, self.num_kv_heads], dim=1
+            )
+        # elif use_triton_attn and self.rotary_emb is not None:
+        elif 0:
```
ChuanLi1101: elif 0: will never execute. That means the entire fused_qk_rope_reshape_and_cache path is effectively dead code.

atom/plugin/vllm/model_wrapper.py
```diff
@@ -0,0 +1,141 @@
...
+
+        logger.info(f"Construct ATOM model {model_arch} for vLLM plugin mode")
+        self.model = model_cls(self.atom_config)
+
+        if self.model is None:
+            model_arch = vllm_config.model_config.architectures[0]
+            raise ValueError(
+                f"The model {model_arch} is not supported by model impl backend atom"
+            )
+
```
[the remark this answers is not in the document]
ChuanLi1101: model_cls(self.atom_config) will either:

Return a valid object instance, or

Raise an exception.

A constructor does not return None unless __new__ is explicitly overridden to do so (which is extremely uncommon and not the case here).

atom/plugin/attention.py
```diff
@@ -0,0 +1,625 @@
...
+            context_batch_size = num_decodes
+        context_graph_bs = context_batch_size
+
+        num_actual_tokens = common_attn_metadata.num_actual_tokens
+        context = Context(
+            positions=None,
+            is_prefill=has_prefill,
+            batch_size=context_batch_size,
+            graph_bs=context_graph_bs,
+        )
```
ChuanLi1101, line 504: At attention_mha.py line 686:

if attn_metadata is not None:
    position = attn_metadata.plugin_metadata.context.positions

When fusion mode is disabled (ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION=False), position becomes None.

Later, calling:

self.rotary_emb(positions, q, k)

will crash because rotary_emb expects valid position indices (typically a torch.Tensor), not None.

atom/plugin/attention.py
```diff
@@ -0,0 +1,625 @@
...
+            sliding_window_sizes.add((layer.impl.sliding_window - 1, 0))
+
+        while len(sliding_window_sizes) > 0:
+            sliding_window_config = sliding_window_sizes.pop()
+            if sliding_window_config is not None and sliding_window_config[0] != -1:
+                assert (
+                    self.aot_sliding_window is None
+                ), "Aiter Backend only support one valid sliding window"
+                self.aot_sliding_window = sliding_window_config
+
```
ChuanLi1101, line 245: sliding_window_sizes is typed to allow None, but in practice add() always receives a tuple.

If layer.impl.sliding_window is None, then at line 232 you’ll hit:

TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'

because the code later assumes it is an int (or tuple of ints) and performs arithmetic on it.

atom/plugin/attention.py
zejunchen-zejun: Thank you Richard for comments. Here we use decorator to wrapper the ATOM attention metadata builder class to make it follow the vLLM inherit convention when running ATOM+vLLM, like ROCM_AITER_FA backend in vLLM. Below is related code. https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/rocm_aiter_fa.py#L395-L397
Here I added check to make sure generic_base can be safely parameterized

atom/plugin/prepare.py
zejunchen-zejun: It makes perfect sense. The `prepare_model` interface is only for SGLang for now, so here I remove the legacy code and add the assertion to make sure only SGLang plugin mode will call `prepare_model` to construct model.

atom/plugin/vllm/platform.py
zejunchen-zejun: Great points. Insightful! I commit the suggest fix.

atom/plugin/attention_mha.py
zejunchen-zejun: Fixed the argument assignment issue.

atom/plugin/attention_mha.py
zejunchen-zejun: Fixed the legacy code!

atom/plugin/attention.py
zejunchen-zejun: Yes. It makes sense. Previously `position` is built during constructing attn metadata, while later I have found the vLLM will pass the positions for the whole model, which is more accurate, especially for the mixed and chunked batch. Under this circumstance, I plan to use `position` passed from vLLM but I have not removed the legacy position in attn metadata context. Here else branch can be safely removed because the ROPE will be calculated outside the attention when set ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION=False, so the positions will not be used under this circumstance

atom/plugin/attention.py
zejunchen-zejun: Make sense. Fixed!

atom/models/qwen3.py
zejunchen-zejun: Thank you for suggestions. Here we need to use `prefix="model."` because we need to record the loaded weight name to match which recorded on vLLM side. This mechanism is developed by vLLM to ensure that all of the required weights would be correctly loaded. Here is the related code: https://github.com/vllm-project/vllm/blob/7e9149d9a9f00f752adb10179d6969acdbc4351b/vllm/model_executor/model_loader/default_loader.py#L290-L301
```
            weights_not_loaded = weights_to_load - loaded_weights
            if weights_not_loaded:
                raise ValueError(
                    "Following weights were not initialized from "
                    f"checkpoint: {weights_not_loaded}"
                )
```
**If the names of loaded weight on ATOM side don't match those on vLLM side, the vLLM will report the error like above.**

Here the loaded weight on ATOM side could be:
```
'model.model.layers.81.mlp.gate.weight',
'model.model.layers.29.mlp.experts.w13_weight'
```
There are 2 model prefixes because the `Qwen3ForCausalLM` is constructed in the `ATOMModelBase`, which works as a model wrapper, so the vLLM assign the 2 model prefix to the required weight. I will leave some comments here to make it more clear.

atom/plugin/attention.py
```diff
@@ -0,0 +1,638 @@
...
+        raise TypeError(
+            f"{self.__class__.__name__} is a utility class and should not be instantiated. "
+            "Its methods are meant to be added to other classes via decorators."
+        )
+
+    @staticmethod
+    def get_supported_kernel_block_sizes():
+        from vllm.v1.attention.backend import MultipleOf
+
+        return [MultipleOf(16)]
```
ganyi1996ppo: `MultipleOf(16)` is problematic in vllm. This might cause the vllm allocate unsupported block size for mamba like model, Qwen3-Next for example. We have changed this to [16, 32]. Maybe you can do it too here, or maybe just inherit this interface from vllm's `rocm_aiter_backend`

atom/plugin/attention_mha.py
```diff
@@ -0,0 +1,891 @@
...
+        if DEQUANT:
+            k_scale = 1.0
+            v_scale = 1.0
+            k_reg = k_reg.to(tl.float32) * k_scale
+            v_reg = v_reg.to(tl.float32) * v_scale
+        tl.store(key_ptr_offset + col_offsets, k_reg)
+        tl.store(value_ptr_offset + col_offsets, v_reg)
+
+
+def cp_mha_gather_cache(
```
valarLip: this one can be moved to base_attention as we do plan to add extend_attn support

atom/plugin/attention.py
zejunchen-zejun: Fixed. Decide to modify here instead of inherit from vLLM as ATOM can have higher freedom.

atom/plugin/attention_mha.py
zejunchen-zejun: ok sure! I will move this triton kernel to base_attention, so that it can be used for both server mode and plugin mode

PR #146, opened 2026-01-16
kv and output scale loading bug -- FIX

amirumoAMD:
## Motivation

Aim was to create a proper solution that didn't just skip over the parameter for kv_scale or output_scale in LLFP4 or LLFP8,
and loaded each parameter properly.

## Technical Details

Small changes to attention_mha to have k_scale and v_scale load as parameters not just NoneType, functions to remap the parameter names for the inputted tensors, and handling weight loading.

## Test Plan

Trace looks normal, and lm_eval results match vllm lm_eval results with the same command used.

## Test Result

Testing passed.

## Submission Checklist

### DIFF
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -1,35 +1,22 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
 
-# from flash_attn import flash_attn_with_kvcache
-from dataclasses import dataclass
+from typing import Optional
 
 import aiter
 import torch
 import triton
 import triton.language as tl
-from aiter.paged_attn import PagedAttention
-from torch import nn
-from typing import Optional
-
-from atom.utils.forward_context import (
-    ForwardContext,
-    get_forward_context,
-)
-from atom.config import get_current_atom_config
-from .attention_mla import MLAModules
-from aiter.ops.triton.unified_attention import unified_attention
-from aiter.ops.triton.gluon.pa_decode_gluon import pa_decode_gluon
-from aiter.ops.triton.fused_kv_cache import fused_qk_rope_reshape_and_cache
 from aiter import fused_qk_norm_rope_cache_quant_shuffle
+from aiter.ops.triton.fused_kv_cache import fused_qk_rope_reshape_and_cache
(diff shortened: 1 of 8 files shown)


atom/model_ops/attention_mha.py
```diff
@@ -58,7 +58,10 @@ def __init__(
         self.k_cache = self.v_cache = torch.tensor([])
         self.kv_cache_dtype = kv_cache_dtype
         self.max_model_len = 0
-        self.k_scale = self.v_scale = None
+        self.k_scale = torch.nn.Parameter(torch.tensor(-1.0),
+                                           requires_grad=False)
+        self.v_scale = torch.nn.Parameter(torch.tensor(-1.0),
```
valarLip, line 52: we never use this ?

atom/model_ops/attention_mha.py
```diff
@@ -58,7 +58,10 @@ def __init__(
         self.k_cache = self.v_cache = torch.tensor([])
         self.kv_cache_dtype = kv_cache_dtype
         self.max_model_len = 0
-        self.k_scale = self.v_scale = None
```
ChuanLi1101, line 60: change to this?
self.k_scale = torch.nn.Parameter(torch.tensor(-1.0),
                                   requires_grad=False)
self.v_scale = torch.nn.Parameter(torch.tensor(-1.0),
                                   requires_grad=False)
 The default value of -1.0 serves as a sentinel to detect if scales were actually loaded.

atom/model_ops/attention_mha.py
```diff
@@ -58,7 +58,10 @@ def __init__(
...
+        self.k_scale = torch.nn.Parameter(torch.tensor(-1.0),
+                                           requires_grad=False)
+        self.v_scale = torch.nn.Parameter(torch.tensor(-1.0),
+                                           requires_grad=False)
```
ChuanLi1101, line 53: Consider adding a comment to document the sentinel value `-1.0`:

```python
# -1.0 indicates scales not loaded from checkpoint; 
# will be populated during KV cache allocation
```

Also, this change from `None` to `nn.Parameter` may break existing code that checks `if self.k_scale is None`. Please verify no such checks exist in the codebase, or update them accordingly.

ChuanLi1101 (commented):
Mostly good with minor suggestions.

PR #158, opened 2026-01-22
Several bugfixes

azaidy:


atom/model_engine/engine_core.py
```diff
@@ -113,20 +112,17 @@ def __init__(self, config: Config, input_address: str, output_address: str):
...
+        # This ensures CoreManager receives READY before any other requests
         self._send_ready_signal()
         logger.info(f"{self.label}: EngineCore fully initialized and ready")
 
+        # Start the input thread after READY signal is sent
+        self.input_thread = threading.Thread(
+            target=self.process_input_sockets, args=(self.input_address,), daemon=True
+        )
+        self.input_thread.start()
+
```
ChuanLi1101, line 125: +assert self.input_thread.is_alive(), "Input thread failed to start"

 Defensive programming - ensures the thread actually started

ChuanLi1101 (dismissed):
race condition in engine_core.
and torch.compile Compatibility: Calling get_pp_group() during forward() breaks torch.compile because it involves dynamic lookups.

Fix makes sense to me. Minor suggestion. Approved for PR merging.

atom/model_engine/engine_core.py
azaidy: Thanks for the suggestion :) Updated to include it

ChuanLi1101 (commented):
LGTM

PR #167, opened 2026-01-27
[Model] add kimi-k2-thinking fp4 support

gbyu-amd:
## Motivation
This PR added support for amd/Kimi-K2-Thinking-MXFP4, which only quantize the linear layers in MoE experts to MXFP4 format, i.e., attn layers, MoE gate, dense MLP layers, MoE shared experts and lm_head remains in bf16 weight.

### TP4 result
```bash
python -m atom.entrypoints.openai_server --model /data/models/Kimi-K2-Thinking-MXFP4 --trust-remote-code -tp 4 --kv_cache_dtype fp8
```
#### Accuracy
image

**NOTE**: torch.compile will hit an error with triton==3.5.1. It's a known issue as detailed in https://github.com/pytorch/pytorch/issues/161618, and already fixed by https://github.com/pytorch/pytorch/commit/05eeb29976fffd8331ea6c7d30960fdf48626294.
Either upgrade torch or downgrade triton can resolve the issue. 
The result above is obtained by downgrading triton to 3.4.0.
image
 

### TP8 result
Since Kimi-k2 has 64 num_heads, each rank will handle 8 heads when running TP8. Some existing kernels are not applicable in this case:
- get_mla_metadata_info_v1() missing support for num_heads=8
- New MLA decode kernel is needed for num_heads=8. One impl is available at https://github.com/ROCm/aiter/tree/zan_triton_mla_ps_mi355, and integration example refer to https://github.com/zhyajie/vllm/blob/fe76b4b7c183af745d656685611d91b297a0eee1/vllm/v1/attention/backends/mla/rocm_aiter_mla.py#L283. Need to confirm if this kernel is performant enough and can be merged to main.
 
 

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -704,14 +704,18 @@ class DeepseekV2MLP(nn.Module):
             hidden_size,
             [intermediate_size] * 2,
             bias=False,
-            quant_config=get_quant_config_for_layer(quant_config, prefix=f"{prefix}.gate_up_proj"),
+            quant_config=get_quant_config_for_layer(
+                quant_config, prefix=f"{prefix}.gate_up_proj"
+            ),
             prefix=f"{prefix}.gate_up_proj",
         )
         self.down_proj = RowParallelLinear(
             intermediate_size,
             hidden_size,
             bias=False,
-            quant_config=get_quant_config_for_layer(quant_config, prefix=f"{prefix}.down_proj"),
+            quant_config=get_quant_config_for_layer(
+                quant_config, prefix=f"{prefix}.down_proj"
+            ),
             reduce_results=reduce_results,
             prefix=f"{prefix}.down_proj",
         )
@@ -780,7 +784,9 @@ class DeepseekV2MoE(nn.Module):
             intermediate_size=config.moe_intermediate_size,
             reduce_results=False,
             renormalize=config.norm_topk_prob,
-            quant_config=get_quant_config_for_layer(quant_config, prefix=f"{prefix}.experts"),
+            quant_config=get_quant_config_for_layer(
+                quant_config, prefix=f"{prefix}.experts"
+            ),
             use_grouped_topk=True,
             num_expert_group=config.n_group,
             topk_group=config.topk_group,
--- a/atom/models/utils.py
+++ b/atom/models/utils.py
@@ -283,7 +283,12 @@ def should_ignore_layer(
                 return True
     return False
 
+
 def get_quant_config_for_layer(
-    quantization_config: Optional[QuantizationConfig], prefix: str    
+    quantization_config: Optional[QuantizationConfig], prefix: str
 ) -> Optional[QuantizationConfig]:
-    return None if should_ignore_layer(quantization_config, prefix) else quantization_config
+    return (
+        None
+        if should_ignore_layer(quantization_config, prefix)
+        else quantization_config
+    )


atom/models/utils.py
```diff
@@ -261,3 +264,21 @@ def fast_topk(values, topk, dim):
     else:
         # Use topk for efficiency with larger k values
         return torch.topk(values, topk, dim=dim)
+
+
+def should_ignore_layer(quantization_config: Optional[QuantizationConfig], prefix: str) -> bool:
+    exclude_layers: List[str] = quantization_config["exclude_layers"]
```
ChuanLi1101: If quantization_config is None, accessing ["exclude_layers"] will raise TypeError.

atom/models/utils.py
```diff
@@ -261,3 +264,21 @@ def fast_topk(values, topk, dim):
...
+    for exclude_layer in exclude_layers:
+        if exclude_layer.startswith("re"):
+            # case "re:model.layers.*self_attn.*", remove the 're:' prefix
+            regex_pattern = exclude_layer[3:]
+            if re.search(regex_pattern, prefix):
+                return True
+        elif exclude_layer.startswith(prefix):
```
ChuanLi1101: exclude_layer.startswith(prefix) is backwards - should be prefix.startswith(exclude_layers)

atom/models/utils.py
```diff
@@ -261,3 +264,21 @@ def fast_topk(values, topk, dim):
...
+            # case "model.layers.0.self_attn.q_a_proj"
+            return True
+        else:
+            # case "lm_head". Common practice won't quant lm_head, however.
+            if prefix.split(".")[-1] == exclude_layer:
+                return True
+    return False
```
ChuanLi1101: Consider sth like this?

def should_ignore_layer(quantization_config: Optional[QuantizationConfig], prefix: str) -> bool:
    if quantization_config is None:
        return False
    exclude_layers: List[str] = quantization_config.get("exclude_layers", [])
    if not exclude_layers:
        return False
    for exclude_layer in exclude_layers:
        if exclude_layer.startswith("re:"):
            # case "re:model.layers.*self_attn.*", remove the 're:' prefix
            regex_pattern = exclude_layer[3:]
            if re.search(regex_pattern, prefix):
                return True
        elif prefix.startswith(exclude_layer):
            # case "model.layers.0.self_attn.q_a_proj"
            return True
        elif prefix.split(".")[-1] == exclude_layer:
            # case "lm_head". Common practice won't quant lm_head, however.
            return True
    return False

atom/models/deepseek_v2.py
```diff
@@ -711,14 +712,14 @@ def __init__(
...
-            quant_config=quant_config,
+            quant_config=None if should_ignore_layer(quant_config, prefix=f"{prefix}.gate_up_proj") else quant_config,
             prefix=f"{prefix}.gate_up_proj",
         )
         self.down_proj = RowParallelLinear(
             intermediate_size,
             hidden_size,
             bias=False,
-            quant_config=quant_config,
+            quant_config=None if should_ignore_layer(quant_config, prefix=f"{prefix}.down_proj") else quant_config,
```
ChuanLi1101: The pattern None if should_ignore_layer(...) else quant_config is repeated 4 times. Consider a helper function or local variable.

gbyu-amd: Thanks for the suggestion! Update the code based on all the comments.

gbyu-amd:
Please let me know if you have any concern @valarLip

PR #169, opened 2026-01-29
[schedule] allow delay in prefill schedule to batch more requests

gbyu-amd:
## Submission Checklist

### DIFF
--- a/atom/model_engine/scheduler.py
+++ b/atom/model_engine/scheduler.py
@@ -298,6 +298,10 @@ class Scheduler:
             return (False, 0)
 
     def _passed_delay(self, now: float) -> bool:
+        # borrowed from https://github.com/vllm-project/vllm/pull/3279
+        # if the earliest arrived request has waited long enough,
+        # i.e., > delay_factor * last_prompt_latency (the latency of last prefill in unit of seconds),
+        # new prefill should be scheduled immediately
         if self.prev_prompt:
             self.last_prompt_latency = now - self.prev_time
         self.prev_time, self.prev_prompt = now, False


atom/model_engine/scheduler.py
```diff
@@ -94,7 +101,11 @@ def schedule(self) -> tuple[ScheduledBatch, dict[int, Sequence]]:
             # self.block_manager.reset()
             return None
 
-        while self.waiting and num_seqs_prefill < self.max_num_seqs:
+        while (
+            self._passed_delay(time.time())
```
valarLip: maybe only call time.time() when self.delay_factor>0?

gbyu-amd: Updated.

atom/model_engine/scheduler.py
```diff
@@ -284,3 +296,17 @@ def get_next_batch_info(self) -> tuple[bool, int]:
         else:
             # No requests
             return (False, 0)
+
+    def _passed_delay(self, now: float) -> bool:
+        if self.prev_prompt:
+            self.last_prompt_latency = now - self.prev_time
```
zejunchen-zejun, line 306: Could you help add comments for the unit of this time duration?

valarLip: good point

atom/model_engine/arg_utils.py
```diff
@@ -156,6 +157,13 @@ def add_cli_args(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
             default=0.9,
             help="GPU memory utilization (0.0 to 1.0)",
         )
+        parser.add_argument(
+            "--scheduler-delay-factor",
+            type=float,
+            default=0.0,
+            help="Apply a delay (of delay factor multiplied by previous"
+            "prompt latency) before scheduling next prompt.",
+        )
```
zejunchen-zejun, line 166: Hi, @gbyu-amd 
Could you add some comments here for guiding users how to set this argument? For example, someone is benchmarking the model under a specific scenario, he/she should know how to set this flag, > 1.0 or < 1.0?

atom/model_engine/scheduler.py
gbyu-amd: Done.

atom/model_engine/arg_utils.py
gbyu-amd: It's not easy to give a guidance. Basically, it's a trade-off between TTFT and ITL/TPOT. In random ISL/OSL scenarios, setting the delay factor to >0 value would allow batch more prefill requests, this may hurt the TTFT (considering the increased waiting time for the earliest arrived request) but improve the TPOT (decoding of some requests will not be interrupted by frequently scheduled prefill with only one or two requests) and the overall throughput.

valarLip (approved):
LGTM

PR #189, opened 2026-02-05
[fix] clean code for qwen3 moe

gbyu-amd:
## Motivation
The kernel fused_rope_rms has been deleted from aiter. Qwen3 moe also don't really need it as well.

## Submission Checklist


atom/models/qwen3_moe.py
valarLip, line 1: maybe refine RotaryEmbeddingQKNormFused and the whole usage of ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION

gbyu-amd: Agree. let me refine it.

PR #209, opened 2026-02-12
Fix exclude layer

ZhangLirong-amd:
## Submission Checklist

### DIFF
--- a/.github/workflows/atom-benchmark.yaml
+++ b/.github/workflows/atom-benchmark.yaml
@@ -81,9 +81,26 @@ jobs:
 
     runs-on: atom-mi355-8gpu.predownload
     steps:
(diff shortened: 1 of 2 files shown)


atom/models/deepseek_v2.py
valarLip: forgot this one?

atom/model_ops/linear.py
```diff
@@ -504,9 +504,12 @@ def __init__(
         bias: bool = False,
         quant_config: Optional[QuantizationConfig] = None,
         source_quant_dtype: torch.dtype = None,
+        prefix: Optional[str] = None,
```
valarLip: maybe we need this to be a "must' instead of optional

PR #212, opened 2026-02-13
Fix CI container name collision for parallel matrix jobs

sunway513:
## Summary
- Multiple matrix jobs sharing the same runner used a hardcoded container name `atom_test`, causing cleanup steps to kill sibling jobs' containers and leading to intermittent GPU OOM errors.
- Use `CONTAINER_NAME=atom_test_${{ strategy.job-index }}` so each job operates on its own container without interference.
- GPU isolation is already handled by `gha-render-devices` — the container name collision was the sole root cause.

## Changes
- Added `CONTAINER_NAME` env var using `strategy.job-index` for unique per-job container names
- Replaced all 12 hardcoded `atom_test` container references with `$CONTAINER_NAME`
- Used exact-match filter `name="^${CONTAINER_NAME}$"` in cleanup to avoid interfering with sibling jobs
- Docker image tag `atom_test:ci` left unchanged (shared image is fine)

## Test plan
- [ ] Verify CI passes with parallel jobs on shared runners (linux-atom-mi355-1)
- [ ] Confirm no container name collisions in job logs
- [ ] Verify cleanup step only stops/removes its own container

### DIFF
--- a/.github/workflows/atom-test.yaml
+++ b/.github/workflows/atom-test.yaml
@@ -190,6 +190,9 @@ jobs:
     if: ${{ needs.pre-checks.result == 'success' && (!github.event.pull_request || github.event.pull_request.draft == false) }}
     runs-on: ${{ matrix.runner }}
 
+    env:
+      CONTAINER_NAME: atom_test_${{ strategy.job-index }}
+
     steps:
       - name: Kill all Docker containers and clean up workspace
         if: (matrix.run_on_pr == true || github.event_name != 'pull_request') && matrix.runner == 'atom-mi355-8gpu.predownload'


.github/workflows/atom-test.yaml
```diff
@@ -169,21 +170,21 @@ jobs:
           -e ATOM_DISABLE_MMAP=true \
           -v "${{ github.workspace }}:/workspace" \
           -w /workspace \
-          --name atom_test \
+          --name $CONTAINER_NAME \
```
ChuanLi1101: Minor suggestion: --name "$CONTAINER_NAME" \ in case for space exceptions in the container_name

ChuanLi1101 (approved):
LGTM, thanks.

.github/workflows/atom-test.yaml
```diff
@@ -78,6 +78,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     env:
       ATOM_BASE_NIGTHLY_IMAGE: rocm/atom-dev:latest
+      CONTAINER_NAME: atom_test_${{ strategy.job-index }}
```
ChuanLi1101: Follow the co-pilot: Consider  change  CONTAINER_NAME: atom_test_${{ strategy.job-index }}
to:
CONTAINER_NAME: atom_test_${{ github.run_id }}_${{ strategy.job-index }}

The reason is strategy.job-index is only unique within a single workflow run. If two PRs trigger CI at the same time and are scheduled on the same runner, both runs may have job-index=0, resulting in both generating atom_test_0 and causing a naming collision. Adding github.run_id. Even if two workflow runs execute concurrently on the same runner, they will not collide.

PR #220, opened 2026-02-16
Enable Triton MXFP4 MoE on gfx950 for GPT-OSS

ChuanLi1101:
## Summary

- Extend the Triton MoE kernel path (matmul_ogs + routing from triton_kernels) to gfx950 (MI355X) when ATOM_USE_TRITON_GEMM=1
- Enables GPT-OSS MXFP4 models on MI355X to use the optimized Triton MoE path with fused routing, Swiglu activation (alpha=1.702, limit=7.0), and matmul_ogs GEMM
- Opt-in only: without ATOM_USE_TRITON_GEMM=1, gfx950 continues to use the CK/ASM MoE path (no behavior change)

## Motivation

The triton_kernels package already supports gfx950 via GFX950MXScaleLayout in _swizzle_mxfp4 (see fused_moe_triton.py), but Mxfp4MoEMethod only enables the Triton path for gfx94x. This PR extends it to gfx950 when explicitly requested via env var.

Builds on #218 (AllReduce+RMSNorm fusion for GPT-OSS).

## Changes

- atom/model_ops/moe.py: Extended use_triton check in Mxfp4MoEMethod to include gfx950 when ATOM_USE_TRITON_GEMM=1

## Precision

- No precision loss: same MXFP4 weight data (just different layout), same Swiglu activation parameters, same softmax routing
- The Triton path uses the same weight data as CK, just swizzled for GPU efficiency via _swizzle_mxfp4

## Test Plan

- [x] Run GPT-OSS-120B MXFP4 inference with ATOM_USE_TRITON_GEMM=1 on MI355X
- [x] Compare output accuracy against CK path (without ATOM_USE_TRITON_GEMM)
- [x] Benchmark throughput on InferenceMax with various ISL/OSL combinations
- [x] Verify no regression when ATOM_USE_TRITON_GEMM is not set (default CK path)

### DIFF
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -43,6 +43,7 @@ from atom.model_ops.utils import (
     per_tensor_dequantize,
     shuffle_weights,
 )
+from atom.utils import envs
 from atom.utils.custom_register import direct_register_custom_op
 from atom.utils.forward_context import get_forward_context
 from torch import nn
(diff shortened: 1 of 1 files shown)


ChuanLi1101:
Good point — moving MoE kernels from triton_kernels into aiter would simplify the deps (one less external pkg).

That said, this PR doesn’t add a new dependency — it just extends the existing triton_kernels path (already used for gfx94x) to gfx950. The has_triton_kernels() guard + import have been there since the original Triton MoE integration.

PR #227, opened 2026-02-20
Refactor ATOM for top-k top-p sampling support

aryaman-gupta:
## Summary

  Adds top-k and top-p sampling support to ATOM, complementing existing temperature-based sampling.

  ## Changes

  | File | Changes |
  |------|---------|
  | `sampling_params.py` | Added `top_k: int = -1` and `top_p: float = 1.0` fields with validation |
  | `sequence.py` | Store `top_k` and `top_p` from sampling params |
  | `scheduler.py` | Added `top_ks` and `top_ps` lists to `ScheduledBatch` |
  | `model_runner.py` | Added GPU buffers, updated `prepare_sample` with CPU-side uniformity optimization |
  | `sampler.py` | Added top-k/top-p filtering with aiter integration. Default temperature-base Gumbel Max path when filtering disabled (no overhead). Native PyTorch fallback when `aiter.ops.sampling` unavailable (marked experimental). |
  | `openai_server.py` | Wired up `top_k` and `top_p` parameters |
  
  ## Basic Testing
  Run a model:
  ```
  python3 -m atom.entrypoints.openai_server --model <model_path> <additional_params> --host 0.0.0.0 --port 8000 
  ```
  Query with top-k and top-p parameters:
  ```python
  import requests

response = requests.post(
    "http://localhost:8000/v1/completions",
    json={
        "model": "/it-share/gpt-oss-120b/",
        "prompt": "The capital city of France is",
        "max_tokens": 32,
        "temperature": 0.8,
        "top_k": 10,
        "top_p": 0.8,
        "stream": True
    },
    stream=True
)
for line in response.iter_lines():
    print(line)
```

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -27,7 +27,7 @@ from atom.model_engine.scheduler import ScheduledBatch, ScheduledBatchOutput
 from atom.model_engine.sequence import Sequence, SequenceStatus, SequenceType
 from atom.model_loader.loader import load_model
 from atom.model_ops.rejection_sampler import RejectionSampler
-from atom.model_ops.sampler import Sampler
+from atom.model_ops.sampler import SAMPLER_EPS, Sampler
 from atom.spec_decode.eagle import EagleProposer
 from atom.utils import (
     CpuGpuBuffer,
@@ -1440,11 +1440,15 @@ class ModelRunner:
 
     def prepare_sample(
         self, batch: ScheduledBatch
-    ) -> tuple[torch.Tensor, torch.Tensor | None, torch.Tensor | None]:
+    ) -> tuple[torch.Tensor, torch.Tensor | None, torch.Tensor | None, bool]:
         bs = batch.total_seqs_num
 
+        # Check on CPU whether all requests are greedy (temperature=0)
+        all_greedy = (batch.temperatures == 0).all()
+
         temp_buffer = self.forward_vars["temperatures"]
-        temp_buffer.np[:bs] = batch.temperatures
+        # Clamp temperatures on CPU to avoid division by zero in sampler
+        temp_buffer.np[:bs] = np.maximum(batch.temperatures, SAMPLER_EPS)
         temperatures = temp_buffer.copy_to_gpu(bs)
 
         # Check on CPU whether filtering is needed to avoid GPU sync in sampler.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: atom/model_ops/sampler.py [hunk 1]
```diff
@@ -16,26 +25,194 @@ def __init__(self):
...
+            )
+        else:
+            return self._native_sample(probs, top_ks, top_ps, temperatures)
+
+    def _to_tensor_scalar(self, x: torch.Tensor):
+        """Convert to (tensor, scalar) tuple for aiter ops."""
+        if x is None:
+            return (None, 0)
+        if (x == x[0]).all():  # Uniform value - use scalar for efficiency
+            return (None, x[0].item())
```
valarLip, line 146: maybe let's just pass scalar from scheduler, intead of uniform value for all reqs if it's all same for every reqs in this batch

aryaman-gupta: This would require changing the API of `prepare_sample` to optionally return a scalar instead of a Tensor. https://github.com/ROCm/ATOM/blob/aryaman/topk-topp-support/atom/model_engine/model_runner.py#L1182

I am not sure whether this is worth it. Do you suggest making this change?

valarLip: it still can be a tensor with shape=[1], it's important to make sure we do our best to reduce the useless data transfor for CPU2CPU(scheduler to all runner) and CPU2GPU(runner to GPU)

aryaman-gupta: That make sense! Done in b74f38295279171f6c86c6dbd9f625d61d13ca23

atom/model_engine/model_runner.py
```diff
@@ -1305,23 +1307,45 @@ def prepare_inputs(self, batch: ScheduledBatch, input_ids: torch.Tensor = None):
...
+        temp_buffer = self.forward_vars["temperatures"]
+        temp_buffer.np[:bs] = batch.temperatures
+        temperatures = temp_buffer.copy_to_gpu(bs)
+
+        # For top_ks and top_ps, check uniformity on CPU before GPU copy.
+        # If all values are the same, only copy a single element to save bandwidth.
+        top_k_buffer = self.forward_vars["top_ks"]
+        top_k_buffer.np[:bs] = batch.top_ks
+        if bs > 1 and all(k == batch.top_ks[0] for k in batch.top_ks):
+            top_ks = top_k_buffer.copy_to_gpu(1)
```
valarLip: a good progress for moving the check here, but we can do more to move the check to scheduler, which can reduce the copy for scheduler to 8 gpu_runners (cpu to cpu copu)

atom/model_ops/sampler.py
```diff
@@ -16,26 +36,220 @@ def __init__(self):
...
+        top_ks: torch.Tensor,
+        top_ps: torch.Tensor,
+    ) -> bool:
+        """Check if any request needs top-k or top-p filtering."""
+        if top_ks is None and top_ps is None:
+            return False
+
+        needs_topk = top_ks is not None and (top_ks != -1).any()
+        needs_topp = top_ps is not None and (top_ps < 1.0).any()
+
```
ChuanLi1101, line 85: .any() on a GPU tensor triggers a CPU-GPU synchronization barrier on every single forward pass, even when no request uses top-k/top-p. This will cause a significant throughput regression for all users.

ChuanLi1101 (commented):
MInor comments FYI.

atom/model_ops/sampler.py
aryaman-gupta: Resolved in 73e788f9fd280e7436961a9e824417585dd854e4

PR #236, opened 2026-02-25
[QUARK-480] Per-layer quantization config support

thpereir:
Introduce a structured per-layer quantization system that replaces ad-hoc dict-based config with:

- LayerQuantSpec: frozen dataclass capturing quant_type, quant_dtype, is_dynamic, quant_method, checkpoint_dtype, and extensible flags. Supports two-phase online quant modeling (init as no_quant, post-load quantize to target spec via needs_online_quant predicate).

- ParsedQuantConfig: holds global_spec, per-layer overrides (layer_specs), and exclude_layers list.

- Parser registry: QuantConfigParser ABC with register_quant_parser() decorator. Built-in parsers: QuarkParser, CompressedTensorsParser, GenericParser (fallback). Extensible via decorator registration.

- QuantizationConfig.resolve(prefix): single resolution point for
  per-layer config. Priority: layer_specs override > exclude-list >
  global spec. Supports exact match, prefix match, and re: regex patterns in exclude lists.

- LinearBase: accepts optional LayerQuantSpec via layer_spec parameter, deriving quant_type, params_dtype, and source_quant_dtype from it. Falls back to building LayerQuantSpec from dict fields for backward compatibility.

- QKVParallelLinear, MergedColumnParallelLinear, RowParallelLinear: use quant_config.resolve(prefix) to get per-layer specs, eliminating the need for get_quant_config_for_layer() at construction time.

- qwen3_moe.py: wires prefix through attention and decoder layers.

- models/utils.py: should_ignore_layer() delegates to resolve() when available, maintaining backward compatibility.

All 34 unit tests pass covering: LayerQuantSpec properties, parser registry, QuarkParser/CompressedTensorsParser parsing, resolve() with exclusions/overrides, backward compat (dict access, should_ignore_layer, get_quant_config_for_layer), LinearBase layer_spec plumbing, and QKV/Row parallel linear resolve integration.

## Submission Checklist

### DIFF
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -1984,6 +1984,13 @@ class FusedMoE(torch.nn.Module):
         ):
             # Use CompressedTensorsFp8MoEMethod for compressed-tensors format
             self.quant_method = CompressedTensorsFp8MoEMethod(quant_config, moe)
+        elif (
+            quant_config["quant_dtype"] == dtypes.fp8
+            and quant_config["quant_type"] == QuantType.per_Token
+        ):
+            # Per-channel FP8 (e.g., Quark per_Token override for MTP layers)
+            # needs CompressedTensors-style weight scale handling
+            self.quant_method = CompressedTensorsFp8MoEMethod(quant_config, moe)
         elif quant_config["quant_dtype"] == dtypes.fp8:
             self.quant_method = Fp8MoEMethod(quant_config, moe)
         elif quant_config["quant_dtype"] == dtypes.fp4x2:
@@ -2100,6 +2107,10 @@ class FusedMoE(torch.nn.Module):
         tp_rank: int,
     ):
         # for per channel weight quantization
+        # When scales are stored as [N,1] (CompressedTensors per-channel)
+        # but loaded from checkpoint as [N], reshape to match.
+        if loaded_weight.dim() == 1 and expert_data.dim() == 2:
+            loaded_weight = loaded_weight.unsqueeze(-1)
         if shard_id == "w2":
             expert_data.copy_(loaded_weight)
         elif shard_id in ("w1", "w3"):


atom/models/qwen3_moe.py
```diff
@@ -153,7 +153,7 @@ def __init__(
         rope_scaling: tuple | None = None,
         kv_cache_dtype: str = "fp16",
         layer_num: int = 0,
-        atom_config: Config = None,
```
haoyangli0109, line 155: This section should be deleted.

atom/model_ops/linear.py
```diff
@@ -200,12 +201,34 @@ def __init__(
         quant_config: Optional[QuantizationConfig] = None,
         reduce_results: bool = False,
         source_quant_dtype: torch.dtype | None = None,
```
haoyangli0109, line 202: It seems odd that linear_base simultaneously possesses both quant_config and qspec.

atom/model_ops/linear.py
```diff
@@ -200,12 +201,34 @@ def __init__(
...
+        layer_spec: Optional[LayerQuantSpec] = None,
     ):
         if quant_config is None:
             quant_config = QuantizationConfig()
-        self.source_quant_dtype = source_quant_dtype
-        quant_type = quant_config["quant_type"]
-        params_dtype = quant_config["quant_dtype"]
+
+        # --- New: prefer LayerQuantSpec if provided ---
+        if layer_spec is not None:
```
haoyangli0109, line 209: We did not apply similar processing to moe.

atom/model_ops/linear.py
thpereir: Agreed — quant_config is kept only for backward compat with callers that don't pass layer_spec yet. Once all model files pass prefix= (so subclasses resolve a layer_spec themselves), we can remove quant_config from LinearBase.__init__ entirely. For now the subclasses (ColumnParallelLinear, RowParallelLinear, etc.) already resolve and pass layer_spec, so quant_config is only used as a fallback.

atom/model_ops/linear.py
thpereir: Correct — MoE experts use FusedMoEQuantConfig (not LinearBase), so the resolve() path doesn't apply there. For this model, MoE layers use the global MXFP4 config which is correct. If we needed per-layer overrides for MoE too, that would be a separate change to the fused MoE kernel config. This PR only addresses the non-MoE linear layers (attention, gate, etc.).

PR #237, opened 2026-02-25
[QUARK-403] Add MiniMax-2.1 support

thpereir:
## Motivation

Add support for MiniMax2.1 into ATOM

## Technical Details

MiniMax2.1 uses sigmoid on the Expert selection instead of just topk

## Test Plan

Run server:
```
python -m atom.entrypoints.openai_server     --model /scratch/models/MiniMax-M2.1-MXFP4/     -tp 4     --kv_cache_dtype fp8     --max-model-len 32768     --gpu-memory-utilization 0.90
```
Run lm-eval:
```
lm_eval --model local-completions --model_args model=/scratch/models/MiniMax-M2.1-MXFP4/,base_url=http://localhost:8000/v1/completions,num_concurrent=100,max_retries=3,tokenized_requests=False --tasks gsm8k --num_fewshot 5
```

## Test Result

Results are below what we obtained on vllm so we need to debug further

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2972|±  |0.0126|
|     |       |strict-match    |     5|exact_match|↑  |0.2585|±  |0.0121|
## Submission Checklist

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -489,17 +489,6 @@ class QuantizationConfig:
     ) -> bool:
         """Match the target string or regular expression"""
 
-        # Replace layer name if packed_modules_mapping is offered
-        proj_name = layer_name.split(".")[-1]
-        if self.packed_modules_mapping is not None:
-            for ori_param, (
-                model_param,
-                shard_id,
-            ) in self.packed_modules_mapping.items():
-                if proj_name in model_param:
-                    layer_name = layer_name.replace(proj_name, ori_param)
-                    break
-
         if ignore_str.startswith("re:"):
             # case "re:model.layers.*self_attn.*", remove the 're:' prefix
             pattern = ignore_str[3:]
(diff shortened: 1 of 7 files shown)


atom/config.py
```diff
@@ -489,17 +489,6 @@ def is_equal_or_regex_match(
     ) -> bool:
         """Match the target string or regular expression"""
 
-        # Replace layer name if packed_modules_mapping is offered
```
ganyi1996ppo, line 492: hi, @thpereir , may I ask why this code piece is deleted? Can I add it back? This will affect qwen3.5's weight loading, and cause serious accuracy regression on qwen3.5

thpereir: @ganyi1996ppo let me take a look. It was affecting MiniMax loading. I will work on a fix to cover both models. 
Can you share the exact Qwen3.5 model you're using o test?

thpereir: Ok, I opened #452 to address this issue. The remap is already in place we just had to add Qwen3.5 to the list

thpereir: Actually, I saw your PR and #452 is not needed. Closed that one.
But I added a comment on your #448

PR #240, opened 2026-02-26
CI: Make ATOM benchmark can be called from other repos

gyohuangxin:


zufayu:
Hi @gyohuangxin, this has been open for ~4 months. Is this still needed, or has the cross-repo benchmark mechanism been handled differently?

gyohuangxin:
I think it's not needed now

PR #252, opened 2026-03-02
CI: Add more tests in nigthly benchmark

gyohuangxin:


.github/workflows/atom-benchmark.yaml
```diff
@@ -41,11 +41,43 @@ jobs:
...
+      # Nightly params aligned with InferenceX dsr1-fp8 config: 1k1k, 1k8k, 8k1k with multiple concurrency levels
+      NIGHTLY_PARAM_LISTS: |
+        1024,1024,1,0.8;
+        1024,1024,2,0.8;
+        1024,1024,4,0.8;
+        1024,1024,8,0.8;
+        1024,1024,16,0.8;
+        1024,1024,32,0.8;
+        1024,1024,64,0.8;
+        1024,1024,128,0.8;
```
valarLip, line 54: let's have bs max to 256

gyohuangxin: Add concurrency=256 to nightly 1k1k benchmark params, do we also need to extend concurrencies to 256 for 1k8k and 8k1k benchmark?

PR #262, opened 2026-03-04
[feat](gpt-oss): support gpt-oss for vllm plugin

PerryZhang01:
## Motivation

This PR is for supporting gpt-oss-120b (https://huggingface.co/openai/gpt-oss-120b) in vllm plugin mode.

## Technical Details

If atom has supported some model, integrating the model in plugin mode is very simple like this PR:
-register the model name
-add weight loader for the model

## Test Plan

export SAFETENSORS_FAST_GPU=1
export VLLM_ROCM_USE_AITER=1

model_path=/data/openai/gpt-oss-120b

vllm serve $model_path \
    --host localhost \
    --port 8000 \
    --tensor-parallel-size 1 \
    --trust-remote-code \
    --disable-log-requests \
    --gpu_memory_utilization 0.8 \
    --async-scheduling \
    --load-format fastsafetensors \
    --kv-cache-dtype fp8 \
    --max-num-batched-tokens 18432 \
    --max-model-len 16384 \
    --no-enable-prefix-caching \
    --compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE"}'

## Test Result

image

### DIFF
--- a/atom/plugin/attention_mha.py
+++ b/atom/plugin/attention_mha.py
@@ -75,7 +75,7 @@ class PagedAttentionImplPluginModeMethods:
             # ATOM: [num_blocks, num_kv_heads, head_size, block_size],
             # vLLM: [num_blocks, num_kv_heads, block_size // x, head_size, x],
             v_cache_template = torch.empty(
-                [num_blocks, num_kv_heads, head_size, block_size],
+                [num_blocks, num_kv_heads, block_size // x, head_size, x],
                 dtype=v_cache.dtype,
                 device="meta",
             )
@@ -673,7 +673,7 @@ class PagedAttentionImplPluginModeMethods:
                 device="meta",
             )
             v_cache_template = torch.empty(
-                [num_blocks, num_kv_heads, head_size, block_size],
+                [num_blocks, num_kv_heads, block_size // x, head_size, x],
                 dtype=v_cache.dtype,
                 device="meta",
             )


atom/models/gpt_oss.py
```diff
@@ -391,3 +393,15 @@ def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
             ckpt_up_proj_name="up_proj",
             num_experts=self.config.num_local_experts,
         )
+
+
+    def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
```
valarLip: let's overide this at top level maybe in loader.py instead of  every model...

wuhuikx: Agree. Then add a new model in OOT mode, we just need to modify the regiter.py and wrapper.py in plugin folder, which will be much more clean.

atom/models/gpt_oss.py
```diff
@@ -391,3 +393,15 @@ def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
...
+        # load weights in plugin mode and discard passed weights generator
+        # here prefix is "model." because GptOssForCausalLM is constructed in model
+        # wrapper class, so the name of loaded weights are prefixed with "model.".
+        # The vLLM will check the name of the loaded weights to make sure all the
+        # weights are loaded correctly
+        loaded_weights_record = load_model_in_plugin_mode(
+            model=self, config=self.atom_config, prefix="model."
```
ChuanLi1101: Minor Issue:

The correctness of `prefix="model."` depends on how the ATOM model is constructed inside **ATOMModelBase**. In `model_wrapper.py` line 91:

```python
self.model = model_cls(self.atom_config)
```

This means `GptOssForCausalLM` is wrapped under `ATOMModelBase.model`. As a result, the weight names seen by vLLM become `model.model.layers.0...`, while the checkpoint names used when ATOM loads the weights are `model.layers.0...`.

Adding `prefix="model."` causes ATOM to prepend `model.` to the keys in `loaded_weights_record`, turning them into `model.model.layers.0...`. This allows the weight name validation in vLLM to pass.

The key point is that this assumption depends on the **context in which vLLM calls `load_weights`**. If vLLM calls `ATOMModelBase.load_weights()` (lines 120–123 in `model_wrapper.py`), the method directly delegates to:

```python
self.model.load_weights(weights)
```

At that point, `self.model` is `GptOssForCausalLM`=。

atom/models/gpt_oss.py
ChuanLi1101: 可以在 ATOMModelBase 或通过 mixin 统一实现，然后每个 model class 不再需要单独 override。但这是架构改进，可以作为 follow-up。如果 reviewer 坚持在这个 PR 做，需要重构。

PerryZhang01: moved it to base class

atom/models/gpt_oss.py
PerryZhang01: "prefix" issue will be fixed in next PR

PerryZhang01: moved it to ATOMModelBase, so it must be called from vllm model.

PR #265, opened 2026-03-04
[feat][plugin] make ATOM mla attention works for vllm

XiaobingSuper:
## Motivation

Following https://github.com/ROCm/ATOM/pull/126, this PR makes ATOM mla attention work for the vLLM plugin model. Note: the sparse mla is not supported now and will be implemented in the next step.

## Technical Details

The design tails can be seen in https://github.com/ROCm/ATOM/pull/126.

## Test Plan

This PR does a test for Kimi-K2-Thinking-MXFP4 mode with TP4 on mi355:

```
export SAFETENSORS_FAST_GPU=1
export VLLM_ROCM_USE_AITER=1
export VLLM_RPC_TIMEOUT=1800000

export VLLM_CACHE_ROOT=/root/.cache/vllm
export TORCHINDUCTOR_CACHE_DIR=/root/.cache/inductor
# quick allreduce
export ATOM_PROFILER_MORE=1

export VLLM_TORCH_PROFILER_RECORD_SHAPES=1

model_path= Kimi-K2-Thinking-MXFP4
vllm serve $model_path \
    --host localhost \
    --port 8001 \
    --tensor-parallel-size 4 \
    --enable-expert-parallel \
    --trust-remote-code \
    --disable-log-requests \
    --gpu_memory_utilization 0.9 \
    --kv-cache-dtype fp8 \
    --max-num-batched-tokens 18432 \
    --max-model-len 16384 \
    --no-enable-prefix-caching
```

## Test Result

gsmk result"
```
|gsm8k|      3|flexible-extract|     3|exact_match|↑  |0.9371|±  |0.0067|
|     |       |strict-match    |     3|exact_match|↑  |0.9363|±  |0.0067|
  
```
## Submission Checklist

### DIFF
--- a/atom/model_ops/attention_mla.py
+++ b/atom/model_ops/attention_mla.py
@@ -102,7 +102,7 @@ class MLAAttention(nn.Module):
     def __init__(
         self,
         num_heads: int,
-        head_size: int,
+        head_dim: int,
         scale: float,
         num_kv_heads: int,
         kv_cache_dtype: str,
@@ -113,7 +113,7 @@ class MLAAttention(nn.Module):
     ) -> None:
         super().__init__()
         self.num_heads = num_heads
-        self.head_dim = head_size
+        self.head_dim = head_dim
         self.scale = float(scale)
         self.num_kv_heads = num_kv_heads
         self.kv_cache_dtype = kv_cache_dtype if kv_cache_dtype == "fp8" else "auto"
--- a/atom/model_ops/paged_attention.py
+++ b/atom/model_ops/paged_attention.py
@@ -174,7 +174,7 @@ class PagedAttention(BaseAttention):
             use_mla=self.use_mla,
         )
         impl_cls = self.attn_backend.get_impl_cls()
-        impl_args = dict(
+        self.impl = impl_cls(
             num_heads=num_heads,
             head_dim=head_dim,
             scale=scale,
@@ -191,8 +191,6 @@ class PagedAttention(BaseAttention):
             k_norm=k_norm,
             **kwargs,
         )
-        impl_args["head_size" if self.use_mla else "head_dim"] = head_dim
-        self.impl = impl_cls(**impl_args)
         compilation_config = atom_config.compilation_config
         default_name = f"MLA_{layer_num}" if self.use_mla else f"MHA_{layer_num}"
         self.layer_name = prefix if prefix is not None else default_name
--- a/atom/plugin/attention_mla.py
+++ b/atom/plugin/attention_mla.py
@@ -922,6 +922,8 @@ def MLAAttentionImplDecoratorForPluginMode(cls):
     orig_init = cls.__init__
 
     def new_init(self, *args, **kwargs):
+        if "head_size" in kwargs and "head_dim" not in kwargs:
+            kwargs["head_dim"] = kwargs.pop("head_size")
         orig_init(self, *args, **kwargs)
         _mla_plugin_mode_init(self, *args, **kwargs)
 
--- a/atom/model_ops/attentions/aiter_attention.py
+++ b/atom/model_ops/attentions/aiter_attention.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: MIT
 # Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
 
-import itertools
 from typing import Type
 
 import aiter


atom/model_ops/linear.py
```diff
@@ -385,8 +385,9 @@ def forward(
                 if self.quant_type.value == QuantType.per_1x128.value:
                     quant_func = functools_partial(quant_func, transpose_scale=True)
                 if self.quant_type.value != QuantType.per_1x32.value:
+                    # quant_func will call view, so we need to call contiguous to avoid view error
                     x, x_scale = quant_func(
-                        x,
+                        x.contiguous(),
```
XiaobingSuper: This is required for the deepseek-r1 model, where x is a sliced tensor that cannot be viewed.

atom/model_ops/attention_mla.py
```diff
@@ -577,7 +609,7 @@ def forward(
         if forward_context.context.is_dummy_run:
             # dummy run: skip real attention and return
             output_shape = list(q.shape)
-            output_shape[-1] = 7168
+            output_shape[-1] = self.num_heads * self.v_head_dim
```
XiaobingSuper: Why do we need this change? We move self.o_proj from attention to alignment with the vLLM pipeline.

atom/model_ops/paged_attention.py
```diff
@@ -88,28 +88,85 @@ def __init__(
...
             self.layer_name = prefix
             if self.layer_name in compilation_config.static_forward_context:
                 raise ValueError("Duplicate layer: {}".format(self.layer_name))
             compilation_config.static_forward_context[self.layer_name] = self
+
+            if self.use_mla:
+                max_num_tokens = (
+                    atom_config.plugin_config.vllm_scheduler_config.max_num_batched_tokens
+                )
+                compilation_config.static_forward_context["positions"] = torch.zeros(
```
ZhangLirong-amd: hi, I remember static_forward_context will save some cls details like moe and attention, for position, our Context in ForwardContext has this variable, https://github.com/ROCm/ATOM/blob/main/atom/utils/forward_context.py#L138, can we use it?

atom/model_ops/attention_mla.py
```diff
@@ -92,11 +97,12 @@ def dynamic_per_batched_tensor_quant(
     return x_scl_sat.to(dtype).contiguous(), scale.float().reciprocal()
 
 
+@MLAAttentionImplDecoratorForPluginMode
```
ganyi1996ppo, line 100: I'm not quite like the current PluginMode attention decorator. We should have better way to implement these works like isolate the attention impl from atom's original one. Maybe we can merge this first and then setup a meeting with @zejunchen-zejun to refine these part. What do you think?

zejunchen-zejun: we indeed need to refine the decorator

atom/model_ops/paged_attention.py
XiaobingSuper: For oot path, it uses the vLLM model runner and does not use the atom model runner, meaning that the forward context will not be initialized, see https://github.com/ROCm/ATOM/blob/main/atom/model_engine/model_runner.py#L1283.

atom/model_ops/attention_mla.py
```diff
@@ -134,7 +140,34 @@ def __init__(
...
+            self.cp_kv_cache_interleave_size: int = (
+                get_current_vllm_config().parallel_config.cp_kv_cache_interleave_size
+            )
+
+            self.is_aiter_triton_fp4_bmm_enabled = (
+                is_rocm_aiter_fp4bmm_enabled()
+                and self.kv_b_proj.weight.dtype == torch.bfloat16
+            )
+            # q_pad_num_heads in kwargs
+            self.q_pad_num_heads = kwargs.get("q_pad_num_heads", None)
```
ChuanLi1101: The __init__ signature at line 97-112 does not include **kwargs. The kwargs parameter is only available in the Attention base class constructor call chain. When is_vllm() is True, this code runs inside MLAAttention.__init__, where kwargs is not defined — this will raise NameError.
Suggested fix: Add **kwargs to MLAAttention.__init__ signature, or explicitly pass q_pad_num_heads as a named parameter.

atom/plugin/attention_mla.py
```diff
@@ -0,0 +1,834 @@
...
+            attn_metadata.plugin_metadata.decode.max_qo_len,
+            sm_scale=self.scale,
+            work_meta_data=work_meta_data,
+            work_indptr=work_indptr,
+            work_info_set=work_info_set,
+            reduce_indptr=reduce_indptr,
+            reduce_final_map=reduce_final_map,
+            reduce_partial_map=reduce_partial_map,
+            q_scale=layer._q_scale,
+            kv_scale=layer._k_scale,
```
ChuanLi1101, line 622: forward_decode_plugin_mode references layer._q_scale and layer._k_scale, but these are only set by process_weights_after_loading in server mode (inside the aiter_mla.py backend's prepare_decode). In plugin mode, process_weights_after_loading now takes an optional act_dtype parameter (line 72), but the scale initialization path for plugin mode is not clearly guaranteed. If these scales are None or missing, mla_decode_fwd will crash.

atom/model_ops/paged_attention.py
```diff
@@ -88,28 +88,86 @@ def __init__(
...
+
+                    def new_forward(*args, **kwargs):
+                        out = orig_impl(*args, **kwargs)
+                        if (
+                            os.getenv("ATOM_DISABLE_VLLM_PLUGIN_ATTENTION", "0").lower()
+                            == "0"
+                        ):
+                            return out
+                        return out, None
+
```
ChuanLi1101: This reads an env var on every forward call, which is in the hot path. Also, the logic is inverted — when plugin is enabled (env=0, default), it returns out; when disabled, it returns (out, None). This seems like debug/fallback code that shouldn't be in production

atom/model_ops/paged_attention.py
```diff
@@ -60,15 +60,15 @@ def __init__(
             **kwargs,
         )
 
+        self.use_mla = use_mla
         # for plugin mode
         if is_vllm():
-            self.use_mla = use_mla
-            self.rotary_emb = rotary_emb
+            self.rotary_emb = mla_modules.rotary_emb
```
ChuanLi1101: The original code (deleted line 303) was self.rotary_emb = rotary_emb (from the parameter). The new code uses mla_modules.rotary_emb. These should be the same object, but if use_mla=False, mla_modules may be None, and accessing mla_modules.rotary_emb will crash.

atom/plugin/attention_mla.py
```diff
@@ -0,0 +1,834 @@
...
+            max_seqlen_q=prefill.max_query_len,
+            max_seqlen_k=prefill.max_query_len,
+            softmax_scale=self.scale,
+            causal=True,
+            return_softmax_lse=return_softmax_lse,
+        )
+
+    def _run_prefill_context_chunk_plugin_mode(self, prefill, chunk_idx, q, k, v):
+        assert prefill.chunked_context is not None
+        assert prefill.chunked_context is not None
```
ChuanLi1101, line 224: duplicated?

atom/model_ops/paged_attention.py
XiaobingSuper: Updated, for mla, rotary_emb is none, now there is a check when using mla_modules.rotary_emb.

atom/plugin/attention_mla.py
XiaobingSuper: removed.

atom/model_ops/paged_attention.py
XiaobingSuper: Yes, this env check is removed.

atom/model_ops/attention_mla.py
XiaobingSuper: I checked the MLAAttention.init, it has **kwargs.

atom/plugin/attention_mla.py
XiaobingSuper: For plugin mode, the scale will be initialized at https://github.com/vllm-project/vllm/blob/v0.15.1/vllm/attention/layer.py#L685, and we use it. the **self.impl.process_weights_after_loadin**g using atom mla attention methond(https://github.com/vllm-project/vllm/blob/v0.15.1/vllm/attention/layer.py#L675C13-L675C52).

XiaobingSuper: For ATOM, path I checked the scales never get updated, always the same as vllm side scale

atom/plugin/attention_mla.py
```diff
@@ -0,0 +1,833 @@
...
+            if self.is_aiter_triton_fp4_bmm_enabled:
+                decode_ql_nope = batched_gemm_a16wfp4(
+                    decode_q_nope,
+                    self.W_K,
+                    self.W_K_scale,
+                    transpose_bm=True,
+                    prequant=True,
+                    y_scale=layer._q_scale if fp8_attention else None,
+                )
+            # elif self.is_aiter_triton_fp8_bmm_enabled:
```
zejunchen-zejun, line 752: what is the default value for is_aiter_triton_fp4_bmm_enabled and is_aiter_triton_fp8_bmm_enabled? Default true right?

XiaobingSuper: yes

atom/model_ops/attention_mla.py
ZhangLirong-amd: And maybe here need to change too? https://github.com/ROCm/ATOM/blob/main/atom/model_ops/base_attention.py#L195

XiaobingSuper: @ZhangLirong-amd  They don't need to change, see 
image
the unified_attention_with_output_base has self.impl.o_proj, and the output_shape[-1] is 7168.

ZhangLirong-amd: ok. Since we want to remove the hardcode 7168 there to support more mla model later

XiaobingSuper: Remove the hardcoded 7168.

atom/models/deepseek_v2.py
```diff
@@ -1899,6 +1900,15 @@ def make_empty_intermediate_tensors(
     def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
         return self.model.get_expert_mapping()
 
+    def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
```
valarLip: refer this one? https://github.com/ROCm/ATOM/pull/262

atom/model_ops/attention_mla.py
```diff
@@ -134,7 +140,34 @@ def __init__(
         )
         self.layer_num = layer_num
 
-    def process_weights_after_loading(self):
+        # for plugin mode(vllm)
```
valarLip: maybe do these stuff in atom/plugin/attention_mla.py

atom/model_ops/attention_mla.py
valarLip: self.num_heads * self.v_head_dim looks like not eaquals to 7168 for deepseek

XiaobingSuper: This is o_proj's input,  see atom path:
image and plugin path: 
image



The reason is that vllm do a_proj outside of the attention backend.

atom/models/deepseek_v2.py
XiaobingSuper: Yes, updated.

atom/model_ops/attention_mla.py
XiaobingSuper: moved.

atom/model_ops/attention_mla.py
valarLip: +1... so many dirty works

atom/model_ops/linear.py
valarLip: we will do something else to avoid contiguous which will introduce mem copy here, all our quant should support non-contiguous tensor already... we hit any issue here?

atom/model_ops/attention_mla.py
valarLip: plugin path is also in our repo... then why we have to move o_proj out of attn

atom/model_ops/linear.py
XiaobingSuper: Yes, I meet this issue for the chunked prefill path. see https://github.com/XiaobingSuper/ATOM/blob/xiaobing/oot_kimi/atom/plugin/attention_mla.py#L345.

atom/model_ops/attention_mla.py
XiaobingSuper: This change is for the fallback path, i.e., for plugin mode, but use vllm attn backend, because we use vllm MLAAttention class(here is self.attn), the forward path doesn't has o_proj, see https://github.com/vllm-project/vllm/blob/v0.15.1/vllm/attention/layer.py#L640, this is only for attention compute.

XiaobingSuper: @valarLip The plugin mode using vllm atten backend will be like(setting ATOM_DISABLE_VLLM_PLUGIN_ATTENTION=1): https://github.com/zejunchen-zejun/ATOM/blob/zejun/plugin_for_atom_1223/recipes/vLLM-ATOM-OOT-Plugin-Backend.md#launching-server-of-vllm-with-atom-oot-plugin-platform

atom/model_ops/linear.py
valarLip: then do contiguous at that place maybe, i don't like loss any perf

XiaobingSuper: Yes, I updated the code, and do contiguous at plugin side.

atom/model_ops/paged_attention.py
```diff
@@ -153,7 +204,8 @@ def __init__(
             k_norm=k_norm,
             **kwargs,
         )
-
+        impl_args["head_size" if self.use_mla else "head_dim"] = head_dim
```
valarLip: this is also comes from vllm? i would like we always use head_dim

atom/model_ops/attention_mla.py
```diff
@@ -134,7 +140,7 @@ def __init__(
         )
         self.layer_num = layer_num
 
-    def process_weights_after_loading(self):
+    def process_weights_after_loading(self, act_dtype: Optional[torch.dtype] = None):
```
valarLip: @zejunchen-zejun  we need add this arg?

atom/model_ops/paged_attention.py
XiaobingSuper: Yes, vllm use head_size, see https://github.com/vllm-project/vllm/blob/v0.15.1/vllm/attention/layer.py#L579. Before this PR https://github.com/ROCm/ATOM/commit/6c402484c8f1ea3d979cc9ce8b46db8f51d990a3, it also use head_size.

atom/model_ops/attention_mla.py
XiaobingSuper: For vLLM side, it's calling path like https://github.com/vllm-project/vllm/blob/1892993bc18e243e2c05841314c5e9c06a80c70d/vllm/attention/layer.py#L675, it needs such a arg.

atom/model_ops/paged_attention.py
valarLip: ic……… @zejunchen-zejun please unify the args…it's begin mess now

atom/model_ops/attention_mla.py
XiaobingSuper: @valarLip, currently, we only move o_proj out of attn only for plugin mode.

atom/model_ops/attention_mla.py
XiaobingSuper: @valarLip, this arg is removed now, and we warp the vllm mathon in plugin mode.

zejunchen-zejun: Hi @valarLip 
The `process_weights_after_loading` is called by vLLM side and vLLM requires that the Attention needs to have `act_type` argument, but don't worry, we can use wrapper here.

zejunchen-zejun (approved):

atom/model_ops/paged_attention.py
XiaobingSuper: @valarLip @zejunchen-zejun , now all use head_size. 1. vllm attention all use head_size, 2. atom other attention also use head_size, such as GDN.

zejunchen-zejun: Thank you @XiaobingSuper 
@valarLip It makes sense to clear the mess here, previously I mess the head_dim and head_size for MHA and MLA, for now Xiaobing cleans the code!

PR #274, opened 2026-03-06
[QUARK-480] Phase 2: replace dict-style quant_config access with LayerQuantConfig

thpereir:
…cross all models and ops

## Motivation

Following up DeepSeek R1 enablement with per layer quantization support this converts all other models to use the same structured quantization config instead of a dict. 

This PR depends on both #236 and #268 

## Technical Details

1. Clean type boundary — LayerQuantConfig is a frozen dataclass; impossible to accidentally mutate or misspell fields.
2. Single resolution API — `resolve(prefix)` handles exclude-list, per-layer overrides, pattern matching, and global fallback in one call with documented priority. Less error prone than previous approach of accessing the dict. What was `quant_config.get_layer_quant_config(f"{prefix}.fused_qkv_a_proj")` becomes `quant_config.resolve(prefix)`
3. Parser registry — @register_quant_parser("quark") means new quantization formats are self-contained modules. In the current approach adding a new quantization method means editing `QuantizationConfig.__init__` directly.

### Qwen3-30B-A3B-FP8-dynamic
ATOM Server:
```
python -m atom.entrypoints.openai_server --model /root/models/RedHatAI/Qwen3-30B-A3B-FP8-dynamic/ --trust-remote-code -tp 4 --kv_cache_dtype fp8
```
ATOM lm-eval GSM8K:
```
python tests/evals/gsm8k/gsm8k_eval.py --host http://127.0.0.1 --port 8000 --num-questions 1000 --save-results logs
Running GSM8K evaluation: 1000 questions, 5-shot

Results:
Accuracy: 0.858
Invalid responses: 0.000
Total latency: 24.870 s
Questions per second: 40.209
Total output tokens: 119342
Output tokens per second: 4798.671
Results saved to logs
```
HuggingFace GSM8K:
```
Accuracy: 0.86
```

### DeepSeek-R1-0528-MoE-MXFP4-Attn-MTP-PTPC-FP8
ATOM Server:
```
python -m atom.entrypoints.openai_server --model /root/models/amd/DeepSeek-R1-0528-MoE-MXFP4-Attn-MTP-PTPC-FP8/ --trust-remote-code -tp 8 --kv_cache_dtype fp8
```
ATOM lm-eval GSM8K:
```
lm_eval --model local-completions   --model_args "model=/root/models/amd/DeepSeek-R1-0528-MoE-MXFP4-Attn-MTP-PTPC-FP8/,base_url=http://0.0.0.0:8000/v1/completions,tokenized_requests=False,tokenizer_backend=None,num_concurrent=32"   --tasks gsm8k   --num_fewshot 5   --batch_size 1

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9386|±  |0.0066|
|     |       |strict-match    |     5|exact_match|↑  |0.9356|±  |0.0068|
```

## Submission Checklist

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -8,11 +8,14 @@ import logging
 import os
 import re
 from dataclasses import dataclass, field
-from typing import Any, cast, Optional, Union
+from typing import Any, Optional, Union
 
 import torch
 from aiter import QuantType
-from aiter.utility.dtypes import d_dtypes
+from atom.quant_spec import (
+    LayerQuantConfig,
+    get_quant_parser,
+)
 from atom.utils import envs, get_open_port
 from atom.utils.distributed.utils import stateless_init_torch_distributed_process_group
 from torch.distributed import ProcessGroup, ReduceOp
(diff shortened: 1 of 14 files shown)


haoyangli0109:
@thpereir 
Thank you very much.
It's great to see that several of the directions we discussed were implemented quickly. If we modify the code for other quantization formats, we should also select the corresponding models for accuracy testing.

atom/quant_spec.py
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    def parse(self, hf_quant_config: dict) -> ParsedQuantConfig:
+        quant_method = hf_quant_config.get("quant_method", "")
+        config_str = str(hf_quant_config).lower()
+
+        quant_dtype = self._infer_dtype(hf_quant_config, config_str)
+        quant_type = self._infer_qtype(hf_quant_config, config_str)
+        is_dynamic = hf_quant_config.get("is_dynamic", True)
+        exclude = list(hf_quant_config.get("exclude", []) or [])
```
haoyangli0109: The exclude_layer name for compressed_tensor is "ignore", while the exclude_layer name for gpt-oss is "modules_to_not_convert".

haoyangli0109: BTW, some quant_config values in the current gpt-oss.py are hard-coded as None, which I believe is unreasonable. For instance, Quark has gpt-oss models configured with fp8 qkv_proj+moe_mxfp4.

thpereir: Rebased and fixed these two points, thanks for the review Haoyang

thpereir:
Rebased again. Fixed a small issue loading gpt-oss-120b. But I found that even main branch has a big issue running inference for gpt-oss-120b! Output is garbage. 

I am investigating and working on a separate fix to not pollute this PR

atom/config.py
```diff
@@ -251,70 +255,95 @@ def set_splitting_ops_for_v1(self):
...
 
-class QuantizationConfig:
     def __init__(self, config: PretrainedConfig = None):
         if config is None:
             self.torch_dtype = torch.bfloat16
             self.hf_quant_config = None
-            self.global_quant_config = LayerQuantConfig()
-            self.layer_quant_config = {}
-            self.exclude_layers = []
+            self._parsed = ParsedQuantConfig()
```
haoyangli0109: Why do we need `ParsedQuantConfig`? The name `self._Parsed` seems a bit odd.

thpereir: `ParsedQuantConfig` exists because parsers (`QuarkParser`, `GenericParser`) need a structured return type. Without it, each parser would need to return a tuple or dict, which is less type-safe. I renamed the class field to `_parsed_quant_config` the name was indeed odd and too short.

atom/quant_spec.py
```diff
@@ -0,0 +1,299 @@
...
+# Structured parsed config
+# ──────────────────────────────────────────────────────────────────────
+
+
+@dataclass
+class ParsedQuantConfig:
+    """Result of parsing a ``quantization_config`` dict."""
+
+    global_spec: LayerQuantConfig = field(default_factory=LayerQuantConfig)
+    layer_specs: dict[str, LayerQuantConfig] = field(default_factory=dict)
```
haoyangli0109: Actually, we’ve never used `layer_specs`. Because it’s never been initialized anywhere.

atom/config.py
```diff
@@ -251,70 +255,95 @@ def set_splitting_ops_for_v1(self):
...
 
-            global_quant_config_dict = cast(
-                dict[str, Any], self.hf_quant_config.get("global_quant_config", {})
-            )
-            self.global_quant_config = self.parse_quark_config_dict(
-                global_quant_config_dict
-            )
+        # Use the parser registry to build a structured ParsedQuantConfig
+        parser = get_quant_parser(self.quant_method)
+        self._parsed_quant_config = parser.parse(self.hf_quant_config)
```
haoyangli0109: If we can pass exclude_layer directly into QuantizationConfig and let it own that field, why can’t we do the same for global_spec and layer_pattern_specs?

Compared to accessing self._parsed_quant_config.layer_pattern_specs, wouldn’t it be more reasonable and straightforward to use self.layer_pattern_specs directly?

Returning a tuple feels somewhat inelegant, but introducing an additional _parsed_quant_config object that doesn’t seem strictly necessary also feels a bit odd.

atom/quant_spec.py
thpereir: Removed!

atom/config.py
thpereir: I will update creating fields, so access is direct but keeping `parsed_quant_config` as a private implementation detail.

haoyangli0109: Hi, @thpereir 
The typed return contract for QuantConfigParser.parse() is more self-documenting than returning a tuple or dict. When adding new parsers (Quark, GenericParser, etc.), it provides a clear "output contract." And if the parsing output ever needs more fields, adding them to ParsedQuantConfig is cleaner than dealing with an ever-growing tuple.

However, as currently implemented there are several issues:

1. QuantizationConfig is the only consumer. ParsedQuantConfig is never used directly anywhere else. It gets created and immediately proxied through properties — essentially a middleman that adds a layer of indirection for no benefit.
2. Inconsistent design. exclude_layers is copied out and stored directly on QuantizationConfig, while global_spec and layer_pattern_specs are accessed via _parsed_quant_config. If one field can be stored directly, why not the other two? This inconsistency is confusing for readers.
Encapsulation is broken. remap_layer_name directly mutates self._parsed_quant_config.layer_pattern_specs, which means ParsedQuantConfig provides no real encapsulation or immutability guarantee — it is neither frozen nor read-only.
3. Extra cognitive load. When reading self._parsed_quant_config.global_spec, you need to understand what _parsed_quant_config is and where it comes from. If it were simply self.global_spec as a direct field, the comprehension cost would be zero.
4. A dedicated class is not necessary for the parser return value. A Python NamedTuple, or simply having the parser return a (global_spec, layer_pattern_specs, exclude_layers) tuple, would be equally clear for this simple scenario without introducing an extra class.

Suggestion: keep ParsedQuantConfig as the parser return type (it's a fine interface contract), but treat it as a transient object — destructure it in __init__ and discard it:
```
parser = get_quant_parser(self.quant_method)
parsed = parser.parse(self.hf_quant_config)
self.global_spec = parsed.global_spec
self.layer_pattern_specs = parsed.layer_pattern_specs
self.exclude_layers = list(parsed.exclude_layers)
```
This way parsers still return a well-typed ParsedQuantConfig, but QuantizationConfig owns all three fields directly — no delegation, no inconsistency, and remap_layer_name just mutates self.layer_pattern_specs without reaching into a private sub-object.
This is just my personal suggestion; I’d like to hear your thoughts.

thpereir: Agree with your point— _parsed_quant_config as a stored member was inconsistent and added indirection for no real benefit. Updated to destructure the result of parser.parse() directly into self.global_spec, self.layer_pattern_specs, and self.exclude_layers.
ParsedQuantConfig is now purely a typed return contract for parsers; QuantizationConfig owns all three fields directly and remap_layer_name mutates self.layer_pattern_specs without reaching into a sub-object.

atom/model_ops/layernorm.py
```diff
@@ -178,6 +182,7 @@ def __init__(
         fused_allreduce: bool = False,
         fused_quant: bool = False,
         quant_config: Optional[QuantizationConfig] = None,
+        prefix: str = "",
```
valarLip, line 185: never assgined any value? like https://github.com/ROCm/ATOM/pull/274/changes#diff-9f6bc3c977817e17870acedee34e792e89194af741292e3f8c814015d6f9610eL261

thpereir: It's there to be consistent with linear.py / moe.py etc

PR #296, opened 2026-03-10
[draft][plugin] sgl radix attn backend

ZhiweiYan-96:


atom/models/qwen3_moe.py
```diff
@@ -227,6 +230,65 @@ def __init__(
 
         self.kv_cache_dtype = kv_cache_dtype
         self.layer_num = layer_num
+        self.k_scale = torch.tensor([1.0], dtype=torch.float32)
+        self.v_scale = torch.tensor([1.0], dtype=torch.float32)
+
+    def forward_sgl_plugin_mode(
```
zejunchen-zejun, line 236: Hi, @ZhiweiYan-96 @zejunchen-zejun 

Here should be refined because the execution details should be contained inside the radix attention instead of the model forward

atom/utils/envs.py
```diff
@@ -42,6 +42,7 @@
         "ATOM_LLAMA_ENABLE_AITER_TRITON_FUSED_SILU_MUL_QUANT", "1"
     )
     == "1",
+    "ATOM_ROPE_FUSED_QKNORM": lambda: os.getenv("AITER_ROPE_FUSED_QKNORM", "0") == "1",
```
zejunchen-zejun, line 45: Is it possible that we remove this env flag because IIRC, there is duplicated

atom/plugin/config.py
```diff
@@ -223,7 +226,6 @@ def generate_atom_config_for_plugin_mode(config: Any = None):
     """
 
     logger.info("Generate atom config for plugin mode from passed config")
-
```
zejunchen-zejun, line 225: no need change oh

atom/config.py
```diff
@@ -799,23 +621,30 @@ def __post_init__(self):
         # assert os.path.isdir(self.model)
 
         assert 1 <= self.tensor_parallel_size <= 8
-        self.hf_config = get_hf_config(self.model)
+        if is_plugin_mode():
```
zejunchen-zejun, line 624: Here we follow the atom config post init?

PR #339, opened 2026-03-16
[Feat][Plugin] improve allreduce performance and apply vLLM fused ops in MLA attention for plugin mode performance

XiaobingSuper:
This PR is about reusing vLLM's tensor parallel (TP) groups and IPC handles to prevent redundant initialization, patching graph capture for efficient fused collectives, which can improve the allreduce performance for plugin mode. and applying vLLM concat_and_cache_mla_rope_fused and _decode_concat_quant_fp8_op fusion ops. After this PR, for deepseek-r1 fp8 with TP8, we can get **8%** performance improvement.

### DIFF
--- a/atom/plugin/attention_mla.py
+++ b/atom/plugin/attention_mla.py
@@ -694,7 +694,8 @@ class MLAAttentionImplPluginModeMethods:
 
         if not decode_only:
             # write the latent and rope to kv cache
-            if kv_cache.numel() > 0:
+            # make sure ops has concat_and_cache_mla_rope_fused
+            if kv_cache.numel() > 0 and hasattr(ops, "concat_and_cache_mla_rope_fused"):
                 ops.concat_and_cache_mla_rope_fused(
                     positions,
                     q[..., self.qk_nope_head_dim :],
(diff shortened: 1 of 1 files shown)


atom/plugin/attention_mla.py
```diff
@@ -688,20 +688,27 @@ def forward_impl_plugin_mode(
...
+        assert (
+            self.rotary_emb is not None
+        ), "Rotary embedding is required for MLAAttentionImplPluginModeMethods"
 
+        if not decode_only:
             # write the latent and rope to kv cache
             if kv_cache.numel() > 0:
-                aiter.concat_and_cache_mla(
-                    k_c_normed,
+                ops.concat_and_cache_mla_rope_fused(
```
wuhuikx, line 699: We reuse the vllm custom ops here. I just wondering, if vllm remove this code, our model will crash? If so, should we add some ut for vllm upgrading?

atom/plugin/attention_mla.py
```diff
@@ -788,29 +795,11 @@ def forward_impl_plugin_mode(
...
-                        dtype=decode_ql_nope.dtype,
-                    )
-                    decode_q0[..., : ql_nope_shape[2]].copy_(decode_ql_nope)
-                    decode_q0[..., ql_nope_shape[2] :].copy_(decode_q_pe)
-
-                    decode_q, _ = ops.scaled_fp8_quant(
-                        decode_q0.view(decode_q_shape[0], -1),
-                        layer._q_scale,
+                    decode_q = layer._decode_concat_quant_fp8_op(
+                        decode_ql_nope, decode_q_pe, layer._q_scale
```
wuhuikx: where's the decode_concat_quant_fp8_op, in vllm or aiter?

XiaobingSuper: The code from vllm.

atom/plugin/attention_mla.py
XiaobingSuper: Yes, it will crash if not have such an op, add check before uss it.

PR #346, opened 2026-03-17
[GPT-OSS](plugin): remove ones kernel in each iteration

PerryZhang01:
## Motivation

This PR removes `ones` in the forward step, only use the initial layer params.

image

### DIFF
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -60,12 +60,12 @@ class PagedAttentionImpl(nn.Module):
         self.k_scale = self.v_scale = None
         self.device = "cuda:" + str(torch.cuda.current_device())
         self.layer_num = layer_num
-        self.one_scale_float = (
+        self.kv_scale_float = (
             torch.finfo(torch.float8_e4m3fn).max / torch.finfo(aiter.dtypes.fp8).max
             if self.kv_cache_dtype == "fp8"
             else 1.0
         )
-        self.one_scale = torch.tensor(self.one_scale_float, dtype=torch.float32)
+        self.kv_scale = torch.tensor(self.kv_scale_float, dtype=torch.float32)
         self.per_token_quant = True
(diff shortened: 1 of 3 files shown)


atom/model_ops/attention_mha.py
```diff
@@ -60,12 +60,13 @@ def __init__(
         self.k_scale = self.v_scale = None
         self.device = "cuda:" + str(torch.cuda.current_device())
         self.layer_num = layer_num
-        self.kv_scale_float = (
+        self.one_scale_float = (
```
valarLip: no name change here, it's not "1"

atom/model_ops/base_attention.py
```diff
@@ -87,8 +88,15 @@ def cp_mha_gather_cache_kernel(
         k_reg = tl.load(key_cache_ptr_offset + col_offsets)
         v_reg = tl.load(value_cache_ptr_offset + col_offsets)
         if DEQUANT:
-            k_scale = tl.load(k_scale_ptr)
-            v_scale = tl.load(v_scale_ptr)
+            if not PER_TOKEN_QUANT:
+                k_scale = 1.0
```
valarLip: never hard code..

atom/model_ops/attention_mha.py
PerryZhang01: there are three `kv_scale`,  one scale, atom initial scale and vllm initial scale, so how to describe these scales.

atom/model_ops/base_attention.py
PerryZhang01: just recover the last commit here, I will try to be more compatible.

atom/model_ops/attention_mha.py
valarLip: atom use "self.kv_scale" only which is a tensor, "self.one_scale_float" is python float

PR #354, opened 2026-03-19
[plugin][OOT CI] refine OOT CI by picking precheck-in as a standalone workflow

zejunchen-zejun:
image

### DIFF
--- a/.github/workflows/pre-checks.yaml
+++ b/.github/workflows/pre-checks.yaml
@@ -4,17 +4,17 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
     types: [opened, synchronize, reopened, ready_for_review]
+    branches: [main]  # Triggers on PRs targeting `main`
     paths-ignore:
       - '**/*.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-checks.yaml
```diff
@@ -1,33 +1,41 @@
...
+      - 'LICENSE'
+      - '.gitignore'
+  schedule:
+    # Nightly at 00:00 Beijing time (16:00 UTC)
+    - cron: '0 16 * * *'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
gyohuangxin, line 21: Please keep the same on trigger condition like https://github.com/ROCm/aiter/blob/main/.github/workflows/aiter-test.yaml

zejunchen-zejun: It makes sense. I will keep the same trigger condition

PR #384, opened 2026-03-23
enable vllm v0.17.0 on qwen3.5 and add qwen3.5 recipe for vllm plugin mode

ganyi1996ppo:
## Submission Checklist

### DIFF
--- a/recipes/atom_vllm/Qwen3.5.md
+++ b/recipes/atom_vllm/Qwen3.5.md
@@ -17,7 +17,7 @@ The ATOM vLLM plugin backend keeps the standard vLLM CLI, server APIs, and gener
 ```bash
 export ATOM_DISABLE_VLLM_PLUGIN_ATTENTION=1
 
-vllm serve Qwen/Qwen3.5-35B-A3B-Instruct-FP8 \
+vllm serve Qwen/Qwen3.5-35B-A3B-FP8 \
     --host localhost \
     --port 8000 \
     --tensor-parallel-size 2 \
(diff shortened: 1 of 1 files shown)


recipes/atom_vllm/Qwen3.5.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Qwen3.5-35B-A3B (TP=2)
+
+```bash
+export ATOM_DISABLE_VLLM_PLUGIN_ATTENTION=1
+
+vllm serve Qwen/Qwen3.5-35B-A3B-FP8 \
+    --host localhost \
+    --port 8000 \
+    --tensor-parallel-size 2 \
```
wuhuikx, line 23: Is TP2 a must? If not, please make this value more general.

wuhuikx: Or you can let users to set this value

recipes/atom_vllm/Qwen3.5.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Step 4: Accuracy Validation
+
+```bash
+lm_eval --model local-completions \
+        --model_args model=Qwen/Qwen3.5-35B-A3B-FP8,base_url=http://localhost:8000/v1/completions,num_concurrent=16,max_retries=3,tokenized_requests=False \
+        --tasks gsm8k \
+        --num_fewshot 3
+```
```
wuhuikx, line 86: I really need your help to add the accuracy test result for this model here. especially on MI355.

recipes/atom_vllm/Qwen3.5.md
ganyi1996ppo: No, tp2 is only for quick verification for small model like 35B. all the tp size should be supported

recipes/atom_vllm/Qwen3.5.md
ganyi1996ppo: sure, I'll update accuracy result on mi355, on 308, the gsm8k score is around 0.86

PR #386, opened 2026-03-23
add ATOM_DUAL_STREAM_MOE_TOKEN_THRESHOLD env to disable deepseek dual stream path

XiaobingSuper:
This PR is about adding env **ATOM_DUAL_STREAM_MOE_TOKEN_THRESHOLD** to disable deepseek dual stream, because we met a hang issue for kimi-2 when capture more graph size with set **AITER_QUICK_REDUCE_QUANTIZATION=INT4**: **"[1, 2, 4, 8, 16, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96, 104, 112, 120, 128, 136, 144, 152, 160, 168, 176, 184, 192, 200, 208, 216, 224, 232, 240, 248, 256, 272, 288, 304, 320, 336, 352, 368, 384, 400, 416, 432, 448, 464, 480, 496, 512]".** those sizes are the defaule graph sizes used in vllm. This issue can be reproduced in the ATOM side or the OOT path.

### DIFF
--- a/.github/scripts/atom_oot_test.sh
+++ b/.github/scripts/atom_oot_test.sh
@@ -174,6 +174,8 @@ launch_one_model() {
   export VLLM_RPC_TIMEOUT=1800000
   export VLLM_CACHE_ROOT=/tmp/.cache/vllm
   export TORCHINDUCTOR_CACHE_DIR=/tmp/.cache/inductor
+  # FIXME here disable the dual stream in OOT CI for avoid the hang issue
+  export ATOM_DUAL_STREAM_MOE_TOKEN_THRESHOLD=0
   rm -rf /tmp/.cache
 
   rm -f "${VLLM_PID_FILE}" || true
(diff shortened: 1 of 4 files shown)


.github/scripts/atom_oot_test.sh
```diff
@@ -174,6 +174,7 @@ launch_one_model() {
   export VLLM_RPC_TIMEOUT=1800000
   export VLLM_CACHE_ROOT=/tmp/.cache/vllm
   export TORCHINDUCTOR_CACHE_DIR=/tmp/.cache/inductor
+  export ATOM_DISABLE_SHARED_EXPERTS_STREAM=1
```
zejunchen-zejun: Let's add a FIXME comment here

```suggestion
  # FIXME here disable the dual stream in OOT CI for avoid the hang issue
  export ATOM_DISABLE_SHARED_EXPERTS_STREAM=1
```

valarLip: i would like to add ATOM_DUAL_STREAM_TOKEN_THRESHOLD

XiaobingSuper: updated.

valarLip (approved):

PR #397, opened 2026-03-24
[plugin] fix dual stream error for vllm plugin path

XiaobingSuper:
## Motivation

For vLLM plugin path, there doesn't have a context attribute for **get_forward_context()**; the dual stream path will have an Attribute error when running the vLLM plugin path.

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -730,13 +730,16 @@ def maybe_dual_stream_forward(
     atom_config = get_current_atom_config()
     self = atom_config.compilation_config.static_forward_context[layer_name]
     # DUAL_STREAM_TOKEN_THRESHOLD = 1024
-    num_tokens, hidden_dim = hidden_states.shape
-    if (
-        self._use_dual_stream
-        and num_tokens > 0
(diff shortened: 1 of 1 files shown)


atom/models/deepseek_v2.py
```diff
@@ -730,13 +730,16 @@ def maybe_dual_stream_forward(
...
-    num_tokens, hidden_dim = hidden_states.shape
-    if (
-        self._use_dual_stream
-        and num_tokens > 0
-        # and num_tokens <= DUAL_STREAM_TOKEN_THRESHOLD
-        and not get_forward_context().context.is_prefill
-    ):
+    num_tokens, _ = hidden_states.shape
+    fwd_ctx = get_forward_context()
+    is_prefill = getattr(getattr(fwd_ctx, "context", None), "is_prefill", None)
```
zejunchen-zejun, line 735: Hi @XiaobingSuper 
It could be better here OOT can pass meta info to fwd_context in ATOM native, otherwise the common code needs to use `getattr` to distinguish if it is OOT or native. The logic could be much more simple here if here directly use fwd_context after OOT translate its own fwd_context to the one used by ATOM native

PR #401, opened 2026-03-24
[plugin][oot] Add Kimi-K2.5 support

gbyu-amd:
## Motivation

This PR added the support for Kimi-K2.5-MXFP4 with vLLM oot path. Functionality and accuracy pass. The recipe is provided as well.

## Submission Checklist


recipes/atom_vllm/Kimi-K2.5.md
```diff
@@ -0,0 +1,148 @@
...
+    "stream": false
+  }' | python3 -m json.tool
+```
+The expected response:
+```bash
+{
+    "id": "chatcmpl-941a3736cc5cce95",
+    "object": "chat.completion",
+    "created": 1774365813,
+    "model": "/workspace/shared/data/models/Kimi-K2.5-MXFP4",
```
wuhuikx: could you please help make this path more general?

atom/plugin/vllm/model_wrapper.py
```diff
@@ -38,6 +38,7 @@
     "Glm4MoeForCausalLM": "atom.models.glm4_moe:Glm4MoeForCausalLM",
     "Qwen3_5MoeForConditionalGeneration": "atom.models.qwen3_5:Qwen3_5MoeForConditionalGeneration_",
     "Qwen3_5ForConditionalGeneration": "atom.models.qwen3_5:Qwen3_5ForConditionalGeneration_",
+    "KimiK25ForConditionalGeneration": "atom.models.kimi_k25:KimiK25ForConditionalGeneration_",
```
wuhuikx: A little bit curious, why we haven't register KimiK2 here, is it because KimiK2 totally reuse DeepseekV3ForCausalLM?

atom/model_config/kimi_k25.py
```diff
@@ -0,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
```
wuhuikx, line 2: I'm not sure, do we need this comment here. how about the other models?

atom/models/kimi_k25.py
```diff
@@ -90,3 +91,315 @@ def compute_logits(
...
+                    exclude_layer[3:], layer_name
+                ):
+                    return None
+                elif exclude_layer == layer_name:
+                    return None
+            return quant_config
+
+        def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
+            # load weights in plugin mode and discard passed weights generator
+            # here prefix is "model." because Qwen3ForCausalLM is constructed in model
```
wuhuikx: is this comment here right?

atom/config.py
```diff
@@ -610,6 +610,9 @@ def _get_hf_token() -> str | None:
             return token
         return None
 
+    # vLLM oot supports text + vision for Kimi_K2.5
+    if is_vllm():
+        _MULTIMODAL_MODEL_TYPES.pop("kimi_k25")
```
wuhuikx: should we change to this pop("kimi_k25", None) to ensure the security.

recipes/atom_vllm/Kimi-K2.5.md
gbyu-amd: Done.

atom/plugin/vllm/model_wrapper.py
gbyu-amd: Yes. Refer to https://huggingface.co/amd/Kimi-K2-Thinking-MXFP4/blob/main/config.json#L4. Kimi K2 will be resolved as DeepseekV3ForCausalLM, thus no need to register.

atom/model_config/kimi_k25.py
gbyu-amd: Same for other models like Qwen3.5, https://github.com/ROCm/ATOM/blob/631483ada7ad27c9429e3726e595f208b8262d03/atom/model_config/qwen3_5.py#L1-L2 . Mainly for compatibility consideration, these files originally comes from vLLM.

atom/models/kimi_k25.py
gbyu-amd: Corrected!

atom/config.py
gbyu-amd: Thanks for catching this. Updated

recipes/atom_vllm/Kimi-K2.5.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+docker pull rocm/atom-dev:vllm-latest
+```
+
+
+## Step 2: Launch vLLM Server
+
+The ATOM vLLM plugin backend keeps the standard vLLM CLI, server APIs, and general usage flow compatible with upstream vLLM. For general server options and API usage, refer to the [official vLLM documentation](https://docs.vllm.ai/en/latest/).
+
+```bash
```
wuhuikx, line 18: Do we need any env var here like quick allreduce to make the performance better and keep accuracy at the same time? If so we can point it out to let users to have a try, but tell them the accuracy risk.

gbyu-amd: Yes, I think we should put such specific env var in our recipes. Maybe we can add it in another pr for all the recipes under atom_vllm, right now all the recipes just provide the basic launch cmd without providing perf-boost env var.

atom/models/kimi_k25.py
```diff
@@ -90,3 +91,315 @@ def compute_logits(
 
     def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
         return self.language_model.get_expert_mapping()
+
+
+if is_vllm():
```
XiaobingSuper: If this is only for vllm plugin, remove it to the plugin folder?

atom/config.py
```diff
@@ -610,6 +610,9 @@ def _get_hf_token() -> str | None:
             return token
         return None
 
+    # vLLM oot supports text + vision for Kimi_K2.5
+    if is_vllm():
+        _MULTIMODAL_MODEL_TYPES.pop("kimi_k25", None)
```
zejunchen-zejun: Can we have a dict or set to store which model should be excluded for processing config here?
So that the new OOT multi-model will not frequently introduce code change here, the new multi-model is only needed to be inserted into a dict or set

atom/models/kimi_k25.py
wuhuikx: Agree. @gbyu-amd we need to make the core files in ATOM clean and try to add all the vllm related code in plugin folder.

atom/config.py
gbyu-amd: Updated.

atom/models/kimi_k25.py
gbyu-amd: Moved to plugin folder done.

recipes/atom_vllm/Kimi-K2.5.md
wuhuikx: agree

atom/plugin/vllm/models/kimi_k25.py
valarLip, line 1: we have to do this for every new model?

gbyu-amd: Not really i think. Here since atom only supported text part of kimi k25 for now and already has a modelling file. To avoid introducing code like if is_vllm() in atom/models/kimi_k25.py, we simply create a new file under plugin. If keep only one modeling file, we need some careful design, which may introduce changes in some common utilities like Config. We still can merge them into a unified modeling file later if needed.

atom/config.py
```diff
@@ -480,6 +485,9 @@ def _get_hf_token() -> str | None:
             return token
         return None
 
+    if is_vllm():
+        for model_name in _PLUGIN_SUPPORTED_MULTIMODAL_MODELS:
```
valarLip: what is this for?

gbyu-amd: ATOM only loads the text config of Kimi k2.5 while OOT needs to load the full config. Updated the code now to be more safe.

PR #402, opened 2026-03-25
[Fix](docker): fix transformers version for atom-vllm

PerryZhang01:
## Motivation

This PR is for fixing the version of transformers. 
`pip install -r requirements/rocm.txt` contains `-r common.txt` and `common.txt` contains `transformers >= 4.56.0, < 5`, so `install transformers==5.2.0` should be after `pip install -r requirements/rocm.txt`.


zufayu:
Hi @PerryZhang01, this Docker/transformers version fix may be outdated after 3 months of dependency updates. Is it still valid against current `requirements.txt`?

PR #419, opened 2026-03-26
[Fix](rope): support transformers 4.57.6 and 5.2.0 for gpt-oss

PerryZhang01:
## Motivation

this pr is for supporting different versions of transformers. we have verified the accuracy of transfomres==4.57.6 and 5.2.0.

### DIFF
--- a/atom/models/gpt_oss.py
+++ b/atom/models/gpt_oss.py
@@ -31,7 +31,6 @@ from aiter.dist.parallel_state import get_pp_group, get_tensor_model_parallel_wo
 
 # from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from aiter.rotary_embedding import get_rope
-from atom.plugin.prepare import is_vllm
 from atom.config import Config, QuantizationConfig
 from atom.model_ops.base_attention import Attention
 from atom.model_ops.embed_head import ParallelLMHead, VocabParallelEmbedding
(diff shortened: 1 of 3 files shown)


atom/models/gpt_oss.py
```diff
@@ -63,20 +64,35 @@ def cdiv(x, y):
...
         super().__init__()
+        config = atom_config.hf_config
         self.layer_idx = layer_num
         self.head_dim = config.head_dim
         self.num_attention_heads = config.num_attention_heads
         self.num_key_value_heads = config.num_key_value_heads
         self.hidden_size = config.hidden_size
 
-        rope_params = config.rope_parameters
+        if is_vllm():
```
wuhuikx: I'm thinking about, whether we can covert the config in the plugin layer and prepare the correct config out of the model file. I prefer not to add the fwk related switch branch in the core model files.

atom/models/gpt_oss.py
```diff
@@ -63,20 +64,35 @@ def cdiv(x, y):
...
+            model_config = atom_config.plugin_config.vllm_config.model_config
+            text_hf_config = (
```
ganyi1996ppo: This line does not required for non-multimodal models

atom/models/gpt_oss.py
ganyi1996ppo: We can override the rope config at initialization phase if we detect this vllm oot mode

ganyi1996ppo: But maybe we can leave this to future PR

PerryZhang01: yea, I move the change to atom config, so both atom and atom-vllm support transformers 4.x and 5.x

atom/models/gpt_oss.py
PerryZhang01: fixed

docker/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN echo "========== [OOT 4/7] Install vLLM ROCm build dependencies =========="
     "${VENV_PYTHON}" -m pip install --no-deps xgrammar==0.1.29 compressed-tensors==0.13.0 loguru && \
     sed -i -e '/peft/d' -e '/tensorizer/d' -e '/runai/d' -e '/timm/d' requirements/rocm.txt && \
     "${VENV_PYTHON}" -m pip install --no-deps peft tensorizer==2.10.1 runai-model-streamer[s3,gcs]==0.15.3 timm>=1.0.17 && \
-    transformers==5.2.0 && "${VENV_PYTHON}" -m pip install -r requirements/rocm.txt
+    "${VENV_PYTHON}" -m pip install -r requirements/rocm.txt
```
wuhuikx, line 55: how about the other workloads? which transformer version the other workloads need?  @zejunchen-zejun

PerryZhang01: cuurently we support both 5.2.0 and 4.57.6. and ` && transformers==5.2.0` does not execute anything because `&&`.  this is I added in my previous PR, so delete here.

ZhangLirong-amd: We use 5.2.0 here since transformer 5.3.0 has some error in tokenizer and got performance regression, https://github.com/huggingface/transformers/issues/44462 , if we use default here, it will install latest  transformer? Maybe we could use default after them release 5.3.1 or later

PerryZhang01: deleting `&& transformers==5.2.0` has no impact to the result.

ZhangLirong-amd (approved):

PR #427, opened 2026-03-27
[feat](a8w4): support a8w4 gpt oss

PerryZhang01:
## Motivation

This PR is for gpt-oss a8w4 weights. compared to the weights of a16w4, the weights of a8w4 have two additional input static quantization parameters in the moe module, we add the input scale in the weight load and kernel forward functions.

image

Note: please use the same format of a8w4 weights as openai/gpt-oss-120b


zejunchen-zejun:
Hi, @PerryZhang01 
If this PR is deprecated, you can close this PR

PR #437, opened 2026-03-28
Remove Qwen3.5 from OOT CI

wuhuikx:
We found an accuracy regression issue on Qwen3.5. We need time for debugging. It may not be an easy case with a quick fix. To avoid influencing the other PRs, we decide to remove it from OOT CI first. When the functionality/accuracy come back and become stable, we will add it back.


wuhuikx:
> we can ignore this ci fail intead introduce useless PRs

@valarLip OK. Then I will close this PR and treat the Qwen3.5 OOT CI failure as a known issue, which shuold be fixed as the first priority. @ganyi1996ppo

PR #449, opened 2026-03-30
[fix](tp8): fix tp8 accuracy for gpt-oss

PerryZhang01:
## Motivation

this PR is for the accuracy of multiple devices in gpt-oss-120b:
- add contiguous before all reduce.
- currently remove rms + allreduce fused op because aiter ops is not supported.

fused ops will be supported in next pr after aiter fixed pr merged,. it`s hard to debug the `contiguous` errors, hope aiter kernel ops add the contiguous check.

image


atom/models/gpt_oss.py
```diff
@@ -222,7 +222,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
         # Remove padding from output
         if self.moe_hidden_pad > 0:
-            x = x[:, : self.hidden_size]
+            x = x[:, : self.hidden_size].contiguous()
```
wuhuikx: The contiguous will introduce D2D copy. Is this necessary for both tp=1 and tp>1?

PerryZhang01: right! I moved it to allreduce so that copy is not existent in tp1.

valarLip (approved):

PR #497, opened 2026-04-07
[fix] add graph capture patch(like vLLM) for sglang+atom path

zhuyuhua-v:
## Motivation
- add a shared graph capture patch that wraps the framework's `GroupCoordinator.graph_capture` with `aiter`'s `ca_comm.capture()`
- enable the patch for the SGLang + ATOM plugin path during model preparation
- refactor the existing ATOM+vLLM-specific implementation to reuse the same shared helper

## Technical Details
When ATOM runs as a plugin backend, the model uses `aiter` collectives such as `tensor_model_parallel_fused_allreduce_rmsnorm`, but the host framework only enters its own graph capture context. As a result, `aiter` does not enter capture mode, falls back to the unregistered path, and triggers extra `hipMemcpyAsync` calls.

This change makes the framework graph capture also enter `aiter`'s `ca_comm.capture()` so the `aiter` collectives can run in the expected capture mode and avoid the extra memory copy overhead.

## Impact
- fixes the graph capture mismatch for the SGLang + ATOM path
- keeps the vLLM path aligned by sharing the same implementation
- reduces unnecessary `hipMemcpyAsync` overhead in SGLang plugin mode

## Submission Checklist


zhuyuhua-v:
merged in https://github.com/ROCm/ATOM/pull/548

PR #503, opened 2026-04-07
[plugin][dashboard] use nightly date tagged docker

zejunchen-zejun:
when running vllm-atom benchmark, only use docker image with the date instead of latest docker
CC: @wuhuikx


wuhuikx:
Low — Reverse lookup in .github/scripts/resolve_oot_image.py performs a linear, per-tag digest fetch for nightly tags. As the number of tags grows, request volume increases and may hit Docker Hub rate limits, which can force a fallback to floating vllm-latest and weaken reproducibility stability.

resolve_oot_image.py
Lines 162–167

candidates = nightly_candidates(list_tags(repository, token), preferred_version)
matched_tag = None
for candidate in candidates:
    if get_manifest_digest(repository, candidate.tag, token) == reference_digest:
        matched_tag = candidate.tag
Suggestion: Limit the scan window (for example, only the most recent N nightly tags), or use a tag-list API/flow that includes digests when possible to reduce N manifest requests.

zejunchen-zejun:
> Low — Reverse lookup in .github/scripts/resolve_oot_image.py performs a linear, per-tag digest fetch for nightly tags. As the number of tags grows, request volume increases and may hit Docker Hub rate limits, which can force a fallback to floating vllm-latest and weaken reproducibility stability.
> 
> resolve_oot_image.py Lines 162–167
> 
> candidates = nightly_candidates(list_tags(repository, token), preferred_version) matched_tag = None for candidate in candidates: if get_manifest_digest(repository, candidate.tag, token) == reference_digest: matched_tag = candidate.tag Suggestion: Limit the scan window (for example, only the most recent N nightly tags), or use a tag-list API/flow that includes digests when possible to reduce N manifest requests.

The docker image is searched from the latest to the oldest according to the digest, so the search overhead is very small.
If the docker cannot be found, we will fallback to the docker with the digest tag instead of vLLM:latest

wuhuikx:
Can we remove the OOT or oot naming?

zejunchen-zejun:
> Can we remove the OOT or oot naming?

Sure. I will remove the OOT terminology in this PR's code change. For other remove work, let's make in PR https://github.com/ROCm/ATOM/pull/541

wuhuikx (dismissed):

wuhuikx (dismissed):

valarLip (dismissed):

valarLip (approved):

PR #525, opened 2026-04-09
[Feat] remove flatten in atom sglang mla like atom vllm mla

ZLkanyo009:
## Motivation

In the decode process of DeepSeek-R1, compared to atom-vLLM, atom-sglang has an extra flatten operator. This PR removes that flatten operator based on the handling in atom-vLLM, thereby accelerating the decode phase.

## Performance
before:
image

after:
image


atom/plugin/sglang/attention_backend/sgl_attention_mla.py
```diff
@@ -237,6 +237,43 @@ def mla_absorbed_bmm(
...
+            YQ=out_3d,
+            transpose_bm=True,
+            transpose_bm_in=False,
+            dtype=torch.bfloat16,
+        )
+        return out
+
+    return mla_absorbed_bmm(
+        attn, inp, weight, weight_scale, weight_scale_k, out_dim
+    ).flatten(1, 2)
```
zhuyuhua-v, line 274: Do we still need to keep the flatten path? Instead of introducing a new function mla_v_up_proj, could we just modify mla_absorbed_bmm directly by passing transpose_bm=True to handle this case — keeping the change minimal and the logic in one place?

ZLkanyo009: The function mla_absorbed_bmm will be called twice: once at q_nope and once at mla_v_up_proj. If you modify it directly inside mla_absorbed_bmm, it will affect the call at q_nope.

zhuyuhua-v: Ok, make sense. 
And we have a longer-term plan to align the SGLang attention backend with atom, which would eventually replace mla_absorbed_bmm / mla_v_up_proj with atom's _v_up_proj_and_o_proj. We can clean this up as part of the larger alignment work.

wuhuikx (approved):

valarLip (approved):

Issue #359, opened 2026-03-19
[RFC] Introduce ATOM as external model package of SGLang


zhuyuhua-v:
## Motivation

ATOM currently integrates with SGLang by maintaining a forked version of SGLang that adds `--model-impl atom` support. This approach requires invasive modifications to SGLang's codebase (6+ places), and imposes a continuous maintenance burden to keep the fork synchronized with upstream SGLang updates.

SGLang [PR #13429](https://github.com/sgl-project/sglang/pull/13429) introduced the `SGLANG_EXTERNAL_MODEL_PACKAGE` mechanism, which allows external packages to register custom model implementations **without modifying upstream SGLang**. This provides a clean, officially supported extension point that perfectly fits ATOM's use case.

By adopting this approach, we eliminate the need for a SGLang fork entirely. ATOM users can run optimized models on AMD GPUs using **unmodified upstream SGLang** with a single environment variable:

```bash
export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
python3 -m sglang.launch_server --model-path <model_path> ...
```

### Comparison with the fork-based approach

| Dimension | `--model-impl atom` (fork) | `SGLANG_EXTERNAL_MODEL_PACKAGE` (now) |
|-----------|---------------------------|---------------------------------------|
| SGLang source | Modified fork, requires continuous sync | **Upstream SGLang, zero modifications** |
| Model registration | Built-in `ATOMForCausalLM` + `ModelImpl.ATOM` | External package with `EntryClass` |
| Supported models | Single wrapper for all architectures | Single base class, dynamic per-arch class generation |
| Maintenance cost | Only maintain ATOM-side code | Only maintain ATOM-side code |
| Adding new models | Append atom plugin registry | Append architecture name to `_MODEL_NAMES` list |

### SGLang's External Model Package Mechanism

SGLang's `SGLANG_EXTERNAL_MODEL_PACKAGE` ([PR #13429](https://github.com/sgl-project/sglang/pull/13429)) works through an automatic module discovery mechanism in `sglang/srt/models/registry.py`:

```
ModelRegistry = _ModelRegistry()
ModelRegistry.register("sglang.srt.models")                     # 1. Register built-in models

if external_pkg := envs.SGLANG_EXTERNAL_MODEL_PACKAGE.get():
    ModelRegistry.register(external_pkg, overwrite=True)         # 2. Register external models (overwrite)
```

[...]

PR #355, opened 2026-03-19
[feat] Make ATOM work with SGLang out-of-tree

zhuyuhua-v:
## Motivation

Detail please refer https://github.com/ROCm/ATOM/issues/359

ATOM currently integrates with SGLang by maintaining a forked version of SGLang that adds `--model-impl atom` support. This approach requires invasive modifications to SGLang's codebase https://github.com/ROCm/ATOM/issues/295, and imposes a continuous maintenance burden to keep the fork synchronized with upstream SGLang updates.

SGLang [PR #13429](https://github.com/sgl-project/sglang/pull/13429) introduced the `SGLANG_EXTERNAL_MODEL_PACKAGE` mechanism, which allows external packages to register custom model implementations **without modifying upstream SGLang**. This provides a clean, officially supported extension point that perfectly fits ATOM's use case.

By adopting this approach, we eliminate the need for a SGLang fork entirely. ATOM users can run optimized models on AMD GPUs using **unmodified upstream SGLang** with a single environment variable:

```bash
export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
python3 -m sglang.launch_server --model-path <model_path> ...
```

## accuracy
```bash
# server
export AITER_QUICK_REDUCE_QUANTIZATION=INT4
export SGLANG_AITER_FP8_PREFILL_ATTN=0
# model_path=deepseek-ai/DeepSeek-R1-0528/ # for ds
# model_path=Qwen/Qwen3-235B-A22B-Instruct-2507-FP8/ # for qwen3-235b

export PYTHONPATH=your_sglang_python_path:your_atom_path

export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models

python3 -m sglang.launch_server \
    --model-path $model_path \
    --host localhost \
    --port 8013 \
    --trust-remote-code \
    --tensor-parallel-size 8 \
    --expert-parallel-size 8 \ # for qwen3-235b only
    --kv-cache-dtype fp8_e4m3 \
    --mem-fraction-static 0.8 \
    --page-size 1 \ # 1024 for qwen3-235b
    2>&1 | tee log.serve.log

# accuracy gsm8k
addr=localhost
port=8013
url=http://${addr}:${port}/v1/completions
# model=Qwen/Qwen3-235B-A22B-Instruct-2507-FP8/
# model=deepseek-ai/DeepSeek-R1-0528/
task=gsm8k
lm_eval --model local-completions \
[...]

PR #532, opened 2026-04-09
[feat] Add support for Qwen3.5 and Qwen3-Next to ATOM-plugined SGLang

wanzhenchn:
## Motivation

Background: [ROCm/ATOM#355](https://github.com/ROCm/ATOM/pull/355) and [ROCm/ATOM#359](https://github.com/ROCm/ATOM/issues/359).

[PR #355](https://github.com/ROCm/ATOM/pull/355)  integrated ATOM with upstream SGLang through the SGLANG_EXTERNAL_MODEL_PACKAGE out-of-tree mechanism, replacing a fork-based workflow and establishing atom.plugin.sglang.models as the external entry package for ATOM-backed architectures.

Building on that foundation, this PR extends the SGLang plugin path so that two major ATOM model families—**Qwen3-next** (**Qwen3NextForCausalLM**) and **Qwen3.5** (**Qwen3_5ForConditionalGeneration/Qwen3_5MoeForConditionalGeneration**)—can run as first-class external models inside SGLang. The goal is parity with prior ATOM-in-SGLang accuracy while improving end-to-end inference performance on the supported paths (e.g. ATOM’s fused kernels, quantization, and MLA / MoE handling tuned for ROCm), without requiring a patched SGLang tree—users continue to point `SGLANG_EXTERNAL_MODEL_PACKAGE` at `atom.plugin.sglang.models` and launch with standard upstream `sglang.launch_server`.

## Technical Details

- Qwen3-next
    - `Qwen3NextForCausalLM` is registered under `atom.plugin.sglang.models` and subclasses `_AtomCausalLMBaseForSglang`, reusing the same SGLang-facing contract as other OOT entry points: the wrapper calls prepare_model(..., engine="sglang") to build the ATOM weight stack, runs the language model forward with pipeline-parallel state mapped from pp_proxy_tensors, applies LogitsProcessor on the last PP rank, and loads weights via load_model_in_plugin_mode. 
    - The linear-attention (GDN) path is wrapped by `Qwen3NextSglangModel` plus sglang_gdn_bridge so GDN layers see the SGLang forward_batch context they expect. At prepare time, apply_qwen3_next_sglang_model_patch swaps `atom.models.qwen3_next.Qwen3NextModel` to that bridged implementation; the shared prepare hook defaults `ATOM_SGLANG_USE_NATIVE_AITER_ATTN_BACKEND` for Qwen3NextForCausalLM before register_ops_to_sglang.

- Qwen3.5
    - Qwen3.5 text / MoE / multimodal stacks reuse SGLang’s in-tree Qwen3_5* container classes while **the language tower is still constructed through ATOM** prepare_model. apply_prepare_model_adaptations applies Qwen3.5-only config fixes (e.g. MoE text fields, quant remaps for ROCm) and _apply_qwen35_sglang_model_patch rebinds atom.models.qwen3_5.Qwen3_5Model to the SGLang-specific implementation. 
    - Weights are not applied through SGLang’s default load_weights iterator; instead load_model_in_plugin_mode drives ATOM’s loader, with _QWEN35_OOT_PACKED_MODULES_MAPPING (and related mapping) kept aligned to SGLang’s packed/fused parameter layout so FP8 and fused checkpoints behave like the in-tree model. 
    - The same prepare hook defaults native Aiter attention backend selection for Qwen3_5* (and Qwen3-next) unless the user has already set the env var.

## How to Run

```bash
export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
python3 -m sglang.launch_server --model-path <model_path> ...
````

The following models have been supported: 
- [Qwen3.5-397B-A17B](https://huggingface.co/Qwen/Qwen3.5-397B-A17B)
- [Qwen3.5-397B-A17B-FP8](https://huggingface.co/Qwen/Qwen3.5-397B-A17B-FP8)
- [Qwen3.5-35B-A3B](https://huggingface.co/Qwen/Qwen3.5-35B-A3B)
- [Qwen3.5-35B-A3B-FP8](https://huggingface.co/Qwen/Qwen3.5-35B-A3B-FP8)
- [Qwen3.5-27B](https://huggingface.co/Qwen/Qwen3.5-27B)
- [Qwen3.5-27B-FP8](https://huggingface.co/Qwen/Qwen3.5-27B-FP8)
- [Qwen3-Next-80B-A3B-Instruct](https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct)

## Accuracy

```bash
# launch server
#! /usr/bin/bash

set -euxo pipefail

export PYTHONPATH=/opt/sglang/python
export SGLANG_DISABLE_CUDNN_CHECK=1

if [ $# -lt 4 ]; then
  echo "Usage: $0 <port> <model_path> <device_id> <enable_atom>"
  exit 1
fi

port=$1
model_path=$2
device_id=$3
enable_atom=$4 # true or false

if [ "${enable_atom}" == true ]; then
  export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
fi

tp=$(echo "$device_id" |grep -o "[0-9]" |grep -c "")

HIP_VISIBLE_DEVICES=${device_id} python3 -m sglang.launch_server \
  --model-path ${model_path} --port ${port} \
  --tensor-parallel-size ${tp} --mem-fraction-static 0.9 \
  --reasoning-parser qwen3 --disable-radix-cache
```

```bash
model_path=$1
port=$2
task=$3

models_url="http://localhost:${port}/v1/models"
echo "Waiting for OpenAI-compatible server at ${models_url} ..." >&2
until curl -sf --connect-timeout 2 --max-time 10 "${models_url}" >/dev/null; do
  sleep 2
done
echo "Server is up; starting lm_eval." >&2

        --model_args model=${model_path},base_url=http://localhost:${port}/v1/completions,num_concurrent=64,max_retries=3,max_gen_toks=2048,tokenized_requests=False \
        --tasks ${task} \
        --batch_size auto \
        --trust_remote_code
```

- Qwen3.5-397B-A17B-FP8
    - ATOM + SGLang
    image

    - SGLang only
    image

- Qwen3-Next-80B-A3B-Instruct
    - ATOM + SGLang
    image

    - SGLang only

    image

## Inference Perf.

- Qwen3.5-397B-A17B-FP8 on MI355X

image

- Qwen3.5-397B-A17B-FP8 on MI308X

image


atom/models/qwen3_next.py
```diff
@@ -420,7 +421,7 @@ def forward(
 
         q, k = self.rotary_emb(positions, q, k)
 
-        attn_output = self.attn(q, k, v)
+        attn_output = self.attn(q, k, v, positions=positions, **model_kwargs)
```
wuhuikx: we should remove the **model_kwargs and refer to DS how to implement the attn part.

atom/models/qwen3_5.py
```diff
@@ -318,7 +317,11 @@ def forward(
         **kwargs: object,
     ):
         hidden_states = self.model(
-            input_ids, positions, intermediate_tensors, inputs_embeds
+            input_ids,
+            positions,
+            intermediate_tensors,
+            inputs_embeds,
+            **kwargs,
```
wuhuikx: Let's remove the **kwargs here and use wrapper in plugin/sglang to handle.

atom/plugin/sglang/models/qwen3_5.py
```diff
@@ -0,0 +1,498 @@
...
+                )
+            hidden_states, _ = self.norm(hidden_states, residual)
+            return hidden_states
+
+
+def _apply_qwen35_sglang_model_patch() -> None:
+    """Swap ``atom.models.qwen3_5.Qwen3_5Model`` for :class:`Qwen3_5SglangModel` (idempotent)."""
+    if _qwen35_shim.Qwen3_5Model is Qwen3_5SglangModel:
+        return
+    _qwen35_shim.Qwen3_5Model = Qwen3_5SglangModel
```
zhuyuhua-v: Does this mean permanently replacing a symbol in the atom.models.qwen3_5 module at runtime? Any other code that later imports atom.models.qwen3_5.Qwen3_5Model will get the SGLang version, even if it's running in a vLLM UT. Can we call  `Qwen3_5SglangModel.forward(self, ...)` in `_AtomQwen35FlatLanguageStack`?

atom/model_ops/attention_gdn.py
```diff
@@ -20,6 +20,25 @@
...
+    need_stride = (d * w, 1, d)
+    t = _sglang_conv_work_cache.get(key)
+    if t is None or t.shape != raw.shape or t.stride() != need_stride:
+        t = torch.empty_strided(
+            (n, d, w),
+            need_stride,
+            dtype=raw.dtype,
+            device=raw.device,
+        )
+        _sglang_conv_work_cache[key] = t
```
zhuyuhua-v: Seems this dict will save gpu tensors with shape, stride, device, sglang server handles requests of varying batch sizes — each unique shape creates a new entry. Over hours of running, will this dict accumulate hundreds of entries, each holding a GPU tensor, slowly leaking GPU VRAM?

atom/model_ops/attention_gdn.py
```diff
@@ -157,14 +176,32 @@ def forward(
...
+        kv_entry = gdn_cache[f"layer_{self.layer_num}"]
+        raw_k = kv_entry.k_cache
+        ssm_state = kv_entry.v_cache
+        conv_layout = getattr(kv_entry, "mamba_conv_layout", "atom")
+        conv_sglang_copy_dst = None
+
+        if conv_layout == "sglang_rowmajor":
+            # SGLang: [slot, conv_dim, kernel-1] row-major; kernel needs stride(-2)==1 on [slot,D,W].
+            conv_state = _sglang_conv_work_tensor(raw_k)
+            conv_state.copy_(raw_k)
```
zhuyuhua-v: This one may leads to perf regression due to extra `copy`?

wuhuikx:
I turn it to draft and will let it go through CI after we finishing the code review.

@ganyi1996ppo  @Yuechguo  please help review this PR.

atom/models/qwen3_next.py
wanzhenchn: yeah, the model wrapper for qwen3.5/qwen3-next has been created in plugin/sglang/models/

atom/models/qwen3_5.py
wanzhenchn: yeah, the model wrapper for qwen3.5/qwen3-next has been created in plugin/sglang/models/

atom/plugin/sglang/models/qwen3_5.py
wanzhenchn: No. We are not permanently replacing atom.models.qwen3_5.Qwen3_5Model at runtime.

In the SGLang plugin we import the core symbol as _Qwen3_5ModelCore and define a separate plugin-local subclass Qwen3_5SglangModel; we never assign it back into atom.models.qwen3_5. So any later code that imports atom.models.qwen3_5.Qwen3_5Model still gets the original core class, including vLLM unit tests. We also added a regression test to lock this in: apply_prepare_model_adaptations() leaves the core symbol unchanged.

And yes, for the VLM language stack this is exactly the approach we use: _AtomQwen35FlatLanguageStack.forward() directly calls Qwen3_5SglangModel.forward(self, ...). That gives us the SGLang-specific forward behavior while keeping the exported symbol in atom.models.qwen3_5 untouched.

atom/model_ops/attention_gdn.py
wanzhenchn: You are right, to avoid that steady-state overhead, i keep one persistent shadow tensor in the ATOM-preferred layout and reuse it across requests. Once created, this shadow cache is exposed through the bridge as `KVCacheTensor.k_cache`, so the attention backend can usually consume it directly without any per-request row-major to ATOM-layout conversion.

atom/model_ops/attention_gdn.py
wanzhenchn: The two extra copies per request (copy in, then copy back out after the kernel updates the state) have been optimized by persistent ATOM-layout shadow cache.

atom/plugin/sglang/models/qwen3_next.py
```diff
@@ -0,0 +1,296 @@
+"""ATOM Qwen3-next model wrappers for SGLang external model loading."""
```
ganyi1996ppo: Why do we need another set of model defination in sglang-atom? I thought the basic idea of atom plugin is about share model defination, rather than maintain separate one. Separate model defination might introduce lots of maintaince effort in the future and I don't recommand that if it's not necessary

atom/plugin/sglang/models/qwen3_5.py
```diff
@@ -0,0 +1,693 @@
+"""ATOM Qwen3.5 VLM for SGLang external model inference.
```
ganyi1996ppo: ditto

atom/model_ops/linear.py
```diff
@@ -372,13 +372,18 @@ def process_weights_after_loading(self):
             )
             self.weight.data = w_q
             self.weight_scale = atom_parameter(w_s)
-            shuffle_weights(self.weight)
+            # Only GEMM weights use aiter block shuffle; e.g. GDN conv1d uses 3D (N,1,K).
```
ganyi1996ppo: I don't think we need this part, conv1d doesn't go through quantization layer. vllm-atom already properly handled this part. So I think we should remove this part

atom/model_ops/linear.py
```diff
@@ -372,13 +372,18 @@ def process_weights_after_loading(self):
...
+            if self.weight.dim() == 2:
+                shuffle_weights(self.weight)
             # self.weight_scale.data = fp4_utils.e8m0_shuffle(self.weight_scale.data)
         else:
             if (
-                self.quant_type == QuantType.per_Token
-                and self.params_dtype == dtypes.fp8
-            ) or (self.quant_type in [QuantType.per_1x32, QuantType.per_1x128]):
+                (
+                    self.quant_type == QuantType.per_Token
```
ganyi1996ppo: ditto, there are too many code path in linear. I don't think it's a good idea

atom/plugin/sglang/models/qwen3_next.py
wanzhenchn: The following points have been considered:
1. ATOM only supports **text-only requests** for the Qwen3.5 series, but the Qwen3.5 models are unified vision-language models by design. The customer likely expects vision-language capabilities (e.g., image understanding) from Qwen3.5, which is not meet the customer’s requirements.
2. Since the principle of the Qwen3.5 plugin code implementation is to minimize changes to the atom core (e.g., the models/ directory), when defining and implementing the model in plugin/sglang, we try to reuse existing classes under atom/models such as `Qwen3_5ForCausalLMBase`, `Qwen3_5Model`, etc. However, these classes have forward functions that **do not pass \*\*keywords**. Therefore, a wrapper needs to be written to implement this functionality.

atom/model_ops/linear.py
wanzhenchn: This check is for fp8-ptpc models quantized by open source quantization tools like [llm-compressor](https://github.com/vllm-project/llm-compressor).

atom/model_ops/linear.py
wanzhenchn: ditto

atom/model_ops/attention_gdn.py
```diff
@@ -154,17 +174,11 @@ def forward(
         core_attn_out: torch.Tensor,
         layer_name: str,
     ):
-        from atom.model_ops.attentions.gdn_attn import GDNAttentionMetadata
-
         fwd_ctx: ForwardContext = get_forward_context()
-        gdn_metadata: GDNAttentionMetadata = fwd_ctx.attn_metadata.gdn_metadata
-
+        gdn_metadata, conv_state, ssm_state = self._resolve_runtime_state(fwd_ctx)
```
ganyi1996ppo: Why abstract this part? former one looks more straight forward. and I don't see any reuse of it.

atom/plugin/sglang/attention_backend/attention_gdn.py
```diff
@@ -0,0 +1,43 @@
...
+
+from atom.model_ops.attentions.gdn_attn import GDNAttentionMetadata
+from atom.model_ops.attention_gdn import GatedDeltaNet as AtomGatedDeltaNet
+from atom.utils.forward_context import ForwardContext
+
+
+class GatedDeltaNet(AtomGatedDeltaNet):
+    """SGLang adapter over the shared ATOM GDN implementation."""
+
+    def _resolve_runtime_state(
```
ganyi1996ppo: Redundant defination seems?

atom/model_ops/attention_gdn.py
wanzhenchn: - SGLang exposes the mamba/GDN conv cache as a row-major tensor shaped
      `[slot, D, W]`.
  - ATOM's `causal_conv1d_*` kernels consume the same logical shape, but they
    require the feature dimension `D` to be contiguous in memory
    (`stride(-2) == 1`).

So we should have tensor layout transformation for conv state.

It has been called [atom/plugin/sglang/attention_backend/attention_gdn.py](https://github.com/ROCm/ATOM/pull/532/changes#diff-95c61fcff6ddce951b0148d1cd28e94df7b1342fbe06b1fd45909d53d3aa558aR12)

atom/plugin/sglang/models/qwen3_next.py
```diff
@@ -0,0 +1,296 @@
...
+    make_empty_intermediate_tensors_factory,
+    make_layers,
+    maybe_prefix,
+)
+from atom.plugin.sglang.models.base_model_wrapper import _AtomCausalLMBaseForSglang
+from atom.plugin.sglang.utils.gdn_forward_helper import sglang_gdn_bridge
+from atom.utils.decorators import TorchCompileWrapperWithCustomDispatcher
+
+
+class Qwen3NextSglangAttention(_CoreQwen3NextAttention):
```
ganyi1996ppo: I think you can trust the modeling already have in atom for the execution part. if there anything varying from want already have, you can just add a if branch, or abstract into a function.

ganyi1996ppo: we shouldn't add this `sglang/models/` folder, you can patch the `qwen3_5.py` or `qwen3_next.py` as you need

atom/plugin/sglang/attention_backend/attention_gdn.py
wanzhenchn: No, it is used to achieve the conv state tensor layout transformation after the `GatedDeltaNet` is initialized.

atom/plugin/sglang/models/qwen3_next.py
ganyi1996ppo: ATOM native is text-only, but vllm-atom support multimodal input. you can refer to vllm-atom's implementation  and put sglang specific part into a code block in `qwen3_5.py`

atom/model_ops/linear.py
ganyi1996ppo: sounds fair

atom/plugin/attention.py
```diff
@@ -611,8 +611,15 @@ def decorator(cls):
                     if callable(method):
                         class_dict[method_name] = method
         elif is_sglang_mode:
-            raise NotImplementedError(
-                "AttentionMetadataBuilder for sglang is not implemented yet"
+            # SGLang uses ATOM's ScheduledBatch + CommonAttentionBuilder pipeline (same as
```
wuhuikx, line 632: I just wonder, for atom/vllm/sglang, will we have standalone attention backend, or we will use the same attention backend for all the frameworks?  Is that possible to use the same attention backend for all frameworks? @ZhiweiYan-96  @ganyi1996ppo

wuhuikx: How about deepseek, all atom/vllm/sglang use the same attention backend or we will refactor to let them use the same attention backend?

atom/plugin/sglang/attention_backend/attention_gdn.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+import torch
+
+from atom.model_ops.attentions.gdn_attn import GDNAttentionMetadata
+from atom.model_ops.attention_gdn import GatedDeltaNet as AtomGatedDeltaNet
+from atom.utils.forward_context import ForwardContext
+
+
+class GatedDeltaNet(AtomGatedDeltaNet):
```
wuhuikx: If we have attention_gdn.py in plugin/sglang/attention_backend, is that mean sglang will use its own attention backend instead of using the one in ATOM or vLLM?  Then why we need to revise the code in plugin/attention.py?  @wanzhenchn  @ZhiweiYan-96

atom/utils/envs.py
```diff
@@ -83,6 +83,13 @@
         "ATOM_DISABLE_VLLM_PLUGIN_ATTENTION", "0"
     ).lower()
     == "1",
+    # SGLang plugin (PR #355 / #532): when True, do not override
+    # ATTENTION_BACKENDS["aiter"] with ATOMAttnBackendForSgl and keep
+    # upstream SGLang's native AiterAttnBackend instead.
+    "ATOM_SGLANG_USE_NATIVE_AITER_ATTN_BACKEND": lambda: os.getenv(
+        "ATOM_SGLANG_USE_NATIVE_AITER_ATTN_BACKEND", "0"
```
wuhuikx: Why we need this flag? is the option for use to use the default upstream sglang attention backend or the ATOM attention backend?  We already have a flag for this, right?  @ganyi1996ppo  @ZhiweiYan-96

atom/plugin/attention.py
ganyi1996ppo: It's possible if we can control the metdata preparation for each backend, then all we need is a extraction layer to get us unified kv cache and metadata to operate

atom/utils/envs.py
ganyi1996ppo: for vllm backend, we won't need fallback when newest optimization merged

atom/plugin/sglang/utils/gdn_context.py
```diff
@@ -0,0 +1,266 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Build SGLang GDN forward context for ATOM plugin execution."""
```
wuhuikx: Is that possible to combine this file with plugin/sglang/attention_backend/gdn_attn.py?

atom/plugin/sglang/utils/forward_context.py
```diff
@@ -0,0 +1,80 @@
...
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from typing import ClassVar, Optional, Union
+
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch, PPProxyTensors
+
+
+@dataclass(frozen=True)
+class SGLangForwardBatchMetadata:
```
wuhuikx: Can we remove this file and combine it with the files in the plugin/sglang/attention_backend?

atom/models/qwen3_5.py
```diff
@@ -967,3 +967,343 @@ class Qwen3_5MoeForConditionalGeneration(Qwen3_5ForConditionalGeneration, IsHybr
...
+    from sglang.srt.model_executor.forward_batch_info import (
+        ForwardBatch,
+        PPProxyTensors,
+    )
+    from sglang.srt.models.qwen3_5 import (
+        Qwen3_5ForConditionalGeneration as _SglQwen35VL,
+        Qwen3_5MoeForConditionalGeneration as _SglQwen35MoeVL,
+    )
+
+    _PACKED_MODULES_MAPPING = {
```
wuhuikx: should the mapping logic in this file? where's the mapping logic for deepseek in sglang-atom? Where we put the code of mapping sglang to atom for deepseek? @ZhiweiYan-96

atom/models/qwen3_5.py
```diff
@@ -967,3 +967,343 @@ class Qwen3_5MoeForConditionalGeneration(Qwen3_5ForConditionalGeneration, IsHybr
 
         def get_expert_mapping(self) -> list[tuple[str, str, int, str]]:
             return self.model.get_expert_mapping()
+
+
+if is_sglang():
```
zhuyuhua-v: can we move these logics to plugin/sglang/ ?

atom/models/qwen3_5.py
zhuyuhua-v: mapping model weight from sglang to atom for deepseek: https://github.com/ROCm/ATOM/blob/main/atom/plugin/sglang/models/base_model_wrapper.py#L143

atom/plugin/sglang/attention_backend/attention_gdn.py
wanzhenchn: > If we have attention_gdn.py in plugin/sglang/attention_backend, is that mean sglang will use its own attention backend instead of using the one in ATOM or vLLM? Then why we need to revise the code in plugin/attention.py? @wanzhenchn @ZhiweiYan-96

Not exactly. `plugin/sglang/attention_backend/attention_gdn.py` does not mean SGLang is using a separate GDN implementation instead of ATOM/vLLM.

That file is only a thin SGLang-side adapter over ATOM's shared GDN implementation. The actual GDN logic still lives in `atom.model_ops.attention_gdn`. Its job is to bridge SGLang runtime state into the format expected by ATOM GDN: it reads SGLang-side `gdn_metadata`, retrieves the per-layer KV/conv state from `ForwardContext`, and normalizes the conv-state layout to the ATOM GDN expectation (either reusing the ATOM-layout shadow cache or transposing from the SGLang layout when needed). So this is an integration/adapter layer, not a separate kernel or backend implementation.

`plugin/attention.py` is at a different layer. It contains the shared plugin-mode attention infrastructure: metadata definitions/builders, backend contracts, and other common attention-related behavior used by the plugin path. So changes there are still needed even if SGLang has its own `attention_backend/attention_gdn.py`.

So these two pieces are solving different problems:
- `plugin/sglang/attention_backend/attention_gdn.py`: SGLang -> ATOM GDN runtime adaptation
- `plugin/attention.py`: shared plugin attention infrastructure / metadata contract

That is why revising `plugin/attention.py` is still necessary. @wuhuikx

atom/utils/envs.py
wanzhenchn: > Why we need this flag? is the option for use to use the default upstream sglang attention backend or the ATOM attention backend? We already have a flag for this, right? @ganyi1996ppo @ZhiweiYan-96

For `Qwen3_5*` and `Qwen3NextForCausalLM`, we intentionally keep upstream SGLang's native `AiterAttnBackend` by default. This is not only a layering/design choice, but also a current implementation constraint: the custom SGLang attention backend in `sgl_attn_backend.py` does not yet fully support these models' attention configuration. In particular, `Qwen3_5*` and `Qwen3NextForCausalLM` use `head_dim=256`, while the relevant custom backend path currently only supports the `head_dim=128` case. Because of that, we cannot switch these models to `ATOMAttnBackendForSgl` yet, so we deliberately leave them on upstream `AiterAttnBackend`. @wuhuikx

atom/plugin/sglang/utils/gdn_context.py
wanzhenchn: > Is that possible to combine this file with plugin/sglang/attention_backend/gdn_attn.py?

`plugin/sglang/utils/gdn_context.py` is responsible for preparing and binding the SGLang runtime state into ATOM's forward-context form. It builds `GDNAttentionMetadata`, maps SGLang-side KV/conv cache state into `KVCacheTensor`, handles conv-state layout adaptation / shadow-cache reuse, and then binds everything through `SGLangGDNForwardContext.bind()`.

By contrast, `plugin/sglang/attention_backend/gdn_attn.py` is only a very thin backend descriptor/registration layer: it defines the backend name and returns the implementation class (`attention_gdn.GatedDeltaNet`).

So they live at different layers:
- `gdn_context.py`: runtime state preparation and binding
- `gdn_attn.py`: backend identity / implementation lookup

Merging them would mix context-preparation logic with backend-registration logic, even though the current GDN context bridge is used by the model-wrapper path independently of the backend descriptor itself.

This distinction is especially important because `Qwen3_5*` / `Qwen3NextForCausalLM` currently still default to upstream SGLang's native `AiterAttnBackend`, not `ATOMAttnBackendForSgl`. Even in that setup, `gdn_context.py` is still needed, because it solves the GDN runtime-state bridging problem rather than the backend-registration problem.

So my view is: combining them is possible mechanically, but keeping them separate is cleaner architecturally.

atom/plugin/sglang/utils/forward_context.py
wanzhenchn: > Can we remove this file and combine it with the files in the plugin/sglang/attention_backend?

`SGLangForwardBatchMetadata` is not just an attention-backend detail. It acts as a small shared forward-metadata contract across several SGLang plugin layers:
- `plugin/sglang/models/base_model_wrapper.py` builds/binds it
- `atom/models/qwen3_5.py` uses it in the Qwen3.5 SGLang adapter path
- `plugin/sglang/utils/gdn_context.py` consumes it to build GDN-specific context
- `plugin/sglang/attention_backend/*` reads from it during attention execution

So semantically it sits at the boundary shared by wrapper/model/attention-backend layers, rather than inside the attention-backend layer itself.

If we move it into `plugin/sglang/attention_backend`, the dependency direction becomes less natural: model-wrapper and model-adapter code would have to depend on an attention-backend-specific location for what is really shared SGLang forward metadata.

So my preference is:
- possible mechanically: yes
- good to merge into `plugin/sglang/attention_backend`: no

Keeping it in `plugin/sglang/utils/forward_context.py` seems cleaner. If we ever want to reduce file count further, a more reasonable merge candidate would be `base_model_wrapper.py`, since this metadata was originally split out from there and is conceptually closer to wrapper-level state than backend-level state.

atom/plugin/attention.py
wanzhenchn: Not yet in the sense of having one identical end-to-end attention backend shared by ATOM / vLLM / SGLang.

What we can realistically share is the core attention implementation itself (e.g. ATOM kernels / operator logic for GDN, MLA, MHA). But the runtime glue around it is still framework-specific today, mainly because the frameworks differ in:
- metadata preparation
- KV-cache layout / ownership
- forward-context binding
- scheduler / batch object integration
- graph-capture/runtime constraints

So in the short term, we still need framework-specific backend adapters, even if the underlying attention computation is shared as much as possible.

For DeepSeek, the direction is the same: the long-term goal is to converge toward a shared attention implementation, but that requires us to first control / unify the metadata and KV-cache adaptation layer. Once each framework can provide a unified view of the runtime state, then a much thinner extraction/adaptation layer would be enough.

There is also one practical constraint on the SGLang side for `Qwen3_5*` and `Qwen3NextForCausalLM`: the current custom backend path in `sgl_attn_backend.py` does not yet fully support their attention configuration (in particular `head_dim=256`; the relevant custom path currently only supports `head_dim=128`). Because of that, these models still keep upstream SGLang's native `AiterAttnBackend` by default.

So the current direction is:
- share the core attention implementation as much as possible
- keep framework-specific adapters where runtime contracts still differ
- gradually reduce those adapters once metadata / KV-cache preparation is unified

atom/models/qwen3_5.py
wanzhenchn: This mapping logic is intentionally kept in `qwen3_5.py` because it is model-specific loader/remap behavior, not a generic SGLang-wrapper concern.

Also, this is not unique to the SGLang path. `qwen3_5.py` already follows the same pattern under `is_vllm()`: it also contains model-specific mapping/remap logic there (for example its own `packed_modules_mapping`, plus the BF16-specific remap to `in_proj_qkvzba`).

DeepSeek follows the same general design. Its mapping logic does not live in `base_model_wrapper.py` either:
- model-side packed mapping is primarily defined in `atom/models/deepseek_v2.py`
- shared remap handling is applied through `quant_config.remap_layer_name(...)` in `atom/config.py`, which already has model-type-specific handling for DeepSeek variants
- the SGLang wrapper layer only orchestrates loading and does not define separate DeepSeek-specific mapping rules

So the split is intentional:
- generic wrapper/loading orchestration in the shared plugin layer
- model-specific mapping/remap rules in the model file and shared remap pipeline

atom/model_ops/attention_gdn.py
```diff
@@ -177,7 +178,14 @@ def forward(
             gdn_metadata.non_spec_state_indices_tensor
         )  # noqa: E501
 
-        conv_state = conv_state.transpose(-1, -2)
+        # `causal_conv1d_*` expects the logical shape [slot, conv_dim, state_len].
+        # ModelRunner stores [slot, state_len, conv_dim], so it needs the
+        # transpose below. SGLang already provides [slot, conv_dim, state_len],
+        # and the Triton kernel consumes the original conv_state strides directly.
+        if conv_state.size(1) != self.conv1d.weight.size(
```
Yuechguo: Maybe it is not a good formatting practice here. Check if there is a formatter available in this project.

atom/model_ops/linear.py
```diff
@@ -372,13 +372,19 @@ def process_weights_after_loading(self):
...
+            # quantized models must keep that tensor unshuffled here.
+            if self.weight.dim() == 2:
+                shuffle_weights(self.weight)
             # self.weight_scale.data = fp4_utils.e8m0_shuffle(self.weight_scale.data)
         else:
             if (
                 self.quant_type == QuantType.per_Token
                 and self.params_dtype == dtypes.fp8
             ) or (self.quant_type in [QuantType.per_1x32, QuantType.per_1x128]):
+                if self.weight.dim() != 2:
```
Yuechguo: It seems like we shouldn't return directly here. Returning directly might cause the last 'if' statement to be skipped.

atom/plugin/vllm/register.py
```diff
@@ -45,7 +45,11 @@ def register_platform() -> Optional[str]:
         logger.info("Disable ATOM OOT plugin platforms")
         return None
 
-    _set_plugin_mode()
+    # Do not call _set_plugin_mode() here. SGLang (and other stacks) discover
+    # vllm.platform_plugins and would set atom's backbone to "vllm" before
+    # importing SGLang OOT modules — then atom.models.qwen3_5's ``if is_vllm():``
```
Yuechguo: rm "OOT" and replace with "plugin"

atom/model_ops/mamba_ops/causal_conv1d.py
```diff
@@ -656,7 +656,9 @@ def causal_conv1d_fn(
         stride_istate_seq = conv_states.stride(0)
         stride_istate_dim = conv_states.stride(1)
         stride_istate_token = conv_states.stride(2)
-        assert stride_istate_dim == 1
+        # Keep this aligned with upstream: the Triton kernel consumes the
```
ZhiweiYan-96, line 659: I think concept upstream is not reasonable to ATOM, due to it is a self-owned project for ROCm. Please change the comments here.

wanzhenchn: The kernel implementation is derived from upstream SGLang, and its implementation supports strided read/write of conv_state. Therefore, this assert is actually unnecessary, and upstream SGLang has already relaxed this restriction [[link](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py#L471)].

ZhiweiYan-96 (approved):

PR #545, opened 2026-04-12
[feat](minimax): support minimax-2.5 in atom-vllm mode

PerryZhang01:
## Motivation

This PR adds support for MiniMax-M2.5 1 in ATOM vLLM plugin mode, following the onboarding flow in add-atom-vllm-model.md. Initial bring-up surfaced gaps between the native ATOM server path and the plugin path that required extra wiring and fixes.
- Fused QKV in plugin mode: The vLLM plugin attention stack expects a fused qkv buffer not only separate q, k, v tensors. 
- Config propagation: Some HF / runtime flags (e.g. trust_remote_code) are applied correctly at the vLLM CLI layer but were not always forwarded into ATOM’s plugin Config.
- Kernel / layout edge cases: Remaining `none` bugs tied to paged-attention assembly kernels.
- server script need add `--trust-remote-code` param.

so, currenly we use eager mode to run minimax-2.5 in atom-vllm mode, and these problems will be done in next prs.

## Test Result

image

## TODO
- remove qkv in atom-vllm plugin mode and use q,k,v, so `qkv cat` can be removed
- add eps in reshape_and_cache_with_pertoken_quant ops to avoid div zero, so plugin can use graph mode
- fix `none` bugs in paged-attention assembly kernels if avaiable.

### DIFF
--- a/.github/benchmark/oot_models_accuracy.json
+++ b/.github/benchmark/oot_models_accuracy.json
@@ -103,6 +103,18 @@
     "_baseline_note": "Reference value from recipes/atom_vllm/Kimi-K2.5.md"
   },
   {
+    "model_name": "MiniMax-M2.5 TP2",
+    "model_path": "MiniMaxAI/MiniMax-M2.5",
+    "extraArgs": "--kv_cache_dtype fp8 -tp 2 --trust-remote-code",
+    "env_vars": "ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION=1",
+    "runner": "linux-atom-mi35x-8",
+    "test_level": "nightly",
+    "accuracy_threshold": 0.92,
+    "accuracy_baseline": 0.9401,
+    "accuracy_baseline_model": "MiniMaxAI/MiniMax-M2.5",
+    "_baseline_note": "HF: amd/MiniMax-M2.5-MXFP4 card shows baseline=0.9401"
+  },
+  {
     "model_name": "DeepSeek-R1-FP8 TP8",
     "model_path": "deepseek-ai/DeepSeek-R1-0528",
     "extraArgs": "--tensor-parallel-size 8",
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -216,7 +216,6 @@ class PagedAttentionImpl(nn.Module):
         elif use_triton_attn and self.rotary_emb is not None:
             self.per_token_quant = False
             k_scale = v_scale = self.kv_scale
-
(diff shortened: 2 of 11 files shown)


wuhuikx:
Please help resolve the conflict and we need code review first then let it go through CI.

atom/model_ops/attention_mha.py
```diff
@@ -205,6 +204,7 @@ def rope_cache(self, q, k, v, qkv, position, fwd_ctx: ForwardContext):
                 q = self.q_norm(q)
             if self.k_norm is not None:
                 k = self.k_norm(k)
+
```
wuhuikx: I'm sorry, Why we need this change here?

atom/models/minimax_m2.py
```diff
@@ -244,8 +247,14 @@ def forward(
             k = (k * torch.rsqrt(k_var + self.rms_norm_eps) * self.k_norm.weight).to(
                 orig_dtype
             )
+            
+            # TODO: is_vllm will be removed after vllm plugin supporting q,k,v rather than qkv
+            if is_vllm():
```
wuhuikx: Can we move this part of code and avoid adding too much vllm/sglang related codes in the ATOM core files?

PerryZhang01: yea, `if is_vllm()` will be deleted after vllm-atom using q,k,v rather than qkv.

atom/model_ops/attention_mha.py
PerryZhang01: will be removed

.github/benchmark/oot_models_accuracy.json
```diff
@@ -90,6 +90,18 @@
     "accuracy_baseline_model": "amd/Kimi-K2.5-MXFP4",
     "_baseline_note": "Reference value from recipes/atom_vllm/Kimi-K2.5.md"
   },
+  {
+    "model_name": "MiniMax-M2.5 TP2",
+    "model_path": "MiniMaxAI/MiniMax-M2.5",
+    "extraArgs": "--kv_cache_dtype fp8 -tp 2 --trust-remote-code",
+    "env_vars": "ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION=1",
+    "runner": "linux-atom-mi35x-8",
```
wuhuikx, line 110: for tp2, we can use the runner of **linux-atom-mi35x-4**, please refer to the other model.

wuhuikx (approved):

zejunchen-zejun (approved):

valarLip (approved):

PR #549, opened 2026-04-13
[feat] Add RLHF rollout integration support (verl)

sijyang:
## Overview                                                                                                                                                                               
                                                            
  This PR enables ATOM to serve as a rollout backend for [verl](https://github.com/sijyang/verl/tree/dev), a distributed RLHF training framework. In RLHF training, the system alternates between two phases:
                                                                                                                                                                                            
  1. **Training phase** — the trainer updates model weights via gradient descent (handled by verl)                                                                                          
  2. **Rollout phase** — the inference engine generates responses using the latest weights (handled by ATOM)
                                                                                                                                                                                            
  This requires ATOM to support a new lifecycle that traditional serving doesn't need: receiving weight updates from an external trainer, dynamically releasing/reclaiming GPU memory between phases, and coordinating these operations across multiple DP ranks and TP ranks.                                                                                                  
                                                                                                                                                                                            
  The integration is designed as a **plugin layer** (`atom/rollout/`) that extends ATOM's existing engine without modifying its core inference path. All changes to existing files are purely incremental additions (new methods, new fields, new message types), no existing behavior is altered.
                                                                                                                                                                                            
  ## Architecture                                           

  verl trainer (PyTorch DDP)                                                                                                                                                                
      │                                                                                                                                                                                    
  ATOMHttpServer (verl side, per-node)                      
      │  ZMQ RPC                                                                                                                                                                            
  AsyncLLMEngine (atom/rollout/async_engine.py)                                                                                                                                                                                                                                                                                                                                 
      ├── sleep()        → release KV cache, free GPU memory for training                                                                                                                   
      ├── wake_up()      → reallocate KV cache, ready for generation                                                                                                                        
      ├── load_weights()  → receive updated weights via CUDA IPC                                                                                                                     
      └── generate()     → standard ATOM inference with logprobs                                                                                                                            
                                                                                                                                                                                            
  ### Weight Synchronization                                                                                                                                                                                                                                                                                                                       

  Weight transfer uses **CUDA IPC** (`weight_sync.py → weight_updater.py`) for zero-copy GPU-to-GPU transfer via `cudaIpcGetMemHandle`/`cudaIpcOpenMemHandle`. Weights are packed into a GPU buffer, and IPC handles are sent to ModelRunner subprocesses. On multi-GPU setups (DP>1), per-GPU buffers ensure same-device IPC.                              
                                                                                                                                                                                            
  Weights are accumulated into fixed-size buckets and flushed incrementally, keeping peak memory overhead bounded regardless of model size.                                                                                                                                                                                      
   
  ### Weight Update Pipeline                                                                                                                                                                
                                                            
  `weight_updater.py` handles the ModelRunner side of weight loading:                                                                                                                       
  - Maps incoming parameter names to ATOM's internal weight names (handling TP sharding, column/row parallel splits)
  - Supports **packed weights** (e.g., QKV fused) by slicing incoming tensors to correct offsets                                                                                            
  - Handles **FP8 requantization** — when the model uses FP8, incoming FP16/BF16 weights are quantized in-place with updated scales                                                         
  - Clears KV cache after weight update to prevent stale cache from previous weights                                                                                                        
                                                                                                                                                                                            
  ### GPU Memory Lifecycle                                                                                                                                                                  
                                                                                                                                                                                            
  `memory_manager.py` manages the sleep/wake cycle:                                                                                                                                         
  - **Sleep**: deallocate KV cache blocks → `torch.cuda.empty_cache()` → memory returned to PyTorch/ROCm for trainer
  - **Wake**: `empty_cache()` → recalculate available blocks → reallocate KV cache → ready for inference                                                                                    
  - Each DP rank manages its own KV cache independently                                                                                                                                     
                                                                                                                                                                                            
  ### DP Isolation                                                                                                                                                                          
                                                                                                                                                                                            
  `model_runner_ext.py` (RLHFModelRunner) extends ATOM's ModelRunner for DP-isolated execution. Each DP rank's ModelRunners form an independent NCCL world scoped to TP only, with correct  physical-to-logical device mapping and NCCL binding patches for ROCm multi-GPU setups.                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                
  ## Changes                                                                                                                                                                                
                                                                                                                                                                                            
  ### New files (`atom/rollout/`)                                                                                                                                              
  | File | Purpose |
  |------|---------|                                                                                                                                                                        
  | `__init__.py` | Package exports |                       
  | `async_engine.py` | AsyncLLMEngine wrapper (sleep/wake/load_weights API) |                                                                                                              
  | `engine_utility.py` | Utility command handlers (update_weights, release/resume_memory) |                                                                                                
  | `memory_manager.py` | GPU memory lifecycle (KV cache alloc/release, weight discard/resume) |                                                                                            
  | `model_runner_ext.py` | RLHFModelRunner with DP isolation, NCCL device binding patch |                                                                                                  
  | `weight_sync.py` | Weight transfer via CUDA IPC (per-GPU buffers) |                                                                                                             
  | `weight_updater.py` | Weight update logic (packed weights, FP8 requantize, TP sharding) |                                                                                               
                                                                                                                                                                                            
  ### Incremental changes to existing files                                                                                                                    
  - **engine_core.py**: utility queue, sleep mode, UTILITY_RESPONSE message type, DP sleep state sync                                                                                       
  - **engine_core_mgr.py**: utility_response_queue, broadcast_utility_command, broadcast_utility_command_sync                                                                               
  - **llm_engine.py**: request_ids and logprobs support in add_request/generate/postprocess                                                                                                 
  - **async_proc.py**: TP-rank barrier for safe weight update buffer reuse                                                                                                                  
  - **scheduler.py**: logprobs tracking in ScheduledBatch/ScheduledBatchOutput                                                                                                              
  - **sequence.py**: request_id, return_logprobs, logprobs fields                                                                                                                           
  - **sampling_params.py**: logprobs parameter                                                                                                                                              
  - **config.py**: runner_qualname, compilation_config dict→object conversion


atom/model_engine/engine_core.py
```diff
@@ -171,9 +186,13 @@ def run_engine(config: Config, input_address: str, output_address: str):
     def busy_loop(self):
         shutdown = False
         while True:
+            self.utility_handler.process_queue(self.utility_queue, self)
             shutdown = shutdown or self.pull_and_process_input_queue()
             if shutdown:
                 break
+            if self._is_sleeping:
```
ZhangLirong-amd: Why we need such sleep? I think it's not a good idea to sleep in our engine core

atom/model_engine/engine_core.py
```diff
@@ -350,18 +390,37 @@ def exit(self):
...
 
-            local_is_prefill, local_num_tokens = self.scheduler.get_next_batch_info()
-            local_unfinished = not self.scheduler.is_finished()
+            # Sleeping cores must still participate in _sync_dp_state (NCCL
+            # all_reduce) to prevent other DP ranks from blocking forever.
+            # The sleep flag is included in the sync tensor so that ALL cores
+            # agree to skip model execution together — MoE expert routing and
+            # dummy_execution also contain DP-wide collectives that would hang
+            # if only some cores participated.
+            local_sleeping = self._is_sleeping
```
ZhangLirong-amd: The same

atom/rollout/memory_manager.py
```diff
@@ -0,0 +1,225 @@
...
+                f"{num_blocks} due to changed GPU memory availability"
+            )
+        self.allocate_kv_cache(num_blocks)
+        logger.info(
+            f"{self.label}: KV cache re-allocated and bound ({num_blocks} blocks)"
+        )
+        # Recapture CUDA graphs after KV cache re-allocation (addresses changed)
+        self._recapture_cudagraphs_if_needed()
+
+    def _recapture_cudagraphs_if_needed(self) -> None:
```
ZhangLirong-amd, line 188: Will be triggered frequently? Cuda Graph capture is overhead and is time-consuming, so it shouldn't be constantly recaptured.

atom/rollout/model_runner_ext.py
```diff
@@ -0,0 +1,378 @@
...
+            if logprobs is not None:
+                logprobs_map = dict(zip(self.prev_req_ids, logprobs))
+
+        self.prev_batch = batch
+        self.prev_token_ids = sampled_token_ids
+
+        return req_ids_out, processed_out, logprobs_map
+
+
+class RLHFModelRunner(ModelRunner, WeightUpdaterMixin, MemoryManagerMixin):
```
ZhangLirong-amd, line 19: Find a better way to reuse parent model_runner in atom

atom/rollout/model_runner_ext.py
```diff
@@ -0,0 +1,378 @@
...
+from atom.model_engine.model_runner import ModelRunner, tokenIDProcessor
+from atom.model_engine.scheduler import ScheduledBatch, ScheduledBatchOutput
+from atom.rollout.memory_manager import MemoryManagerMixin
+from atom.rollout.weight_updater import WeightUpdaterMixin
+from aiter.dist.parallel_state import get_tp_group
+
+logger = logging.getLogger("atom")
+
+
+class LogprobsTokenIDProcessor(tokenIDProcessor):
```
ZhangLirong-amd: Use atom TokenIDProcessor it's enough, can we move logprobs in TokenIDProcessor  and remove this class?

atom/model_engine/model_runner.py
```diff
@@ -201,20 +211,39 @@ def prepare_sampled_ids(
...
                 processed = [(tid,) for tid in token_ids]
-            return req_ids, processed
+            logprobs_out = None
+            if sampled_logprobs is not None:
+                lp = sampled_logprobs.tolist()
+                logprobs_out = dict(zip(req_ids, lp))
+            return req_ids, processed, logprobs_out
 
         token_ids = self.recv_async_output(self.token_ids_cpu).tolist()
-        self.send_to_cpu_async(sampled_token_ids, self.token_ids_cpu, sync_event)
```
ZhangLirong-amd, line 243: why not added logprobs in send_to_cpu_async directly?

sijyang: `send_to_cpu_async` requires a real tensor, but `sampled_logprobs` is `Optional`, it's `None` when no sequence in the batch requested `logprobs`. We also need `token_ids` and `logprobs` to share the same `copy_done` event for correct pairing. Inlining the copy in the same stream context handles both concerns cleanly.

atom/model_engine/engine_core.py
sijyang: Renamed to `_is_rl_weights_offloaded` to clarify intent.

atom/model_engine/engine_core.py
sijyang: Renamed to `local_offloaded` = `self._is_rl_weights_offloaded`. This flag is only used in the RL training workflow, where weights are temporarily offloaded to free GPU memory for the training step.

atom/rollout/memory_manager.py
sijyang: Not frequent. This is only triggered during RL training's weight offload/reload cycle, once per training iteration, not per batch or per request.

atom/rollout/model_runner_ext.py
sijyang: RLHF-specific logic lives in rollout/model_runner_ext.py as Mixin composition (`WeightUpdaterMixin` + `MemoryManagerMixin`). The base `ModelRunner` stays clean. Only `logprobs` and DP isolation were merged back because they are also useful for inference. `RLHFModelRunner` only overrides `_setup_device_and_distributed` for DP-isolated NCCL worlds. See commit 5bdc160 for the refactor history.

atom/rollout/model_runner_ext.py
sijyang: Done. Logprobs merged into base TokenIDProcessor in commit 5bdc160.

atom/model_engine/engine_core.py
```diff
@@ -98,20 +109,12 @@ def __init__(self, config: Config, input_address: str, output_address: str):
 
             config.num_kvcache_blocks = num_blocks
             if not config.enforce_eager:
-                # Start profiler before cudagraph capture only if mark-trace is enabled.
```
ZhangLirong-amd, line 105: recover it, we need support it

atom/model_engine/engine_core.py
```diff
@@ -84,9 +94,10 @@ def __init__(self, config: Config, input_address: str, output_address: str):
             self.runner_mgr = AsyncIOProcManager(
                 self._finalizer,
                 config.tensor_parallel_size,
-                "atom.model_engine.model_runner.ModelRunner",
+                config.runner_qualname,
```
ZhangLirong-amd, line 98: do we have to change it?

atom/model_engine/engine_core.py
sijyang: Recovered.

atom/model_engine/engine_core.py
sijyang: Yes, this is needed for veRL rollout. `AsyncLLMEngine` uses `RLHFModelRunner` instead of the default `ModelRunner`.

valarLip (approved):

PR #571, opened 2026-04-15
[fix](metadata): optimize attention metadata

PerryZhang01:
## Motivation

This PR refactors and optimizes attention metadata. It can bring a 1% to 2% performance improvement on gpt-oss-120b in low concurrency.

image

### DIFF
--- a/atom/plugin/attention.py
+++ b/atom/plugin/attention.py
@@ -26,7 +26,6 @@ disable_vllm_plugin_attention = envs.ATOM_DISABLE_VLLM_PLUGIN_ATTENTION
 @dataclass
 class AiterFlashAttentionPhaseMetadata:
     max_query_len: int
-    min_query_len: int
     max_seq_len: int
     query_start_loc: torch.Tensor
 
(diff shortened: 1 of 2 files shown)


atom/plugin/attention.py
```diff
@@ -324,35 +322,75 @@ def build(
             num_prefill_tokens,
         ) = split_ret
 
-        query_start_loc_cpu = common_attn_metadata.query_start_loc_cpu
+        prefill_only = num_decodes == 0 and num_extends == 0 and num_prefills > 0
+        decode_only = num_decodes > 0 and num_extends == 0 and num_prefills == 0
+        mixed = not (prefill_only or decode_only)
+
+        # common_attn_metadata._seq_lens_cpu is equal to common_attn_metadata.seq_lens.cpu(),
```
wuhuikx, line 329: Do we still need the code commented here?

PerryZhang01: In vllm0.19.0, we don`t need to introduce D2H in the attention metadata. but it`s so strange in the real test of gpt-oss-4-conc, copying seqs_len to cpu brings shorter TTFT and TPOT, so we resolve it here.

zejunchen-zejun (approved):

valarLip (approved):

PR #579, opened 2026-04-16
[atom-vllm] by default disable vllm mori because the atom mori will be used for atom-vllm

zejunchen-zejun:
Disable the vllm mori import because when atom-vllm mode, the atom mori will be used for model implementation. When atom upgrades the mori version, there could be ABI issue if importing vllm mori. Meanwhile warn users when set the flag `ATOM_DISABLE_VLLM_MORI` to 0 because there could be high risky to encounter an ABI issue.


zejunchen-zejun:
Close this PR because patch `has_mori` to be False in vLLM is not reasonable way to avoid ABI issue or compatability issue

PR #582, opened 2026-04-16
[Acc] remove flatten in dpsk fp4 by quantizing mla weights && a…

ZLkanyo009:
remove flatten in dpsk fp4 by quantizing mla weights && acc kv_b_proj by changing weight layout like vllm atom'

## flatten op
The flatten operation was previously removed in DeepSeek FP8, but in FP4, the weights in MLA are BF16. Because the type check was not performed, the wrong branch was taken, and the flatten operator still appeared in that branch. In vLLM, the MLA weights are directly quantized to FP8 right after loading (regardless of the original type of the weights). Therefore, the modification was made here to mimic vLLM.
This modification also successfully changed the ck operator back to the triton operator in the vLLM trace.

before:
image

after:
image

## acc kv_b_proj
After changing the ck operator back to the triton operator, the performance of kv_b_proj decreased. It was found that the reason is that the layout of the weights is different from that in vLLM.

before:
image

after:
image

### DIFF
--- a/atom/plugin/sglang/attention_backend/sgl_attention_mla.py
+++ b/atom/plugin/sglang/attention_backend/sgl_attention_mla.py
@@ -23,7 +23,10 @@ from aiter import dtypes
 from aiter.dist.parallel_state import get_tensor_model_parallel_world_size, get_tp_group
 from aiter.ops.triton.fused_fp8_quant import fused_rms_fp8_group_quant
 from atom.model_ops.base_attention import Attention
-from atom.model_ops.attention_mla import fused_qk_rope_concat_and_cache_mla
+from atom.model_ops.attention_mla import (
+    dynamic_per_batched_tensor_quant,
+    fused_qk_rope_concat_and_cache_mla,
+)
 from atom.models.utils import maybe_prefix
 
 # sglang imports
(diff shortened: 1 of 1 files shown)


atom/plugin/sglang/attention_backend/sgl_attention_mla.py
```diff
@@ -230,13 +253,54 @@ def mla_absorbed_bmm(
...
         )
-        out = bmm_fp8(val, weight, scale, weight_scale, torch.bfloat16)
+        out = bmm_fp8(val, weight, scale, effective_weight_scale, torch.bfloat16)
         return out.transpose(0, 1)
 
     # bf16 fallback
     return torch.bmm(inp.transpose(0, 1), weight).transpose(0, 1)
 
 
+def mla_v_up_proj(
```
valarLip, line 310: why not use atom's original impl

ZLkanyo009: I used the built-in mla_absorbed_bmm from this file and optimized it further. This PR is based on https://github.com/ROCm/ATOM/pull/525/changes. We can discuss it in PR #525.

zhuyuhua-v (approved):

valarLip (approved):

PR #583, opened 2026-04-16
[atom-vllm] rebuild mori because torch will be downgrade to a low version, and mori needs to be built with this low version torch to avoid the ABI issue

zejunchen-zejun:


docker/Dockerfile
```diff
@@ -103,6 +103,22 @@ RUN echo "========== [OOT 7/7] Install vLLM runtime dependencies ==========" &&
 RUN echo "========== [VLLM-ATOM] Validate vision/audio wheels ==========" && \
     "${VENV_PYTHON}" -c "import torch, torchvision, torchaudio; from torchvision.transforms import InterpolationMode; from transformers.models.auto.image_processing_auto import get_image_processor_config; print(f'torch: {torch.__version__}'); print(f'torchvision: {torchvision.__version__}'); print(f'torchaudio: {torchaudio.__version__}'); print(f'InterpolationMode: {InterpolationMode.BILINEAR}'); print(f'get_image_processor_config: {get_image_processor_config.__name__}')"
 
+# Rebuild mori against the current torch after vLLM deps may have changed it.
+# The ATOM base image compiled mori against its own torch (e.g. 2.10.0), but
+# vLLM's requirements/rocm.txt can downgrade torch (e.g. to 2.9.1), causing
+# ABI mismatches (missing symbols in libc10.so). Re-running setup.py install
+# links mori against the torch actually present in the OOT image
+RUN echo "========== [VLLM-ATOM] Rebuild mori against current torch ==========" && \
```
valarLip, line 111: why we will get torch downgrade..

zejunchen-zejun: good question, and I find here remain torch downgrade doesn't really make sense for atom-vllm. Let me keep the torch version 2.10

zejunchen-zejun:
Close this PR as it is not needed. Why today's atom-vllm CI has mori ABI conflict? It is because the torch had been downgraded to 2.9.1 because the torchvision was previously pinned to 2.9.1 as a WA in yesterday night's docker construction. For now yuhua has a PR: https://github.com/ROCm/ATOM/pull/575(merged yesterday night), it removed the torchvision version pin, so the torch would not be downgraded anymore since now, so there will not be mori ABI conflict.

PR #595, opened 2026-04-17
[fix](dashboard): Add dashboard_model for SGLang deepseek TP4

zhuyuhua-v:
## Motivation
  TP4 benchmark results for `DeepSeek-R1-0528` (FP8 and MXFP4) were not visible in the benchmark dashboard because they shared the same derived model name as their TP8 counterparts, causing the TP8 entry to silently overwrite the TP4 results.

  This PR adds explicit `dashboard_model` names for the two TP4 cases, following the same convention used in the vLLM benchmark config (e.g. `Kimi-K2-Thinking-MXFP4-tp4`, `Qwen3.5-397B-A17B-FP8-tp4`):                                                                                                                                                                                      
  - `DeepSeek-R1-0528 FP8 TP4` → `dashboard_model: "DeepSeek-R1-0528-tp4"`
  - `DeepSeek-R1-0528-MXFP4 FP4 TP4` → `dashboard_model: "DeepSeek-R1-0528-MXFP4-tp4"`                                                                                                                                                             

TP8 entries continue to use their existing derived names (no `-tp8` suffix), consistent with the convention that TP8 is the default configuration.

### DIFF
--- a/.github/benchmark/sglang_benchmark_models.json
+++ b/.github/benchmark/sglang_benchmark_models.json
@@ -22,8 +22,8 @@
   },
   {
     "display": "DeepSeek-R1-0528-MXFP4 FP4 TP8",
-    "source_path": "amd/DeepSeek-R1-0528-MXFP4",
-    "path": "amd/DeepSeek-R1-0528-MXFP4",
+    "source_path": "amd/DeepSeek-R1-0528-MXFP4-MTP-MoEFP4",
+    "path": "amd/DeepSeek-R1-0528-MXFP4-MTP-MoEFP4",
     "prefix": "deepseek-r1-fp4-tp8",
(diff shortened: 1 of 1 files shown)


.github/benchmark/sglang_benchmark_models.json
```diff
@@ -31,6 +32,7 @@
   },
   {
     "display": "DeepSeek-R1-0528-MXFP4 FP4 TP4",
+    "dashboard_model": "DeepSeek-R1-0528-MXFP4-tp4",
```
wuhuikx, line 35: let's use the -mtp model to align with ATOM and vLLM-ATOM.

zhuyuhua-v: updated

valarLip (approved):

PR #613, opened 2026-04-20
[feat](minimax): refactor rmsnorm for minimax

PerryZhang01:
## Motivation

This PR refactors rmsnorm in minimax_v2.py. In minimax norm operation uses whole q dim and k dim rather than one head dim, so it should introduce allreduce in tp mode. this pr algins vllm code. (https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/minimax_m2.py#L222)

## TEST
image


atom/model_ops/attention_mha.py
```diff
@@ -77,6 +78,9 @@ def __init__(
         self.rotary_emb = rotary_emb
         self.q_norm = q_norm
         self.k_norm = k_norm
+        self.norm_accross_all_heads = False
```
valarLip, line 81: urgly..

valarLip (changes requested):

atom/model_ops/attention_mha.py
PerryZhang01: variable name or other something else?

PR #636, opened 2026-04-23
fix(plugin): guard mixed attention metadata reductions

XiaobingSuper:
## Summary
- guard mixed attention metadata reductions so empty decode/prefill slices do not call `.max()`
- handle mixed batches that only contain a subset of decode, extend, and prefill request types
- prevent attention metadata build failures in mixed edge cases such as decode+extend without prefills

## Test plan
- Not run
- Reproduce the mixed-batch case that previously hit `query_lens_cpu[num_decodes + num_extends :].max()`
- Verify metadata build succeeds when `num_prefills == 0` or `num_decodes == 0`

## Submission Checklist

### DIFF
--- a/atom/plugin/attention.py
+++ b/atom/plugin/attention.py
@@ -345,17 +345,18 @@ class vllmAttentionMetadataBuilderMethods:
         )
 
         if mixed:
-            prefill_max_query_len = (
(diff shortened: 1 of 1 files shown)


PerryZhang01:
approved

zejunchen-zejun (approved):

valarLip (approved):

PR #643, opened 2026-04-24
[ATOM-SGLang][Feat] Enable Deepseek v3 MTP

ZhiweiYan-96:
# Proposed Design

## 1. MTP module creation: Override the draft architecture through the external model package

As background knowledge, it is helpful to first detail how `SGLang` loads the DeepSeek MTP module in its native flow. From `SGLang`'s point of view, DeepSeek MTP is not an auxiliary block hidden inside the target model. It is a standalone draft architecture. The loading path is roughly:

1. The user enables speculative decoding through server arguments such as `--speculative-algorithm NEXTN`
2. `SGLang` normalizes `NEXTN` into the `EAGLE` runtime family in server args
3. The draft `ModelConfig` rewrites the DeepSeek V3 draft architecture to `DeepseekV3ForCausalLMNextN` inside `_config_draft_model()`
4. `ModelRegistry` resolves the model class by that architecture name
5. The resolved class is instantiated as the draft model and then used by the speculative worker in the propose / verify / extend lifecycle

In other words, `SGLang` first interprets "DeepSeek MTP" as "a separately loaded draft model", and only then enters the runtime phase. The external model package hook works exactly at this architecture-resolution stage.

For MTP side, SGLang uses `DeepseekV3ForCausalLMNextN` as MTP model architecture. 

- `DeepseekV3ForCausalLMNextN`

The following diagram shows the native `SGLang` view of how the MTP module is loaded:

```mermaid
flowchart TD
    subgraph SGL["SGLang domain"]
        A["CLI / server args<br/>--speculative-algorithm NEXTN"]
        B["Normalize algorithm<br/>NEXTN -> EAGLE"]
        C["Build draft ModelConfig"]
        D["_config_draft_model()<br/>rewrite architecture to<br/>DeepseekV3ForCausalLMNextN"]
        E["ModelRegistry.resolve_model_cls(...)"]
        F["Instantiate draft model class"]
        G["Speculative worker uses draft model<br/>propose / verify / extend"]
    end

    A --> B
    B --> C
    C --> D
    D --> E
    E --> F
    F --> G
```

`SGLang` allows external model packages to register architectures through `SGLANG_EXTERNAL_MODEL_PACKAGE` and module-level `EntryClass`. This is also the core mechanism for ATOM SGLang plugin. The plugin uses this mechanism to expose a class with the exact architecture name expected by `SGLang`:

Once this class is available in the plugin package, `SGLang` resolves the draft architecture to the plugin implementation instead of the upstream one in `sglang/srt/models/deepseek_nextn.py`.

The following diagram illustrates what "overriding the draft architecture" means in practice:

```mermaid
flowchart TD
    subgraph SGL["SGLang domain"]
        A["launch server<br/>SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models"]
        B["Import external model package"]
        C["Read module EntryClass"]
        D["Register architecture:<br/>DeepseekV3ForCausalLMNextN"]
        E["ModelRegistry.resolve_model_cls(...)"]
        H["upstream draft implementation<br/>sglang/srt/models/deepseek_nextn.py"]
    end

    subgraph PLUGIN["ATOM SGLang Plugin domain"]
        F["plugin wrapper<br/>DeepseekV3ForCausalLMNextN"]
    end

    subgraph CORE["ATOM Core domain"]
        G["DeepSeekMTP"]
    end

    A --> B
    B --> C
    C --> D
    D --> E
    E --> F
    F --> G

    H -. "same architecture name is overridden" .-> D
```

The important point is that architecture resolution and `ModelRegistry` selection still happen inside the `SGLang` domain. The `ATOM SGLang Plugin` domain only contributes a same-name wrapper through the external package entry point, while the actual draft computation is delegated to `DeepSeekMTP` in the `ATOM Core` domain. This makes it easier to separate who owns scheduling and model resolution from who owns the draft implementation.

## 2. MTP Warrper in ATOM:  A thin wrapper as the compatibility bridge

The plugin adds a lightweight wrapper named `DeepseekV3ForCausalLMNextN`. Externally, it matches the draft-model interface expected by `SGLang`. Internally, it delegates the actual draft computation to `ATOM DeepSeekMTP`.

The wrapper is responsible for:

- creating the plugin-mode `atom_config`
- rewriting configuration semantics so they match `DeepSeekMTP`
- instantiating `ATOM/atom/models/deepseek_mtp.py::DeepSeekMTP`
- exposing `get_embed_and_head()`, `set_embed_and_head()`, and `set_embed()` so speculative workers can share embeddings and LM head weights with the target model
- consuming `forward_batch.spec_info.hidden_states` in `forward()`
- loading weights through the `spec_decode=True` path

The design principle is:

- keep the `SGLang` architecture name and draft-worker contract unchanged at the top layer
- reuse `ATOM DeepSeekMTP` as the implementation at the lower layer

This minimizes duplication, avoids recreating the upstream NextN hierarchy inside the plugin, and makes future improvements to `ATOM`'s native MTP implementation reusable in plugin mode.

### Intrusive change to formal runtime variable control codes

`ATOM` core code currently relies on some process-global runtime/config state. In speculative mode, target and draft wrappers coexist. Without isolation, initializing or running the draft wrapper may overwrite global state used by the target path, leading to subtle cross-contamination in MoE or attention behavior.

To address this, the plugin introduces a runtime scope that explicitly binds and restores the proper runtime context around wrapper `__init__`, `forward()`, and `load_weights()`. This allows target and draft instances to coexist safely.

However, this also makes an architectural issue visible: the current plugin system still has meaningful complexity around process-global state management. In order to let multiple wrappers coexist, the plugin must repeatedly switch and restore global runtime state at execution boundaries. In that sense, `runtime scoping` should be understood as a containment mechanism for the current global-state model, not as the ideal long-term abstraction. It solves the correctness problem for this branch, but it also suggests a future direction toward fewer implicit globals and more explicitly instantiated runtime state.

### Direct attn backend replacment `sglang_aiter_backend.AiterAttnBackend = ATOMAttnBackendForSgl`

The reason can be summarized by the key `SGLang` call chain:

```python
# eagle_worker.py
self.draft_attn_backend = DraftBackendFactory(...).create_decode_backend()

# draft_utils.py
def _create_aiter_decode_backend(self):
    return AiterMultiStepDraftBackend(...)

# aiter_backend.py
for i in range(self.speculative_num_steps - 1):
    self.attn_backends.append(AiterAttnBackend(model_runner, ...))
```

In other words, `EAGLE` draft multi-step decode actually goes through:

```mermaid
flowchart LR
    A["EAGLEWorker"] --> B["DraftBackendFactory"]
    B --> C["AiterMultiStepDraftBackend"]
    C --> D["AiterAttnBackend(...)"]
    R["ATOM-sglang attention registry"] -. "not used on this path" .-> D
```

So if the plugin only overrides the `"aiter"` registry entry, but does not also rewrite:

- `sglang.srt.layers.attention.aiter_backend.AiterAttnBackend`

then `EAGLE` draft decode still directly constructs the upstream `AiterAttnBackend`. That is why this monkeypatch is hacky, but still practically necessary on the current branch.

The plugin is mutating an upstream module symbol directly. This is not a clean extension point.

### Others changes

Complete the radix attention forward in specualtive mode, like

- `TARGET_VERIFY`
- `DRAFT_EXTEND`

###  Accuracy

image

Acceptance ratio

image


atom/plugin/sglang/attention_backend/radix_attention.py
```diff
@@ -88,17 +88,20 @@ def __init__(
                 self.attn.k_scale = atom_parameter(
                     torch.tensor([1.0], dtype=torch.float32, device="cuda")
                 )
+            elif not self.attn.k_scale.is_cuda:
+                self.attn.k_scale = torch.nn.Parameter(
```
wanzhenchn: Please use the centralized `atom_parameter()` factory in `atom/model_ops/utils.py` introduced by PR https://github.com/ROCm/ATOM/pull/574 to replace `nn.Parameter(...)` constructor.

atom/plugin/sglang/attention_backend/radix_attention.py
```diff
@@ -88,17 +88,20 @@ def __init__(
...
                 )
-            # Some SGLang attention backends consume the host-side float scales
-            # directly. Keep them in sync with the device-side defaults so the
-            # plugin path works even when checkpoint loading never populates them.
-            if self.attn.k_scale_float is None:
-                self.attn.k_scale_float = 1.0
-            if self.attn.v_scale_float is None:
-                self.attn.v_scale_float = 1.0
+            elif not self.attn.v_scale.is_cuda:
+                self.attn.v_scale = torch.nn.Parameter(
```
wanzhenchn: Ditto

atom/plugin/sglang/models/deepseek_nextn_wrapper.py
```diff
@@ -0,0 +1,199 @@
...
+            attn_obj = getattr(self_attn, attr_name, None)
+            if attn_obj is None:
+                continue
+            _set_runtime_layer_id(attn_obj, local_layer_id)
+            nested_attn = getattr(attn_obj, "attn", None)
+            if nested_attn is not None:
+                _set_runtime_layer_id(nested_attn, local_layer_id)
+
+
+class DeepseekV3ForCausalLMNextN(nn.Module):
```
wanzhenchn, line 79: Perhaps you can implement your custom `DeepseekV3ForCausalLMNextN` by inheriting from the `_AtomCausalLMBaseForSglang` base class and adapting `prepare_model()` in `plugin/prepare.py`. This would help avoid a lot of redundant initialization code(like self.xx=xx, register_ops_to_sglang(), set_attn_cls(), init_aiter_dist(() and setting up engine type with plugin_runtime_scope()).

ZhiweiYan-96: Thanks for suggestions. Currently, the most complexity exists at forward and load methods. On the contrary, init is not a big burden. This module is only used for MTP while the _AtomCausalLMBaseForSglang is assumed as target model.  I prefer not to mix the responsibility for these two classes currently.

atom/plugin/sglang/attention_backend/radix_attention.py
ZhiweiYan-96: modifed, thanks

atom/plugin/sglang/attention_backend/radix_attention.py
ZhiweiYan-96: modified, thx

atom/plugin/sglang/models/deepseek_nextn_wrapper.py
```diff
@@ -0,0 +1,199 @@
...
+    @torch.no_grad()
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        positions: torch.Tensor,
+        forward_batch: ForwardBatch,
+        input_embeds: torch.Tensor = None,
+        **kwargs,
+    ):
+        del kwargs
```
zhuyuhua-v: why are we del kwargs?

atom/plugin/sglang/attention_backend/radix_attention.py
```diff
@@ -88,17 +88,20 @@ def __init__(
...
-            # plugin path works even when checkpoint loading never populates them.
-            if self.attn.k_scale_float is None:
-                self.attn.k_scale_float = 1.0
-            if self.attn.v_scale_float is None:
-                self.attn.v_scale_float = 1.0
+            elif not self.attn.v_scale.is_cuda:
+                self.attn.v_scale = atom_parameter(
+                    self.attn.v_scale.detach().to(device="cuda"),
+                    requires_grad=False,
+                )
```
zhuyuhua-v, line 102: why are we del self.attn.k_scale_float=1.0 and self.attn.v_scale_float = 1.0?

atom/plugin/sglang/attention_backend/sgl_attn_backend.py
```diff
@@ -872,40 +1271,240 @@ def init_forward_metadata_replay_cuda_graph(
         seq_lens_cpu: Optional[torch.Tensor],
         out_cache_loc: Optional[torch.Tensor] = None,
     ):
-        if not forward_mode.is_decode_or_idle():
-            raise ValueError("Invalid forward mode")
+        num_kv_splits = None
```
zhuyuhua-v, line 1283: a lot of same code with init_forward_metadata_capture_cuda_graph?

qichu-yun (dismissed):

atom/plugin/sglang/attention_backend/radix_attention.py
ZhiweiYan-96: v_scale_float and k_svale_float is not used in sglang side. All usage is attn.v_scale and attn.k_scale

atom/plugin/sglang/models/deepseek_nextn_wrapper.py
ZhiweiYan-96: added back

atom/plugin/sglang/attention_backend/sgl_attn_backend.py
ZhiweiYan-96: yes, this is due to different forward mode requires diverse metadata. no need to extract common code i think.

ZhiweiYan-96:
Hi, @valarLip could you please take a look on this PR on sglang side MTP support?  Appreciation for all suggestions.

atom/plugin/sglang/attention_backend/sgl_attn_backend.py
```diff
@@ -429,6 +438,186 @@ def _init_forward_metadata_extend(self, forward_batch: ForwardBatch):
         else:
             self._init_extend_mha(bs, forward_batch)
 
+    def _init_draft_extend_mla(self, bs, forward_batch):
+        """Init MLA metadata for speculative draft_extend."""
```
whx-sjtu, line 436: It looks like we re-implement the whole metadata build methods. Why can't we reuse sglang's original metadata?

valarLip (dismissed):

atom/plugin/sglang/attention_backend/sgl_attn_backend.py
ZhiweiYan-96: yes, we can reuse, but sglang side would try uses kernel same as ATOM when mature. Add new fields or removing fileds is expected (already met for non speculative case). Generally, it is in my intention for the metadata no reusing here.

atom/plugin/sglang/attention_backend/radix_attention.py
ZhiweiYan-96: added back, these codes are necessary to triton backend....

valarLip (approved):

PR #644, opened 2026-04-24
[vLLM-ATOM] Add profile trace parsing tool for vLLM-ATOM

kliuae-amd:
## Motivation
This PR adds a profile trace parsing tool that breaks down vLLM-ATOM traces for hot kernel analysis.

## Technical Details
Provide kernel breakdowns of mixed/prefill and decode batches for kernel runtime analysis.
It leverages vLLM's trace markers enabled by `VLLM_CUSTOM_SCOPES_FOR_PROFILING=1` to gather kernels in batches.
Currently it uses input layernorms as boundaries to separate layers from one another.

Usage:
```
python parse_vllm_trace.py <trace.json.gz>
    [--layer N]            Target transformer layer index.
    [--percentile P]       The X-percentile batches to extract kernel stats from.
    [--eager-trace <path>] Companion eager trace for extracting host module attributes.
                           For graph-captured batches like decode workloads in FULL_AND_PIECEWISE settings,
                           GPU kernels in a graph are packed into a single graph replay event.
                           An eager trace can be provided alongside with graphed traces to establish kernel
                           launches with host side launch modules from eager traces.
    [--output-prefix <str> Prefix for output XLSX filename.
```

## Submission Checklist

### DIFF
--- a/tools/vllm_atom/parse_vllm_trace.py
+++ b/tools/vllm_atom/parse_vllm_trace.py
@@ -1,0 +1,1703 @@
+#!/usr/bin/env python3
+"""Parse vLLM profiler traces to extract per-layer kernel breakdowns.
+
+Important: vLLM traces must be captured with `VLLM_CUSTOM_SCOPES_FOR_PROFILING=1` to enable user annotation
+           markers that provide scopes for gpu kernel events.
+
+Usage:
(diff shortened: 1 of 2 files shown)


wuhuikx:
can we also extract the shape? or the token numbers for each forward?

PR #692, opened 2026-05-05
ci(dashboard): use amd-smi for GPU name to fix MI355X identification

ChuanLi1101:
## Summary

The benchmark dashboard at <https://rocm.github.io/ATOM/benchmark-dashboard/> identifies each run by the GPU name collected via `rocm-smi --showproductname` from inside the container. On MI355X (and other recently-launched ASICs) `rocm-smi`'s product-name table is missing the marketing string, so the dashboard shows runs as `8 x Radeon Graphics` instead of `MI355X`, which has been confusing customers (e.g. Neysa).

This PR replaces the duplicated inline `rocm-smi`/`rocminfo` blocks in **six** benchmark and test workflows with a single shared helper at `.github/scripts/collect_gpu_info.sh` that:

1. **Prefers `amd-smi static --asic`** — the supported successor to `rocm-smi`, which correctly reports MI300X / MI325X / MI355X market names because new ASIC enablement lands here first.
2. **Falls back to `rocm-smi --showproductname`** for older container images that do not yet ship `amd-smi`.
3. **Falls back to `rocminfo` "Marketing Name"** as a last resort, matching the existing benchmark-pipeline behaviour.

VRAM detection follows the same prefer-amd-smi-with-fallback pattern (`amd-smi static --vram --json`, schema-tolerant via python).

The PR also adds a top-level `.gitattributes` rule pinning shell scripts to LF so Windows contributors cannot accidentally introduce CRLF line endings that would break the script on Linux runners.

Net diff: **+98 / -57** lines (8 files changed). Pure CI/dashboard plumbing — no runtime / model code touched.

## Why now

This is the workaround Mrinal (Karvir) asked for in the Teams thread: the dashboard label is wrong on MI355X hardware, the rocm-smi gap will eventually be patched upstream but in the meantime amd-smi already has the correct mapping. Migrating to amd-smi is also the recommended direction going forward (rocm-smi is deprecated).

## Files changed

| File | Change |
| --- | --- |
| `.github/scripts/collect_gpu_info.sh` | **new** — shared helper, prefers amd-smi |
| `.gitattributes` | **new** — pin `*.sh` to `eol=lf` |
| `.github/workflows/atom-benchmark.yaml` | call helper |
| `.github/workflows/atom-vllm-benchmark.yaml` | call helper |
| `.github/workflows/atom-sglang-benchmark.yaml` | call helper |
| `.github/workflows/atom-vllm-accuracy-validation.yaml` | call helper |
| `.github/workflows/atom-sglang-accuracy-validation.yaml` | call helper |
| `.github/workflows/atom-test.yaml` | call helper |

## Test plan

- [x] `bash -n .github/scripts/collect_gpu_info.sh` — syntax OK
- [x] Smoke test on a host with **no** GPU tools — emits `gpu_name=unknown / gpu_vram_gb=0 / rocm_version=unknown` cleanly (no errors)
- [x] Smoke test with a **mocked `amd-smi`** returning MI355X — produces `gpu_name=AMD Instinct MI355X`, `gpu_vram_gb=284`
- [x] Smoke test with **only `rocm-smi`** mocked (legacy MI300X path) — produces `gpu_name=Instinct MI300X`, `gpu_vram_gb=192`
- [ ] Trigger an `atom-benchmark` / `atom-vllm-benchmark` run on an MI355X self-hosted runner after merge and confirm the new dashboard entry shows `MI355X` instead of `Radeon Graphics`
- [ ] Confirm legacy MI300X / MI250X dashboard entries still render the correct product name (fallback path)

## Notes / follow-ups

- This PR intentionally only touches the dashboard-facing GPU **identification** path. The other `rocm-smi` calls in these workflows (`--showmemuse`, `--showpidgpus`, `--showtemp`, `--showid`) are runtime monitoring — they don't affect the dashboard label and migrating them to `amd-smi` can be a follow-up.
- Long-term `rocm-smi` should be patched to recognize MI355X too; @Mrinal Karvir is following up with the ROCm team on that in parallel.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/.github/scripts/collect_gpu_info.sh
+++ b/.github/scripts/collect_gpu_info.sh
@@ -2,20 +2,28 @@
 # Collect GPU info from a (running) container or the local host and emit
 # `gpu_name`, `gpu_vram_gb`, and `rocm_version` to $GITHUB_OUTPUT (when set).
 #
-# Prefers `amd-smi` because it correctly identifies recent ASICs (e.g. MI355X)
-# whose marketing name is missing from `rocm-smi`'s product table. Falls back
-# to `rocm-smi` and `rocminfo` for older container images that do not yet ship
-# `amd-smi`.
+# Probing order for the GPU marketing name:
+#   1. `amd-smi static --asic` MARKET_NAME
+#   2. `rocm-smi --showproductname` Card Series
+#   3. `rocminfo` Marketing Name
+#   4. <runner_hint> pattern match (mi355 / mi325 / mi300 / mi250)
+#
+# Step 4 is needed because on freshly-released ASICs (currently MI355X) every
+# in-container SMI tool can still report "Radeon Graphics" until the
+# marketing-name table is patched. The CI runner name is operator-asserted
+# and reliable, so we use it as the final tie-breaker for the dashboard label.
 #
 # Usage:
-#   collect_gpu_info.sh                              # run directly on the host
-#   collect_gpu_info.sh <container>                  # docker exec <container>
-#   collect_gpu_info.sh <container> <engine>         # custom engine, e.g. podman
+#   collect_gpu_info.sh                                          # local host
+#   collect_gpu_info.sh <container>                              # docker exec
+#   collect_gpu_info.sh <container> <engine>                     # custom engine
+#   collect_gpu_info.sh <container> <engine> <runner_hint>       # + runner hint
 
 set -uo pipefail
 
 CONTAINER="${1:-}"
 ENGINE="${2:-docker}"
+RUNNER_HINT="${3:-${RUNNER_HINT:-}}"
 
 if [ -n "$CONTAINER" ]; then
     exec_in() { "$ENGINE" exec "$CONTAINER" bash -lc "$1" 2>/dev/null; }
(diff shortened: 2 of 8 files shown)


ChuanLi1101:
### Update — pushed `97e21ce0` to address what the first run revealed

The first PR run surfaced two things:

**1. The accuracy CI failures are pre-existing, not caused by this PR.**

Spot-checked four other recent PRs on completely unrelated branches; the same `Accuracy (...)` jobs fail on all of them, and the `ATOM Test` workflow has been red on every scheduled `main` run for the past 6+ days:

| Branch | Failing accuracy jobs (sample) |
| --- | --- |
| `main` (nightly, 5/1–5/5) | `Model Accuracy (...)` × 5–8 every night |
| `guanbao/fix_kimi_acc` | `Accuracy (Qwen3.5-35B-A3B-FP8 TP2)`, `Accuracy (gpt-oss-120b TP1)` |
| `mooncake_pd` | `Accuracy (Qwen3.5-35B-A3B-FP8 TP2)` |
| `ganyi/remove_chunk_split` | `Accuracy (Kimi-K2-Thinking-MXFP4 TP4)`, `Accuracy (Qwen3.5-35B-A3B-FP8 TP2)` |
| `support_kimi_draft`, `sijyang/verl_dev`, `ci/dockerfile-triton-kernels` | failures |

Underlying error signatures from the logs are runtime / model-layer issues, not CI plumbing:

- `[AITER] hipIpcOpenMemHandle ... [HIP error](invalid device pointer)` — DeepSeek-R1-FP8 TP8
- `Cannot send a request, as the client has been closed` — Kimi-K2-Thinking-MXFP4 TP4
- `triton.compiler.errors.CompilationError` — Qwen3-Next-80B-A3B-Thinking on `atom-mi355-8gpu`

The `Collect GPU info` step itself ran cleanly on every job. On the MI355X runner it produced:

```
GPU: AMD Radeon Graphics, VRAM: 288GB, ROCm: 7.2.3
```

…which leads to the second observation:

**2. `amd-smi` alone doesn't actually fix the MI355X label on these hosts.**

Even on `ROCm 7.2.3` the in-container `amd-smi static --asic` reports `MARKET_NAME: AMD Radeon Graphics` for MI355X — same gap as `rocm-smi` and `rocminfo`. So the original PR (commit `8fb33be5`) didn't fully solve the dashboard problem.

`97e21ce0` fixes this by adding a **4th-tier runner-name hint** fallback: if all three SMI tools come back empty or with `Radeon Graphics`, the script consults the GitHub matrix runner expression (already passed as the 3rd argument, e.g. `atom-mi355-8gpu.predownload`) and pattern-matches the chip family:

```
*mi355*  -> "AMD Instinct MI355X"
*mi325*  -> "AMD Instinct MI325X"
*mi300x* -> "AMD Instinct MI300X"
*mi250x* -> "AMD Instinct MI250X"
*mi210*  -> "AMD Instinct MI210"
```

The hint is **only** applied when the SMI tools fail; if `amd-smi` reports a real marketing name it wins and the hint is ignored (verified locally with mocked SMI binaries).

So: please don't be alarmed by the red `Accuracy (...)` jobs — they are pre-existing on `main` and will be addressed separately. The actual change in this PR is small, contained to the `Collect GPU info` step, and is now provably enough to make the dashboard render `MI355X` correctly on the affected runners.

valarLip (approved):

PR #749, opened 2026-05-11
Add Mistral-3-8B + Qwen3-8B-FP8 + native triton attention backend for gfx1201 (RDNA4 / RX 9070 XT)

carlushuang:
## Summary

Bring up two natively-FP8 8B models on a single **AMD RX 9070 XT (gfx1201, RDNA4, 16 GB)**: Mistral's **Ministral-3-8B-Instruct-2512** (per-Tensor FP8) and Qwen's **Qwen3-8B-FP8** (block-128 FP8). Same all-Triton stack, cudagraph on, no torch fallback.

- New `Mistral3` text-only model (strips Pixtral vision tower).
- New `NativeTritonBackend` — JIT triton kernels in place of AITER's prebuilt HIP `.so`s (gfx1201 has no prebuilt code objects). Auto-routed on gfx1201; opt-in elsewhere via `ATOM_NATIVE_TRITON_ATTN=1`.
- CUDAGraph-correct decode at all bs (incl. fix for NaN-from-padding in `prepare_decode` that was silently producing wrong logits when `scheduled_bs < captured_bs`).
- Triton-only path — no torch fallback (every fallback we removed contained `.item()` / `.cpu()` syncs that would silently break cudagraph capture on ROCm).
- Per-shape `gemm_a8w8` configs (`GROUP_SIZE_M=1`) + per-token FP8 quant (1 kernel, no atomic).
- **Qwen3-8B-FP8 (block-128) support** — three-line patch routes `weight_block_size=[128,128]` through aiter's triton `gemm_a16w8_blockscale` (PREQUANT=False), which casts FP8 weight → BF16 inside the kernel and runs `tl.dot(bf16, bf16)`. Sidesteps the fact that Triton on this gfx1201 build does not implement `tl.dot(fp8, fp8)`.

## Side-by-side: Ministral-3-8B vs Qwen3-8B-FP8

Same env, same flags, GPU 1, BF16 KV, cudagraph, conc=1, OSL=256:

| | Ministral-3-8B (per-Tensor FP8) | **Qwen3-8B-FP8 (block-128)** |
|---|---:|---:|
| Quant scheme | per-Tensor + static activation | **block-128 + dynamic activation** |
| Triton kernel | aiter `gemm_a8w8` (per-tensor triton) | **aiter `gemm_a16w8_blockscale` (PREQUANT=False)** |
| TPOT | **22.0 ms** | **21.6 ms** |
| Output tok/s | 43.2 | 44.7 |
| TTFT (ISL≈1k) | 280 ms (ISL=1076) | 185 ms (ISL=549, Qwen3 tokenizer is denser) |
| gsm8k 5-shot flex (n=50) | 0.72 ± 0.064 | **0.86 ± 0.050** |
| gsm8k 5-shot flex (n=200, recipe) | 0.815 | 0.86 (n=50) |
| Cudagraph capture | all 11 batch sizes | all 11 batch sizes |
| Chat template | strict alternation (breaks OpenClaw etc.) | **lenient + native tool calling** |
| VRAM (cudagraph) | ~13.5 GB | ~14 GB |

Both models hit ~22 ms TPOT (≈ same memory roofline, similar 8B shape). Qwen3 is the recommended agent-stack backend — beats Mistral on accuracy (+5%) and accepts multi-system / tool-call payloads.

## Performance + accuracy at ISL/OSL = 1024 / 1024 (Ministral-3-8B detail)

Cudagraph default capture set `[1,2,4,8,16,32,48,64,128,256,512]`,
BF16 KV, `--max-model-len 4096`, `--gpu-memory-utilization 0.85`,
single GPU:

| conc | TTFT (ms) | TPOT (ms) | Out tok/s | Total tok/s | gsm8k strict / flex (n=200) |
|---:|---:|---:|---:|---:|:---:|
| **1** | 172 | **21.9** | 45 | 81 | — |
| 2 | 222 | 22.6 | 85 | 220 | **0.765 / 0.765** |
| 4 | 310 | 22.8 | 119 | 262 | **0.780 / 0.785** |
| 8 | 502 | 24.3 | 276 | 510 | — |
| 16 | 1,251 | 28.1 | 355 | 795 | **0.715 / 0.725** |
| 32 | 12,323 | 42.8 | 398 | 789 | **0.735 / 0.740** |
| 64 | 36,675 | 40.4 | 418 | 849 | — |
| 128 | 99,119 | 42.9 | 440 | 870 | — |
| 256 | 195,066 | 41.5 | **440** | **916** | — |
| 512¹ | 26,398 | 45.8 | 413 | 819 | — |

¹ conc=512 with num_prompts=64 instead of 1024 (per-test timeout cap).

- **Eager baseline**: 0.785 / 0.785. All cudagraph results within ±0.030 stderr.
- **No OOM at any conc 1..512** — paged KV manager queues sequences when in-flight cap is hit.
- **In-flight cap = 7 active sequences at 1k/1k**: KV pool (941 blocks × 16 tokens = 15k slots) ÷ 128 blocks/seq → 7. Past conc=8 the engine queues; aggregate output throughput plateaus at ~440 tok/s while TTFT inflates with queue depth.
- **TPOT @ conc=1**: 21.9 ms = 45.6 tok/s = **53% of the 86 tok/s memory roofline** (8 GB FP8 weights ÷ 640 GB/s). Beats published llama.cpp Q4 numbers on the same GPU (~30–50 tok/s bs=1) despite reading 2× as much weight per step (per-byte ~2× more efficient).

## Qwen3-8B-FP8 details

Three small additions to enable `weight_block_size=[128,128]` on gfx1201 — all Triton, no torch reference:

1. **`atom/quant_spec.py`** — `weight_block_size=[128,128]` now maps to `QuantType.per_1x128` (the `per_128x128` enum has zero consumers in `linear.py`; the existing `per_1x128` path already allocates the right `(out//128, in//128)` scale-grid).
2. **`atom/model_ops/linear.py`** — new gfx1201 branch in the `per_1x128` dispatch that calls aiter's Triton `gemm_a16w8_blockscale` (PREQUANT=False) with a custom config (BLOCK_N=64 to fit gfx1201's 64 KiB shared mem; the shipped `gfx1201-GEMM-A16W8_BLOCKSCALE.json` picks 256 and OOMs). Disables `shuffle_weights()` for `per_1x128` on gfx1201 since this kernel wants the plain `(N, K)` layout.
3. **`recipes/Qwen3-8B-FP8.md`** — serve cmd, env vars, perf+accuracy table, side-by-side, debug journey notes.

Why `gemm_a16w8_blockscale` and not the obvious `gemm_a8w8_blockscale_preshuffle`? Triton on this gfx1201 build does not implement `tl.dot(fp8, fp8)` — the assert `only int8 supported!` fires for FP8 lhs. The a16w8 path casts FP8 weight to BF16 inside the kernel, then `tl.dot(bf16, bf16)` (which is supported). Weight stays FP8 in DRAM, x stays BF16, no activation quant needed.

Three gotchas worth flagging for reviewers:
- `aiter.utility.dtypes.d_dtypes['fp8'] == torch.uint8` — FP8 weights are stored as raw uint8 bytes with e4m3fn semantics. Always `.view(torch.float8_e4m3fn)` before passing to a kernel that does `b.to(bf16)`, or `.to(float32)` decodes byte values 0–255 as integers and you get garbage outputs that *look* numerically reasonable.
- For `per_1x128` on gfx1201, weight preshuffle MUST stay off — the a16w8 kernel wants plain `(N, K)` layout (not the `(N//16, K*16)` of the preshuffle GEMM).
- Shipped `gfx1201-GEMM-A16W8_BLOCKSCALE.json` config picks `BLOCK_N=256` → ~98 KiB shared mem → JIT-fails. Override at the call site with `BLOCK_N=64` → ~57 KiB.

## Required setup (reproducibility note)

Reviewers reproducing these numbers must run `bash scripts/gfx1201/setup_aiter_configs.sh` once after starting the container. **aiter ships zero gfx1201 GEMM tuned configs** as of `rocm/atom-dev:latest` digest `sha256:b704d9a8...`; without aliasing the gfx1250 ones to gfx1201 the autotuner falls back to a default that is **30–50% slower at 8B-class shapes** — verified end-to-end on Mistral-3-8B (32.5 ms TPOT without the script, 22.0 ms with). Qwen3 is unaffected because we override its `gemm_a16w8_blockscale` config in code. Both recipes flag this as a required step. The script is idempotent and creates 24 symlinks in `/app/aiter-test/aiter/ops/triton/configs/gemm/`.

## Test plan

- gsm8k 5-shot, n=200, conc=2/4/16/32 on Ministral-3: all within ±0.030 of the 0.785 eager baseline.
- gsm8k 5-shot, n=50 on Qwen3-8B-FP8: 0.86 / 0.86 (eager + cudagraph both).
- 1485-token prompt, 128-token decode: TPOT 0.023 s/tok on Mistral (5% degradation for ~300× more context — matches KV-bandwidth roofline).
- GPU-0 vs GPU-1 perf parity verified.
- Per-kernel + 36-layer-chained cudagraph standalone capture-replay pass bitwise.
- Determinism check: 3 back-to-back curl batches give bitwise identical outputs.
- `black --check` and `ruff check` clean on changed files.
- **End-to-end reproducibility**: spun up a fresh `rocm/atom-dev:latest` container with no caches, ran the setup script, and re-ran both recipes. Mistral-3 TPOT reproduced at 22.0 ms (3 runs), Qwen3-8B-FP8 TPOT reproduced at 21.7 ms, gsm8k accuracies identical (0.72 / 0.86 respectively).

## Known caveats

- **TP=2 blocked at host kernel level**: HIP IPC needs `iommu=pt amd_iommu=on` on the GRUB cmdline. Both RCCL/PyNccl AND aiter CustomAllreduce fail on the same root cause. Fix is host-side (reboot). Once unblocked, TP=2 lets the BF16 8B Reasoning variant fit (16.6 GB → 8.3 GB / GPU).
- **FP8 KV cache TODO** — `_kv_dtype()` hardcoded BF16. At our contexts (~700 tok gsm8k), FP8 KV would save < 1% TPOT; becomes meaningful past ~32k context.
- 238 `activation_scale` checkpoint tensors silently dropped at load (Mistral-3 ships `activation_scheme: "static"` but 0 actual `input_scale` tensors in the safetensors, so the static fast path isn't reachable for this checkpoint anyway).
- **Qwen3 a16w8 BLOCK_N=64 is a correctness-first config** — there's likely a better-tuned config for gfx1201 that fits 64 KiB shared mem. Closing this gap (the bench shows we already match Ministral's 22 ms TPOT, so the headroom is small) is a follow-up.

## Note on diff size

Branch is 34 commits ahead of merge-base; real changes are ~13 files / +1.9k lines. Diff vs current `origin/main` looks larger because main has drifted; happy to rebase if reviewers prefer a cleaner diff. Per-model recipes in `recipes/Ministral-3-8B.md` and `recipes/Qwen3-8B-FP8.md`.


chuanbowang2026:
I also tried an experimental M=1 decode-specialized GEMM path, but did not include it in this PR. It used two custom Triton kernels for single-token decode: one FP8xFP8 path after per-token activation quantization, and one BF16xFP8 path that skips activation quantization. In local RX 9070 XT testing on Ministral-3-8B with 1024 input / 256 output tokens, neither path beat the tuned default aiter gemm_a8w8 path: default was ~18.38 ms TPOT, FP8xFP8 was ~19.34 ms, and BF16xFP8 was ~18.46 ms.

I removed these experimental kernels to keep this PR smaller and lower-risk. 
**In short, the M=1 experimental kernel did not produce a successful optimization. I plan to open a separate PR later to specifically optimize the decode batch=1 GEMM path.**

PR #789, opened 2026-05-14
fix(openai): harden chat request handling

san-tian:
## Summary
- Normalize OpenAI JSON-string tool call arguments before chat template rendering.
- Decode streaming responses with cumulative token context so partial UTF-8/token-boundary output does not emit replacement characters.
- Reject requests whose prompt length or requested prompt+generation length exceeds `max_model_len` before they reach scheduling/model-runner internals.

## Tests
- `python -m pytest -q -p no:cacheprovider tests/entrypoints/test_protocol.py` in `rocm/atom-dev:latest` on the AMD host: 27 passed.
- Compile check for `atom/entrypoints/openai/protocol.py`, `atom/entrypoints/openai/api_server.py`, `atom/model_engine/llm_engine.py`, and `tests/entrypoints/test_protocol.py` in `rocm/atom-dev:latest` on the AMD host.
- GLM-5.1 service smoke on AMD host: normal completion returns `5`; oversized prompt above `max_model_len` returns HTTP 400 instead of being scheduled.


valarLip:
could you please fix the Code Style issue

carlushuang:
@san-tian  thank you for contribute. Can you please elaborate more on what cases do you see this PR needed, is some bug, or some deployment hang?
And please address the copilot 's comment and fix the format with `black` (please see CI)

san-tian:
@carlushuang This PR includes two features:

1. **Tool parser**: adds support for function calling in the OpenAI format.
2. **Detokenizer for streaming requests**: fixes an issue where vanilla streaming output directly converts tokens into text, which can produce invalid characters when a token represents only part of a character.

valarLip:
please pay some attention on the format..

san-tian:
@valarLip thank u for your advise.. I will do it

PR #805, opened 2026-05-15
[vLLM-ATOM] Fix GLM-4.7 MTP in vLLM plugin

kliuae:
## Motivation
In the current vLLM-ATOM upstream, GLM-4.7 crashes with arguments like lm_head and embed_tokens not found and conflicting compile cache key, when MTP is enabled. This PR adds fixes to MTP to address these issues.

## Technical Details
- Expose speculative decode attributes, which is fine for qwen3next mtp too as its lm_head lives in model, not inner
- Mask inputs for glm4-mtp
- Separate main model and draft model configs so the draft model doesn't load from the main model's compiled artifact

## Test Plan
`lm_eval` with gsm8k

## Test Result
Server command
```
ATOM_ENABLE_QK_NORM_ROPE_CACHE_QUANT_FUSION=1 \
AITER_QUICK_REDUCE_QUANTIZATION=INT4 \
SAFETENSORS_FAST_GPU=1 \
ATOM_DISABLE_VLLM_PLUGIN=0 \
ATOM_DISABLE_VLLM_PLUGIN_ATTENTION=0 \
VLLM_LOGGING_LEVEL=INFO \
VLLM_USE_V1=1 \
VLLM_ROCM_USE_AITER=1 \
vllm serve zai-org/GLM-4.7-FP8 \
  -tp 4 \
  --gpu-memory-utilization 0.9 \
  --no-enable-prefix-caching \
  --disable-uvicorn-access-log \
  --trust-remote-code \
  --kv-cache-dtype fp8 \
  --speculative-config.method mtp \
  --speculative-config.num_speculative_tokens 1
```

Eval command
```
lm_eval --model local-completions   --model_args model=zai-org/GLM-4.7-FP8,base_url=http://localhost:8000/v1/completions,num_concurrent=64,tokenized_requests=False  --tasks gsm8k --num_fewshot 5
```

Eval score
|gsm8k|      3|flexible-extract|     5|exact_match|_  |0.9462|_  |0.0062|
|     |       |strict-match    |     5|exact_match|_  |0.9454|_  |0.0063|

## Submission Checklist


atom/plugin/vllm/model_wrapper.py
```diff
@@ -155,7 +156,12 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
         model_arch = _select_model_arch(vllm_config)
         self.is_mtp_draft_model = self.is_mtp and model_arch != main_model_arch
         if self.is_mtp_draft_model:
-            self.atom_config = get_current_atom_config()
```
ganyi1996ppo, line 158: Can't we use the same config build from the main model? I mean vllm config should already contains the two config

kliuae: I think we need to build the draft config separately as the draft model can `@support_torch_compile` and need its own compilation config. If we share them the draft model would try to load from the main model's compiled cache, and in the case of glm4, this leads to cache key errors.

ganyi1996ppo: `generate_atom_config_for_plugin_mode` will override the global atom config, which might corrupt the models that stores the layer into it during intialization phase. I met that issue in Qwen3Next, so I add this condition branch to prevent this issue happens. Is it possible to build the spec config separately for draft model?

kliuae: I see. I think building only the speculative config would not suffice here as the violated compilation config is one level above. I've tried calling the intermediate config builder `_generate_atom_config_from_vllm_config` that doesn't alter atom_config, but it can throw framework issues `AssertionError: VllmBackend can only be called once`.

Despite writing to the global atom config, this `if` condition retains `hf_config` from the old atom config and does not override it so that changes happen during models' __init__ can be carried over. May I know what were the configs corrupted in the qwen3next case? Maybe we can do something similar for those configs that could get corrupted.

whx-sjtu: I agree with @kliuae and also curious about what were the configs corrupted in the qwen3next case. Please provide detailed error logs @ganyi1996ppo . Actually I think main model and draft model should have different atom configs containing different compilation configs, because at least they should have their separate `static_forward_context` in compilation config, otherwise we can't get module by layer name here: https://github.com/ROCm/ATOM/blob/6619cc7d33e3bb1f5a8f2f9966b2c4cb3a20e494/atom/plugin/attention.py#L2467

wuhuikx: @ganyi1996ppo  any further comment?

zejunchen-zejun (dismissed):

valarLip (dismissed):

wuhuikx:
can we add this to recipe/nightly/benchmark. The label should be AW P0.  @junyyang-amd  cc @zejunchen-zejun

PR #819, opened 2026-05-18
[ATOM SGL] Qwen3.5 mha support

ZhiweiYan-96:
## Motivation

This PR adds a Qwen3.5 head_dim=256 dense MHA path for the ATOM SGLang plugin, mirroring upstream SGLang Aiter dense attention behavior.

Adds an upstream-mimic style path for non-MLA full-attention layers with head_dim/qk_head_dim/v_head_dim == 256.
1. Uses raw dense KV cache layout with mha_batch_prefill_func for extend and paged_attention_ragged for decode.
2. For avoiding unnecessary kv cache cast copy. Keeps FP8 KV decode efficient by passing the FP8 KV cache dtype to paged_attention_ragged instead of upcasting the full KV cache.

# Accuracy test

image


ZhiweiYan-96:
@zhuyuhua-v  @qichu-yun  @wuhuikx @wanzhenchn   mind take a look?

ZLkanyo009 (dismissed):

valarLip (approved):

PR #862, opened 2026-05-21
[atom-vllm benchmark] remove quick allreduce for dsv3.2

zejunchen-zejun:


wuhuikx:
Please change the nightly/benchmark/recipe at the same time.

PR #873, opened 2026-05-21
remove tensor contiguous before indexer_qk_rope_quant_and_cache_kernel

XiaobingSuper:
## Motivation

indexer_qk_rope_quant_and_cache_kernel will support non-contiguous k input, which reduce tensor memory copy operations. see https://github.com/ROCm/aiter/pull/3301.

## Test Result

before:
image

after:

image

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -1051,6 +1051,8 @@ def sparse_attn_indexer(
         weights = weights_out
     else:
         q_fp8 = q_input
+        k = k.contiguous()
+        weights = weights.contiguous()
         indexer_k_quant_and_cache(
             k,
             kv_cache,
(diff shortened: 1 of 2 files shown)


atom/models/deepseek_v2.py
```diff
@@ -1416,10 +1418,10 @@ def forward(
         else:
             k = self.wk(hidden_states)
             weights = self.weights_proj(hidden_states)
-        k = k.contiguous()
-        weights = weights.contiguous()
 
         if not self.use_qk_rope_cache_fusion:
+            k = k.contiguous()
```
valarLip: so... "use_wk_weights_proj_fusion" introduced contiguous() for both "k" and "weights", feel like there is no gain from this fusion

valarLip (changes requested):

atom/models/deepseek_v2.py
XiaobingSuper: Updated. We only do wk_weights_proj_fusion when use_qk_rope_cache_fusion=True now.

valarLip (approved):

PR #882, opened 2026-05-22
Fix Eagle3 prefix cache hidden state flow

yhl-amd:
## Summary
- Keep Eagle3 draft MHA block tables aligned with target metadata during prefix-cache decode.
- Move Eagle3 draft normalization back into `compute_logits` so the draft loop passes prenorm hidden states between steps.

## Test plan
- ReadLints on `atom/models/eagle3_llama.py` and `atom/spec_decode/eagle.py`.
- Python syntax compile for the same files.

### DIFF
--- a/atom/spec_decode/eagle.py
+++ b/atom/spec_decode/eagle.py
@@ -372,16 +372,14 @@ class EagleProposer:
 
         for i in range(self.mtp_k):
             with record_function(f"draft[{i}/{self.mtp_k} bs={bs}]"):
-                model_output = self.model(
+                ret_hidden_states = self.model(
                     input_ids=input_ids,
                     positions=positions,
                     hidden_states=hidden_states,
                 )
 
-                sample_hidden_states = (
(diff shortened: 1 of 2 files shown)


atom/spec_decode/eagle.py
```diff
@@ -378,19 +377,11 @@ def propose(
...
-                if draft_uses_mha:
-                    ret_hidden_states, ret_hidden_prenorm = model_output
-                else:
-                    ret_hidden_states = model_output
-                    ret_hidden_prenorm = None
 
                 sample_hidden_states = (
-                    torch.index_select(ret_hidden_states, 0, last_token_indices)
+                    torch.index_select(model_output, 0, last_token_indices)
                     if i == 0
```
valarLip, line 383: keep the name ret_hidden_states, and remove "if i ==0"...

valarLip (changes requested):

valarLip (approved):

PR #904, opened 2026-05-25
ci: preserve runner groups in docker containers for atom-vllm and atom-sglang

junyyang-amd:
## Motivation

Resolve the error when run jobs with  runner ‘mia1-p02-g47’.
image

### DIFF
--- a/.github/scripts/gpu_load_test.sh
+++ b/.github/scripts/gpu_load_test.sh
@@ -63,7 +63,6 @@ docker run \
   --device=/dev/kfd \
   --device=/dev/dri \
   --group-add video \
-  --group-add keep-groups \
   --cap-add=SYS_PTRACE \
   --security-opt seccomp=unconfined \
   -v /data:/data \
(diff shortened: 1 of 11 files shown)


zejunchen-zejun (dismissed):

zejunchen-zejun:
Hi @junyyang-amd 
Could you help confirm if this argument can be get by docker or podman?
I remember this argument belongs to the podman, for docker, there is no such argument

gyohuangxin (changes requested):
It will fail if there is no keep-groups group on the runner: https://github.com/ROCm/ATOM/actions/runs/26387852575/job/77670575701?pr=904#step:11:44

junyyang-amd:
> It will fail if there is no keep-groups group on the runner: https://github.com/ROCm/ATOM/actions/runs/26387852575/job/77670575701?pr=904#step:11:44

Yes, I use a flag for '--group-add', if it is podman, --group-add=keep-groups, else it will be gid.  @zejunchen-zejun

zejunchen-zejun (approved):

PR #933, opened 2026-05-26
[atom-vllm MTP] remove seq_len D2H for spec decode

whx-sjtu:
## Summary
Remove the seq_len device-to-host sync for spec decoding draft model (MHA/MQA, pure decode scenario).

## Test plan
Before:
image

After:
image

### DIFF
--- a/atom/plugin/attention.py
+++ b/atom/plugin/attention.py
@@ -378,6 +378,10 @@ class vllmAttentionMetadataBuilderMethods:
         decode_only = num_decodes > 0 and num_extends == 0 and num_prefills == 0
         mixed = not (prefill_only or decode_only)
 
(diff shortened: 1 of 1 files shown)


wuhuikx:
Could you please help show the accuracy results for both MTP and non-MTP?

whx-sjtu:
> Could you please help show the accuracy results for both MTP and non-MTP?

I found that the baseline accuracy is around 85% which is not good. And my PR didn't fix accuracy issue, it is just a performance optimization.

zejunchen-zejun (approved):

zejunchen-zejun:
This PR makes sense, it provided the `build_for_drafting` fast pass method to vllm for building attn metadata for draft model.
BTW @whx-sjtu , this PR will not affect the acc and accept ratio right? It is a pure perf optimization

ganyi1996ppo (approved):

PR #939, opened 2026-05-27
scheduler: fix dead HBM prefix cache under deferred-output mode (regression from #902)

yhl-amd:
## Summary

Restores HBM prefix-cache hit-path for the default deployment (deferred-output mode). Regression introduced by #902.

- #902 moved hash registration from `BlockManager.allocate()` into `BlockManager.hash_blocks()` called from `Scheduler.postprocess` under the gate `if seq.type == SequenceType.PREFILL`.
- The motivation was correct (chunked prefill requires deferring registration until KV is computed), but the new gate is incompatible with deferred-output mode, which is the default — `tokenIDProcessor.is_deferred_out = True` is hardcoded in `model_runner.py:89`.
- Under deferred mode, the prefill step's `postprocess` sees `idx is None` and skips the seq; by the time the prefill output surfaces one step later, the next `schedule()` has already flipped `seq.type` to `DECODE`. The PREFILL gate never fires → `hash_to_block_id` stays empty for prompt blocks → HBM prefix cache silently dead.
- Fix: gate on a per-seq one-shot flag `Sequence.prefix_hashes_published`, set on the first postprocess pass that has an `idx` for the seq (works in both eager and deferred modes).
- Also subtracts `num_placeholder` from `num_new` when `need_placeholder` is active, so partial-block hashes aren't minted from mixed prompt+placeholder content.

## Reproduction

MiniMax-M2.5, TP=2 on MI300X, fp8 KV cache (default deferred-output mode):

```bash
  --model /shared_models/MiniMax-M2.5 --kv_cache_dtype fp8 -tp 2 \
  --trust-remote-code --gpu-memory-utilization 0.78
```

Send the same 360-token prompt 3 times. Grep `Scheduled prefill batch` in the server log.

**Before this PR (origin/main @ 3ceb3f4):**
```
Scheduled prefill batch: 1 reqs, 360 new tokens (cached: [0], new: [360]), req_ids: (0,)
Scheduled prefill batch: 1 reqs, 360 new tokens (cached: [0], new: [360]), req_ids: (1,)
Scheduled prefill batch: 1 reqs, 360 new tokens (cached: [0], new: [360]), req_ids: (2,)
```
Every identical-prompt repeat re-prefills all 360 tokens.

**After this PR:**
```
Scheduled prefill batch: 1 reqs, 360 new tokens (cached: [0],   new: [360]), req_ids: (0,)
Scheduled prefill batch: 1 reqs,  12 new tokens (cached: [288], new: [12]),  req_ids: (1,)
Scheduled prefill batch: 1 reqs,  12 new tokens (cached: [288], new: [12]),  req_ids: (2,)
```
288 cached tokens = 18 blocks × 16 tokens/block, exactly the full-block prefix of the 360-token prompt.

## Why #902 CI did not catch this

- The new tests in `tests/test_block_manager.py` from #902 call `bm.hash_blocks(...)` directly, bypassing scheduler/postprocess/deferred-output entirely.
- End-to-end serving regressions for short-prompt benchmarks (GSM8K, throughput sweeps with unique prompts) do not surface the issue. At 360 tokens on MI300X TP=2 the prefill compute is dominated by per-request scheduling overhead — TTFT differences from cache hits are within noise.
- Becomes user-visible on long-context workloads (8K+ token prompts) where prefill compute dominates, and especially on agentic workloads with shared prefixes (e.g. `kv-cache-tester` 100K-context traces).

## Test plan

- [x] black + ruff pass on changed files
- [x] Live server reproduction (above): `cached:[0,0,0]` → `cached:[0,288,288]` confirmed
- [ ] Existing `tests/test_block_manager.py` continues to pass (they call `hash_blocks` directly, unaffected by gate change)
- [ ] CI

### DIFF
--- a/atom/model_engine/scheduler.py
+++ b/atom/model_engine/scheduler.py
@@ -878,12 +878,31 @@ class Scheduler:
             # Register prefix-cache hashes for blocks the prefill step just
             # finalized. Deferred from BlockManager.allocate() so a hash is
             # only published after the block's KV has actually been computed
-            # by the forward. Must run before any seq state update so
-            # num_cached_tokens and block_table still reflect the pre-step view.
-            if seq.type == SequenceType.PREFILL:
-                self.block_manager.hash_blocks(
-                    seq, seq.num_tokens - seq.num_cached_tokens
-                )
+            # by the forward — keeps the block manager correct under future
+            # chunked-prefill scheduling where one block may span multiple
+            # steps. Must run before any seq state update so num_cached_tokens
+            # and block_table still reflect the pre-step view.
+            #
+            # Gate is `not prefix_hashes_published`, not `seq.type ==
+            # PREFILL`: ModelRunner runs in deferred-output mode by default
+            # (tokenIDProcessor.is_deferred_out), so the prefill step's
+            # postprocess sees idx=None and skips this seq (above). By the
+            # time the prefill output surfaces, the next step's schedule
+            # has already flipped seq.type to DECODE — the old PREFILL
+            # gate never fires and `hash_to_block_id` stays empty for
+            # prompt blocks (HBM prefix cache silently dead). The flag
+            # gate fires once per seq at the first postprocess with idx.
+            #
+            # `num_new` subtracts `num_placeholder` when deferred-output is
+            # active: those slots are filled with the real prefill output
+            # later in this loop, so they're not part of the prompt hash
+            # chain — leaving them in would mint a stale partial-block hash.
+            if not seq.prefix_hashes_published:
+                _num_new = seq.num_tokens - seq.num_cached_tokens
+                if need_placeholder:
+                    _num_new -= num_placeholder
+                self.block_manager.hash_blocks(seq, _num_new)
+                seq.prefix_hashes_published = True
             token_ids = prev_token_ids[idx]
             num_new_token = len(token_ids)
             if is_deferred_out or self.use_spec:
--- a/atom/model_engine/sequence.py
+++ b/atom/model_engine/sequence.py
@@ -84,6 +84,14 @@ class Sequence:
         self.stop_strings = sampling_params.stop_strings
         self.stop_token_sequences = stop_token_sequences or []
         self.is_first_decode = False
+        # Set to True by Scheduler.postprocess after BlockManager.hash_blocks
+        # has registered the prompt blocks for prefix caching. The trigger has
+        # to be per-seq because in deferred-output mode the prefill step's
+        # postprocess has no fwd_output entry for the seq (idx is None) — the
+        # prefill output surfaces one step later, at which point seq.type has
+        # already been flipped to DECODE. A seq.type / len(output_tokens) gate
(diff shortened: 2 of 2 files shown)


yhl-amd:
## Follow-up: 8K-token TTFT comparison

The 360-token reproduction in the PR description proves the mechanism (`cached:[0]` triple), but the TTFT impact is buried under per-request scheduling overhead at that prompt size. Re-ran on the same hardware with a 7820-token prompt to surface the user-visible perf delta. Identical config (MiniMax-M2.5 TP=2, fp8 KV, gpu-memory-utilization 0.78, default deferred-output mode), same prompt sent 3x sequentially, compile cache cleared between branches.

### Scheduler log (mechanism)

**Without this PR (origin/main @ 3ceb3f4):**
```
Scheduled prefill batch: 1 reqs, 7820 new tokens (cached: [0], new: [7820]), req_ids: (0,)
Scheduled prefill batch: 1 reqs, 7820 new tokens (cached: [0], new: [7820]), req_ids: (1,)
Scheduled prefill batch: 1 reqs, 7820 new tokens (cached: [0], new: [7820]), req_ids: (2,)
```
Every identical-prompt repeat re-prefills all 7820 tokens — `hash_to_block_id` stays empty for prompt blocks.

**With this PR:**
```
Scheduled prefill batch: 1 reqs, 7820 new tokens (cached: [0],    new: [7820]), req_ids: (0,)
Scheduled prefill batch: 1 reqs,   12 new tokens (cached: [7808], new: [12]),   req_ids: (1,)
Scheduled prefill batch: 1 reqs,   12 new tokens (cached: [7808], new: [12]),   req_ids: (2,)
```
7808 cached tokens = 488 blocks × 16 tokens/block — the exact full-block prefix of the 7820-token prompt (the trailing 12-token partial block correctly stays uncached).

### TTFT comparison

|         | req 1   | req 2   | req 3       |
|---------|---------|---------|-------------|
| no-fix  | 287 ms  | 271 ms  | 271 ms      |
| with-fix| 290 ms  | 329 ms  | **100 ms**  |

**req 3 steady state: −171 ms, −63%** at 7820 tokens.

### Heads-up on req 2 (with-fix)

req 2 is slower than req 1 despite the cache hit (329 ms vs 290 ms). The likely cause is first-time CUDA-graph capture for the 12-token prefill shape (warmup uses 16384, req 1 hits 7820 — neither matches 12). req 3 reuses the shape captured during req 2 and drops to the true steady state. A benchmark that compares req 1 to req 2 alone would understate the fix's benefit — req 3+ is the right reference.

### Why the 360-token repro showed near-identical TTFT (~70 ms both sides)

At 360 tokens on MI300X TP=2, prefill compute is dominated by per-request scheduling/sampling/serialization overhead. The cache hit saves only a few ms of GPU compute — invisible in TTFT. At 7820 tokens prefill compute dominates and the 171 ms gap appears cleanly. Bug impact scales with prompt length; agentic workloads with shared 100K-context prefixes will see seconds saved per request.

valarLip (approved):

PR #957, opened 2026-05-28
[Bugfix] eagle: fix MTP draft warmup crash on mrope-config target models

yhl-amd:
## Summary

Fix `RuntimeError: Index tensor must have the same number of dimensions as input tensor` raised from `EagleProposer.propose` during MTP draft warmup whenever the target model's HF config carries `mrope_section` in `rope_parameters` (e.g. Qwen3.5-397B-A17B-FP8).

`atom/spec_decode/eagle.py` previously did:

```python
positions = torch.gather(positions, 0, last_token_indices)
```

`torch.gather` requires `input.ndim == index.ndim`. After #831 (mrope path), `context.positions` becomes 2D `[3, num_tokens]` for any model whose rope config contains `mrope_section`, while `last_token_indices` stays 1D `[bs]` → ndim mismatch → crash on MTP warmup.

Fix: switch to ellipsis indexing, which produces the same shape semantics for both 1D and 2D positions.

```python
positions = positions[..., last_token_indices]
```

| `positions.shape`  | `last_token_indices.shape` | result shape | works? |
|---|---|---|---|
| `[num_tokens]`     | `[bs]`                     | `[bs]`       | ✅ (same as before) |
| `[3, num_tokens]`  | `[bs]`                     | `[3, bs]`    | ✅ (mrope) |

Identical to the pre-#254 form (`positions[last_token_indices]`) on 1D inputs.

## Failing CI

https://github.com/ROCm/ATOM/actions/runs/26529268074/job/78141372143

Affected jobs (Qwen3.5-397B-A17B-FP8 MTP3, schedule on 378715d8):

- `Qwen3.5-397B-A17B-FP8 MTP3 (isl=1024 osl=1024 c=4)`
- `Qwen3.5-397B-A17B-FP8 MTP3 (isl=1024 osl=1024 c=8)`
- `Qwen3.5-397B-A17B-FP8 MTP3 (isl=1024 osl=1024 c=16)`
- … (rest cancelled after the first three failed)

Stack trace from the server log:

```
File "/workspace/atom/spec_decode/eagle.py", line 431, in propose
    positions = torch.gather(positions, 0, last_token_indices)
RuntimeError: Index tensor must have the same number of dimensions as input tensor
```

Last green schedule for the same job: 2026-05-22 (`4181e4249`). First red: 2026-05-24 (`d55a233ae`). DeepSeek-V4-Pro MTP3 keeps passing because V4 targets don't opt into mrope.

## Test plan

- [ ] Bench `Qwen3.5-397B-A17B-FP8 MTP3` ISL=1024 OSL=1024 c=4 — server reaches ready, no warmup crash
- [ ] Bench `DeepSeek-V4-Pro MTP3` ISL=1024 OSL=1024 c=4 — no regression (1D path is bit-identical)
- [ ] (Optional) full nightly atom-benchmark MTP matrix

### DIFF
--- a/atom/spec_decode/eagle.py
+++ b/atom/spec_decode/eagle.py
@@ -428,7 +428,14 @@ class EagleProposer:
                             kv_indptr[1 : bs + 1] -= torch.cumsum(
                                 num_reject_tokens, dim=0
                             )
-                        positions = torch.gather(positions, 0, last_token_indices)
+                        # Index along the last dim so this works for both
+                        # 1D text positions [num_tokens] and 2D mrope
+                        # positions [3, num_tokens] (Qwen3.5 sets
+                        # `mrope_section` in `rope_parameters`, which flips
+                        # `use_mrope=True` and makes context.positions 2D).
+                        # `torch.gather` requires input.ndim == index.ndim
+                        # and crashes on the 2D case during MTP warmup.
+                        positions = positions[..., last_token_indices]
                         context.is_prefill = False
 
                     # update metadata


yhl-amd:
## Root cause walkthrough (with a minimal repro)

This regression was introduced by #831 (`add qwen35 multimodel`), which added the mrope detection / 2D positions plumbing but did not update the spec-decode draft path. The interaction is what bit Qwen3.5-397B-A17B-FP8 MTP3.

### Why a *text-only* model trips the mrope path

Qwen3.5-397B-A17B-FP8 ships with this in `config.json` even though it is a pure-text checkpoint (the schema is shared with the multimodal variant):

```jsonc
\"rope_parameters\": {
  \"mrope_interleaved\": true,
  \"mrope_section\": [11, 11, 10],
  \"partial_rotary_factor\": 0.25,
  \"rope_theta\": 10000000,
  \"rope_type\": \"default\"
}
```

After #831, `ModelRunner.__init__` does:

```python
rope_parameters = getattr(self.hf_text_config, \"rope_parameters\", None) or {}
self.use_mrope = \"mrope_section\" in rope_parameters   # True for Qwen3.5
```

so the field-presence check flips `use_mrope=True` on this model, even though no multimodal request will ever be served.

### How that leaks into `EagleProposer.propose`

`atom/model_ops/attentions/backends.py::prepare_prefill` then returns a 2D positions tensor for the whole forward:

```python
mrope_positions = self._build_mrope_prefill_positions(batch)   # [3, num_tokens] when use_mrope
if mrope_positions is not None:
    positions = mrope_positions
else:
    positions = var[\"positions\"].copy_to_gpu(sum_scheduled_tokens)   # [num_tokens]
return attn_metadata, positions
```

That tensor is stashed on `context.positions`, picked up by `ModelRunner.postprocess` → `propose_draft_token_ids` → `EagleProposer.propose(target_positions=…)`. Inside `propose`:

```python
positions = target_positions + 1                                # still 2D for mrope
…
positions = torch.gather(positions, 0, last_token_indices)      # ← crashes
```

`torch.gather` enforces `input.ndim == index.ndim`. `last_token_indices` is always 1D `[bs]`. With `positions=[3, num_tokens]`, PyTorch raises:

```
RuntimeError: Index tensor must have the same number of dimensions as input tensor
```

This fires on the very first `warmup_model()` call because `is_prefill=True` + `mtp_k=3` enters the `if i == 0` branch unconditionally.

### Minimal repro (no model needed)

```python
import torch

num_tokens, bs = 16384, 1

# Pre-#831 (or non-mrope models): 1D positions, gather works
positions_1d = torch.arange(num_tokens)
last_token_indices = torch.tensor([num_tokens - 1])
torch.gather(positions_1d, 0, last_token_indices)
# tensor([16383])  ✅

# Post-#831 with Qwen3.5-style mrope_section config: 2D positions
positions_2d = torch.arange(num_tokens).unsqueeze(0).expand(3, -1).contiguous()
torch.gather(positions_2d, 0, last_token_indices)
# RuntimeError: Index tensor must have the same number of dimensions as input tensor

# This PR: works for both
positions_1d[..., last_token_indices]   # tensor([16383])           [bs]
positions_2d[..., last_token_indices]   # tensor([[16383],[16383],[16383]])  [3, bs]
```

### Why DeepSeek-V4-Pro MTP3 still passes

DeepSeek-V4 (and every other model currently exercised by `atom-benchmark`) does **not** have `mrope_section` in its `rope_parameters`, so `use_mrope=False` and `context.positions` stays 1D. The gather is bit-identical to the new ellipsis-indexing form on 1D inputs, so V4 MTP3 / DSR1 MTP3 / etc. are unaffected by this PR.

### Scope of the surgical fix vs. tightening `use_mrope`

I deliberately kept this PR to the spec-decode call site only. The broader question — *should `use_mrope` flip on schema presence, or only when a multimodal request actually appears?* — is #831's territory and would have a larger blast radius. If maintainers prefer that route, happy to file a follow-up; for now this restores Qwen3.5-FP8 MTP3 on nightly without touching the multimodal path.

PR #963, opened 2026-05-28
[Feat] Fused qknorm + quant for dpsk v2 model

qichu-yun:
## Motivation

Improve DeepSeek V2 performance by enabling fused q/k RMSNorm + quantization for both FP8 and MXFP4 paths.

## Technical Details

This commit enables fuse_qknorm_quant for MXFP4, preserves unshuffled MXFP4 attention projection weights for raw-scale GEMM, and routes SGLang q_b_proj through the proper FP4 GEMM path when fused quantization is used. Existing non-fused and fallback paths remain unchanged.

## Test Result
before：
image

after:
image

## Submission Checklist


atom/plugin/sglang/models/deepseek_mla_forward.py
```diff
@@ -158,6 +166,55 @@ def _fuse_qk_rmsnorm_and_q_quant(
...
+            q_2d.view(torch.uint8),
+            raw_weight.view(torch.uint8),
+            q_scale.view(torch.uint8),
+            raw_weight_scale.view(torch.uint8),
+            torch.bfloat16,
+            y,
+        )
+        return y
+
+    q_scale = fp4_utils.e8m0_shuffle(q_scale.view(torch.float8_e8m0fnu))
```
valarLip, line 171: we should better remove e8m0_shuffle by other fusion

valarLip (changes requested):

atom/plugin/sglang/models/deepseek_mla_forward.py
qichu-yun: This e8m0_shuffle() is only on the fallback path. The native SGLang-style FP4 GEMM path consumes raw q_scale directly and returns before this line, so the optimized path has no post-shuffle. It is only used when we fall back to ATOM q_b_proj, which still requires shuffled scale layout. Removing it without removing/replacing the fallback would reintroduce the scale-layout correctness issue.

valarLip (dismissed):

valarLip (dismissed):

valarLip (approved):

PR #1008, opened 2026-06-01
[atom-sgl-benchmark] Release mi308 benchmark results to dashboard

junyyang-amd:
## Motivation

The atom-sgl-benchmark results for the MI308 will be automatically pushed to the dashboard. A machine filtering option has been added to the dashboard page, allowing users to filter by machine type.


.github/workflows/atom-sglang-benchmark.yaml
Yuechguo, line 1842: The previous data was uploaded in data.js format. After this modification, can forward compatibility of the data be guaranteed?Can you run the dashboard locally and fully to see if the data is correct?

junyyang-amd: Yes. This change keeps the existing `benchmark-dashboard/data.js` path and schema unchanged. The dashboard still loads legacy data through `window.BENCHMARK_DATA` from `data.js`; the modification only adds hardware/GPU metadata parsing and filtering on top of the existing format. Existing data without the new GPU metadata falls back to `unknown`, so old `data.js` entries remain loadable.

I also validated it locally with the live published legacy `data.js` from `https://rocm.github.io/ATOM/benchmark-dashboard/data.js`. The dashboard assets were served locally, `index.html` and `data.js` loaded successfully, and the parser handled 90 runs / 19,920 benchmark rows into 713 config keys without schema issues. I also checked an MI308-shaped legacy row and it is classified as `mi308x` while preserving throughput, latency, `_gpu_count`, and `_tp` values.

junyyang-amd:
Added GPU selection options (MI355X and MI308X) to the webpage, with MI355X as the default. Header information is now displayed according to the selected GPU.   commit:[20c6690](https://github.com/ROCm/ATOM/pull/1008/commits/20c6690e5a8e1d2dcec2c0bfdf1c233b7b34faf4)

**Default display interface：**
image
**GPU selection options (MI355X and MI308X)** 
image
**Select MI308X:**
image
cc: @wuhuikx, @Yuechguo

wuhuikx (dismissed):

.github/benchmark/sglang_benchmark_models.json
```diff
@@ -23,23 +23,97 @@
       "source_path": "deepseek-ai/DeepSeek-R1-0528",
       "path": "deepseek-ai/DeepSeek-R1-0528",
       "prefix": "deepseek-r1-fp8-tp4",
-      "extra_args": ["trust_remote_code", "--tensor-parallel-size 4", "aiter_runtime"],
+      "extra_args": [
+        "trust_remote_code",
```
zhuyuhua-v: What's these format changes for...? I see a lot of format changes like this one..

valarLip (approved):

PR #1017, opened 2026-06-01
Enable DPA + TBO Support for Kimi K2.5 Inference

ftyghome:
## Motivation

Kimi K2.5 inference under Data-Parallel Attention (DPA) combined with Two-Batch Overlap (TBO) exposed several gaps that either crashed the engine or left performance on the table. This PR enables the DPA + TBO path end-to-end: it fixes the fused-MoE fallback and tensor lifetimes in the TBO overlap, aligns cross-DP prefill admission with TBO's two-batch requirement, and extends persistent MLA to multi-rank DP.

## Technical Details

- **Persistent MLA for DP attention** (`attention_mla.py`): relax `use_persistent_mode` from "single-rank only" (`not (dp_size > 1)`) to `dp_size <= 8`, so persistent MLA also runs in the multi-rank DP configuration used by Kimi K2.5.

- **Fix fused MoE on the DPA fallback path** (`moe.py`, `topK.py`):
  - In the DP-attn fallback (`dp_size > 1`, no MORI all2all), MoE runs after `all_gather_with_padding`, so the token dim can grow to `dp_size ×` the per-rank max. Scale `max_num_tokens` for the topK / fused-MoE metadata accordingly to avoid undersized buffers.
  - Only select the MORI all2all path when expert parallel is actually enabled (`enable_expert_parallel`), so DPA-without-EP correctly falls back instead of assuming all2all.

- **Fix TBO tensor live range** (`moe.py`): add a per-(role, ubatch) `_TBO_KEEPALIVE` holder around the all-gather and reduce-scatter comm/compute switches. Under TBO the source/output tensors of in-flight collectives could be freed before the overlapping ubatch waited on the comm; the keepalive defers release to the next same-role hold, which is past the wait point.

- **Two-batch-aware prefill alignment** (`scheduler.py`, `prefill_delayer.py`): TBO prefill splitting needs at least two local prefill requests per DP rank. Replace `_can_admit_head_prefill` (boolean) with `_count_admittable_head_prefills(limit)` and a `_prefill_delayer_readiness()` helper that reports both "has any prefill" and "alignment-ready" (`>= 2` requests when TBO is on, `>= 1` otherwise). `PrefillDelayer` gains a 4th MAX-reduce slot (`local_alignment_ready`) so prefill is delayed until every DP rank can launch a full two-batch, not just until one rank has a request.

- **Fix delayer batch padding** (`model_runner.py`): treat a non-uniform-decode batch as eager (`runs_eager = enforce_eager or not dp_uniform_decode`) when picking the padded graph batch size, so mixed / delayed batches don't pad up to a CUDAGraph batch size they can't actually use.

## Test Plan

We tested Kimi K2.5 MXFP4 end-to-end inference on MI355X with ROCm 7.2.3, TP4.

The comparison includes:

- baseline without DPA / TBO
- DPA only
- DPA + TBO

## Test Results

Numbers in parentheses are throughput/GPU changes relative to the baseline.

| Conc | baseline throughput/GPU | baseline interactivity | DPA throughput/GPU | DPA interactivity | DPA+TBO throughput/GPU | DPA+TBO interactivity |
|---:|---:|---:|---:|---:|---:|---:|
| 4 | 898.9 | 104.95 | 611.7 (-32.0%) | 73.53 | 610.2 (-32.1%) | 73.63 |
| 8 | 1502.8 | 89.61 | 1061.4 (-29.4%) | 65.48 | 1077.1 (-28.3%) | 65.67 |
| 16 | 2045.6 | 61.85 | 1653.5 (-19.2%) | 51.93 | 1677.6 (-18.0%) | 52.86 |
| 32 | 2964.5 | 43.46 | 2542.4 (-14.2%) | 38.51 | 2588.0 (-12.7%) | 39.21 |
| 64 | 3946.8 | 29.06 | 3761.2 (-4.7%) | 28.52 | 3961.7 (+0.4%) | 30.09 |
| 128 | 4984.5 | 19.81 | 5133.1 (+3.0%) | 21.04 | 5745.4 (+15.3%) | 23.55 |

At higher concurrency, DPA and TBO show a higher throughput ceiling, with DPA+TBO reaching +15.3% throughput/GPU over the baseline at conc=128.

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -1664,9 +1664,10 @@ class ModelRunner:
             if hasattr(self, "drafter"):
                 mtp_step = self.drafter.mtp_k + 1
                 padded_scheduled_bs = (padded_scheduled_bs + mtp_step - 1) // mtp_step
+            runs_eager = self.enforce_eager or not dp_uniform_decode
             bs = (
                 padded_scheduled_bs
-                if self.enforce_eager
+                if runs_eager
                 else next(
                     (x for x in self.graph_bs if x >= padded_scheduled_bs),
                     padded_scheduled_bs,
--- a/atom/model_engine/prefill_delayer.py
+++ b/atom/model_engine/prefill_delayer.py
@@ -7,11 +7,12 @@ the "mixed prefillable status → delay" core that fixes our 1k/1k workload's
 
 Mechanism (per scheduler tick):
   1. Each DP rank reports its local state via cpu all_gather:
-       (local_prefillable, watermark_force_allow)
+       (local_prefillable, local_alignment_ready, watermark_force_allow)
   2. Compute `prefillable_status` ∈ {all, none, mixed}:
-       - "all"   → every rank has a new prefill ready  → allow (8-way aligned)
+       - "all"   → every rank has enough prefill ready → allow (8-way aligned)
        - "none"  → no rank has any prefill             → allow (vacuous)
-       - "mixed" → only some ranks have prefill ready  → DELAY
+       - "mixed" → at least one rank has prefill, but
+                   not all ranks are alignment-ready    → DELAY
   3. In "mixed", refuse the prefill (return False from `should_allow_prefill`)
      for up to `max_delay_passes` consecutive ticks (default 30, ≈ 255ms at
      8.5ms/decode-tick) OR `max_delay_ms` wall-clock (default 5000ms),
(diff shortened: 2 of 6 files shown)


ftyghome:
Hi @wuhuikx ,

I have added the full end-to-end performance results for this PR.

Could you please take another look when you get a chance? Happy to provide any additional data if needed.

Thanks!

ftyghome:
Closing because code needs refactoring to adapt changes in upstream.

PR #1018, opened 2026-06-01
Add HIP MLA reduce kernel dispatch logic

ftyghome:
Add HIP MLA reduce kernel dispatch logic. Rely on https://github.com/ROCm/aiter/pull/3468.

## Performance Result

- E2E (rocprof, Kimi K2.5, conc128): the decode reduce drops from **7997 ns**
  (stock `kn_mla_reduce_v1_ps`) to **6176 ns** (ours), ~1.3×.
- Microbench (conc64, plain TP4, H=16 / K=512): stock `mla_reduce_v1` ~4.7 µs →
  ours (adaptive) ~2.5 µs, ~1.9×. Correctness checked bit-faithful against a
  torch reference (atol/rtol 2e-2).


ftyghome:
Closing as we found this change can be fully contained in AITER.

PR #1024, opened 2026-06-02
fix: warmup uses full token budget for DP

ZhangLirong-amd:
Warmup now uses the full `max_num_batched_tokens` instead of dividing by `dp_size`. Under DP attention each rank's MoE sees up to `dp_size * local_tokens` after the all-gather, so warmup must exercise the full token budget to capture the true peak activation / CUDA-graph footprint; dividing by dp_size under-sized warmup and let decode OOM later. Also updated the warning message accordingly.

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -1056,7 +1056,11 @@ class ModelRunner:
             self.config.max_num_batched_tokens,
             self.config.max_model_len,
         )
-        warmup_max_tokens = max_num_batched_tokens
+        dp_size = get_dp_group().world_size
+        if self.config.enable_dp_attention:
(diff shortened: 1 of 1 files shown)


atom/model_engine/model_runner.py
```diff
@@ -1056,8 +1056,7 @@ def warmup_model(self):
             self.config.max_num_batched_tokens,
             self.config.max_model_len,
         )
-        dp_size = get_dp_group().world_size
-        warmup_max_tokens = max_num_batched_tokens // dp_size
+        warmup_max_tokens = max_num_batched_tokens
```
valarLip: only for dp attn maybe

ZhangLirong-amd: done

valarLip (approved):

PR #1057, opened 2026-06-03
perf: add preshuffle logic for HIP fused_moe

ftyghome:
## Motivation
ROCm/aiter#3470 adds `mxfp4_moe`, a native HIP MXFP4 (a4w4) MoE backend for gfx950 that beats the baseline path. To use it, weights must be shuffled into the layout its kernels expect; `fused_moe` dispatches by `shuffle_kind`, so old-layout weights never reach the HIP backend.

## Technical Details
Add a preshuffle pass for MXFP4 MoE weights (packed fp4 + e8m0 group scales) matching aiter's `mxfp4_moe` layout, applied at model load and gated on gfx950 + an aiter build with #3470. Tagging weights with the matching `shuffle_kind` is enough for `fused_moe` to auto-dispatch to the HIP backend.

## Test Plan
See ROCm/aiter#3470.

## Test Result
See ROCm/aiter#3470.

### DIFF
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -920,60 +920,36 @@ class Mxfp4MoEMethod(FusedMoEMethodBase):
             layer.w2_weight_scale = None
             return
 
-        E13, N13, K13 = layer.w13_weight_scale.data.shape
+        # shuffle weight
         layer.w13_weight.data = shuffle_weight(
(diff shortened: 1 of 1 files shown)


ftyghome:
Closing since we found a more elegant way to make this change fully contained in aiter.

PR #958, opened 2026-05-28
[Feat]  Support layerwise shared expert fusion for all MoE models

ZLkanyo009:
## Motivation
In DeepSeek-R1-0528-MXFP4-v2, layer 61 has a situation where shared experts and routed experts are of different types, while the other layers still have the same type. The original is_rocm_aiter_fusion_shared_expert_enabled() function will cause all layers to skip shared expert fusion. This PR enables shared expert fusion for the compatible layers in DeepSeek-R1-0528-MXFP4-v2. For other models, the original behavior is preserved to ensure compatibility.

## Command

```
export CUDA_VISIBLE_DEVICES=0,1,2,3

export AITER_QUICK_REDUCE_QUANTIZATION=INT4
export SGLANG_AITER_FP8_PREFILL_ATTN=0
export SGLANG_USE_AITER=1
export ATOM_ENABLE_DS_QKNORM_QUANT_FUSION=1

model_path=/models/deepseek-ai/DeepSeek-R1-0528-MXFP4-v2

export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
 
TORCHINDUCTOR_COMPILE_THREADS=128 python3 -m sglang.launch_server \
    --model-path $model_path \
    --host localhost \
    --port 8000 \
    --trust-remote-code \
    --tensor-parallel-size 4 \
    --mem-fraction-static 0.9 \
    --disable-radix-cache \
    --attention-backend aiter \
    --kv-cache-dtype fp8_e4m3 \
    --max-running-requests 128
```

```
export CUDA_VISIBLE_DEVICES=0,1,2,3

export AITER_QUICK_REDUCE_QUANTIZATION=INT4
export SGLANG_AITER_FP8_PREFILL_ATTN=0
export SGLANG_USE_AITER=1
export ATOM_ENABLE_DS_QKNORM_QUANT_FUSION=1

export MORI_SHMEM_MODE=ISOLATION

model_path=/models/deepseek-ai/DeepSeek-R1-0528-MXFP4-v2

export SGLANG_EXTERNAL_MODEL_PACKAGE=atom.plugin.sglang.models
 
TORCHINDUCTOR_COMPILE_THREADS=128 python3 -m sglang.launch_server \
    --model-path $model_path \
    --host localhost \
    --port 8000 \
    --trust-remote-code \
    --tensor-parallel-size 4 \
    --data-parallel-size 4 \
    --expert-parallel-size 4 \
    --enable-dp-attention \
[...]

PR #1098, opened 2026-06-05
Fix DSV4 MTP shared expert loading

yhl-amd:
## Summary
- prevent DSV4 MTP standalone shared expert weights from being rewritten into routed expert slots
- include fused shared experts in the MTP expert mapping when shared experts are actually fused

## Root cause
PR #1076 broadened loader matching to include `ffn.shared_experts.*`. The target DSV4 model disables fused-shared loading when shared experts are standalone, but `DeepseekV4MTP` did not expose the same signal. As a result, MTP shared expert checkpoint tensors were rewritten to `ffn.experts.{n_routed_experts}.*`, silently dropped, and the draft model accept rate collapsed.

## Validation
- `python3 -m py_compile atom/models/deepseek_v4_mtp.py atom/model_loader/loader.py atom/models/deepseek_v4.py`
- `git diff --check`
- Docker checkpoint loader smoke against `/data/amd_int/models/deepseek-ai/DeepSeek-V4-Pro`: MTP shared expert tensors load into standalone params; no missing or silently dropped MTP shared expert tensors
- 100-sample GSM8K smoke, 3-shot, MTP3: flexible-extract `0.97`, strict-match `0.88`; final smoke MTP stats `Average toks/fwd: 2.94`, `Acceptance rate: 64.52%`

### DIFF
--- a/atom/models/deepseek_v4_mtp.py
+++ b/atom/models/deepseek_v4_mtp.py
@@ -227,15 +227,28 @@ class DeepseekV4MTP(nn.Module):
         """
         return name if "mtp." in name else None
 
+    @property
+    def disable_fused_shared_loading(self) -> bool:
+        """True when MTP shared experts are standalone, not fused into FusedMoE."""
+        for m in self.model.modules():
+            if m.__class__.__name__ == "MoE":
+                return not getattr(m, "_fuse_shared_into_routed", True)
+        return False
+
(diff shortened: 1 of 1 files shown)


yhl-amd:
Closing because this DSV4 MTP shared expert loading fix is now covered by PR #958, including the MTP disable_fused_shared_loading guard and actual fused shared slot mapping.

PR #1119, opened 2026-06-07
fix: fix quant config read logic in model loading

Phi-C:
## Motivation

Fix quant config read procedure in https://github.com/ROCm/ATOM/pull/958. Without this modification, ATOM SGLang benchmark will fail (e.g. https://github.com/ROCm/ATOM/actions/runs/27054539290/job/79856431418). 

## Submission Checklist

### DIFF
--- a/atom/model_loader/loader.py
+++ b/atom/model_loader/loader.py
@@ -310,10 +310,18 @@ def load_model(
         module_prefix = matching_name.split("shared_expert", 1)[0]
         shared_expert_prefix = layer_prefix + matching_name.rstrip(".")
         routed_expert_prefix = layer_prefix + f"{module_prefix}experts"
-        model_quant_config = getattr(getattr(model, "args", None), "quant_config", None)
+        model_quant_config = getattr(
+            getattr(model, "atom_config", None), "quant_config", None
+        )
+        if model_quant_config is None:
+            model_quant_config = getattr(
+                getattr(model, "args", None), "quant_config", None
+            )
         if model_quant_config is None:
             model_quant_config = getattr(model, "quant_config", None)
-        if model_quant_config is not None:
+        if model_quant_config is not None and hasattr(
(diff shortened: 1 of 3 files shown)


ZLkanyo009 (dismissed):

atom/model_loader/loader.py
```diff
@@ -310,10 +310,18 @@ def should_fuse_shared_expert_weight(name: str, matching_name: str) -> bool:
         module_prefix = matching_name.split("shared_expert", 1)[0]
         shared_expert_prefix = layer_prefix + matching_name.rstrip(".")
         routed_expert_prefix = layer_prefix + f"{module_prefix}experts"
-        model_quant_config = getattr(getattr(model, "args", None), "quant_config", None)
```
valarLip, line 313: maybe let's force all models have "quant_config"

Phi-C: It seems not easy to unify the read logic in one path, since 1) for plugins, "model.quant_config" means vllm/sglang's quant_config, which is different from "model.atom_config.quant_config";  2) for dsv4, it uses "model.args.quant_config".

valarLip: how about unified to "model.atom_config.quant_config", you can feel free to change dsv4's code for this target

Phi-C: > how about unified to "model.atom_config.quant_config", you can feel free to change dsv4's code for this target

For models in ATOM/atom/models, we use "self.config" for atom config, if we want to unified to "model.atom_config.quant_config", it means  we have to change all these into "self.atom_config". Maybe we can keep "model.atom_config.quant_config" and "model.quant_config" in this PR to avoid too many modifications, and change all models's "config" to "atom_config" in another PR to unify the read path if necessary?

valarLip (approved):

PR #1120, opened 2026-06-07
refactor: unify cudagraph-vs-eager dispatch via ForwardMode

JiaoliangYu:
prepare_inputs and run_model used to each derive eager-vs-cudagraph from the same four-OR chain and drifted under PR #930's dp_uniform_decode path: prepare_inputs rounded bs up to the next captured graph size while run_model took the eager branch with local-real input_ids, leaving slot_mapping at the rounded length. Downstream this tripped aiter's fused_qk_rope_reshape_and_cache assert (t_slot=48 > t=33).

Introduce ForwardMode (frozen dataclass) owning the decision and its data:
- ForwardMode.decide() is the single rule (all four force-eager conditions live here).
- prepare_inputs uses effective_bs to size attn_metadata.
- run_model strict-asserts forward_mode is set, then dispatches off use_cudagraph.
- assert_shape_contract() guards the slot_mapping <-> input_ids invariant (skips prefill / cudagraph internally so callers stay decision-free).

## Motivation
Consolidate eager-vs-cudagraph dispatch and padding decisions into a single ForwardMode.

## Test Plan

the same with ci on gpt-oss-120b

## Test Result

 flex-extract:0.8832       strict-extract: 0.3806

### DIFF
--- a/atom/utils/forward_context.py
+++ b/atom/utils/forward_context.py
@@ -160,6 +160,143 @@ class SpecDecodeMetadata:
     bonus_logits_indices: torch.Tensor
 
 
+@dataclass(frozen=True)
+class ForwardMode:
+    """Per-step dispatch decision: cudagraph vs eager + per-rank attention bs
+    + cross-DP MoE-pad bs.
+
+    Two distinct sizes because they answer different questions:
+      - ``effective_bs`` sizes per-rank attention tensors (slot_mapping /
+        cu_seqlens_q). Must match ``input_ids.shape[0]`` (= local real tokens)
+        in eager so aiter's ``t == t_slot`` invariant holds.
+      - ``moe_pad_bs`` sizes ``context.graph_bs`` which MoE's
+        ``pad_for_all_gather`` reads to pad ``hidden_states`` before the
+        cross-DP ``all_gather``. Must be identical on every DP rank or the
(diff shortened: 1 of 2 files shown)


atom/utils/forward_context.py
```diff
@@ -160,6 +160,105 @@ class SpecDecodeMetadata:
...
+        if is_prefill:
+            # Prefill is always eager. effective_bs is the sequence count;
+            # per-token sums are tracked separately via cu_seqlens_q.
+            return cls(
+                use_cudagraph=False,
+                effective_bs=total_seqs_num,
+                is_prefill=True,
+            )
+
+        if enforce_eager or not dp_uniform_decode:
```
ZhangLirong-amd: In decode, we must unify graph_bs in every ranks since moe all_gather https://github.com/ROCm/ATOM/blob/main/atom/model_ops/moe.py#L185, although `dp_uniform_decode`=False, it only means decode fallback to eager, but the graph_bs we need to use the max_tokens in each decode ranks.

JiaoliangYu: when dp_uniform_decode =False, we use all_gatherrv to deal with variable-length

ZhangLirong-amd (approved):

valarLip (approved):

PR #1128, opened 2026-06-08
[fix](moe): fix fused shared expert in qwen3.5-fp4

PerryZhang01:
## Motivation

Fix a Qwen3.5-397B-A17B-MXFP4 accuracy regression introduced by shared expert fusion changes. In this case, shared expert fusion was disabled in the runtime path, but the weight loader still rewrote shared expert checkpoint weights into the fused MoE expert slot. This made the loaded weights inconsistent with the execution path and caused GSM8K accuracy to drop near zero.

This change keeps shared expert weight remapping behind the same runtime fusion gate used by MoE/topK execution, so Qwen3.5 MXFP4 loads shared expert weights through the standalone path when fusion is disabled. Validation on Qwen3.5-397B-A17B-MXFP4 recovers exact_match,flexible-extract above the required threshold.

## TEST

image

### DIFF
--- a/atom/model_loader/loader.py
+++ b/atom/model_loader/loader.py
@@ -38,10 +38,7 @@ from atom.model_loader.weight_utils import (
 )
 from atom.model_ops.base_config import QuantizeMethodBase
 from atom.model_ops.moe import FusedMoEMethodBase
-from atom.model_ops.topK import (
-    is_rocm_aiter_fusion_shared_expert_enabled,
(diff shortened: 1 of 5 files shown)


PerryZhang01:
fp4 routed weights can not fuse fp8 shared experts weights.

ZLkanyo009 (commented):
is_rocm_aiter_fusion_shared_expert_enabled() should not use again

yhl-amd:
plz fix confict

ZLkanyo009:
this pr should be closed

PR #1137, opened 2026-06-09
benchmarks: standalone model-loading (safetensors) speed benchmark

carlushuang:
## Summary

Two complementary, **standalone** benchmarks for model-loading (safetensors) speed under `atom/benchmarks/`. Neither requires installing ATOM.

| script | deps | measures |
|---|---|---|
| `benchmark_model_loading.py` | `torch` + `safetensors` | full load of a model dir incl. deserialize/copy, mirroring ATOM's loader paths |
| `benchmark_disk_loading.py` | **stdlib only** (runs on base `python3`) | raw safetensors read bandwidth across **every disk**, dummy-only |

Both follow the published safetensors-load studies: **GB/s throughput** (fastsafetensors, NVIDIA Run:ai Model Streamer), **cold vs warm cache + major page-fault counts** (ServerlessLLM, OSDI'24), and **median over N runs / mmap vs read / 1 vs N threads** (HuggingFace official comparison).

---

## 1. `benchmark_model_loading.py` (torch + safetensors)

Mirrors ATOM's actual read path in `model_loader/loader.py::safetensors_weights_iterator`:
- default `safe_open` **mmap** + `posix_fadvise` read-ahead,
- the **`ATOM_DISABLE_MMAP`** read-whole-file → `safetensors.torch.load` path,
- optional **thread pool** (`ATOM_LOADER_USE_THREADPOOL`),
- registers the `F8_E8M0` MX-scale dtype like the loader, so MXFP8 checkpoints work,
- forces the actual read (clone) since `get_tensor()` is zero-copy on CPU.

**Dummy weights** (`--make-dummy`) so loading can be benchmarked without the real checkpoint:
- generic: `--size-gb / --shards / --dtype {bf16,fp16,fp32,fp8_e4m3}`
- `--like REAL_DIR`: clone a real model's exact name/shape/dtype/shard layout (incl. MXFP8 `F8_E4M3` weights + `F8_E8M0` scales), byte-faithful even if the host torch lacks the exact dtype.

```bash
python atom/benchmarks/benchmark_model_loading.py --model /path/to/snapshot_dir
python atom/benchmarks/benchmark_model_loading.py --make-dummy --out /nvme/dummy --size-gb 24 --shards 12
python atom/benchmarks/benchmark_model_loading.py --model /nvme/dummy --device cuda --cache warm
```

## 2. `benchmark_disk_loading.py` (stdlib only, all disks)

The safetensors container is `<u64 len><json header><raw bytes>`, and a load is dominated by reading those bytes — so this reads them with only `mmap` / `os.posix_fadvise` / `read()` (no torch, no safetensors lib, no ATOM). It **auto-discovers every real local block device** (`/proc/mounts`, dedup by physical disk, excludes pseudo/loop/network FS, labels HDD vs SSD/NVMe via `/sys/block/.../rotational`, resolves a writable subdir per filesystem), writes a dummy safetensors file on each, and reports cold/warm × mmap/read × 1/N-thread GB/s + page faults.

```bash
python3 atom/benchmarks/benchmark_disk_loading.py                       # all disks
python3 atom/benchmarks/benchmark_disk_loading.py --list-disks
python3 atom/benchmarks/benchmark_disk_loading.py --size-gb 8 --disks / /mnt/sda1
```

---

## Measured (agent-tr9980x-01: Samsung 9100 PRO Gen5 NVMe + WDC 24 TB HDD)

`benchmark_disk_loading.py`, 4 GiB dummy/disk:
```
disk         kind       mmap c x1  mmap c xN  read c xN  mmap w xN
/            SSD/NVMe        3.75       6.39       7.17      15.55
/mnt/sda1    HDD             0.23       0.23       0.28      20.67
```
`benchmark_model_loading.py` (torch, NVMe, 24 GiB bf16): mmap **5–7 GB/s cold / 11–16 warm**; `ATOM_DISABLE_MMAP` **2–3 GB/s** (deserialize-bound).

**Takeaways:** the default mmap path reaches **~half the NVMe's rated ~14 GB/s cold** (OSDI'24 page-fault effect); **warm hits the RAM ceiling**; `ATOM_DISABLE_MMAP` is **~3× slower**; **concurrent reads help on NVMe but not HDD** (seek-bound); **NVMe cold ≈ 25× faster than HDD** — so weights on the HDD vs NVMe is ~28 min vs ~1.2 min for a ~430 GB model. A concurrent/direct-IO loader (fastsafetensors / Run:ai-style) is where the remaining ~2× on NVMe lives.

## Testing
Validated in `rocm/atom:latest` (torch script: sweep, cache modes, all `--make-dummy` incl. `--like` MXFP8 clone) and on the **base `python3`** (stdlib script: multi-disk discovery + benchmark). All numbers above are real runs.

### DIFF
--- a/atom/benchmarks/benchmark_disk_loading.py
+++ b/atom/benchmarks/benchmark_disk_loading.py
@@ -1,0 +1,298 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+"""Minimal-dependency safetensors load-speed benchmark across every local disk.
+
+Pure Python **stdlib only** — no torch, no safetensors, no ATOM install. The
+safetensors container is just `<u64 header_len><json header><raw tensor bytes>`,
+and a load is dominated by *reading those bytes*, so we generate a dummy
+safetensors file and time reading it back with the two strategies ATOM's loader
+uses: `mmap` (+`posix_fadvise` read-ahead, like
+`model_loader/loader.py`) and a plain `read()` (the `ATOM_DISABLE_MMAP` path).
+
+It auto-discovers all real local filesystems (from /proc/mounts), labels each
+HDD vs SSD/NVMe (/sys/block/.../queue/rotational), writes a dummy on each, and
+reports GB/s cold (page cache evicted -> real disk) and warm (cached -> RAM
+ceiling), single- and multi-threaded, plus major page-fault counts (the
+ServerlessLLM/OSDI'24 mmap effect). GB/s is the metric fastsafetensors / NVIDIA
+Run:ai Model Streamer report.
+
+Run (no deps):  python3 benchmark_disk_loading.py
+                python3 benchmark_disk_loading.py --list-disks
+                python3 benchmark_disk_loading.py --size-gb 8 --disks / /mnt/sda1
+"""
+
+import argparse
+import json
+import mmap
+import os
+import re
+import shutil
+import struct
+import time
+import zlib
+from concurrent.futures import ThreadPoolExecutor
+
+GiB = 1024**3
+CHUNK = 16 << 20
+_PSEUDO = {
+    "tmpfs", "devtmpfs", "proc", "sysfs", "cgroup", "cgroup2", "overlay",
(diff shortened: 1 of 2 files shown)


valarLip:
format...

PR #1144, opened 2026-06-09
[Triton] [Gluon] [GFX12] ATOM_USE_UNIFIED_ATTN=1 force block_size for Unified Attention and resolve bug in prefill

k50112113:
This PR does the following for `ATOM_USE_UNIFIED_ATTN=1` case:

1. patches block_size to the most performant one when `ATOM_USE_UNIFIED_ATTN=1`
2. make the prefill part of Unified Attention aware of the shuffling format depending on if it's pure prefill or prefill from cache
3. perform blind downcast for q right before the unified attention if kv_cache is fp8


atom/utils/tbo/ubatch_splitting.py
```diff
@@ -387,6 +387,7 @@ def split_attn_metadata(
         total_kv=ub_total_kv,
         num_cached_tokens=ub_num_cached_tokens,
         seq_starts=ub_seq_starts,
+        prefill_from_cache=attn_metadata.prefill_from_cache,
```
valarLip: use num_cached_tokens instead maybe

k50112113: removed

valarLip (approved):

PR #1151, opened 2026-06-10
[atom-vllm-benchmark] Add temperature=0 for vLLM bench MTP models

junyyang-amd:
## Motivation

Add "--temperature 0" for vLLM bench MTP models

### DIFF
--- a/.github/workflows/atom-vllm-benchmark.yaml
+++ b/.github/workflows/atom-vllm-benchmark.yaml
@@ -1659,6 +1659,11 @@ jobs:
             fi
           fi
 
+          TEMPERATURE_ARG=""
+          if [[ "${MODEL_NAME}" == *"MTP"* ]]; then
+            TEMPERATURE_ARG="--temperature 0"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/atom-vllm-benchmark.yaml
```diff
@@ -1659,6 +1659,11 @@ jobs:
             fi
           fi
 
+          TEMPERATURE_ARG=""
+          if [[ "${MODEL_NAME}" == *"MTP"* ]]; then
```
whx-sjtu, line 1663: What's the MODEL_NAME here? Normally MTP cases run with server arg like: `--speculative-config '{"method": "mtp", "num_speculative_tokens": 3}' `. There will be no "MTP" string in model name/path.

zejunchen-zejun: @junyyang-amd  MODEL NAME means the model case name right? I remember
for example, here
image

zejunchen-zejun (approved):

PR #1153, opened 2026-06-10
[ATOM SGL] fp8 prefill for fp4 v2 model

ZhiweiYan-96:
## Motivation

DeepSeek MLA prefill in the ATOM SGLang plugin uses BF16 attention. This PR let FP8 prefill can be exercised for DeepSeek-R1-0528-MXFP4-v2 model..

## Technical Details

* Enable ATOM SGLang DeepSeek MLA prefill to use the FP8 prefill path when SGLANG_AITER_FP8_PREFILL_ATTN is enabled.
* Align MLA prefill metadata with upstream by using KV indptr for prefix-aware prefill, allowing both no-prefix and prefix + non-absorbed MLA cases to reach the FP8 prefill kernel.
* Add an MXFP4 kv_b_proj fused split/cat path to emit FP8 K/V directly before FP8 prefill attention.

###  semantic of `fused_gemm_afp4wfp4_preshuffle_split_cat` 
For the DeepSeek MLA non-absorbed prefill path, `kv_b_proj` projects the latent KV tensor into the per-head K/V representation. With MXFP4 weights and FP8 prefill attention enabled, we can use `fused_gemm_afp4wfp4_preshuffle_split_cat` to fuse the following steps into one Triton/AITER kernel:

1. Quantize `kv_c_normed` activation to MXFP4/FP8-compatible layout.
2. Run the MXFP4 `kv_b_proj` GEMM.
3. Split the GEMM output into `k_nope` and `v`.
4. Concatenate `k_nope` with the RoPE part `k_pe` to form the full MHA K.
5. Return `k` and `v` directly in the layout consumed by FP8 prefill attention.

This avoids materializing the full `kv_b_proj(kv_a)` output and then doing separate `view/split/cat` operations in Python/Torch.

## Test Result

image

accuracy gsm8k 

image

## Submission Checklist


valarLip:
normal acc for this weight？

ZhiweiYan-96:
@valarLip  Thanks for remind, I used relatively low num_fewshot(=3). 

When using num_fewshot=5, the acc is 
image

Besides, I tested acc w/o the change in this PR, the result is also 0.94(num_fewshot=5).

image

valarLip (approved):

PR #1161, opened 2026-06-10
Route GPT-OSS full prefill to sink ASM

carlushuang:
Review-only PR. This routes GPT-OSS full-attention prefill on gfx1250/bf16/head_dim=64/sink/no-SWA/no cached prefix to AITER flash_attn_varlen_func so it can hit the sink varlen ASM path. SWA layers and decode/PA decode are intentionally untouched. Commit: da777e9464b10ae2d66c9f03620e0737440cbc20


carlushuang:
Closed intentionally after creation so the diff remains available for review. Branch preserved.

PR #1166, opened 2026-06-10
[atom-vllm] enable multistream for dsv4

whx-sjtu:
## Summary
- Enables multi-stream for vllm dsv4 by pass `in_hipgraph=True` to atom context in FULL_GRAPH mode
- Patch vllm's `_warmup_and_capture` to also pass `in_hipgraph=True` in vllm's profile run and graph capture, in order to warm up gemm workspace.

## Test plan
image

```
local-completions ({'model': '/home/models/deepseek-ai/DeepSeek-V4-Pro', 'base_url': 'http://localhost:8001/v1/completions', 'num_concurrent': 64, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 20, batch_size: 1
|gsm8k|      3|flexible-extract|    20|exact_match|↑  |0.9447|±  |0.0063|
|     |       |strict-match    |    20|exact_match|↑  |0.9454|±  |0.0063|
```


atom/plugin/vllm/deepseek_v4_bridge.py
```diff
@@ -1322,28 +1348,9 @@ def atom_deepseek_v4_forward_context(
...
-                None
-                if common_attn_metadata is None
-                else common_attn_metadata.seq_lens[:8].detach().cpu().tolist()
-            )
-            n_csa = getattr(attn_metadata, "n_committed_csa_per_seq_cpu", None)
-            print(
-                f"[ATOM_VLLM_V4_DEBUG] state={attn_metadata.state} max_q={attn_metadata.max_seqlen_q} ids={ids} pos={pos} seq={seq} n_csa={None if n_csa is None else n_csa[:8].tolist()}",
-                flush=True,
-            )
+    in_hipgraph = bool(getattr(attn_metadata, "in_hipgraph", False)) or (
```
zejunchen-zejun, line 1379: Hi @whx-sjtu  Is it possible to get the `cuda graph capture enabled or not` through the vllm official status context: https://github.com/vllm-project/vllm/blob/v0.22.0/vllm/v1/worker/gpu_model_runner.py#L6298 instead of patch a thread local depth

zejunchen-zejun: It looks like the graph capture status is set by vllm before and after the warmup and capture
```
set_cudagraph_capturing_enabled(True)
self._warmup_and_capture
set_cudagraph_capturing_enabled(False)
```

whx-sjtu: sure, I will try.

atom/plugin/vllm/deepseek_v4_bridge.py
```diff
@@ -1280,6 +1284,38 @@ def get_deepseek_v4_proxy_metadata_from_vllm_context():
...
+            return False
+        vllm_ctx = get_forward_context()
+        batch_descriptor = getattr(vllm_ctx, "batch_descriptor", None)
+        is_uniform_decode_bucket = bool(
+            batch_descriptor is not None and getattr(batch_descriptor, "uniform", False)
+        )
+        is_single_query_decode = int(getattr(attn_metadata, "max_seqlen_q", 0)) == 1
+        if not (is_uniform_decode_bucket or is_single_query_decode):
+            return False
+        return bool(getattr(vllm_monitor, "cudagraph_capturing_enabled", False))
```
zejunchen-zejun, line 1332: More clear now! But for enforce-eager scenario, the `_is_vllm_decode_graph_phase` could return True because the default value of `cudagraph_capturing_enabled` is True, while multi stream could lead to hang in eager mode according to the atom comments.

Could we use atom_config from parent caller to exclude this condition(enforce-eager) ? I remember the enforce eager or cuda graph mode are in this config.
```
def atom_deepseek_v4_forward_context(
    *,
    atom_config,  <----------------- use this config
    input_ids,
    positions,
    common_attn_metadata=None,
    force_dummy: bool = False,
    state_model=None,
    meta_params=None,
    slot_allocator=None,
):
```

zejunchen-zejun (dismissed):

.github/workflows/atom-vllm-test.yaml
```diff
@@ -249,7 +249,7 @@ jobs:
               AITER_BF16_FP8_MOE_BOUND=0
               ATOM_MOE_GU_ITLV=1
             lm_eval_num_fewshot: 20
-            accuracy_test_threshold: 0.94
+            accuracy_test_threshold: 0.93
```
zejunchen-zejun, line 252: @whx-sjtu you can remove this change to the acc threshold
I will have a PR to change the acc threshold for all needed models

zejunchen-zejun (approved):

valarLip (approved):

PR #1190, opened 2026-06-12
[atom-vllm nightly acc] remove config in workflow file and fetch config from JSON file

zejunchen-zejun:


zejunchen-zejun:
Hi, @junyyang-amd 
Could you help review? I change the atom-vllm nightly workflow and make the config of model case coming from the JSON file

junyyang-amd (dismissed):

PR #1209, opened 2026-06-14
Fix MHA PA ASM metadata alignment

yhl-amd:
## Summary
- allow head_dim=128 MHA decode to dispatch to AITer PA ASM when no sliding window is used
- align context_lens and cu_seqlens_q with active block_tables rows before calling pa_fwd_asm
- fail fast if PA ASM metadata is shorter than the active block table batch

## Testing
- python3 -m py_compile ATOM/atom/model_ops/attention_mha.py
- end-to-end lm_eval GSM8K with ATOM server + eagle3, num_concurrent=64: completed 1319 requests without PA_ASM_SYNC_FAIL, illegal memory access, traceback, or server death
- GSM8K result: flexible-extract exact_match 0.8560 +/- 0.0097; strict-match exact_match 0.8491 +/- 0.0099

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -2013,8 +2013,10 @@ class ModelRunner:
                 num_tokens = context.batch_size * max_q_len
                 hidden_states = self.forward_vars["outputs"][:num_tokens]
                 if graph_key in self.graph_aux_hidden:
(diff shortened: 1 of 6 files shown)


yhl-amd:
context-len的计算改为按照q的shape计算

PR #1210, opened 2026-06-15
(eplb): EPLB_V_0.9

JiaoliangYu:
## Summary

  This PR adds the core EPLB runtime plumbing for ATOM MoE models, including expert-location metadata, load monitoring, runtime rebalance orchestration, expert migration support, and MoE routing integration  for the currently supported MXFP4 path.

### EPLB runtime core
`atom/model_ops/eplb.py`

- `ExpertLocationMetadata` — tracks logical↔physical expert placement (physical→logical map, logical→physical map, per-logical replica counts, per-rank dispatch maps).
- `ExpertLoadMonitor` — collects per-physical-expert load statistics over a sliding window.
- `EPLBManager` — owns runtime metadata, rebalance scheduling/gating, placement computation, and the migrate→commit flow.
- Helper APIs: `initialize_eplb_runtime`, `eplb_map_logical_to_physical`, `record_eplb_expert_load`.

### MoE routing integration
`atom/model_ops/moe.py`

- Adds `select_experts_with_record()` in `FusedMoEMethodBase`, which performs expert selection, logical→physical remapping, and EPLB load recording.
- Currently wired into `Mxfp4MoEMethod`. FP8 paths are not enabled for EPLB routing/recording in this PR.

### Model runner lifecycle
`atom/model_engine/model_runner.py`

- Initializes the EPLB runtime before model warmup.
- Wraps forward execution with EPLB forward-start / forward-end hooks.

### Configuration
`atom/config.py`

- EPLB configuration knobs: enablement, redundant-expert count, load-window size, rebalance interval, rebalance balancedness threshold/aggregation, P2P batch chunk size, and placement policy.

## Load monitoring & rebalance scheduling

`ExpertLoadMonitor` collects the per-physical-expert load that drives placement decisions.

- **Sliding window** — a fixed-size ring buffer of per-physical-expert token counts (window length = `--eplb-load-window-size`); each forward pass writes one slot, overwriting the oldest.
- **On-forward recording** — load is recorded through the forward-start / forward-end hooks. Counting uses `scatter_add` with only fixed-shape elementwise ops (no `bincount`, boolean-mask indexing, or `.item()` host syncs), so it is safe to run inside a HIP/CUDA graph capture (decode path).
- **Global aggregation** — `dump_global_physical_load()` all-reduces the window across EP ranks to a global per-physical load; `dump_global_logical_load()` folds physical→logical (summing replicas) to produce the per-logical load the placement policy consumes.

Rebalancing is scheduled by a generator advanced once per forward pass:

- Fires every `--eplb-rebalance-interval` forward passes, with a vLLM-style warm start (first rebalance at ¼ of the interval) so balancing on real traffic kicks in early. The step counter is advanced on every forward (including dummy/padding steps) to keep all ranks in lockstep.
- A **balancedness gate** (`--eplb-rebalance-min-balancedness`, aggregation `min`/`mean`) can skip a scheduled rebalance when the current placement is already balanced enough, avoiding low-value migrations.

## Placement policies

Selectable via `--eplb-placement-policy {naive,biased}`. A policy is a single callable `(weight_l, num_physical, num_gpus, old_p2l_layer) -> (p2l, phyrank, logcnt)`; K is always derived from the redundant budget (no extra CLI knob). The hierarchical (multi-node) path applies the policy per node for within-node replication.

- **`naive`** — greedy per-replica-load replication + balanced packing (spread the redundant budget thinly over the hottest experts).
- **`biased`** — spend the whole redundant budget fully replicating the top-K hottest experts to every GPU (`K = num_redundant // num_gpus`); cold experts are LPT-packed into the remaining slots.
- **Sticky (shortest-path) migration for `biased`** — the policy receives the current per-layer placement (`old_p2l_layer`). When the new top-K hot set equals the set already fully replicated in the live placement, it returns the existing placement unchanged, so the per-slot migration diff is zero (no P2P) for that layer. Hot experts are placed in stable (expert-id) order so an unchanged hot expert keeps its slot even when the layout is rebuilt. On a converged, stationary load the highest-traffic experts are no longer re-migrated every interval; only genuine hot-set changes trigger movement.

## Runtime correctness & robustness

These pieces are what make runtime rebalancing safe on gfx950 / RCCL.

### Redundancy-aware initialization (`fill_redundant`)
The checkpoint loader fills only base physical slots (it iterates `num_logical`), so redundant physical slots are uninitialized after load. `EPLBManager.fill_redundant()` performs a one-time P2P copy from the base holders into the redundant slots at init (reusing the migrate+commit path, `old = base-only`, `new = live`), so dispatch to a replica never reads uninitialized memory before the first rebalance. No-op when `num_redundant == 0`. This is a temporary bridge until the loader itself fills redundant slots at load time (see TODO).

### Migration communicator isolation
Weight-migration P2P runs on a dedicated process group (`torch.distributed.new_group` over the EP ranks), separate from the EP group used by the forward pass. NCCL/RCCL matches P2P per `(peer, per-communicator op order)`; sharing the EP communicator between migration and the forward MoE all-to-all lets a migration op cross-match an in-flight forward op to the same peer and stall the collective. A dedicated migration group removes this coupling.

### Chunked P2P migration
Migration uses chunked `batch_isend_irecv` (unchunked P2P triggers RCCL GPU-side accumulation hangs). Chunk boundaries are keyed by expert id so a send and its matching recv always land in the same batch.

## Configuration / CLI

- `--enable-eplb`
- `--eplb-num-redundant-experts <N>`
- `--eplb-placement-policy {naive,biased}`
- `--eplb-rebalance-interval <steps>`
- `--eplb-load-window-size <steps>`
- `--eplb-rebalance-layers-per-chunk <N>`
- `--eplb-rebalance-min-balancedness <float>`

## Validation

8×MI355 (gfx950), DeepSeek-V4-Pro, EP=8, DP-attention, `--max-num-batched-tokens 8192`, prefill (random 8192-in / 1-out), concurrency 128, warmup 1024 + measured 4096 prompts.

### Throughput & accuracy

| Config | Throughput | Δ vs no-EPLB | gsm8k (flex / strict) | Extra VRAM |
|tp+dpa+tbo, no ep|6.97 req/s, 7.2 req/s for 16k chunk|—|—|—|
|tp+dpa+ep+tbo| 5.71 req/s, 5.87 req/s for 16k chunk|— | —|—|
| no-EPLB | 6.48 req/s | — | — | — |
| EPLB, 0 redundant (pure rearrange) | 7.56 req/s | **+16.7%** | 0.9560 / 0.9568 | none |
| EPLB `naive`, 64 redundant | 8.06 req/s | **+24.4%** | 0.956 / 0.956 | +64 slots/layer |
| EPLB `biased`, 64 redundant | 8.68 req/s | **+34.0%** | 0.9553/ 0.9553 | +64 slots/layer |

Pure rearrange (`num_redundant = 0`) already recovers most of the imbalance at **zero extra memory**; replication (`naive`/`biased`) adds dispatch locality on top, with `biased` (top-K fully replicated) the strongest.

### Rebalance overhead (throughput-bench phase only)

Per-rebalance cost measured during the throughput bench (ep_rank 0; the gsm8k phase is excluded). Each config performed 5 rebalances in the bench window (rebalance interval 200). `rearrange` = plan computation (module-C), `migrate` = P2P weight migration + commit (modules D/E). The first migration after startup is a one-time cold cost (NCCL/buffer warmup); steady-state excludes it.

| Config | rearrange (steady) | migrate (steady) | migrate #1 (cold) |
| --- | --- | --- | --- |
| 0 redundant (pure rearrange) | 55 ms | ~2.0 s | 4.33 s |
| `naive`, 64 redundant | 169 ms | ~2.1 s | 4.64 s |
| `biased`, 64 redundant | 65 ms | **~0.73 s** | 4.72 s |

- `biased` + sticky migration moves far less in steady state — once the hot set converges, most intervals are near-zero P2P (e.g. 68 ms). `naive` re-spreads the redundant budget every interval; pure-rearrange relocates experts across GPUs without replication.
- `naive`'s plan computation is the most expensive (balanced-packing over all 448 physical slots with heavy replication); `biased` (top-K) and pure-rearrange are ~3× cheaper.

### Correctness

- No accuracy regression on any config — gsm8k was measured with rebalancing active, confirming migration installs the correct weights.
- **No deadlock or crash on any config across the full run** — init trivial→placement rebalance (incl. `fill_redundant`) plus 9 runtime rebalances each. Runtime rebalancing on gfx950 requires the migration communicator to be initialized with an all-rank barrier before the asymmetric `fill_redundant` P2P (see Runtime correctness).

## Tests

Files: `tests/test_eplb_metadata.py`, `tests/test_eplb_module_a.py`, `tests/test_eplb_module_b.py`, `tests/test_eplb_module_c.py`, `tests/test_eplb_module_d.py`, `tests/test_eplb_module_e.py`.

Cover metadata construction, load monitoring, scheduling/gating, rebalance behavior, migration helpers, and module-level EPLB behavior. 38 passing.

## TODO

- Remove `fill_redundant` at EPLB init; fill redundant physical slots while loading weights (make the loader redundancy-aware — iterate physical experts and map each back to its logical checkpoint weight).
- Gate rebalancing on per-GPU balancedness rather than per-slot balancedness, so a converged placement stops triggering low-value rebalances.
- Avoid copy on local rank
- sync copy from temp buffer to rank+ event
- Analyze trace with decode
- accuracy ci case
- add recipe


atom/model_engine/arg_utils.py
```diff
@@ -318,6 +314,64 @@ def add_cli_args(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
                 '\'{"use_index_cache": true, "index_topk_freq": 4}\''
             ),
         )
```
JiaoliangYu, line 355: 优化传参为map

atom/model_ops/moe.py
```diff
@@ -2415,19 +2455,33 @@ def __init__(
...
         if self.use_ep:
             self.local_num_experts, self.expert_map = determine_expert_map(
                 ep_size=self.ep_size,
                 ep_rank=self.ep_rank,
-                global_num_experts=self.global_num_experts,
+                global_num_experts=self.num_physical_experts,
             )
         else:
-            self.local_num_experts = self.global_num_experts
+            self.local_num_experts = self.num_physical_experts
```
JiaoliangYu: 控制命名style的统一

atom/model_ops/moe.py
JiaoliangYu, line 2481: expert_mask 更新

atom/model_ops/eplb.py
```diff
@@ -0,0 +1,2354 @@
...
+                )
+            self._expert_map_tails[layer_id] = tail
+            runtime_map = torch.empty(
+                num_physical + int(tail.numel()),
+                dtype=torch.int32,
+                device=self.live_metadata.expert_map.device,
+            )
+            runtime_map[:num_physical].copy_(self.live_metadata.expert_map[layer_id])
+            if tail.numel() > 0:
+                runtime_map[num_physical:].copy_(tail.to(runtime_map.device))
```
JiaoliangYu, line 1860: 后面只用expert map一份

atom/model_ops/eplb.py
```diff
@@ -0,0 +1,2354 @@
...
+    stream_ctx = (
+        torch.cuda.stream(cuda_stream) if cuda_stream is not None else nullcontext()
+    )
+    with stream_ctx:
+        for src_slot, dst_slot in plan:
+            assert 0 <= src_slot < num_slots
+            assert 0 <= dst_slot < num_slots
+            for b, w in zip(temp_buffers, expert_weights):
+                # Keep fixed tensor addresses for cudagraph compatibility.
+                w[dst_slot].copy_(b[src_slot])
```
JiaoliangYu, line 1236: 后续性能优化用异步拷贝+event

atom/models/deepseek_v4.py
JiaoliangYu, line 2771: 这里可能有layer id

JiaoliangYu: 是字符串，不用

atom/model_ops/eplb.py
```diff
@@ -0,0 +1,2354 @@
+"""EPLB module-A runtime helpers (statistics only)."""
+
+from __future__ import annotations
+
+from contextlib import nullcontext
+from dataclasses import dataclass
+from functools import wraps
+from typing import Any, Optional
+
```
JiaoliangYu, line 9: pure prefill用

JiaoliangYu: done

atom/model_ops/eplb.py
JiaoliangYu: todo

ZhangLirong-amd:
add a recipe as example to tell user how to config it?

JiaoliangYu:
> add a recipe as example to tell user how to config it?

I'd show it on another pr

PR #1243, opened 2026-06-17
[atom-vllm-nightly] Modify qwen3-next-80b-a3b-fp8 threshold from 0.83 to 0.81

junyyang-amd:
## Motivation

Modify qwen3-next-80b-a3b-fp8 threshold from 0.83 to 0.81.


PerryZhang01:
image

zejunchen-zejun (approved):

PR #1244, opened 2026-06-17
[ci][mesh] add model P/D benchmark workflow

wanzhenchn:
## Atomesh CI 

image

### Summary
image


gyohuangxin:
@wanzhenchn Converting it as a draft to skip running all other tests in every commit, please make it open when it's ready to merge. Can we also add the path ignores in other test yaml files to skip tests when only modifying the mesh related files and to run mesh CI only changing the mesh codes? Thanks!

valarLip (approved):

PR #1289, opened 2026-06-18
Feat: RTP-LLM plugin GLM5-FP8 support

zhaoan12-prc:
## Motivation
Add RTP + ATOM integration for GLM5-FP8 so GLM5 can run through the RTP plugin path with ATOM model loading and attention/MoE execution.

## Technical Details
- Added GLM5-FP8 RTP plugin integration on top of ATOM model construction and weight loading.
- Wired RTP plugin attention/context handling for GLM5 MLA and sparse attention paths.
- Kept RTP-specific import and patch behavior isolated under `atom/plugin/rtpllm` to avoid affecting the main ATOM execution path.
- Added/updated RTP plugin tests for GLM5 lifecycle and sparse backend behavior.
- RTP-side enablement is controlled through the GLM5 FP8 server environment, including:
  - `RTP_LLM_EXTERNAL_MODEL_PACKAGES=atom.plugin.rtpllm.models`
  - `LOAD_PYTHON_MODEL=1`
  - `ENABLE_CUDA_GRAPH=1`
  - `FP8_KV_CACHE=1`
  - `MODEL_TYPE=glm_5`
  - `ACT_TYPE=BF16`

## Test Plan
- Verified RTP + ATOM GLM5 with BF16 MLA.
- Verified RTP + ATOM GLM5 with FP8 MLA.
- Benchmarked pure ATOM FP8 MLA as the baseline for comparison.

## Test Result
image

- Verified GLM5 RTP + ATOM with both FP8 MLA and BF16 MLA.
- Compared RTP + ATOM FP8 MLA against pure ATOM FP8 MLA baseline.
- RTP + ATOM FP8 MLA performance is broadly on par with pure ATOM:
  - RTP + ATOM FP8 MLA: TTFT 2290.07 ms, TPOT 34.67 ms
  - Pure ATOM FP8 MLA: TTFT 2231.86 ms, TPOT 33.53 ms


atom/plugin/rtpllm/models/qwen3_5.py
ZhiweiYan-96, line 1: Does change in this file necessary?

zhaoan12-prc: Thanks, this change is necessary, because GLM5 and Qwen3.5 are both registered under the same RTP plugin package, adding GLM5 implement might make the shared import or forward-context code easier to affect Qwen3.5 during package loading.

Compared with main branch, I moved Qwen3.5’s heavy imports to lazy imports and changed it to use `RTPForwardQwen35HybridContext` instead of the shared base `RTPForwardContext`, so Qwen3.5 keeps its existing metadata behavior isolated from the new GLM5 path.

I also removed some extra Qwen3.5 buffer changes that are not needed now.

atom/plugin/rtpllm/models/glm5.py
qichu-yun, line 1: Since you already have base_model_wrapper.py, why does glm5.py need to be so long? Usually, a simple hook of a few lines is enough for the GLM5 model.

zhaoan12-prc: Thanks for the review! You're right that the registration itself is just a few lines. However, glm5.py isn't only a registration hook, it's the GLM5 RTP-LLM runtime adapter.

In the RTP plugin path, ATOM GLM5 has to fit into RTP's runtime objects, and several of those are RTP-defined, like RTP's KV-cache layout and RTP's own CUDA Graph capture/replay flow (not torch.cuda.graph). So the wrapper has to bridge ATOM GLM5 into them: model-specific position reconstruction, forward-context binding, sparse-MLA metadata construction, CUDA Graph prewarm, and weight-loading adaptation. That's what makes glm5.py this long.

base_model_wrapper.py is only the registration. It doesn't run the model forward path or manage RTP KV-cache / CUDA Graph metadata, so we keep a self-contained adapter per model.

valarLip (approved):

PR #1300, opened 2026-06-20
fix(minimax): restore qkv<=256 shape guard + harden compile-cache frozen-path guard

sunway513:
## Summary
MiniMax-M2.5 hard-fails at engine init under torch.compile at TP>1 on `main` (and the v0.1.5 release line). Two surgical fixes. Companion of #1299 (release/v0.1.5).

## Root cause
- `#999` ("Remove qkv 256 tok limitation") dropped the `qkv.shape[0] <= 256` guard in `MiniMaxM2Attention.forward`. The fused custom collective `tensor_model_parallel_fused_qknorm_allreduce` then runs unconditionally at TP>1 inside the compiled forward → top-level graph break → re-enters `VllmBackend` → `AssertionError: VllmBackend can only be called once`. The revert lives only on the v0.1.4 line; `main` carries the un-reverted change.
- The compile-cache hash guard in `backends.py` only skipped `"<string>"`/`"<frozen os>"`; a traced `"<frozen posixpath>"` reached `open()` → `FileNotFoundError` → `BackendCompilerFailed`.

## Changes
1. `atom/models/minimax_m2.py` — restore `qkv.shape[0] <= 256 and` guard on the fused qknorm-allreduce path; non-fused `else` already handles larger token counts.
2. `atom/utils/backends.py` — generic `not os.path.exists(filepath)` instead of enumerated frozen-path special-cases.

## Validation
mi355-gpu-15, GSM8K 3-shot flexible-extract: MiniMax-M2.5 **3/3 mean 0.929** (was hard-failing); DSR1 0.952 / GLM5 0.947 / Kimi 0.940 / Qwen3 mean 0.871 unaffected.

### DIFF
--- a/atom/models/minimax_m2.py
+++ b/atom/models/minimax_m2.py
@@ -233,7 +233,7 @@ class MiniMaxM2Attention(nn.Module):
             # TP-aware RMSNorm: all-reduce variance across TP ranks so
             # normalization uses the global variance (over 6144/1024 dims)
             # rather than per-rank variance (768/128 dims).
-            if self.tp_size > 1:
+            if qkv.shape[0] <= 256 and self.tp_size > 1:
                 q, k, v = tensor_model_parallel_fused_qknorm_allreduce(
                     qkv, self.q_norm.weight, self.k_norm.weight, self.rms_norm_eps
                 )
--- a/atom/utils/backends.py
+++ b/atom/utils/backends.py
@@ -575,9 +575,9 @@ class VllmBackend:
             hash_content = []
             for filepath in forward_code_files:
                 hash_content.append(filepath)
-                if filepath == "<string>" or filepath == "<frozen os>":
-                    # This means the function was dynamically generated, with
-                    # e.g. exec() or frozen os module. We can't actually check these.
+                if filepath == "<string>" or not os.path.exists(filepath):
+                    # Dynamically generated (exec()) or frozen stdlib module
+                    # (<frozen os>, <frozen posixpath>, ...): no readable source.
                     continue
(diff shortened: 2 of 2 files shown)


valarLip:
something mismatch? our nightly benchmark still works well..
https://github.com/ROCm/ATOM/actions/runs/27840294756/job/82397709250#step:7:41

sunway513:
@valarLip @carlushuang — review request.

This carries the MiniMax-M2.5 torch.compile fix onto `main` (companion of #1299 which is already merged to `release/v0.1.5`, shipped in ATOM v0.1.5). `main` still has the bug at `atom/models/minimax_m2.py:236` (unguarded fused qknorm-allreduce → top-level graph break → `VllmBackend can only be called once`). Restores the `qkv.shape[0] <= 256` guard (re-applies the revert of #999, which only ever landed on the v0.1.4 line) + a generic frozen-path guard in `backends.py`.

**On the 3 red accuracy checks (DeepSeek-V4-Flash TP4 / Kimi-K2.5-MXFP4 TP4 / Qwen3.5-35B-A3B-FP8 TP2):** they are **pre-existing on `main`**, not introduced by this PR — this diff only touches `minimax_m2.py` + `backends.py` (unrelated to those 3 models). Same 3 fail on plain `main` (2026-06-24 `083551ec`, 2026-06-26 `4b40ede3`). Tracked separately in ROCm/frameworks-internal#17096.

Validated: MiniMax-M2.5 GSM8K 3-shot mean 0.929 (was hard-failing at init); DSR1/GLM5/Kimi/Qwen3 unaffected.

PR #1305, opened 2026-06-22
feat(minimax_m3): add MXFP4 native support

wuhuikx:
## Summary
- Add the minimal native ATOM MiniMax-M3 MXFP4 model path, including sparse attention/indexer support and MXFP4 MoE loading/runtime plumbing.
- Register MiniMax-M3 model architectures and normalize the text config for MiniMax-M3 checkpoints.

## MiniMax-M3 MXFP4 Accuracy Validation

Using the recipe GSM8K 5-shot flow with `local-chat-completions`, `num_concurrent=32`, `max_gen_toks=16384`, and `batch_size=65`.

### FP4 TP4

```text
local-chat-completions ({'model': 'amd/MiniMax-M3-MXFP4', 'base_url': 'http://127.0.0.1:8000/v1/chat/completions', 'num_concurrent': 32, 'max_gen_toks': 16384}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: 65
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9469|±  |0.0062|
|     |       |strict-match    |     5|exact_match|↑  |0.9477|±  |0.0061|
```

### FP4 TP8

```text
local-chat-completions ({'model': 'amd/MiniMax-M3-MXFP4', 'base_url': 'http://127.0.0.1:8000/v1/chat/completions', 'num_concurrent': 32, 'max_gen_toks': 16384}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: 65
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9393|±  |0.0066|
|     |       |strict-match    |     5|exact_match|↑  |0.9401|±  |0.0065|
```

### DIFF
--- a/atom/model_ops/attentions/aiter_attention.py
+++ b/atom/model_ops/attentions/aiter_attention.py
@@ -10,7 +10,10 @@ import torch
 from aiter.dist.parallel_state import get_tp_group
 from atom.model_engine.scheduler import ScheduledBatch
 from atom.utils import CpuGpuBuffer
-from atom.utils.block_convert import kv_indices_generate_triton
+from atom.utils.block_convert import (
+    block_table_convert_triton,
+    kv_indices_generate_triton,
+)
 from atom.model_ops.attention_mha import PagedAttentionImpl, use_pa_decode_bf16_asm
 from atom.utils.forward_context import AttentionMetaData, Context
 from atom.utils.tbo import TokenSplitPrefillState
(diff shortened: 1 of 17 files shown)


atom/model_ops/moe.py
```diff
@@ -3158,26 +3161,7 @@ def select_experts(
                     num_routing_experts=num_routing_experts,
                     fused_shared_experts_scoring_func=fused_shared_experts_scoring_func,
                 )
-            elif scoring_func == "sigmoid":
```
ZhangLirong-amd, line 3173: Could changing the sigmoid path here have any impact on other models?

atom/model_ops/attentions/aiter_attention.py
```diff
@@ -49,27 +49,23 @@ def __init__(
...
-                    "with the requested block size.",
-                    expected,
-                    model_runner.kv_cache_dtype,
-                    model_runner.block_size,
+        hf_config = model_runner.config.hf_config
+        text_config = getattr(hf_config, "text_config", hf_config)
+        sparse_cfg = getattr(text_config, "sparse_attention_config", None)
+        self._has_sparse_attention = bool(sparse_cfg)
+        if sparse_cfg and (required_block_size := sparse_cfg.get("sparse_block_size")):
+            if model_runner.block_size != required_block_size:
```
valarLip: here we have 2 block_size， runner block size used for manage kv cache and prefix for scheduler， attn block size for kernel point of view，we don't have to make these two same

atom/model_ops/attentions/aiter_attention.py
```diff
@@ -440,6 +465,23 @@ def build_kv_cache_tensor(self, layer_id: int, module):
         from atom.config import KVCacheTensor
         from aiter import dtypes
 
+        if getattr(module, "is_indexed_sparse_attention", False):
+            runner = self.model_runner
+            config = runner.config
+            sparse_idx = runner._sparse_attention_cache_next
+            runner._sparse_attention_cache_next += 1
+            module.kv_cache = runner.kv_cache[:, layer_id].permute(1, 0, 2, 3, 4)
```
valarLip: why we need permute here

atom/model_ops/moe.py
XiaobingSuper: I don't think this change will impact other models. The fused op has the same functionality as the non-fused path, so I think we can remove the non-fused path.

atom/model_ops/attentions/aiter_attention.py
XiaobingSuper: For minimax-m3 attention, it needs` [num_blocks, 2, block_size, num_kv_heads, head_dim]`, I move this permute to minimax-m3 model file.

atom/model_ops/attentions/aiter_attention.py
XiaobingSuper: Yes, updated.

valarLip (approved):

PR #1323, opened 2026-06-23
[atom-vllm-ci] Move atom vllm CI from mi355 to mi350

junyyang-amd:
## Motivation

Move gpt-oss and kimi2.5 CI from mi355 to mi350 to resolve oom issue.


.github/workflows/atom-vllm-test.yaml
gyohuangxin: Add mi350 runners here.

gyohuangxin: Ditto

gyohuangxin (changes requested):

gyohuangxin (approved):

PR #1336, opened 2026-06-24
Feat/low bit ep

JiaoliangYu:
## Submission Checklist

### DIFF
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -3257,6 +3257,20 @@ class FusedMoE(torch.nn.Module):
                     f"Unsupported scoring function for non-grouped topk: {scoring_func}"
                 )
 
+        # [balance exp] env-gated: force perfectly balanced round-robin routing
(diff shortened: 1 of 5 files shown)


atom/model_ops/moe.py
```diff
@@ -201,63 +180,51 @@ def naive_multicast(
...
-    The padding MUST be zeros, never uninitialized memory: padded rows are
-    all-gathered across DP ranks and fed straight into the aiter fused-MoE
-    expert GEMM, where garbage values leak into real tokens' outputs.
-    Bisection traced a ~0.7pp GSM8K drop at large batch to a bare
-    ``torch.empty`` pad here; explicitly zeroing the pad rows fixes it.
-
-    Returns the (possibly padded) tensor and the original row count so the
-    caller can unpad after reduce-scatter.
-    """
+def pad_for_all_gather(x: torch.Tensor):
```
ZhangLirong-amd, line 183: Why we need change the old logic in all_gather and reduce scatter, I think quant dis/com may not influence such old logic

PR #1334, opened 2026-06-24
gluon pa with shuffle layout for minimax m3 

ganyi1996ppo:
## Motivation

serving script:
```
main_model=/workspace/shared/data/amd_int/models/MiniMax-M3-MXFP4
export AITER_LOG_LEVEL=WARNING ATOM_FORCE_ATTN_TRITON=1 HIP_VISIBLE_DEVICES=0,1,2,3
python -m atom.entrypoints.openai_server --model $main_model \
  -tp 4 --server-port 8014 --trust-remote-code --gpu-memory-utilization 0.8 --block-size 128 --no-enable_prefix_caching \
  --max-num-batched-tokens 32768 --max-model-len 32768 --max-num-seqs 128 --kv_cache_dtype fp8 \
```

accuracy result
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9318|±  |0.0069|
|     |       |strict-match    |     5|exact_match|↑  |0.9325|±  |0.0069|
```

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #1342, opened 2026-06-24
[MiniMax-M3] optimize sparse vLLM metadata for improved performance of vllm-atom

lirui927:
## Motivation

Improve MiniMax-M3 sparse attention performance in the vLLM plugin path by reducing per-step metadata overhead and enabling CUDA Graph capture for the uniform single-token decode path.

## Technical Details

- Marked `MinimaxM3SparseAttentionMetadataBuilder` as `UNIFORM_SINGLE_TOKEN_DECODE` instead of disabling CUDA Graph support entirely.
- Added a decode-only metadata fast path for `max_query_len == 1`, avoiding decode/prefill splitting and CPU-side max reductions for plain decode batches.
- Reused preallocated `qo_indptr` for prefill metadata and replaced exact CPU-side max calculations with existing common metadata upper bounds.
- Moved MiniMax-M3 sparse block-size validation to builder initialization.
- Reused page-16 KV cache views between sparse KV/index insertion and sparse attention execution to avoid duplicate view setup.
- support index cache of vLLM-ATOM

## Test Plan

- Launch MiniMax-M3 vLLM service with ATOM adapters enabled with index cache.
- Run GSM8K accuracy validation:
  - MIXFP8
  - MIXFP4
  - MIXFP8 + kv_fp8
  - MIXFP4 + kv_fp8
 - Performance Test
```
MODEL=/shared/data/amd_int/models/MiniMax-M3-MXFP8
vllm serve "$MODEL" \
    --dtype auto \
    --load-format auto \
    --host localhost \
    --port 8001 \
    --tensor-parallel-size 4 \
    --max-model-len 32768 \
    --max-num-batched-tokens 32768 \
    --block-size 128 \
    --no-async-scheduling \
    --kv-cache-dtype fp8 \
    --no-enable-prefix-caching \
    --language-model-only \
    --no-trust-remote-code \
    --hf-overrides '{"use_index_cache": true, "index_topk_freq": 4}' \
    --additional-config '{"online_quant_config": {"global_quant_config": "ptpc_fp8", "exclude_layer": ["lm_head", "model.embed_tokens", "vision_tower", "multi_modal_projector", "patch_merge_mlp", "*block_sparse_moe"]}}' \
    2>&1 | tee log_m3_mxfp8_vllm.log
```

```
MODEL=/shared/data/amd_int/models/MiniMax-M3-MXFP4
vllm serve "$MODEL" \
    --dtype auto \
    --load-format auto\
    --host localhost \
    --port 8001 \
    --tensor-parallel-size 4 \
    --max-model-len 32768 \
    --max-num-batched-tokens 32768 \
    --block-size 128 \
    --kv-cache-dtype fp8 \
    --no-enable-prefix-caching \
    --language-model-only \
    --no-trust-remote-code \
    --hf-overrides '{"use_index_cache": true, "index_topk_freq": 4}' \
    2>&1 | tee log_m3_mxfp4_vllm_0625.log
```

### GSM8K accuracy
**MIXFP4** 
image

**MIXFP4-kv_fp8** 
image

**MIXFP8** 
image

**MIXFP8-kv_fp8** 
image

### Performance 
**MIXFP4-8192-1024-16**
image

## Submission Checklist

### DIFF
--- a/atom/plugin/vllm/attention/metadata.py
+++ b/atom/plugin/vllm/attention/metadata.py
@@ -434,9 +434,11 @@ class MinimaxM3SparseMetadata:
 
 
 class MinimaxM3SparseAttentionMetadataBuilder(AttentionMetadataBuilder):
-    # MiniMax-M3 sparse attention owns dynamic index/topk/cache state that must
-    # be refreshed outside full cudagraph capture.
-    _cudagraph_support = AttentionCGSupport.NEVER
+    # Only uniform single-token decode is safe to capture. Prefill/mixed batches
+    # still use build(), where variable query lengths and CPU-side max reduction
+    # are allowed. The decode kernels consume per-step seq_lens/block_table from
+    # vLLM's fixed metadata buffers and keep their grids shape-constant.
+    _cudagraph_support = AttentionCGSupport.UNIFORM_SINGLE_TOKEN_DECODE
     reorder_batch_threshold = 1
 
     def __init__(
@@ -450,6 +452,7 @@ class MinimaxM3SparseAttentionMetadataBuilder(AttentionMetadataBuilder):
         del model_runner
         super().__init__(kv_cache_spec, layer_names, config, device)
         logger.info("init MinimaxM3SparseAttentionMetadataBuilder")
+        from atom.model_ops.minimax_m3.sparse_attn import SPARSE_BLOCK_SIZE
         from vllm.config import VllmConfig
 
         assert isinstance(config, VllmConfig)
@@ -458,6 +461,14 @@ class MinimaxM3SparseAttentionMetadataBuilder(AttentionMetadataBuilder):
         self.cache_config = config.cache_config
         self.scheduler_config = config.scheduler_config
         self.block_size = kv_cache_spec.block_size
+        if self.block_size != SPARSE_BLOCK_SIZE:
+            raise ValueError(
+                f"MiniMax-M3 sparse block size must be {SPARSE_BLOCK_SIZE}."
+            )
+        max_num_batched_tokens = self.scheduler_config.max_num_batched_tokens
+        self.prefill_qo_indptr = torch.arange(
+            max_num_batched_tokens + 1, dtype=torch.int32, device=device
+        )
         self._init_reorder_batch_threshold(1, supports_spec_as_decode=True)
 
     def build(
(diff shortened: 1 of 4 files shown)


atom/models/minimax_m3.py
```diff
@@ -446,6 +440,7 @@ def forward(
         return self.o_proj(attn_output)
 
 
+@MiniMaxM3SparseAttentionDecoratorForPluginMode
```
zejunchen-zejun: Hi @lirui927 
After attn refactor(happened in May and merged), we plan to deprecate the decorator to path atom-vllm specific method to frontend attention class, @ganyi1996ppo has a PR to have an abstract attention class, https://github.com/ROCm/ATOM/pull/1334, you can follow this PR

atom/plugin/vllm/attention/layer_mha.py
```diff
@@ -923,3 +953,377 @@ def get_kv_cache_spec(self, vllm_config):
...
+            num_kv_heads=1,
+            head_size=self.head_size,
+            dtype=self.kv_cache_torch_dtype,
+        )
+
+
+AttentionLayerBase.register(SparseMHAIndexerCache)
+
+
+class SparseMHAPagedAttentionImplForVllm(AttentionForVllmMHA):
```
ganyi1996ppo: for this class, can we reuse the atom's already intree impl? maybe inherit from atom's impl, and impl only forward to prevent duplicate code.

lirui927: It should be possible. The vLLM version also not uses Preshuffle now; I’ll adapt it so we can reuse Atom’s implementation.

ganyi1996ppo: we don't need vllm to allocate shuffle kv cache, when we write it in shuffle layout. It is should layout. I think you can fully reuse the atom's impl. you can use prefill path for mix input, and decode path for decode only

lirui927: modified

XiaobingSuper (dismissed):

atom/plugin/vllm/attention/metadata.py
```diff
@@ -471,7 +482,11 @@ def build(
             raise ValueError("ATOM does not support cascade attention yet")
         assert common_attn_metadata is not None
 
-        from atom.model_ops.minimax_m3.sparse_attn import SPARSE_BLOCK_SIZE
+        # Plain decode can have many concurrent requests; each request contributes
+        # exactly one query token, so max_query_len stays 1.
+        if common_attn_metadata.max_query_len == 1:
```
ZhangLirong-amd: Could this path also work with MTP?

lirui927: the previous approach wasn't compatible with MTP. I've revised the implementation so that both regular decode and MTP decode now go through `_build_uniform_decode_metadata`.

lirui927: I've also rerun the accuracy and performance tests, and everything looks good.

ZhangLirong-amd (approved):

valarLip (approved):

zejunchen-zejun (approved):

PR #1350, opened 2026-06-25
[fix](ci): mv plugin test to new node

PerryZhang01:
## Motivation

change the test nodes of vllm-plugin and sgalng plugin to new nodes


gyohuangxin (approved):

gyohuangxin:
@PerryZhang01  Can you fix the failures in https://github.com/ROCm/ATOM/actions/runs/28149992397/job/83394249502?pr=1350? The new runner label should be also included in https://github.com/ROCm/ATOM/blob/main/.github/actionlint.yaml#L16

PR #1353, opened 2026-06-25
fix: convert weight scale shape[N, K/128]  to [ceil(N/128), ceil(K/128)]

JiaoliangYu:
## Motivation

  per_block_fp8 online quantization in ATOM can produce degenerate generation (all-f outputs or immediate EOS) because the weight-scale layout emitted by the online quant path does not match the layout expected by the existing linear FP8 blockscale GEMM backend.

## Technical Details

 In LinearBase.online_quantize_weight for online_quant_type == QuantType.per_1x128, we no longer
  use the generic per-group quant output directly (weight_scale shaped like [N, K/128]). Instead,
  we explicitly quantize weights into 128x128 blockscale format and emit:

  - q_weight: shape [N, K] (same logical weight shape),
  - weight_scale: shape [ceil(N/128), ceil(K/128)].

  This aligns online-quantized weights with the current linear compute path
  (gemm_a8w8_blockscale*), which consumes 128x128 block scales.

Additional fix included:
  Added float8_e4m3fnuz compatibility in online quant dtype checks and avoided redundant normalize
  marking when quant output is already fnuz (relevant on gfx942).

## Submission Checklist

### DIFF
--- a/atom/model_ops/linear.py
+++ b/atom/model_ops/linear.py
@@ -92,6 +92,44 @@ def divide(numerator, denominator):
     return numerator // denominator
 
 
+def quantize_weight_to_fp8_128x128_blockscale(weight, quant_dtype):
+    """Quantize a 2D weight to FP8 with 128x128 block scales.
+
+    Returns:
+        q_weight: quantized weight with the same shape as input ``weight``.
+        scale: per-block scale with shape ``(ceil(N/128), ceil(K/128))``.
+    """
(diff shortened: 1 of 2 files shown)


lihaoyang-amd (approved):
LGTM

JiaoliangYu:
reference https://github.com/ROCm/ATOM/pull/1411

PR #1369, opened 2026-06-26
Enable TBO Support & Fix Accuracy Regressions for Kimi K2.5

jpy794:
## Motivation

Kimi K2.5 inference under Data-Parallel Attention (DPA) combined with Two-Batch Overlap (TBO) exposed several gaps that either crashed the engine or left performance on the table. This PR enables the DPA + TBO path end-to-end: it fixes the fused-MoE fallback and tensor lifetimes in the TBO overlap, aligns cross-DP prefill admission with TBO's two-batch requirement, and extends persistent MLA to multi-rank DP.

## Technical Details

- **Persistent MLA for DP attention** (`attention_mla.py`): relax `use_persistent_mode` from "single-rank only" (`not (dp_size > 1)`) to `dp_size <= 8`, so persistent MLA also runs in the multi-rank DP configuration used by Kimi K2.5.

- **Fix fused MoE on the DPA fallback path** (`moe.py`, `topK.py`):
  - In the DP-attn fallback (`dp_size > 1`, no MORI all2all), MoE runs after `all_gather_with_padding`, so the token dim can grow to `dp_size ×` the per-rank max. Scale `max_num_tokens` for the topK / fused-MoE metadata accordingly to avoid undersized buffers.
  - Only select the MORI all2all path when expert parallel is actually enabled (`enable_expert_parallel`), so DPA-without-EP correctly falls back instead of assuming all2all.

- **Fix TBO tensor live range** (`moe.py`): add a per-(role, ubatch) `_TBO_KEEPALIVE` holder around the all-gather and reduce-scatter comm/compute switches. Under TBO the source/output tensors of in-flight collectives could be freed before the overlapping ubatch waited on the comm; the keepalive defers release to the next same-role hold, which is past the wait point.

- **Two-batch-aware prefill alignment** (`scheduler.py`, `prefill_delayer.py`): TBO prefill splitting needs at least two local prefill requests per DP rank. Replace `_can_admit_head_prefill` (boolean) with `_count_admittable_head_prefills(limit)` and a `_prefill_delayer_readiness()` helper that reports both "has any prefill" and "alignment-ready" (`>= 2` requests when TBO is on, `>= 1` otherwise). `PrefillDelayer` gains a 4th MAX-reduce slot (`local_alignment_ready`) so prefill is delayed until every DP rank can launch a full two-batch, not just until one rank has a request.

- **Fix TBO prefill ubatch DP offsets** by propagating per-ubatch per-rank token counts through ForwardContext, then rebuilding ubatch-local DPMetadata inside UBatchWrapper. This prevents DP all_gatherv/reduce_scatterv from using full-batch offsets for individual ubatches.

- **Zero MoE all-gather padding rows** before fused-MoE routing/sort/dispatch. Padding rows are later sliced away, but they still participate in fused MoE internals; leaving them uninitialized can introduce NaN/Inf garbage, perturb expert buckets/shared scratch, and corrupt real tokens.

## Bugfix Validation

Bad TBO run before MoE padding fix: GSM8K flexible 0.8999, with 125 invalid responses and 238 corrupted outputs.
After fix: GSM8K flexible 0.9742, invalid down to 1, corrupted outputs down to 0.

## Perf Benchmark Plan

We tested Kimi K2.5 MXFP4 end-to-end inference on MI355X with ROCm 7.2.3, TP4.

The comparison includes:

- baseline without DPA / TBO
- DPA only
- DPA + TBO

## Test Results

Numbers in parentheses are throughput/GPU changes relative to the baseline.

| Conc | baseline throughput/GPU | baseline interactivity | DPA throughput/GPU | DPA interactivity | DPA+TBO throughput/GPU | DPA+TBO interactivity |
|---:|---:|---:|---:|---:|---:|---:|
| 4 | 898.9 | 104.95 | 611.7 (-32.0%) | 73.53 | 610.2 (-32.1%) | 73.63 |
| 8 | 1502.8 | 89.61 | 1061.4 (-29.4%) | 65.48 | 1077.1 (-28.3%) | 65.67 |
| 16 | 2045.6 | 61.85 | 1653.5 (-19.2%) | 51.93 | 1677.6 (-18.0%) | 52.86 |
| 32 | 2964.5 | 43.46 | 2542.4 (-14.2%) | 38.51 | 2588.0 (-12.7%) | 39.21 |
| 64 | 3946.8 | 29.06 | 3761.2 (-4.7%) | 28.52 | 3961.7 (+0.4%) | 30.09 |
| 128 | 4984.5 | 19.81 | 5133.1 (+3.0%) | 21.04 | 5745.4 (+15.3%) | 23.55 |

At higher concurrency, DPA and TBO show a higher throughput ceiling, with DPA+TBO reaching +15.3% throughput/GPU over the baseline at conc=128.

## Submission Checklist

### DIFF
--- a/atom/model_engine/prefill_delayer.py
+++ b/atom/model_engine/prefill_delayer.py
@@ -7,7 +7,7 @@ the "mixed prefillable status → delay" core that fixes our 1k/1k workload's
 
 Mechanism (per scheduler tick):
   1. Each DP rank reports its local state via cpu all_gather:
-       (local_prefillable, watermark_force_allow)
+       (local_prefillable, local_prefill_sufficient, watermark_force_allow)
   2. Compute `prefillable_status` ∈ {all, none, mixed}:
        - "all"   → every rank has a new prefill ready  → allow (8-way aligned)
        - "none"  → no rank has any prefill             → allow (vacuous)
@@ -97,10 +97,11 @@ class PrefillDelayer:
         # Encoding:
         #   slot 0 = local_prefillable          (MAX → "any rank prefillable")
         #   slot 1 = local_force                (MAX → "any rank forces allow")
-        #   slot 2 = NOT local_prefillable      (MAX → "any rank lacks prefill")
+        #   slot 2 = NOT local_prefill_sufficient
+        #            (MAX → "any rank lacks the configured dense batch")
         # Then prefillable_status:
-        #   any_prefillable AND any_not_prefillable → "mixed"
-        #   any_prefillable AND NOT any_not_prefillable → "all"
+        #   any_prefillable AND any_not_sufficient → "mixed"
+        #   any_prefillable AND NOT any_not_sufficient → "all"
         #   NOT any_prefillable → "none"
         # Single all_reduce, 3 int64s on cpu — negligible overhead.
         self._reduce_buf = torch.zeros(3, dtype=torch.int64, device="cpu")
@@ -132,6 +133,7 @@ class PrefillDelayer:
         self,
         local_prefillable: bool,
         token_usage: float,
+        local_prefill_sufficient: Optional[bool] = None,
     ) -> bool:
         """
         Returns True iff this rank is allowed to admit new prefills this tick.
@@ -139,10 +141,16 @@ class PrefillDelayer:
         Args:
             local_prefillable: this rank has at least one new prefill ready
                 (i.e. self.waiting non-empty and admission would succeed).
+            local_prefill_sufficient: this rank also meets the configured
+                token/request density target. Defaults to local_prefillable
+                for compatibility with existing callers.
             token_usage: fraction of KV cache blocks currently in use
                 (used_blocks / total_blocks ∈ [0, 1]). Used by the
                 low-watermark safety valve.
         """
+        if local_prefill_sufficient is None:
+            local_prefill_sufficient = local_prefillable
+
         # Local "force allow" if KV cache is underutilized — don't delay
         # when GPU is starving. Only meaningful if this rank actually has
         # a prefill to push through (otherwise force_allow is a no-op).
@@ -157,7 +165,7 @@ class PrefillDelayer:
         # Cross-DP MAX-reduce: 3 booleans encoded as int64.
         self._reduce_buf[0] = 1 if local_prefillable else 0
         self._reduce_buf[1] = 1 if force else 0
-        self._reduce_buf[2] = 0 if local_prefillable else 1
+        self._reduce_buf[2] = 0 if local_prefill_sufficient else 1
         torch.distributed.all_reduce(
             self._reduce_buf,
             op=torch.distributed.ReduceOp.MAX,
@@ -165,11 +173,11 @@ class PrefillDelayer:
         )
         any_prefillable = int(self._reduce_buf[0].item()) > 0
         force_max = int(self._reduce_buf[1].item())
-        any_not_prefillable = int(self._reduce_buf[2].item()) > 0
+        any_not_sufficient = int(self._reduce_buf[2].item()) > 0
 
         # Derive 3-way status: all / none / mixed.
         prefillable_max = 1 if any_prefillable else 0
-        prefillable_min = 0 if any_not_prefillable else 1
+        prefillable_min = 0 if any_not_sufficient else 1
 
         # Watermark short-circuit: ANY rank below the watermark forces all
         # ranks to allow this tick. Without this the delayer can stall a
@@ -211,7 +219,7 @@ class PrefillDelayer:
                     f"[PrefillDelayer] DELAY: count={self._delayed_count} "
                     f"elapsed={elapsed_ms:.1f}ms "
                     f"any_prefillable={any_prefillable} "
-                    f"any_not_prefillable={any_not_prefillable}"
+                    f"any_not_sufficient={any_not_sufficient}"
                 )
             self._maybe_log()
             return False
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -2477,6 +2477,17 @@ class FusedMoE(torch.nn.Module):
                 ),
                 dim=0,
             )
+        # In the DP-attn fallback path (dp>1, no MORI all2all), MoE runs
+        # after all_gather_with_padding, so the token dim can be dp_size times
+        # the per-rank max.
+        moe_max_num_tokens = atom_config.max_num_batched_tokens
+        if (
+            self.moe_parallel_config.dp_size > 1
+            and not self.moe_parallel_config.use_all2all_kernels
+            and atom_config.enable_dp_attention
+        ):
+            moe_max_num_tokens *= self.moe_parallel_config.dp_size
+
         if fuse_shared_experts and self.num_fused_shared_experts > 0:
             init_aiter_topK_meta_data(
                 n_routed_experts=self.global_num_experts,
@@ -2489,7 +2500,7 @@ class FusedMoE(torch.nn.Module):
                     if is_rocm_aiter_fuse_routed_scaling_factor()
                     else 1 / self.routed_scaling_factor
                 ),
-                max_num_tokens=atom_config.max_num_batched_tokens,
+                max_num_tokens=moe_max_num_tokens,
                 is_EP=self.use_ep,
             )
         if fuse_shared_experts:
@@ -2529,7 +2540,7 @@ class FusedMoE(torch.nn.Module):
             moe_parallel_config=self.moe_parallel_config,
             in_dtype=atom_config.torch_dtype,
             a_quant_dtype=a_quant_dtype,
-            max_num_tokens=atom_config.max_num_batched_tokens,
+            max_num_tokens=moe_max_num_tokens,
             has_bias=self.has_bias,
             # is_act_and_mul=True,
             is_lora_enabled=False,
--- a/atom/model_ops/moe.py
+++ b/atom/model_ops/moe.py
@@ -87,6 +87,9 @@ class MoEActivationQuant(Enum):
         return MoEActivationQuant.BF16
 
 
+_TBO_KEEPALIVE: dict[tuple[str, int], tuple[torch.Tensor, ...]] = {}
+
+
 class FusedMoeWeightScaleSupported(Enum):
     """Supported quantization strategies for MoE weight scales."""
 
@@ -3459,6 +3462,26 @@ class FusedMoE(torch.nn.Module):
             hidden_states, router_logits, self.layer_name
         )
 
+    def _tbo_keepalive_slot(self) -> int:
+        try:
+            from atom.utils.tbo.ubatching import tbo_current_ubatch_id
+
+            return tbo_current_ubatch_id()
+        except Exception:
+            return 0
+
+    def _hold_tbo_keepalive(self, role: str, *tensors: torch.Tensor) -> None:
+        tensors = tuple(tensor for tensor in tensors if tensor is not None)
+        if tensors:
+            # Keep one previous tensor set per ubatch/role alive globally.
+            # The next same-role hold, often in the next MoE layer, happens
+            # after this ubatch has waited on the prior comm work, so
+            # overwriting here is the delayed safe release point.
+            key = (role, self._tbo_keepalive_slot())
+            if key in _TBO_KEEPALIVE:
+                del _TBO_KEEPALIVE[key]
+            _TBO_KEEPALIVE[key] = tensors
+
     def forward_impl_graph(
         self, hidden_states: torch.Tensor, router_logits: torch.Tensor
     ):
@@ -3489,6 +3512,7 @@ class FusedMoE(torch.nn.Module):
                 )
 
                 tbo_yield_and_switch_from_compute_to_comm()
+                self._hold_tbo_keepalive("ag_source", hidden_states, router_logits)
 
             (
                 hidden_states,
@@ -3501,6 +3525,7 @@ class FusedMoE(torch.nn.Module):
 
             if _tbo:
                 tbo_switch_to_compute_sync()
+                self._hold_tbo_keepalive("ag_output", hidden_states, router_logits)
 
         # Matrix multiply.
         final_hidden_states = self.quant_method.apply(
@@ -3526,6 +3551,7 @@ class FusedMoE(torch.nn.Module):
         if use_dp_gather_scatter:
             if _tbo:
                 tbo_yield_and_switch_from_compute_to_comm()
+                self._hold_tbo_keepalive("rs_source", final_hidden_states)
             if dp_eager_mode:
                 final_hidden_states = reduce_scatterv(
                     final_hidden_states, sizes, dp_group
@@ -3536,6 +3562,7 @@ class FusedMoE(torch.nn.Module):
                 )
             if _tbo:
                 tbo_switch_to_compute_sync()
+                self._hold_tbo_keepalive("rs_output", final_hidden_states)
 
         if self.reduce_results and (self.tp_size > 1 or self.ep_size > 1):
             # Default set to False. (May have to add shared expert outputs.)
--- a/atom/model_ops/topK.py
+++ b/atom/model_ops/topK.py
@@ -25,7 +25,13 @@ def is_rocm_aiter_fusion_shared_expert_enabled_for_quant_config(
     # layout (set by the vLLM plugin under DP+EP); disable it there.
     if dp_size > 1 and config.moe_ep_flatten_tp_across_dp:
         return False
-    if dp_size > 1 and _has_module("mori") and config.enable_dp_attention:
+    use_mori_all2all = (
+        dp_size > 1
+        and _has_module("mori")
+        and config.enable_dp_attention
+        and config.enable_expert_parallel
+    )
+    if use_mori_all2all:
         return False
 
     if quant_config is not None and shared_expert_prefix is not None:
@@ -66,15 +72,6 @@ def is_rocm_aiter_fusion_shared_expert_enabled_for_quant_config(
                 return False
             break
 
-    dp_size = config.parallel_config.data_parallel_size
-    use_mori_all2all = (
-        dp_size > 1
-        and _has_module("mori")
-        and config.enable_dp_attention
-        and config.enable_expert_parallel
-    )
-    if use_mori_all2all:
-        return False
     return True
 
 
--- a/atom/model_ops/topK.py
+++ b/atom/model_ops/topK.py
@@ -66,6 +66,15 @@ def is_rocm_aiter_fusion_shared_expert_enabled_for_quant_config(
                 return False
             break
 
+    dp_size = config.parallel_config.data_parallel_size
+    use_mori_all2all = (
+        dp_size > 1
+        and _has_module("mori")
+        and config.enable_dp_attention
+        and config.enable_expert_parallel
+    )
+    if use_mori_all2all:
+        return False
     return True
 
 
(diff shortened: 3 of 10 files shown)


atom/model_ops/topK.py
```diff
@@ -62,6 +62,15 @@ def is_rocm_aiter_fusion_shared_expert_enabled_for_quant_config(
                 return False
             break
 
+    dp_size = config.parallel_config.data_parallel_size
```
ZhangLirong-amd: This is a duplicate? You can check line 24. 
`    if dp_size > 1 and _has_module("mori") and config.enable_dp_attention:
        return False`

atom/model_ops/moe.py
```diff
@@ -2375,6 +2268,17 @@ def __init__(
...
+        # In the DP-attn fallback path (dp>1, no MORI all2all), MoE runs
+        # after all_gather_with_padding, so the token dim can be dp_size times
+        # the per-rank max.
+        moe_max_num_tokens = atom_config.max_num_batched_tokens
+        if (
+            self.moe_parallel_config.dp_size > 1
+            and not self.moe_parallel_config.use_all2all_kernels
+            and atom_config.enable_dp_attention
+        ):
+            moe_max_num_tokens *= self.moe_parallel_config.dp_size
```
ZhangLirong-amd, line 2492: I don't understand why we need moe_max_num_tokens *= self.moe_parallel_config.dp_size here.. In all_gahter and model runner, we have padded, * dp_size  here will make BS large and kernel bad perf

atom/model_ops/topK.py
jpy794: Thanks for pointing out. That's a rebase error (fixed now). Here we try to enable shared expert fusion for DPA for allgather/reducescatter MoE path (not mori all2all).

atom/model_ops/moe.py
jpy794: Here, we only increase the size of the preallocated internal buffer in FusedMoE, not the actual batch size used in the forward pass. This internal buffer needs to be large enough to accommodate tokens from all DP ranks, so we multiply by dp_size, similar to what we've already done for the all-gather / reduce-scatter buffers.

atom/model_engine/prefill_delayer.py
```diff
@@ -7,11 +7,12 @@
 
 Mechanism (per scheduler tick):
   1. Each DP rank reports its local state via cpu all_gather:
-       (local_prefillable, watermark_force_allow)
+       (local_prefillable, local_alignment_ready, watermark_force_allow)
```
ZhangLirong-amd: Your changes seem to require >=2 bs for TBO to be ready; does this approach has performance improvement? Or whether it will affect old performance.

jpy794: I’ve made this behavior configurable via an environment variable (disabled by default to avoid affecting existing performance).

Below is a Kimi K2.5 performance comparison (conc=128, isl=8k, osl=1k) with `ATOM_PREFILL_DELAYER_REQUIRED_PREFILLS=0/2`, about 25% throughput gain observed.


`ATOM_PREFILL_DELAYER_REQUIRED_PREFILLS=0`
```
============ Serving Benchmark Result ============
Successful requests:                     512       
Benchmark duration (s):                  235.88    
Total input tokens:                      3775394   
Total generated tokens:                  473911    
Request throughput (req/s):              2.17      
Output token throughput (tok/s):         2009.10   
Total Token throughput (tok/s):          18014.52  
---------------Time to First Token----------------
Mean TTFT (ms):                          2994.52   
Median TTFT (ms):                        786.19    
P99 TTFT (ms):                           17830.07  
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          59.09     
Median TPOT (ms):                        61.84     
P99 TPOT (ms):                           77.20     
---------------Inter-token Latency----------------
Mean ITL (ms):                           59.33     
Median ITL (ms):                         27.85     
P99 ITL (ms):                            626.02    
----------------End-to-end Latency----------------
Mean E2EL (ms):                          57912.88  
Median E2EL (ms):                        58740.95  
P99 E2EL (ms):                           81474.65  
==================================================
```

`ATOM_PREFILL_DELAYER_REQUIRED_PREFILLS=2`
```
============ Serving Benchmark Result ============
Successful requests:                     512       
Benchmark duration (s):                  178.86    
Total input tokens:                      3775394   
Total generated tokens:                  473911    
Request throughput (req/s):              2.86      
Output token throughput (tok/s):         2649.66   
Total Token throughput (tok/s):          23758.06  
---------------Time to First Token----------------
Mean TTFT (ms):                          3435.93   
Median TTFT (ms):                        1459.46   
P99 TTFT (ms):                           16975.80  
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          43.29     
Median TPOT (ms):                        44.66     
P99 TPOT (ms):                           54.32     
---------------Inter-token Latency----------------
Mean ITL (ms):                           43.39     
Median ITL (ms):                         27.81     
P99 ITL (ms):                            742.68    
----------------End-to-end Latency----------------
Mean E2EL (ms):                          43596.09  
Median E2EL (ms):                        43810.70  
P99 E2EL (ms):                           60233.29  
==================================================
```

ZhangLirong-amd: sure, that's good news. We will test `ATOM_PREFILL_DELAYER_REQUIRED_PREFILLS=2` on deepseek v4 and other models if it's indeed effective

ZhangLirong-amd:
And could you solve the conflicts and we continue the next step？

jpy794:
@ZhangLirong-amd Hello, I've rebased to main. I'll add more performance data tomorrow.

ATOM config: `--no-enable_chunked_prefill --enable-tbo --enable-dp-attention`
Benchmark dataset: isl=8k osl=1k random ratio=0.8 conc=128 prompt=512

branch | total tok/s | output tok/s | median TPOT | interactivity tok/s/user | GSM8K flexible | GSM8K strict
-- | -- | -- | -- | -- | -- | --
rf-dpa-tbo-rebase (delay factor=1) | 25424.70 | 2835.53 | 34.46 ms | 29.02 | 0.9515 | 0.9469
main | 20367.48 | 2271.52 | 50.90 ms | 19.65 | 0.9136 | 0.9128

atom/model_ops/moe.py
```diff
@@ -3386,6 +3420,7 @@ def forward_impl_graph(
                 )
 
                 tbo_yield_and_switch_from_compute_to_comm()
+                self._hold_tbo_keepalive("ag_source", hidden_states, router_logits)
```
ZhangLirong-amd, line 3515: Hi, I have a question, why we need this in all_gather/reduce_scatter with TBO, other models we enabled before didn't meet issues in old logic

jpy794: This fixes the use-after-free race for the tensor allocated in stream A and used in stream B. Without this fix the intermediate tensor could be reused by pytorch in the allocating stream before its real use by kernels in the other stream. 

You can see the difference in gsm8k (0.9136 vs 0.9515) for Kimi k2.5.

I think other models should have the same race issue, not sure if it's because some minor difference in code path hide this race condition.

```
Race without tbo_keepalive
==========================

Time ─────────────────────────────────────────────────────────────────────>

Compute stream:  produce T ─────────────── drop ref ───── alloc U / reuse T storage
                                      │                         │
                                      │ CPU enqueues AG/RS(T)   │
                                      ▼                         ▼
Comm stream:                         AG/RS reads T ─────────────X
                                                               corrupted / UAF
```

jpy794:
@ZhangLirong-amd Hi, we have refactored the TBO scheduling optimization based on PR #1437 and included a bug fix for an issue in #1437. With this fix, we observed a 13.6% throughput improvement.

We have also added more ablation results below and provided the corresponding [microbenchmark scripts](https://gist.github.com/jpy794/ac099e85b7e1273ec8bee65faa7fc079) for the tbo keepalive race condition fix. We hope these additions make the changes easier to evaluate and review.

Recently, however, we noticed that PR #1537 reverted #1437 (`tbo dp delay gate`), while PR #1503 reverted #1474 (`dp uniform decode`). Since these PRs explore optimization ideas similar to those in our PR, their recent reverts have made it somewhat difficult for us to understand the current upstream direction and how we should proceed.

It is possible that the bug fix in commit `2a5517ea527` could help address the performance issue that led to the revert of #1437.

We would greatly appreciate **any guidance** on how we could adjust or restructure this PR to make it easier to review and merge. We are also happy to **split the changes into multiple smaller PRs** if that would better align with the upstream development process.

Please also feel free to reach out to us for an online discussion if that would be more convenient—we would be happy to walk through the design, implementation, and evaluation results in more detail.

## Motivation

This PR contains several related TBO changes. If the combined scope is too large for one review, we are happy to discuss splitting it into smaller PRs in the following order:

1. TBO DP-prefill batch scheduling improvements based on PR #1437 .
2. TBO multi-stream race correctness fix.
3. The remaining TBO performance fixes and Kimi K2.5 support.

This PR completes the TBO work in four areas: DP-prefill scheduling, multi-stream correctness, avoiding an unintended padded all-gather path, and Kimi K2.5 enablement.

PR #1437 has already upstreamed a cross-DP prefill batching strategy similar to the one proposed here, so this PR does not duplicate that work. Instead, this PR provides the following improvements:

- **DP-prefill scheduling:** the configurable 10k threshold plus the cross-rank sufficiency fix reaches 25,517 tok/s, 13.6% above the original delay policy's 22,466 tok/s. Compared with the 10k threshold alone, the sufficiency fix improves throughput by 3.45%-4.37% and reduces P99 TTFT by about 25.5% across three paired reruns.
- **Multi-stream correctness:** keeping TBO intermediates alive eliminates allocator-reuse corruption in the microbenchmark and restores GSM8K exact match from 91.74% to 97.50%, a 5.76 percentage-point improvement.
- **TBO performance path:** preserving `dp_uniform_decode` prevents uniform decode ubatches from unnecessarily taking the padded all-gather path.
- **Kimi K2.5 TBO support:** the persistent MLA and corrected DPA MoE buffer sizing remove the GQA-ratio-64 initialization failure and the 16,384-versus-40,960-token metadata overflow; fused shared experts are also enabled for the non-MORI intra-node DPA path.

## Technical Details

### 1. TBO scheduling optimization

- `970a64ab` adds `--prefill-batch-token-threshold`, allowing the token threshold used by the prefill delayer to be tuned independently of `max_num_batched_tokens`. A value of `0` preserves the upstream behavior and uses `max_num_batched_tokens` as the threshold.
- `2a5517ea527` fixes incomplete cross-DP alignment in the delayer introduced by #1437.

Previously, the scheduler folded two different states into `local_prefillable`: whether the rank could admit any prefill and whether it had accumulated enough waiting tokens for a dense batch. The delayer's cross-rank reduction therefore only distinguished ranks with and without a prefill. Once every rank had at least one admissible request, it released all ranks even if some ranks had only a sparse batch, which left single-request prefills in the schedule.

The fix reports both states explicitly:

- `local_prefillable`: at least one prefill can be admitted on this rank.
- `local_prefill_sufficient`: an admissible prefill exists and the locally waiting tokens reach the configured dense-batch threshold.

The cross-DP reduction now delays while any rank is not sufficient, while retaining the existing watermark and timeout escape paths. This aligns the release condition across all DP ranks around batch sufficiency rather than mere request availability.

Benchmark configuration: 512 prompts at concurrency 128, random ISL/OSL 8192/1024 with range ratio 0.8, `max_num_batched_tokens=16384`, chunked prefill and prefix caching disabled, TP=4, seed 0.

| Case | Duration (s) | Total throughput (tok/s) | Mean TTFT (ms) | P99 TTFT (ms) | Mean TPOT (ms) |
|---|---:|---:|---:|---:|---:|
| No delay | 312.37 | 13,603.44 | 6,670.00 | 44,822.00 | 75.68 |
| Original delay, threshold 16,384 | 189.14 | 22,466.08 | 3,724.99 | 16,042.04 | 45.61 |
| Threshold 10k (3-run mean) | 173.05 | 24,555.72 | 3,094.82 | 19,415.30 | 42.09 |
| Threshold 10k + sufficient fix (3-run mean) | 166.53 | 25,517.01 | 3,048.23 | 14,468.06 | 40.32 |

The 10k rows report the arithmetic mean of three interleaved reruns. In particular, they supersede an earlier 186.52s sufficient-fix run that was affected by run order / machine state.

Across the three paired reruns, the sufficient fix reduced duration by 3.33%-4.19%, increased total throughput by 3.45%-4.37%, and reduced P99 TTFT by about 25.5% relative to threshold-only. The batching trace also shows the scheduling effect directly:

| Case | Prefill batches | Requests per batch | Mean tokens/batch | Median tokens/batch |
|---|---:|---|---:|---:|
| No delay | 437 | 1: 362, 2: 75 | 8,639.3 | 7,560.0 |
| Original delay | 280 | 1: 48, 2: 232 | 13,483.5 | 14,606.5 |
| Threshold 10k | 273 | 1: 34, 2: 239 | 13,829.3 | 14,602.0 |
| Threshold 10k + sufficient fix | 256 | 2: 256 | 14,747.6 | 14,808.5 |

The sufficient check eliminates the remaining single-request prefill batches and produces 256 consistently dense two-request batches.

### 2. TBO correctness bug fix

`fa5100c5733e` fixes a multi-stream lifetime race in the TBO MoE path. TBO switches work between compute and communication streams, but returning from a collective or switching the active stream does not by itself guarantee that every asynchronous consumer has finished reading the input storage. If the last Python reference to an intermediate tensor is released too early, PyTorch's caching allocator may immediately recycle that storage for a new tensor while another stream or peer is still reading it. The result is silent corruption rather than necessarily a launch failure.

The fix keeps the previous all-gather/reduce-scatter input and output tensors alive per TBO ubatch and role. They are released only at a later same-role hold, after the ubatch has crossed the synchronization point for the prior communication work.

The [`multistream_keepalive_race_microbench.py`](https://gist.github.com/jpy794/ac099e85b7e1273ec8bee65faa7fc079) repro isolates this lifetime rule. It creates an intermediate tensor on the producer/default stream, queues a delayed pointer-based read on a consumer stream, drops the Python reference, and immediately allocates a same-shape poison tensor. The keepalive variant retains the source until an event recorded on the consumer stream completes.

```text
unsafe    ptr_reuse=100/100 corrupt_iters=100/100 corrupt_elements=25600
keepalive ptr_reuse=0/100 corrupt_iters=0/100 corrupt_elements=0
PASS: delayed keepalive prevents cross-stream allocator-reuse corruption
```

The end-to-end GSM8K result confirms that this is observable model corruption even though the server starts successfully without the fix. On the full 1,319-example, 5-shot, temperature-0 evaluation at concurrency 128:

| Version | Correct | Exact match |
|---|---:|---:|
| With `fa5100c` | 1,286 / 1,319 | 97.4981% +/- 0.4302% |
| Revert `fa5100c` | 1,210 / 1,319 | 91.7362% +/- 0.7584% |

Reverting the fix loses 5.7619 percentage points and 76 net correct answers (80 correct-to-wrong and 4 wrong-to-correct).

This race is independent of [ROCm/aiter#4082](https://github.com/ROCm/aiter/pull/4082). That issue concerns synchronization inside custom collective kernels before callers reuse peer-read input buffers; this PR fixes a separate ATOM/TBO ownership problem where Python tensor references can expire while asynchronous work on another stream still uses their storage.

### 3. TBO performance bug fix

`906179c63af7` propagates `dp_uniform_decode` into each TBO ubatch `ForwardContext`. Without this field, a globally uniform decode batch can be misclassified inside the split TBO context and take the all-gather-with-padding path. Preserving the parent context's uniform-decode state keeps prefill/decode routing consistent and avoids the unnecessary padded all-gather code path.

### 4. Kimi K2.5 TBO support

- `cfe57e2d` enables persistent MLA for DPA configurations up to DP=8. Kimi K2.5 uses GQA ratio 64, which the non-persistent AITER kernel does not support. Reverting this change fails during initialization:

  ```text
  AITER: fp8/fp8 with gqa_ratio=64 only supports persistent mode
  RuntimeError: Engine Core Mgr received SHUTDOWN during initialization
  ```

- `ff364f91c728` sizes fused-MoE and top-k metadata buffers for the actual DPA gather/scatter token domain. In the DP-attention fallback without MORI all-to-all, MoE receives the padded all-gather result, whose token dimension can be `dp_size * max_num_batched_tokens`; reserving only the per-rank budget under-allocates metadata. Reverting the fix reproduces:

  ```text
  AssertionError: AITER topK meta data support 16384 tokens, but got 40960 tokens
  ```

- `1e3ee2e6a0` enables fused shared experts for the intra-node DPA gather/scatter path when MORI all-to-all is not actually in use. The MORI DP+EP layout remains excluded, while the non-MORI path can use the faster fused implementation. The revert ablation did not produce a valid performance comparison because it stalled in the AITER JIT/build-lock path before server readiness.

## Test Plan

Run the prefill-delay performance ablation with the following setup and compare both throughput/latency and emitted prefill batch shapes:

- Concurrency: 128
- Number of prompts: 512
- Input/output lengths: random ISL=8192 and OSL=1024, range ratio 0.8
- `max_num_batched_tokens=16384`
- Chunked prefill: disabled
- Prefix caching: disabled
- TP=4, seed=0

Cover no delay, the original 16,384-token threshold, the custom 10k threshold, and the 10k threshold with the sufficient fix. Interleave repeated runs of the last two cases to avoid run-order bias.

Run the tensor-lifetime microbenchmark on an otherwise idle GPU:

```bash
HIP_VISIBLE_DEVICES=0 python tools/multistream_keepalive_race_microbench.py
```

Run the full GSM8K evaluation with and without `fa5100c`, and verify Kimi K2.5 initialization plus inference under TBO/DPA.

ZhangLirong-amd:
@jpy794 ,sure, please rebase to main and solve the conflict, we will test this branch on dsv4 tbo to make sure its performance.

jpy794:
@ZhangLirong-amd Hi, I've rebased to main with changes from #1437 included. I'm also glad to help run some dsv4 TBO benchmark to verify the performance, if you could provide some detailed benchmark setup.

ZhangLirong-amd:
> @ZhangLirong-amd Hi, I've rebased to main with changes from #1437 included. I'm also glad to help run some dsv4 TBO benchmark to verify the performance, if you could provide some detailed benchmark setup.

Sure, you can rty
```
GPU_MAX_HW_QUEUES=5 ATOM_DISABLE_MMAP=true AITER_BF16_FP8_MOE_BOUND=0 ATOM_MOE_GU_ITLV=1 python -m atom.entrypoints.openai_server   --model /data/models/DeepSeek-V4-Pro/   --kv_cache_dtype fp8   -tp 8   --gpu-memory-utilization 0.85   --server-port 7777  --torch-profiler-dir ./log --enable-dp-attention --enable-tbo


python -m atom.benchmarks.benchmark_serving   --model=/data/models/DeepSeek-V4-Pro/ --backend=vllm --base-url=http://localhost:7777   --dataset-name=random --random-input-len=8192 --random-output-len=1024 --random-range-ratio 0.8   --num-prompts=10240 --max-concurrency=1024   --request-rate=inf --ignore-eos
```

ZhangLirong-amd:
@jpy794 ,hi, seems I meet regression, Total Token throughput (tok/s): 42820.94, in nightly benchmark ,it's about 48000`49000

benenzhu:
For kimi for conc128 + TP4, though haven't tested on other models.
https://inferencex.semianalysis.com/inference?unofficialRun=29234093839&i_disagg=agg
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/29234093839/job/86764699211

Metric | Before | After | Improvement
-- | -- | -- | --
Throughput | 5370 | 5971 | +11.19%
Interactivity | 19.16 | 25.20 | +31.52%

With configs: https://github.com/SemiAnalysisAI/InferenceX/compare/main...amd/zty_test3
```
    --enable-dp-attention \
    --enable-tbo \
    --prefill-batch-token-threshold 10240 \
```

image

jpy794:
I could also reproduce the regession in dsv4. Total throughput dropped from 19,345.53 to 18,940.36 with tp8, conc=128, prompts=512. I'm currently investigating the root cause.

PR #1370, opened 2026-06-26
support online quant for quark models

gbyu-amd:
## Motivation

There are cases where we want to online quant the quark models for some specific modules. For example, quant the bf16 attn linear layers to PTPC fp8 for https://huggingface.co/amd/MiniMax-M3-MXFP4, which is already quark quant model.

image

## Submission Checklist

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -300,7 +300,7 @@ class QuantizationConfig:
             self.quant_method = self.hf_quant_config.get("quant_method", "")
 
         # Online quantization: re-quantize float / FP8 / MXFP4 / MXFP8 / Quark
-        # models at load time. 
+        # models at load time.
         self.online_quant = False
         self.online_quant_config_raw = online_quant_config
         self.online_global_spec: LayerQuantConfig = LayerQuantConfig()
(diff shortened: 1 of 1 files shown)


haoyangli0109:
Hi, 
@gbyu-amd  
It makes sense that bf16 runs successfully, since it doesn’t require dequantization of the weights.
Quark supports multiple quantization formats, and completely opening up the Quark options could pose risks. For example, if we try to perform mxfp4 quantization on a ptpc_fp8 model, problems will arise.

Merging this PR carries some risk, but as long as you confirm there won’t be any misuse, I believe it can be merged.

Based on this PR, we’ll submit another PR to handle online quantization of Quark models in common scenarios next week.

gbyu-amd:
> Hi, @gbyu-amd It makes sense that bf16 runs successfully, since it doesn’t require dequantization of the weights. Quark supports multiple quantization formats, and completely opening up the Quark options could pose risks. For example, if we try to perform mxfp4 quantization on a ptpc_fp8 model, problems will arise.
> 
> Merging this PR carries some risk, but as long as you confirm there won’t be any misuse, I believe it can be merged.
> 
> Based on this PR, we’ll submit another PR to handle online quantization of Quark models in common scenarios next week.

Thanks @haoyangli0109 , it would be great if you could support more general cases regarding quark models.

valarLip (approved):

PR #1101, opened 2026-06-05
[plugin][vllm] support dp and ep

gbyu-amd:
## Motivation

This PR added DP + EP support for vllm plugin.

## Test Plan
Functionality and accuracy under DP + EP.

### DeepSeek-R1-0528 DP4 + EP4

```bash
local-completions ({'model': '/workspace/shared/data/amd_int/models/deepseek-ai/DeepSeek-R1-0528', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 65, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 3, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     3|exact_match|↑  |0.9469|±  |0.0062|
|     |       |strict-match    |     3|exact_match|↑  |0.9424|±  |0.0064|
```

### DeepSeek-R1-0528 DP8 + EP8
```bash
local-completions ({'model': '/workspace/shared/data/amd_int/models/deepseek-ai/DeepSeek-R1-0528', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 65, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 3, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     3|exact_match|↑  |0.9462|±  |0.0062|
|     |       |strict-match    |     3|exact_match|↑  |0.9462|±  |0.0062|
```

### DeepSeek-R1-0528-MXFP4-MTP-MoEFP4 DP4 + EP4
```bash
local-completions ({'model': '/workspace/shared/data/amd_int/models/deepseek-ai/DeepSeek-R1-0528-MXFP4-MTP-MoEFP4', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 65, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 3, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
[...]

PR #1401, opened 2026-06-29
[Fix] Fix DeepSeek-V4 DP+EP on gfx942 (MI308X)

yitingw1:
## Motivation

On gfx942 (MI308X), DeepSeek-V4 with `--enable-dp-attention --enable-expert-parallel` never reached `Application startup complete`. This PR fixes **two** independent root causes:

1. **MoE warmup deadlock** (GPU 100% spin): mori IntraNode combine launched `block_num=128`, but its grid-wide barrier (`CrossDeviceBarrierIntraNodeKernel`) needs all blocks co-resident. MI308X has 80 CU and the combine block gets ~1 block/CU → 48 blocks never schedule → barrier hangs. (Dispatch survived at ~2 blocks/CU; gfx950 ≥128 CU never hit it.)
2. **mori buffer inconsistency**: after #1101, `max_num_inp_token_per_rank` was still hardcoded `16384` while the other two params used `max_num_batched_tokens`, truncating routed output at `mbt>16384` on the TBO path.

## Submission Checklist

### DIFF
--- a/atom/model_ops/fused_moe/mori_prepare_finalize.py
+++ b/atom/model_ops/fused_moe/mori_prepare_finalize.py
@@ -27,6 +27,19 @@ _NUM_TBO_UBATCHES = 2
 
 
 @lru_cache(maxsize=8)
+def _device_cu_count(device_index: int | None = None) -> int:
+    """Compute-unit (multiprocessor) count of the current CUDA/HIP device.
+
+    Used to cap mori IntraNode block_num so the kernels' grid-wide barrier
+    never asks for more co-resident blocks than the GPU has CUs (see
+    _get_dispatch_config). MI308X=80, MI300X=304, MI355X=256.
+    """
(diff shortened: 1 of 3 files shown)


atom/model_ops/fused_moe/mori_prepare_finalize.py
```diff
@@ -26,6 +26,19 @@
 _NUM_TBO_UBATCHES = 2
 
 
+@lru_cache(maxsize=8)
+def _device_cu_count(device_index: int | None = None) -> int:
```
valarLip: we should already have one in aiter

yitingw1: Done.

ZhangLirong-amd (approved):

valarLip (approved):

PR #1406, opened 2026-06-29
[Feature] OFFLOAD: MultiConnector — run P/D (mooncake/moriio) + LMCache offload together

yhl-amd:
## Summary

Adds a **MultiConnector** so a single ATOM engine can run a P/D-disaggregation connector (moriio / mooncake) **and** the LMCache CPU/NVMe offload connector **at the same time**. The canonical use case is a prefill node that both (a) RDMA/TCP-sends KV to a remote decode node and (b) offloads the same KV to CPU so a future request sharing the prefix can skip recompute.

> Note on scope: this branch builds on the LMCache offload subsystem; the diff therefore includes that stack. The new piece in this PR is `multi/multi_connector.py` (+ factory registration + unit tests).

## What's added

- `atom/kv_transfer/disaggregation/multi/multi_connector.py` — `MultiConnector` (worker), `MultiConnectorScheduler` (scheduler), `MultiConnectorMetadata`
- `atom/kv_transfer/disaggregation/factory.py` — registers `kv_connector: "multi"`
- `tests/test_multi_connector.py` — pure-mock unit tests (no GPU / lmcache / moriio), **16/16 passing**

## Connector vs engine (where this sits)

The engine computes tokens and **owns** the KV in HBM; a connector only **moves** KV bytes (to a remote decode node, or to CPU/NVMe) and never computes. They are decoupled by a narrow set of hooks in `base.py`.

```
                        +------------------+
                        |    API Server    |
                        +------------------+
                                 |
                                 v
+=================================================+
| INFERENCE ENGINE - computes tokens, owns KV     |
| EngineCore -> Scheduler -> ModelRunner (GPU)    |
| BlockManager -> [ HBM KV Cache = real KV bytes ]|
+=================================================+
                 ^               |
                 |   hooks (engine calls connector)
                 |               v
+=================================================+
| KV CONNECTOR - moves KV bytes, no compute       |
| KVConnectorBase (7 hooks) -> pick ONE:          |
| moriio | mooncake | lmcache_offload | multi     |
+=================================================+
            |                          |
            v                          v
     +-------------+            +-------------+
     | Decode node |            |  CPU / NVMe |
     +-------------+            +-------------+
       (P/D xfer)                (LMCache)
```

The connector is split across two processes/layers — a scheduler-side object (no GPU) and a worker-side object (holds HBM KV pointers); `meta` flows down, `KVConnectorOutput` flows up:

```
SCHEDULER LAYER (main, no GPU)        | hook                      -> policy in MultiConnector
  Scheduler + scheduler-side conn     | get_num_new_matched_tokens-> first-hit-wins (one sub owns it)
                                      | update_state_after_alloc  -> fan-out to ALL subs
                                      | build_connector_meta      -> pack -> MultiConnectorMetadata
                                      | request_finished          -> fan-out to ALL subs
        | build_connector_meta() => MultiConnectorMetadata(metas=[meta_sub0, meta_sub1])
        v  (IPC down: scheduler -> workers)
WORKER LAYER (GPU TP-worker, per rank)| register_kv_caches        -> fan-out to ALL subs
  ModelRunner + worker-side conn      | start_load_kv(meta)       -> route metas[i] -> subs[i]
  (holds HBM KV)                      | get_finished()            -> UNION + send/save gating
        ^  (IPC up: workers -> scheduler, merged across TP by KVOutputAggregator)
```

## Merge strategy (mirrors vLLM's MultiConnector, adapted to ATOM `base.py`)

- **`get_num_new_matched_tokens`** — first-hit-wins by `connectors` list order (the first sub reporting a match owns the load).
- **`update_state_after_alloc` / `request_finished`** — fan out to all subs.
- **`build_connector_meta`** — produces `MultiConnectorMetadata(metas=[...])`; the worker de-multiplexes by index in `start_load_kv` (each sub only sees its own metadata).
- **`get_finished`** — unions the four completion sets across subs.
- **offload-specific scheduler methods** (`should_park_*`, `adjust_prefill_chunk_after_alloc`, `should_defer_free`, `save_finished`, `load_failed`) are forwarded to the owning sub. The scheduler already guards each call with `hasattr`, so **the engine is unchanged**.

## Block-free correctness: send ∧ save gating

On a producer node, prefilled KV blocks must survive until **both** the P/D send **and** the offload save have read them. The scheduler frees a deferred block on `finished_sending` (and may free on `finished_saving` when not deferred) — so if one transfer finishes first, the block could be freed while the other is still reading it.

MultiConnector solves this at the connector layer (engine untouched): when a request needs both a send and a save, it **withholds both completion signals until the pair is done, then emits them in the same step**. Within one `_update_from_kv_xfer_finished`, the `finished_sending` loop runs first (pops + frees), and the `finished_saving` loop then finds nothing to free (no-op). Both arrival orders (send-first / save-first) are covered, and it holds across TP ranks (each rank gates; `KVOutputAggregator` reaches quorum on both sets in the same cycle). Unit-tested in both orders.

## Usage

Prefill node (decode node stays a plain mooncake/moriio consumer):

```bash
--kv-transfer-config '{
  "kv_connector": "multi",
  "connectors": [
    {"kv_connector": "mooncake", "kv_role": "kv_producer", "protocol": "tcp",
     "proxy_ip": "127.0.0.1", "proxy_ping_port": 36367, "http_port": 8000, "handshake_port": 6700},
    {"kv_connector": "lmcache_offload", "kv_role": "offload"}
  ]
}'
```

## Test plan

- [x] `pytest tests/test_multi_connector.py` — 16/16 (first-hit-wins, fan-out, metadata routing, get_finished union, send∧save gating in both orders, attribute aggregation). Pure mock, no GPU.
- [x] End-to-end on MI325X (Llama-3.1-8B, mooncake `protocol=tcp`): proxy + producer `multi[mooncake-producer + lmcache_offload]` + plain mooncake consumer.
  - P/D transfer through MultiConnector: `do_remote_prefill=true`, producer `All 1 consumers served`.
  - offload save through MultiConnector on the producer: `LMCache ... Stored 2304 tokens` for a 2509-token prompt, **concurrent with** the P/D send, 0 errors / 0 asserts.
  - HBM-hit second request: send proceeds, offload reload correctly skipped (HBM satisfies), no redundant re-store.

### DIFF
--- a/atom/kv_transfer/disaggregation/multi/multi_connector.py
+++ b/atom/kv_transfer/disaggregation/multi/multi_connector.py
@@ -290,9 +290,7 @@ class MultiConnectorScheduler(KVConnectorSchedulerBase):
         )
         # Opt into the scheduler's offload suffix-prefill path if any sub is the
         # offload backend (Scheduler._is_offload_connector reads this).
-        self.is_offload = any(
-            getattr(c, "is_offload", False) for c in self._connectors
-        )
+        self.is_offload = any(getattr(c, "is_offload", False) for c in self._connectors)
 
     # -- base interface -----------------------------------------------------
 
@@ -329,13 +327,13 @@ class MultiConnectorScheduler(KVConnectorSchedulerBase):
 
     def adjust_prefill_chunk_after_alloc(self, seq: Any, chunk: int) -> int:
         c = _first_with(self._connectors, "adjust_prefill_chunk_after_alloc")
-        return c.adjust_prefill_chunk_after_alloc(seq, chunk) if c is not None else chunk
+        return (
+            c.adjust_prefill_chunk_after_alloc(seq, chunk) if c is not None else chunk
+        )
 
     def should_park_partial_prefill_for_load(self, seq: Any) -> bool:
         c = _first_with(self._connectors, "should_park_partial_prefill_for_load")
-        return (
-            c.should_park_partial_prefill_for_load(seq) if c is not None else False
-        )
+        return c.should_park_partial_prefill_for_load(seq) if c is not None else False
 
     def should_defer_free(self, seq: Any) -> bool:
         # Defer if ANY sub wants to defer (so neither a pending save nor a
--- a/atom/kv_transfer/disaggregation/multi/multi_connector.py
+++ b/atom/kv_transfer/disaggregation/multi/multi_connector.py
@@ -1,0 +1,356 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2025, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Composite KV connector — run several sub-connectors behind one interface.
+
+The canonical use case is a prefill node that must do two things with the same
+KV at once:
+
+* **moriio** (``kv_role: kv_producer``) — RDMA-send the KV to a remote decode
+  node for P/D disaggregation;
+* **lmcache_offload** (``kv_role: offload``) — save the KV to CPU/NVMe so a
+  future request that shares the prefix can skip recompute.
+
+A single engine selects exactly one connector (``KVConnectorFactory`` reads one
+``kv_connector`` name). ``MultiConnector`` is that one connector; it owns a list
+of real sub-connectors and merges their results so the engine, scheduler, and
+output aggregator stay unchanged.
+
+Config::
+
+    --kv-transfer-config '{
+      "kv_connector": "multi",
+      "connectors": [
+        {"kv_connector": "moriio", "kv_role": "kv_producer", "proxy_ip": "...", ...},
+        {"kv_connector": "lmcache_offload", "kv_role": "offload"}
+      ]
+    }'
+
+Merge strategy mirrors vLLM's ``MultiConnector``, adapted to ATOM's
+``base.py`` interface:
+
+* ``get_num_new_matched_tokens`` — **first-hit-wins**: the first sub-connector
+  that reports a prefix match owns the load for that request.
+* ``update_state_after_alloc`` / ``request_finished`` — fan out to **all** subs
+  (moriio sets up its send, offload sets up its save; both must run).
+* ``build_connector_meta`` — returns :class:`MultiConnectorMetadata` carrying one
+  sub-metadata per connector, in connector order. The worker de-multiplexes by
+  index in ``start_load_kv``.
+* ``get_finished`` — union the completion sets, **but** see the send/save
+  pairing below.
+
+Send/save pairing (the one tricky correctness point)
+----------------------------------------------------
+On a producer node the scheduler frees a finished request's blocks as soon as it
+sees ``finished_sending`` (``scheduler.py``: producer path), and it can *also*
+free on ``finished_saving`` when the connector does not defer. If offload is
+still reading those blocks for its save when the moriio send completes (or vice
+versa), the free would corrupt the in-flight transfer. So when a request needs
+**both** a send and a save, ``MultiConnector`` withholds *both* completion
+signals until the pair is done, then emits them together. The scheduler's
+``finished_sending`` handler frees first; the ``finished_saving`` handler then
+finds nothing to free and no-ops. This is the analogue of vLLM's
+``_extra_async_saves`` refcount.
+"""
+
+from __future__ import annotations
+
+import copy
+import logging
+from typing import Any
+
+from atom.kv_transfer.disaggregation.base import (
+    KVConnectorBase,
+    KVConnectorSchedulerBase,
+)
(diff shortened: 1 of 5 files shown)


atom/kv_transfer/disaggregation/multi/multi_connector.py
```diff
@@ -0,0 +1,354 @@
...
+    connectors = []
+    for i, sub in enumerate(subs):
+        if not isinstance(sub, dict) or "kv_connector" not in sub:
+            raise ValueError(
+                f"connectors[{i}] must be a dict with a 'kv_connector' key, "
+                f"got {sub!r}"
+            )
+        if sub["kv_connector"] == "multi":
+            raise ValueError("multi connector cannot nest another 'multi'")
+        cfg_i = copy.copy(config)
```
ZhangLirong-amd, line 105: it's safe here?

atom/kv_transfer/disaggregation/multi/multi_connector.py
```diff
@@ -0,0 +1,354 @@
...
+``kv_connector`` name). ``MultiConnector`` is that one connector; it owns a list
+of real sub-connectors and merges their results so the engine, scheduler, and
+output aggregator stay unchanged.
+
+Config::
+
+    --kv-transfer-config '{
+      "kv_connector": "multi",
+      "connectors": [
+        {"kv_connector": "moriio", "kv_role": "kv_producer", "proxy_ip": "...", ...},
```
ZhangLirong-amd, line 24: The previous `{ "kv_connector": "moriio", "kv_role": "kv_producer" }`, they are the same level, if you want to change to 
```
{
  "kv_connector": "multi",
  "connectors": [
    { "kv_connector": "moriio", "kv_role": "kv_producer" },
    { "kv_connector": "lmcache_offload", "kv_role": "offload" }
  ]
}
```
But the engine/scheduler site still reading the top-level dict gets None under multi... I think you should test first and not influence previous logic

atom/kv_transfer/disaggregation/multi/multi_connector.py
yhl-amd: Yes, it's safe. `copy.copy(config)` is a shallow copy, and `cfg_i.kv_transfer_config = sub` only **rebinds** the attribute on the copy — it does not mutate the original `config.kv_transfer_config` (rebind, not in-place mutation). Each loop iteration gets its own fresh `cfg_i`, so the two subs never alias each other's `kv_transfer_config` either. The remaining attributes are shared by reference, but the sub-connectors only *read* them (never mutate), so there's no leakage to siblings or to the original config.

This mirrors vLLM's `MultiConnector`, which uses the same `copy.copy(config)` + swap-`kv_transfer_config` pattern. I'll add a one-line comment explaining the rebind-is-safe reasoning so it's clear to future readers.

atom/kv_transfer/disaggregation/multi/multi_connector.py
yhl-amd: Good point to check — I went through every engine/scheduler site that reads the top-level `kv_transfer_config` to confirm nothing breaks under `multi`:

- **enable gate** — `engine_core.py:141` `self.kv_transfer_enabled = bool(config.kv_transfer_config)` only checks truthiness; the multi dict is non-empty, so KV transfer is enabled. ✅
- **producer/consumer/offload routing** — the scheduler decides roles via `self.kv_connector.is_producer` / `is_offload` (`scheduler.py:1553/1730/1746/1581`), **not** by reading top-level `kv_role`. `MultiConnector` aggregates these from its subs (`any(sub.is_producer)` / `any(sub.is_offload)`), so the role is resolved correctly. ✅
- **proxy handshake** — the proxy reads `handshake_port` from the node's ZMQ **push-registration** payload (`proxy.py:161/340`), which mooncake fills from its own sub-config (`handshake_port:6700`), never from the top-level dict. ✅

So the only place that actually returns `None`/default under `multi` is the **`/kv_transfer_info` debug endpoint** (`api_server.py:1531-1532`) — `kv_role: null`, `handshake_port: 6301`. The proxy does not call that endpoint, so it's purely cosmetic, but I agree it's misleading and I'll fix it to surface the producer sub's role/handshake_port.

On "test first" — this path is already validated end-to-end on MI325X (8B, mooncake `tcp`, prefill via `multi`): P/D handshake + transfer go through (`All 1 consumers served for transfer_id=0`, consumer side `remote_handshake_port=6700` resolved correctly) and the offload save runs concurrently (`LMCache … Stored 2304`). Existing single-connector behavior is untouched because the engine reads roles through the connector object, not the raw dict.

ZhangLirong-amd (approved):

valarLip (approved):

PR #1409, opened 2026-06-30
fix(kimi): align input norm quant with attention quant

qichu-yun:
## Motivation

Fix input RMSNorm quant fusion gating for Kimi/DeepSeek-style attention paths. The previous logic could infer quantization from the decoder layer prefix instead of the actual attention input projection, enabling fused RMSNorm quantization even when that projection could not consume the quantized activation. The PR also keeps non-Triton FP8 support for Kimi while avoiding unsupported block/group FP8 paths such as GLM-5.1-FP8.

## Technical Details

- Resolve the real attention input projection (fused_qkv_a_proj for LoRA-Q MLA, otherwise q_proj) and use its quant config to decide whether input RMSNorm quant fusion is allowed.
- Support online quant configs when the offline projection config does not accept quantized input.
- Enable fusion only for real FP8/FP4 quantized attention inputs.
- Allow non-Triton FP8 fusion only for per-token FP8, preventing unsupported FP8 block/group quant kernels from being used.
- Preserve the existing non-Triton MXFP4 gating path.

## Test Plan
**1. fix the bug:**
launch server:
```
export AITER_AR_1STAGE_MAX_KB=512

model_path=/shared/data/amd_int/models/Kimi-K2.5-MXFP4
vllm serve $model_path \
    --host localhost \
    --port 9000 \
    --tensor-parallel-size 4 \
    --trust-remote-code \
    --async-scheduling \
    --gpu_memory_utilization 0.9 \
    --kv-cache-dtype fp8 \
    --max-model-len 16384 \
    --profiler-config '{"profiler": "torch", "torch_profiler_dir": "kimi_k2_5_mxfp4_attnfp8_tp4_profile", "torch_profiler_record_shapes": "True"}' \
    --no-enable-prefix-caching
```

before：
image

after:
image

image

**2.input rmsnorm and quant fused for kimi2.5**
before：
image

after：
image

## Test Result
**accuracy:**
image

## Submission Checklist


ZhangLirong-amd (dismissed):

atom/models/deepseek_v2.py
```diff
@@ -2079,28 +2079,48 @@ def __init__(
...
-            else quant_config.get_layer_quant_config(prefix).quant_type.value
-        )
+        attn_input_proj_name = (
+            "fused_qkv_a_proj"
+            if getattr(config, "q_lora_rank", None) is not None
+            else "q_proj"
+        )
+        self.quant_dtype = None
+        self.input_norm_quant_type = None
+        if quant_config is not None:
```
valarLip, line 2096: more clean up needed..

qichu-yun: Moved the definitions of self.quant_dtype and self.input_norm_quant_type into __init__ to centralize initialization.

valarLip (approved):

zhuyuhua-v (approved):

PR #1411, opened 2026-06-30
[Enhancement] Supports per_block_fp8 format for online quantization

haoyangli0109:
1. For the `weight` layer, perform per_block_fp8 quantization by grouping according to a 128×128 shape; the input must be quantized to 1×128.
In offline mode, `atom` sets the `quant_type` of layers quantized to `per_128×128` to `per_1×128`; online quantization follows this approach.

2.
```
  --model /shareddata/deepseek-ai/DeepSeek-R1-0528 \
  -tp 4 \
  --server-port 01346 \
  --online_quant_config '{"global_quant_config":"per_block_fp8","exclude_layer":["lm_head","*.gate.*"]}'

  --model_args "model=/shareddata/deepseek-ai/DeepSeek-R1-0528,base_url=http://localhost:01346/v1/completions,num_concurrent=65,max_retries=3,tokenized_requests=False,trust_remote_code=True" \
  --tasks gsm8k \
  --num_fewshot 3

```
|gsm8k|      3|flexible-extract|     3|exact_match|↑  |0.6907|±  |0.0127|
|     |       |strict-match    |     3|exact_match|↑  |0.7331|±  |0.0122|

### DIFF
--- a/atom/quant_spec.py
+++ b/atom/quant_spec.py
@@ -212,39 +212,31 @@ class QuarkOnlineParser(QuantConfigParser):
         and ``online_exclude_layers_list``.
 
         Supported format strings:
-        - ``"ptpc_fp8"``  — per-tensor-per-channel FP8
-        - ``"mxfp4"``     — microscaling FP4 (block size 32)
+        - ``"ptpc_fp8"``        — per-tensor-per-channel FP8
+        - ``"per_block_fp8"``   — per-block FP8 (128x128; alias of per_block128_fp8)
+        - ``"per_block128_fp8"``— per-block FP8, 128x128 block
+        - ``"mxfp4"``           — microscaling FP4 (block size 32)
         """
         if not isinstance(online_quant_config, dict):
             raise TypeError("online_quant_config must be a dict parsed from JSON.")
 
-        scheme_map = {
-            "ptpc": QuantType.per_Token,
-            "per_block": QuantType.per_1x128,
+        # Explicit table of supported online quant formats -> (QuantType, dtype_str).
+        # per_block / per_block128 are aliases (128 is the default block size).
+        FORMAT_MAP = {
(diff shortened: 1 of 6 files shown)


atom/rollout/weight_updater.py
```diff
@@ -231,27 +231,21 @@ def _requantize_fp8_weight(
...
-            blocks = padded.reshape(N, K_blocks, block_k)
-            block_amax = blocks.abs().amax(dim=-1)
-            scale = (block_amax / fp8_max).clamp(min=1e-12)
-            quantized = (blocks / scale.unsqueeze(-1)).to(fp8_dtype)
-            quantized = quantized.reshape(N, K_blocks * block_k)[:, :K].contiguous()
-            param.data.copy_(quantized)
-            ws = weight_scale.data
-            weight_scale.data.copy_(
-                scale[: ws.shape[0], : ws.shape[1]].contiguous().to(ws.dtype)
+            # Must match the load-time online_quantize_weight layout: a true
```
haoyangli0109, line 234: A colleague reported that when using RL with Atom's `per_block_fp8` for online quantized inference, precision issues arise here, so this has been modified accordingly. The colleague has verified that the fix works correctly.

atom/quant_spec.py
```diff
@@ -218,8 +218,9 @@ def parse(self, online_quant_config: dict) -> ParsedQuantConfig:
         if not isinstance(online_quant_config, dict):
             raise TypeError("online_quant_config must be a dict parsed from JSON.")
 
-        SCHEME_MAP = {
+        scheme_map = {
             "ptpc": QuantType.per_Token,
+            "per_block": QuantType.per_1x128,
```
valarLip: how deal with 1x32 or 1x64

haoyangli0109: Use a dictionary to enforce matching

valarLip (approved):

PR #1421, opened 2026-06-30
feat(prezero): wire split-K GEMM prezero into MLA / MoE decode

ColorsWind:
## Motivation

The split-K bf16 GEMMs on the MoE decode critical path — the fused `q_a + kv_a`
down-projection (N=2112, K=7168), the `q_b` up-projection, and the MoE router
gate — accumulate cross-block partial sums into the output `C` through
packed-bf16 global atomics, which requires `C` to be zero-initialized first. To
stay CUDA-graph safe, that zeroing happens inside the GEMM kernel, costing two
device-wide synchronizations per launch. In decode this is a **latency** cost —
on the order of 1-2us of pure latency, paid every layer, every step, largely
independent of problem size.

The companion aiter PR ([ROCm/aiter#4023](https://github.com/ROCm/aiter/pull/4023))
makes that zeroing relocatable: a preceding kernel can
zero the GEMM's output buffer, and the kernel-launch boundary already orders the
GEMM after it, so the GEMM atomic-adds straight in with no zero-init and no
handshake — no extra synchronization at all. The kernel that precedes these
GEMMs is the fused allreduce + RMSNorm that produces their input; it launches
one CTA per token, so in decode it leaves most CUs idle and can ride zero-fill
blocks along on them for free.

This PR is the **ATOM-side integration** of that capability: it allocates the
GEMM output buffers, has the fused AR-RMSNorm zero them on its idle CUs, and runs
the GEMMs as a `zero_init=False` atomic-add into them — for `qkv_a`, `q_b`, and
the MoE gate in DeepSeek-V3/V4 and Kimi-K2 decode.

> Stacked on the aiter prezero PR
> [ROCm/aiter#4023](https://github.com/ROCm/aiter/pull/4023) (the `zero_fill`
> ride-along, the `ZERO_INIT` skip path, and the `tgemm.mm` `zero_init`/`out=`
> passthrough live there). Please merge that first.

## Technical Details

Gated behind `ATOM_ENABLE_SPLITK_PREZERO` (default **off**); when off, the path
is byte-identical to before. Gating reads envs directly and derives buffer widths
from weights — no new stored module state.

**1. Wire prezero into the fused allreduce-RMSNorm (qkv_a).**
`RMSNorm.forward` gains an optional `zero_fill` argument forwarded to the fused
AR op, so the decoder layer can hand it a buffer to zero on idle CUs.
`_mm_prezero_` is added as a `torch.compile`-safe custom op
(`tgemm.mm(a, w, bias, zero_init=False, out=buf)`; mutates `out`, returns
`None`, since a custom op may not return an alias of a mutated input).
`MLAAttention` runs the bf16 `qkv_a` GEMM as a prezero atomic-add into the
`input_layernorm`-zeroed buffer. The path is taken only when
`is_prezero_free(graph_bs, n)` reports the zero-fill rides for free
(AR CTAs + zero CTAs < CU count), using the padded cudagraph token count,
requiring the fused-allreduce norm and `tp_size > 1`, and skipped on prefill.

**2. MoE router gate prezero.**
The single-stream MoE path runs the router gate GEMM as a prezero atomic-add into
a buffer zeroed by the `post_attention_layernorm` fused AR-RMSNorm, under the
same `is_prezero_free` gating. Dual-stream is left unchanged: it runs inside an
opaque custom op, so no gate buffer is threaded there.

**3. Extend prezero to the `q_b` up-projection via a shared buffer.**
`q_b` has no fused AR-RMSNorm of its own (`q_a_layernorm` is a plain local
RMSNorm), so it cannot get a free-zeroed buffer directly. Instead the
`input_layernorm` buffer is widened to `[M, N_qkva + N_qb]` and split into two
contiguous blocks: `qkv_a` atomic-adds into the first, `q_b` into the second, so
`q_b` rides the same `input_layernorm` zeroing. The wider half is gated on
`is_prezero_free(combined width)` and a bf16 (non-quantized) `q_b_proj`, falling
back to qkv_a-only when the wider zeroing would not fit — no regression vs. the
qkv_a path. `qb_prezero` is threaded as an optional mutable tensor arg through
`MLAAttention.forward -> unified_attention_with_output_base -> MLA.forward_impl
-> _q_proj_and_k_up_proj`, and passed only when allocated, keeping the
prezero-off path and non-ATOM attention backends byte-identical.

## Test Plan

End-to-end evaluation on the full model (the per-kernel halves are covered by the
aiter PR's standalone microbench).

Test environment:

- GPU: MI355X (gfx950)
- ROCm: 7.2.3

**Accuracy.** Prezero is mathematically identical to the baseline — it only
relocates where `C` is zeroed — so model output must be unchanged. Validated
end-to-end (GSM8K via `lm_eval`) with `ATOM_ENABLE_SPLITK_PREZERO=1` vs the
default, expecting no accuracy regression.

**End-to-end serving.** Kimi-K2.5 MXFP4 decode serving on the random dataset,
isl=8192 / osl=1024, random range ratio=0.8, concurrency 4 to 128, comparing the
baseline config against the prezero-enabled config. We report decode throughput /
per-token latency across the concurrency sweep, confirming the in-kernel
zero-init synchronizations are removed from the decode critical path with no
quality loss.

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -237,6 +237,7 @@ class CompilationConfig:
         factors.append(self.local_cache_dir)
         factors.append(self.cudagraph_capture_sizes)
         factors.append(self.cuda_graph_sizes)
+        factors.append(envs.ATOM_ENABLE_SPLITK_PREZERO)
 
         return hashlib.sha256(str(factors).encode()).hexdigest()
 
--- a/atom/model_ops/attention_mla.py
+++ b/atom/model_ops/attention_mla.py
@@ -32,6 +32,8 @@ except ImportError:
 from aiter.dist.parallel_state import get_dp_group
 from aiter.mla import mla_decode_fwd, mla_prefill_fwd
 from aiter.tuned_gemm import tgemm
+
+from atom.model_ops.prezero import prezero_active
 from aiter.ops.triton.attention.mla import (
     mla_decode_fwd as triton_shuffle_mla_decode_fwd,
 )
@@ -371,10 +373,8 @@ class MLAAttention(nn.Module):
     @mark_trace(prefix="q_proj_and_k_up_proj", torch_compile=False)
     def _q_proj_and_k_up_proj(self, x, x_scale=None, qb_prezero=None):
         if qb_prezero is not None:
-            # The q up-proj output buffer was pre-zeroed for free by the layer's
-            # fused allreduce-rmsnorm; run the bf16 q_b GEMM as a split-K
-            # atomic-add into it (zero_init=False) instead of allocating fresh.
-            tgemm.mm(x, self.q_proj.weight, None, zero_init=False, out=qb_prezero)
+            active = prezero_active(self.q_proj.prezero_n_total)
+            tgemm.mm(x, self.q_proj.weight, None, zero_init=not active, out=qb_prezero)
             q_proj_out = qb_prezero
         else:
             q_proj_out = self.q_proj(x, x_scale)
--- a/atom/model_ops/attention_mla.py
+++ b/atom/model_ops/attention_mla.py
@@ -31,6 +31,7 @@ except ImportError:
     fused_qk_rope_concat_and_cache_mla_seg = None
 from aiter.dist.parallel_state import get_dp_group
 from aiter.mla import mla_decode_fwd, mla_prefill_fwd
+from aiter.tuned_gemm import tgemm
 from aiter.ops.triton.attention.mla import (
     mla_decode_fwd as triton_shuffle_mla_decode_fwd,
 )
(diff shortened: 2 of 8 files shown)


ftyghome:
Hi @ZhangLirong-amd ,

Here are the performance test results:

**Model**: Kimi-K2.5-MXFP4
**TP**: 4
**ISL/OSL**: 8192 / 1024, random ratio 0.8

## 1. Serving throughput (ON vs OFF)

| CONC | throughput ON | OFF | Δ% | median TPOT ON | OFF | Δ% | e2e ON | OFF | Δ% |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 4   | 3743.4 | 3662.7 | **+2.2%** | 9.19  | 9.38  | **−2.0%** | 88.4  | 90.3  | **−2.2%** |
| 8   | 5709.1 | 5634.8 | **+1.3%** | 12.04 | 12.21 | **−1.4%** | 115.4 | 116.9 | **−1.3%** |
| 16  | 8635.8 | 8576.8 | **+0.7%** | 16.02 | 16.24 | **−1.4%** | 153.0 | 154.1 | −0.7% |
| 32  | 12034.3 | 11831.7 | **+1.7%** | 23.03 | 23.43 | **−1.7%** | 220.1 | 223.8 | **−1.7%** |
| 64  | 15963.6 | 15400.9 | **+3.7%** | 35.37 | 35.90 | **−1.5%** | 333.1 | 345.3 | **−3.5%** |
| 128 | 19974.0 | 20025.1 | −0.3% | 56.81 | 56.58 | +0.4% | 532.3 | 531.0 | +0.3% |

throughput = total_token_throughput (tok/s); TPOT in ms; e2e in s.

**Result:** prezero gives **+1–3.7% decode throughput** and **−1–2% TPOT** at CONC 4–64;

## 2. Decode kernel medians

Positive = slower under prezero. `qkv_a`/`q_b`/gate are the three prezero'd GEMMs;
`ar1`/`ar2` are the two per-layer AR+RMSNorm.

| role | 4 | 8 | 16 | 32 | 64 | 128 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| qkv_proj (`qkv_a`)      | −19.1% | −13.7% | −12.8% | −13.1% | −11.1% | +0.5% |
| up_b_proj (`q_b`)       | −3.4%  | −5.9%  | −14.4% | −15.9% | −16.9% | +3.5% |
| moe_route_proj (gate)   | −12.4% | −11.5% | −13.3% | −20.4% | −11.8% | −12.9% |
| ar1                     | −5.9%  | +2.7%  | −6.4%  | −0.6%  | −1.5%  | +1.8% |
| ar2                     | −5.1%  | −6.5%  | −5.8%  | −2.4%  | +1.0%  | −2.7% |

## 3. Conclusion

- **The prezero'd GEMMs drop 10–20%** (qkv_a / q_b / gate) at CONC 4–64.
- **AR + RMSNorm does not get slower**: the
  relocated zero-fill rides on the AR's idle CUs essentially for free.
- **CONC 128 gates off**: qkv_a/q_b use non-SplitK kernels there (no zero-fill to move),
  so only the gate benefits.

The kernel table is produced by the trace-parsing script [prezero_diff.py](https://github.com/user-attachments/files/29620678/prezero_diff.py).

Please feel free to let me know if there are anything needed from my end.

Thanks!

PR #1423, opened 2026-07-01
[Bugfix] DeepSeek-V4: content-addressed paged SWA fixes prefix-cache corruption (#1417)

### LINKED ISSUE #1417: [Bug] DeepSeek-V4: prefix caching corrupts output on cross-request hit (SWA window read from stale per-request ring)
## Summary

With `--enable_prefix_caching`, **DeepSeek-V4 produces corrupted output on a cross-request prefix-cache hit**: the reused output collapses (immediate EOS / empty completion). Root cause is that V4's **sliding-window (SWA) KV is stored in a per-request ring buffer that is NOT part of the (content-addressed) prefix cache**, so on a cross-request hit the SWA window for the cached region is read from a stale/empty ring.

Disabling prefix caching (`--no-enable_prefix_caching`) makes every request a full recompute and the output is healthy. So the only trigger is prefix caching.

## Repro (native: no offload, no PD/disaggregation)

Bare `atom.entrypoints.openai_server`, only toggling `--enable_prefix_caching`. Send the same prompt (one cold + several repeats); a repeat is a cross-request HBM prefix-cache hit.

Serve (DeepSeek-V4-Flash-FP8, TP=4, eager):
```
  --model <DeepSeek-V4-Flash-FP8> -tp 4 --kv_cache_dtype fp8 --trust-remote-code \
  --enforce-eager --enable_prefix_caching \
  --max-model-len 12288 --max-num-batched-tokens 8192 --gpu-memory-utilization 0.6 \
  --port 8009 --server-port 8000
```
Then POST the same ~7.7k-token prompt repeatedly with `max_tokens=24, temperature=0`.

### Reproduction request (copy-paste; hits the running server)

Sends 1 cold + 3 repeats of the **same** prompt and prints completion length / finish_reason.
With prefix caching ON the repeats are cross-request HBM hits; the bug = a repeat collapsing
(`ctoks` tiny / `finish=eos` / empty) while COLD stayed full.

```python
import json, time, urllib.request
BASE  = "http://127.0.0.1:8000/v1/completions"
MODEL = "<DeepSeek-V4-Flash-FP8 path>"
PROMPT = "the archival index records that fact again " * 1100   # ~7.7k tokens (> window, < max-model-len)

def gen(p):
    body = json.dumps({"model": MODEL, "prompt": p, "max_tokens": 24, "temperature": 0}).encode()
    req  = urllib.request.Request(BASE, data=body, headers={"Content-Type": "application/json"})
    d = json.loads(urllib.request.urlopen(req, timeout=600).read())
    c, u = d["choices"][0], d["usage"]
    return u["completion_tokens"], c.get("finish_reason"), c["text"]

ct, fin, txt = gen(PROMPT)
print(f"COLD     : ctoks={ct:3d} finish={fin!r}  out={txt[:60]!r}")
for i in range(1, 4):
    time.sleep(3)
    ct, fin, txt = gen(PROMPT)
    flag = "COLLAPSED <-- bug" if ct <= 2 else "full"
    print(f"REPEAT {i} : ctoks={ct:3d} finish={fin!r}  {flag}  out={txt[:60]!r}")
```

Equivalent single `curl` (run twice; 2nd is the prefix-cache hit — observe the 2nd returns an empty `text` with `finish_reason":"eos"` and `completion_tokens":1`):
```bash
P=$(python3 -c 'print("the archival index records that fact again "*1100)')
curl -s http://127.0.0.1:8000/v1/completions -H 'Content-Type: application/json' \
  -d "{\"model\":\"<DeepSeek-V4-Flash-FP8 path>\",\"prompt\":\"$P\",\"max_tokens\":24,\"temperature\":0}" \
  | python3 -c 'import sys,json;d=json.load(sys.stdin);c=d["choices"][0];print("ctoks=",d["usage"]["completion_tokens"],c["finish_reason"],repr(c["text"][:60]))'
```

Full A/B harness (restarts the server for prefix-cache ON vs OFF and asserts ON collapses / OFF does not)
is available on request.

### Field glossary (these are TWO different phases — don't conflate them)

| Field | Phase | Meaning |
|---|---|---|
| `cached:[X], new:[Y]` | **PREFILL** (input) | scheduler log: of the **prompt** (7701 input tokens), `Y` are (re)computed this step and `X` are reused from the prefix cache. This is about *reading the input*, not generation. |
| `ctoks` (`completion_tokens`) | **DECODE** (output) | how many tokens were **generated**. We requested `max_tokens=24`, so a healthy request produces 24. |
| `finish` (`finish_reason`) | **DECODE** (output) | why generation stopped: `max_tokens` = kept producing until the 24 cap (healthy); `eos` = model emitted end-of-sequence immediately (gave up). |
| `out` | DECODE (output) | the generated text; `''` = empty. |

So `new:[149]` (how much of the **input** was recomputed) and `ctoks=1` (how many **output** tokens were generated) are unrelated numbers from different phases. The bug: a small prefill recompute (`new:[149]` ≪ SWA cone ~5632) on a cache hit leaves the SWA state corrupted, which then makes **decode** collapse (`ctoks` 24→1, `finish` max_tokens→eos, empty output).

### Observed

**prefix caching ON** — 2nd+ requests are prefix-cache hits (**prefill**: `cached:[7552], new:[149]`):
```
            (decode output below)
COLD     : ctoks=24 finish=max_tokens          (full, normal)
REPEAT 1 : ctoks=24 finish=max_tokens          (full)
REPEAT 2 : ctoks=1  finish=eos        out=''   <-- COLLAPSED (decode gave up immediately)
REPEAT 3 : ctoks=1  finish=eos        out=''   <-- COLLAPSED
```
**prefix caching OFF** — every request is a full **prefill** (`cached:[0], new:[7701]`):
```
COLD + REPEAT 1/2/3 : all ctoks=24 finish=max_tokens   (all healthy, never collapses)
```

### Expected
Prefix-cache hits should produce the same (well-formed, ~`max_tokens`) output as the cold/recompute path.

> Note on testing: ATOM V4 appears **non-deterministic even at `temperature=0` with no caching** (two identical cold prefills produce different text). So a byte-exact `shot1==shot2` check is not a valid criterion; the repro instead checks for **output collapse** (a prefix-cache hit degenerating to empty/EOS while cold/recompute stays full). The two configs differ unambiguously: ON → collapses, OFF → never collapses.

## Root cause (per source)

V4 splits KV into a paged classical cache (CSA/HCA compressed entries) + a **per-request state cache** for the SWA window:

- **SWA storage is a per-slot ring**: `swa_kv = unified_kv[:swa_pages].view(num_slots, win_with_spec, head_dim)`, written by `swa_write` as `swa_kv[slot, pos % cache_size] = kv` (`atom/model_ops/v4_kernels/state_writes.py`). Only the last window is retained; it is **per-request, position-indexed, not content-addressed**.
- **The SWA slot is a free-list group, not prefix-cached**: claimed/released in `atom/model_engine/block_manager.py` (`free_per_req_cache_groups.pop()` on allocate / `.append()` on free). The generic prefix cache only hashes/reuses the paged (compressed) blocks; the per-request SWA ring is a fresh slot on every request.
- **On a prefix-cache hit, prefill reads the SWA prefix from this (now stale/empty) ring**: in `_build_paged_prefill_meta` (`atom/model_ops/attentions/deepseek_v4_attn.py`),
  `prefix_swa_count[t] = max(0, chunk_start - max(0, p_global - win + 1))`, where `chunk_start = num_cached_tokens` (the hit boundary). These `prefix_swa` entries index into the SWA ring (`kv_indices_prefix_swa` → `unified_kv` SWA region). The cached region's SWA was written by a *different* request (or never), so the window is garbage → degenerate output.

Within a single request, chunked prefill is fine (each chunk's SWA is written to the same slot before the next chunk reads it). The problem is specifically **cross-request reuse** (prefix cache, and equivalently KV offload), where the SWA ring is not the one that produced the cached compressed blocks.

## Suggested direction
Treat SWA like a content-addressed cache on the reuse path (snapshot the boundary window and key it by the prefix, and bound the reuse length to `min(compressed_hit, swa_hit)` so a hit never reuses compressed-without-SWA), or — at minimum — disable / guard prefix caching for V4 until SWA reuse is handled, to avoid silent output corruption.

## Environment
- Model: DeepSeek-V4-Flash-FP8 (also applies to V4-Pro by construction), TP=4, MI3xx (ROCm), `rocm/atom-dev:latest`.
- ATOM commit `80c66f15`. Local tree has **offload-subsystem-only** changes (`atom/kv_transfer/offload/*`, a `transfer_tensors.offload_segments` field, and a default-off `ATOM_V4_DISABLE_AITER_MHC` env); none are on the prefill / prefix-cache / SWA path exercised here (repro runs with offload disabled and the env unset).

yhl-amd:
## Summary

Fixes #1417. DeepSeek-V4 stored its sliding-window (SWA) KV in a **per-request ring** (`swa_kv[slot, pos % cache_size]`) that is **not** part of the content-addressed prefix cache. On a cross-request prefix-cache hit, the SWA window for the cached region was read from a stale/empty ring → decode collapses to empty/EOS. Disabling prefix caching (full recompute) hid the bug.

**Fix:** make SWA content-addressed via `block_tables`, like the compressed KV.
- **M1** — SWA paged into `unified_kv`, addressed by `block_tables` (shares the compressed phys). Correctness fix, but `block_bytes` grows so `num_kvcache_blocks` drops; kept only as a fallback.
- **M2** — a **separate windowed SWA pool** (own free-list + content hash + sliding-window freeing) so SWA is prefix-cacheable **and** memory-bounded; restores KV capacity while fixing the bug.

Toggle `ATOM_M2_PAGED_SWA` (default **on** for V4; `=0` forces the M1 page-everything fallback).

## Test plan / validation (DeepSeek-V4-Flash-FP8, TP4)

- [x] **eager**: #1417 repro **0/3 collapsed**, gsm8k 5-shot **0.952** (1319 @ nc=64), perf parity (decode ~502 vs ~500 tok/s), `num_kvcache_blocks` 14385→**108464**.
- [x] **non-eager (CUDAGraph)**: capture ok, repro 0/3, gsm8k 0.94, capacity restored (108232 vs baseline 113503).
- [x] **MTP/spec** (`--method mtp`): repro 0/3, gsm8k 0.94.
- [x] **BlockManager** dual-pool unit tests **40/40**; SWA index kernels have offline kernel-vs-reference tests (`tests/test_{swa_write,decode_indices,prefill_indices}_paged.py`).
- [x] Full test suite: **no regression** vs main.

## Known follow-ups (not blocking the fix)

- **PD/disaggregation**: forces the M1 fallback (SWA transfers as a paged `block_region` keyed by the compressed `block_table`; no connector change). M2 separate-pool PD transfer needs a second block-table in the connector. Not validated end-to-end (no 2-server env).
- **sglang/vllm plugin bridges**: `deepseek_v4_bridge.py` still bind `attn.swa_kv` with the old layout; needs the paged layout + cross-framework validation.


jiayyu:
**B' tail re-forward is now redundant — gate it off (or remove) when paged SWA is active**

`atom/model_engine/scheduler.py:861-868` (and the sizing at `:476,496`)

This PR deletes the per-request SWA ring entirely (`state_writes.py`'s `swa_write` now addresses via `block_tables` only), and the hit path reuses SWA correctly from the paged pool (M1 shared compressed block_table / M2 separate window pool). B' existed solely to "re-forward the tail block to refill the per-req ring"; with the ring gone, the `_v4_swa_warmup_blocks` rollback no longer provides correctness — it only re-forwards an extra `ceil(win_with_spec/block_size)` blocks per hit.

Measured on this branch (`97ed98c1`, V4-Flash TP4 eager, M2 default, 7701-token prompt, 3 hits):
```
Scheduled prefill batch: ... (cached: [7552], new: [149])   # B' dropped 1 block
```
Without the rollback it would be `new:[21]` — i.e. **~128 extra tokens re-forwarded per hit** (~7x the necessary tail). On long-hit workloads this is a persistent TTFT/throughput loss that works against this PR's stated "perf parity".

Suggestion: set `self._v4_swa_warmup_blocks = 0` when paged SWA is active (or drop B' plus its `__init__` computation at `:480-497`).

⚠️ Note: this PR was validated *with* B' in place (0/3 collapse). After removing it, please re-verify still 0/3 with `atom/examples/probe_prefix_hit_collapse.py --corpus issue` before merging.

jiayyu:
**CUDAGraph note contradicts the commit message and `state_writes.py` — comment looks stale**

`atom/model_ops/attentions/deepseek_v4_attn.py:2435-2438`

This comment says:
> `update_compressor_states` / `swa_write` ... currently use variable kernel grids ... model.forward inside torch.cuda.graph **will likely fail at the first such kernel launch**

But elsewhere in the same PR:
- the commit message states "non-eager (CUDAGraph): capture ok, repro 0/3, gsm8k 0.94, capacity restored";
- `state_writes.py:286` says "Fixed-grid + sentinel for CUDAGraph compat", and `swa_write`'s grid is `(bs, write_per_batch)` with `write_per_batch` a `constexpr` and `bs` fixed at capture.

These three don't agree. Please clarify and pick one:
1. If the kernels are now fixed-grid (which the code suggests), **remove/update this stale comment** and state explicitly in the docs/PR description that **V4 prefix caching is supported under the default level-3 (piecewise + CUDAGraph)**.
2. If it genuinely still fails under CG, then "CUDAGraph capture ok" in the commit message is an overclaim and should be corrected to "eager-only", with the real CG status documented.

This matters because level-3 is the default compilation level — users hit the CG path by default, so the support status can't be ambiguous.

yhl-amd:
@jiayyu thanks for both catches — addressed:

**1. B' `_v4_swa_warmup_blocks` (redundant tail re-forward)**

Removed entirely in `4a964777` (scheduler `__init__` computation + the `schedule()` rollback), as part of switching the SWA prefix-cache hit to a paged tail-window gate. Correctness is now enforced in `BlockManager`:
- `can_allocate` → `_swa_bounded_hit`: compressed prefix `P` stays a full chained match; SWA then scans right-to-left within `[0,P)` for the rightmost contiguous window of `ceil((window-1+mtp_k)/bs)` present blocks and returns that boundary `L`. Bounding by `P` guarantees `L` has both compressed[0,L) and SWA[L-window,L) present, so a hit never reuses a block whose in-window SWA is gone.
- `allocate`: out-of-window front blocks are marked `-1` (never read), only the trailing window is claimed from the paged pool.

Re-verified with the #1417 repro (1 cold + 3 repeats of the ~7.7k-token corpus, prefix caching ON), **0/3 collapsed** on both:
- `--enforce-eager`: 0/3 (ctoks 24/24/24)
- default level-3 (piecewise + CUDAGraph): 0/3

No more `~ceil(win_with_spec/bs)`-block re-forward penalty per hit.

**2. Stale CUDAGraph note (`deepseek_v4_attn.py` build_for_cudagraph_capture)**

You're right, the comment was stale — the follow-up it referenced has landed. Both kernels are already fixed-grid + sentinel (`state_writes.py`):
- `swa_write`: `grid=(bs, write_per_batch)`, `bs` baked at capture, `write_per_batch` a `constexpr`; rows past each seq's token count sentinel-skip.
- `update_compressor_states`: `grid=(write_plan.shape[0],)` = plan-buffer capacity fixed at builder init (not per-fwd `num_write`); inactive rows carry `position=-1` and bail.

So `model.forward` under `torch.cuda.graph` does not hit a variable-grid launch. Updated the comment in `c3ff3779` to match, resolving the contradiction. Picking the "supported under default level-3" option: CUDAGraph capture succeeds (8.34s over sizes [1..512]) and the #1417 repro is 0/3 on that path (above), so V4 prefix caching works at the default compilation level, not eager-only.

ZhangLirong-amd:
please also test on dp_attn and tbo, also `--enable-tbo all`, since tbo also involves kv cache

yhl-amd:
@ZhangLirong-amd ran the #1417 repro (1 cold + 3 repeats of the ~7.7k-token corpus, prefix caching ON) under both modes — **0/3 collapsed** in each. DeepSeek-V4-Flash-FP8, `--enforce-eager` (isolates the KV/SWA logic), dp=2 × tp=4 on 8 GPUs.

**1. DP attention** — `--data-parallel-size 2 -tp 4 --enable-dp-attention --enable-expert-parallel`
```
COLD     : ctoks=24
REPEAT 1..3 : ctoks=24 (full)   ==> collapsed 0/3
```
paged-SWA pool sized normally (num_swa_blocks=1154); no errors.

**2. TBO all** — same + `--enable-tbo all` (`enable_tbo=True, enable_tbo_decode=True`, dp_attn + EP on)
- prefill TBO (the repro's long prompt is split into ubatches): **0/3 collapsed**.
- decode TBO: fired 6 concurrent decodes (bs>2 → `enable_tbo_decode` path, which indexes `swa_block_tables` per ubatch) — all 6 produce the identical, coherent continuation, no garble/collapse.
- no Traceback / No free SWA / HIP error throughout.

So the separate SWA pool + tail-window hit gate + chunked-prefill window-freeing all hold under DP + EP + TBO (both prefill and decode ubatch paths, which are the ones that touch the SWA KV indexing).

Note: run on the current branch tip, which also includes the SWA extraction into `SlidingWindowPool` — so this doubles as end-to-end coverage of that refactor. Tested eager to isolate the KV logic; happy to also run non-eager (CUDAGraph) + TBO/DP if you'd like that combination covered too.

yhl-amd:
@ZhangLirong-amd correction to my earlier TBO-decode result above — my "6 concurrent decodes, all coherent" check was **too weak** and gave a false pass. 6-way concurrency did not reliably drive the `scheduled_bs > 2` ubatch-decode staging, and "output looks coherent" is not a real correctness signal. A proper **gsm8k conc32** run under `--enable-tbo all` told a very different story, so I want to set the record straight.

**What actually happened at conc32 (dp2×tp4 EP, eager):** first a cascade of crashes, then — once those were cleared — **silently wrong output**: gsm8k dropped `0.955 → 0.665`. I isolated it:

| config | prefill-TBO | decode-TBO | gsm8k (limit 200, 5-shot) |
|---|---|---|---|
| baseline (no TBO) | — | — | 0.955 |
| conc1 | fires | not (bs=1) | 0.95 |
| conc32 `--enable-tbo prefill` | fires (batched) | **off** | 0.93 |
| conc32 `--enable-tbo all` (before fix) | fires | on | **0.665** |
| conc32 `--enable-tbo all` (after fix) | fires | on | **0.965** |

Turning decode-TBO on is what regressed accuracy → the bug was squarely in the **decode ubatch path**, not prefill/DP-attn.

**Root cause (the numerical one):** in `_attach_v4_paged_decode_meta`, the decode index kernel `write_v4_paged_decode_indices` was passed the **global** `var["swa_block_tables"]`, but in a TBO ubatch `batch_id_per_token` holds **ubatch-local** req indices `[0, ub_real_reqs)`. So ubatch1 (`req_start > 0`) indexed *other* requests' SWA blocks → cross-request KV contamination, wrong output with no crash (ubatch0 happens to be correct since local==global there). Fixed to use the ubatch-prefixed `var[f"{p}swa_block_tables"]`, matching the HCA `block_tables` path in the same function.

Three more issues had to be cleared before TBO-decode even ran to completion (so this path was effectively never exercised before):
1. `_prepare_ubatch_decode` never sliced/attached `swa_block_tables` per ubatch → `swa_write(None)` crash after the M1 fallback removal (this PR's gap).
2. `ub*_v4_batch_id_per_token` buffer was `int64` while the staged data / every consumer is `int32` → `_stage` dtype assert (pre-existing on main: `2742` i64 vs `2630` i32).
3. `local_tbo_precompute` gated decode eligibility on `scheduled_bs >= 2`, but V4 `_prepare_ubatch_decode` + the decode CUDAGraph capture only prepare meta for `> 2`; under DP the AND-reduce can pull a `bs==2` rank into a ubatch step with no prepared meta → `build_ubatch_metadata` assert (pre-existing on main).

All four fixed in `1bc4acfa`. **Re-verified: conc32 `--enable-tbo all` gsm8k = 0.965 (flexible + strict), no crash.** #1/#3 are paged-SWA ubatch wiring (this PR); #2/#4 are pre-existing V4 TBO-decode dtype/threshold inconsistencies on main that had to be cleared for the path to run — flagged separately in the commit body in case you want them upstreamed on their own.

Thanks for pushing on `--enable-tbo all` — the light concurrency check would have shipped a real correctness bug.

yhl-amd:
## Full regression update — CUDAGraph, MTP, PD, TBO-decode

Following the reviews (eager-only coverage earlier), here is the complete regression on the rebased branch. **DeepSeek-V4-Flash-FP8, gsm8k limit 200 / 5-shot, conc32** unless noted.

| scenario | mode | result |
|---|---|---|
| baseline | tp4, eager, prefix ON | gsm8k **0.955** |
| **paged-SWA default path** | tp4, **CUDAGraph (level-3)**, prefix ON, no TBO | gsm8k **0.945** ✅ |
| **MTP** | tp4, **CUDAGraph**, `--method mtp --num-speculative-tokens 1`, prefix ON | gsm8k **0.965**, MTP acceptance **89.5%** (1.90 toks/fwd) ✅ |
| DP attention | dp2 tp4 EP, eager | #1417 repro **0/3** ✅ |
| TBO prefill | dp2 tp4 EP `--enable-tbo all`, eager | repro **0/3** ✅ |
| **TBO decode** | dp2 tp4 EP `--enable-tbo all`, eager | gsm8k **0.965** ✅ (after fixes below) |
| **PD disagg (1P1D)** | mooncake, prefill/decode split, eager | gsm8k **0.92** @conc4 (0 transfer failures), **0.950** @conc32 ✅ |
| benchmark | 96 req / conc32 / out128 | decode **425.9** tok/s, total **3483.7** tok/s |

### CUDAGraph coverage (addressing the earlier eager-only caveat)
- **paged-SWA under the default level-3 CUDAGraph works** (no-TBO 0.945; the "CUDAGraph capture ok" claim holds).
- **MTP under CUDAGraph works** — capture clean, 0 errors, 0.965 with a genuine 89.5% draft acceptance (so the multi-query decode + SWA spec-step path is actually exercised, not falling back to 1 token).

### TBO-decode correctness fix (`1bc4acfa`)
conc32 gsm8k under `--enable-tbo all` initially regressed 0.955 → 0.665. Root cause: in `_attach_v4_paged_decode_meta` the decode index kernel used the **global** `swa_block_tables` while `batch_id_per_token` holds ubatch-**local** req indices, so ubatch1 read other requests' SWA blocks (cross-request KV contamination, no crash). Fixed to the ubatch-prefixed table; plus three crash fixes that had prevented TBO-decode from ever running (swa_block_tables not wired into `_prepare_ubatch_decode`; int64/int32 `batch_id` buffer mismatch; `local_tbo_precompute` decode eligibility `>=2` vs the `>2` capture/prepare threshold). After the fix: **conc32 all-TBO gsm8k 0.965**, no crash. Details in the correction comment above.

### Known limitation (orthogonal, tracked separately): decode-TBO + CUDAGraph
`--enable-tbo all` under CUDAGraph aborts during **capture** with `HIP 900 'operation not permitted when stream is capturing' @ hipblaslt.cpp:147`. Isolated to decode-ubatch capture (CG+no-TBO ✅, CG+prefill-TBO ✅, only CG+decode-TBO ❌). Root cause is hipBLASLt running a per-shape heuristic query on the ubatch streams inside capture (the pre-capture warmup only initializes cuBLAS, not hipBLASLt) — a **TBO + hipBLASLt + CUDAGraph infra issue, independent of paged-SWA**. Verified empirically: at the commit before the TBO-decode fix, CG+decode-TBO crashes even earlier at the int64 `batch_id` assert (0 occurrences of the hipBLASLt error), i.e. this path never ran before and this PR neither introduces nor touches the affected code (`capture_tbo_graph`/warmup/hipBLASLt). Filed as **ROCm/ATOM#1506** with repro + suggested fix (warm the ubatch shapes on their streams before capture). decode-TBO is correct in eager (0.965); the default non-TBO CUDAGraph path is unaffected.

ZhangLirong-amd:
@yhl-amd , thanks for the test and fix! And can you help test --enable-tbo all on cudagraph? I think eager and cudagraph are different route, and better to use `-tp 8 --enable-dp-attention --enable-tbo all` we focus on tp 8 -> dp 8, and conc use 1024, we used to meet conc 1024 accuracy issue

atom/models/deepseek_v4.py
```diff
@@ -1841,15 +1852,22 @@ def forward_impl(
...
-            swa_write_per_batch=(
-                min(attn_md.max_seqlen_q, cache_size) if is_decode else None
-            ),
+            swa_kv=None,
         )
+        # Decode SWA write (paged, content-addressed). Prefill writes its SWA
+        # AFTER sparse_attn (see below) so chunked prefix reads see prior-chunk
+        # contents; decode has no such ordering hazard, so write here before the
+        # decode attention reads the window.
+        if is_decode:
```
valarLip: ...

atom/model_ops/attentions/deepseek_v4_attn.py
```diff
@@ -417,20 +419,55 @@ def compute_per_req_cache_bytes(self) -> int:
...
             + len(self.hca_layers) * hca_main
-            + self.num_layers * swa_per_layer
         )
 
+    def swa_pool_block_bytes(self) -> int:
+        """paged-SWA: bytes of ONE SWA physical block across all layers
+        (full-resolution, ratio-1). This is exactly the SWA term that
+        `compute_block_bytes` adds; the budget moves it to a separate
+        `num_swa_blocks`-sized pool instead of charging it per compressed block."""
+        elem_bf16 = self._classical_dtype.itemsize
```
valarLip: not always bf16

atom/model_ops/attentions/deepseek_v4_attn.py
```diff
@@ -522,7 +566,9 @@ def allocate_per_req_cache(self, num_slots: int) -> dict[str, object]:
         num_blocks = self.model_runner.num_physical_kvcache_blocks
         n_csa = len(self.csa_layers)
         n_hca = len(self.hca_layers)
-        swa_pages = num_slots * self.win_with_spec
+        # paged-SWA: SWA lives in its own num_swa_blocks pool, content-
+        # addressed by swa_block_tables. Size = num_swa_blocks * block_size.
+        swa_pages = self.model_runner.num_swa_blocks * self.block_size
```
valarLip, line 588: swa_pages = self.model_runner.num_swa_blocks  without "* self.block_size" ?

atom/model_ops/attentions/deepseek_v4_attn.py
```diff
@@ -775,15 +826,20 @@ def get_kv_transfer_tensors(self):
...
 
-        # Slot regions: SWA per layer
-        swa_slot_bytes = self.win_with_spec * self.head_dim * elem_bf16
+        # paged-SWA: SWA region [0, swa_pages) is the SEPARATE window-freed
+        # pool, content-addressed by seq.swa_block_table (NOT the compressed
+        # block_table). Emit it as swa_block_regions so the connector keys it by
+        # swa_block_table — window-freeing leaves only the live tail (the last
+        # ~128-token block) as non-(-1) entries, so only that gets transferred.
+        # block b's SWA lives at uv[0] + b*block_size*head_dim*elem.
+        swa_block_bytes = self.block_size * self.head_dim * elem_bf16
```
valarLip: "def swa_pool_block_bytes" we have new util func, should use that one?

atom/utils/tbo/ubatching.py
```diff
@@ -92,7 +92,13 @@ def local_tbo_precompute(
...
         return False, 0, 0
     scheduled_bs = batch.total_seqs_num_decode
-    if scheduled_bs < 2:
+    # Eligibility must be at least as strict as the strictest backend's
+    # per-ubatch metadata / CUDAGraph-capture threshold, or a rank can report
+    # "eligible" for a batch size that has no captured graph / no prepared
+    # ubatch meta, tripping build_ubatch_metadata's assert. Decode ubatch
+    # graphs are captured only for bs > 2 (model_runner) and V4 prepares meta
+    # only for scheduled_bs > 2, so gate on > 2 here too.
+    if scheduled_bs <= 2:
```
ZhangLirong-amd, line 107: here, all the tbo decode are uniformly set to >2. If you want to change here, pls, modify all the related place such as deepseek_v4_attn.py ..... And please give me a reason we need this modification, I used to meet =2 accuracy issue.
```
        self._ubatch_decode_meta = None
        if (
            self.model_runner.config.enable_tbo_decode
            and scheduled_bs > 2
            and not batch.is_dummy_run
        ):
```

yhl-amd: This change actually **tightens** ubatching to match the `> 2` convention you're describing — it doesn't introduce a new threshold. Before this PR, the two sides disagreed:

- `local_tbo_precompute` (this file) gated decode eligibility on `scheduled_bs < 2` → **eligible for `bs >= 2`** (includes `bs == 2`).
- V4 only prepares/captures for `> 2`: `deepseek_v4_attn.py:1283` (`_ubatch_decode_meta`, `scheduled_bs > 2`) and `:2678` (decode CUDAGraph capture, `bs > 2`).

So `bs == 2` was reported "eligible" here but had **no prepared ubatch meta / no captured graph** on the V4 side. Single-rank/non-DP hides it (all ranks agree), but under `--enable-dp-attention` the AND-reduce can pull a `bs == 2` rank into a collective ubatch decode step → `build_ubatch_metadata` asserts ("meta not prepared").

The fix is one line, `< 2` → `<= 2`, i.e. eligibility becomes `bs > 2`, which is now **uniform** with `deepseek_v4_attn.py`. Nothing else needs to change — `deepseek_v4_attn.py` and the capture path are already `> 2`; this file was the only outlier. So this removes the mismatch rather than adding a divergent gate.

On the `= 2` accuracy issue you hit before: this is consistent with it — excluding `bs == 2` from decode-TBO is exactly what avoids that path. This was one of the fixes in `1bc4acfa` that let `--enable-tbo all` run to completion; after it, conc32 `--enable-tbo all` gsm8k = **0.965** (no crash).

If you'd prefer, I can instead flip the V4 side to `>= 2` and make everything uniformly `>= 2` (prepare meta + capture graphs for `bs == 2` too) — but given your prior `= 2` accuracy experience, gating everything on `> 2` seems safer. Happy to go whichever way you prefer.

valarLip (dismissed):

valarLip (approved):

PR #1428, opened 2026-07-01
[Bugfix] openai: parse DeepSeek-V4 DSML tool calls + hoist system for V4

yhl-amd:
## Summary

Parse DeepSeek-V4 **DSML** tool-call format in the OpenAI server and hoist the system message for V4.

- `tool_parser.py`: add DSML detection + parsing (`_is_dsml` / `_parse_dsml`) with type coercion, so V4 tool calls surface as proper OpenAI `tool_calls`.
- `chat_encoders.py`: hoist system message into the V4 prompt layout (`_normalize_for_v4`).

## Test plan

- V4 tool-call round-trip: OpenAI `tool_calls` and Anthropic `tool_use` both parse cleanly from DSML output.
- Non-tool chat unchanged.

## Note

Previously this PR also carried the client-disconnect cancellation + non-stream leak fix; that has been **split into #1441** so each fix reviews independently. This PR is now DSML-only.

### DIFF
--- a/atom/entrypoints/openai/chat_encoders.py
+++ b/atom/entrypoints/openai/chat_encoders.py
@@ -85,6 +85,50 @@ def load_custom_message_encoder(model_path: str) -> Optional[MessageEncoder]:
     return _load_encoder_from_dir(_resolve_model_path(model_path))
 
 
+def _content_str(c: Any) -> str:
+    if isinstance(c, list):
+        return "\n".join(
+            b.get("text", "")
+            for b in c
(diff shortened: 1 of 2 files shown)


yhl-amd:
Closing this one. The client-disconnect / non-stream-leak fix has been split into #1441 (reviewed independently). The DSML tool-call changes here are being set aside for now; branch `fix/dsv4-tooluse-and-abort` is kept if we want to revisit.

PR #1431, opened 2026-07-01
feat(openai): add tool calling support with GPT-OSS Harmony parser

seungrokj:
## Summary

  - Add typed Pydantic models for `tools` and `tool_choice` in `ChatCompletionRequest` with input validation
  - Forward `tool_choice` to chat templates on both `/v1/chat/completions` and `/v1/messages` endpoints
  - Add GPT-OSS Harmony tool calling support via `openai_harmony` library — token-level prompt rendering and output parsing

  ## Changes

  | File | Description |
  |------|-------------|
  | `protocol.py` | Typed tool models + `check_tool_usage` validator |
  | `harmony_utils.py` | **New** — Harmony encoding, message building, prompt rendering for GPT-OSS |
  | `harmony_parser.py` | **New** — Token-level streaming parser via `openai_harmony.StreamableParser` |
  | `api_server.py` | Auto-detect GPT-OSS, route to Harmony on both endpoints |
  | `serving_chat.py` | Add `use_harmony` flag for token-level tool call parsing |

  ## Test plan

  - [ ] GPT-OSS tool calls via `/v1/messages` (Claude Code)
  - [ ] GPT-OSS tool calls via `/v1/chat/completions`
  - [ ] Qwen3/Kimi-K2 regression (existing tool parsers unaffected)

  ## Reproduce (on gfx942)

export ATOM_USE_TRITON_MOE=1
python -m atom.entrypoints.openai_server --model  openai/gpt-oss-20b --server-port 8006 --kv_cache_dtype fp8  --gpu-memory-utilization 0.7 

claude

image

### DIFF
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -1791,7 +1791,7 @@ async def stop_profile():
 def main():
     """Main entry point for the server."""
     global engine, tokenizer, model_name, default_chat_template_kwargs, _request_logger
-    global custom_message_encoder
+    global custom_message_encoder, use_harmony
 
     parser = argparse.ArgumentParser(description="ATOM OpenAI API Server")
     EngineArgs.add_cli_args(parser)
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -959,6 +959,9 @@ async def chat_completions(request: ChatCompletionRequest):
         merged_kwargs = dict(default_chat_template_kwargs)
         if request.chat_template_kwargs:
             merged_kwargs.update(request.chat_template_kwargs)
+        tool_choice_val = request.get_tool_choice_value()
+        if tool_choice_val is not None:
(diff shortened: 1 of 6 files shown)


carlushuang:
code format please (black/ruff), AGENT: put to memory!

carlushuang:
**`HarmonyStreamParser.parse_full` can duplicate output when generation stops on `max_tokens`.**

Normal completions are fine: the scheduler keeps the harmony stop token (`<|return|>`/`<|call|>`) in `output_tokens`, so the final message completes during `process_tokens` and `flush()` polls nothing new.

But when generation is cut by `max_tokens` there's no stop token, so the final message never completes in `process_tokens` (its content is emitted incrementally as deltas), then `flush()` runs `process_eos()` and re-emits that same message in full — yielding duplicated `content`, or a spurious second tool call carrying the full arguments on top of the partial-arg deltas.

Suggest tracking whether the current in-progress message already produced deltas and skipping the re-emit in `flush()` for that case.

PR #1441, opened 2026-07-02
[Bugfix] Cancel inference on client disconnect + fix non-stream request leak

yhl-amd:
## Summary

Two independent bugs in the OpenAI server, both surfaced by retrying clients (opencode / Claude Code) which pile up unbounded `pending requests` and waste GPU on output nobody reads.

**1. No disconnect cancellation.** Neither the streaming generators nor the non-stream handlers aborted the engine sequence when the client hung up, so an abandoned request ran to `max_tokens`.
- Add an engine abort path: `Sequence.aborted` → Scheduler reuses the normal finish/deallocate stop path; `EngineUtilityHandler.abort_request`; `CoreManager.abort_request` (broadcast to all DP ranks).
- Streaming generators (`stream_chat_response` / `_fanout`) now run under try/finally so `cleanup_streaming_request` (which aborts the seq) fires on `GeneratorExit`.
- Starlette does **not** cancel the handler on disconnect for non-stream, so add `_run_nonstream_with_disconnect()` which races generation against `raw_request.is_disconnected()` and cancels on disconnect. `chat_completions` / `completions` gain a `raw_request` param.

**2. Non-stream request leak.** `generate_async` / `_multimodal` / `_fanout` only removed the request from `io_processor.requests` on a setup exception, never on normal completion → every completed non-stream request leaked a `Sequence` (pending grew forever). Wrap the token loops in try/finally that always pops (and aborts on early exit). Streaming already pops via `cleanup_streaming_request`.

Also add `"choices": []` to the streaming usage chunk so AI-SDK clients (opencode) stop rejecting the final usage frame.

## Why

Under an agentic client that retries / cancels mid-generation, the pre-fix server leaks a Sequence per completed non-stream request and keeps abandoned requests running to `max_tokens`. `pending requests` grows without bound and GPU stays pinned on discarded work.

## Test plan

Verified on DeepSeek-V4, TP8:
- 6 sequential non-stream completions → `pending requests` stays flat (was 1→8).
- 3 mid-generation client disconnects → `abort_request ... found=True`, GPU utilization drops immediately, pending flat (was +3 leak with GPU pinned to `max_tokens`).
- Normal streaming + non-stream answers unchanged.

Reproduce:
```bash
PORT=9700; M=<served-model-id>
# (1) non-stream leak: fire 6, pending must stay flat
for i in $(seq 6); do
  curl -s localhost:$PORT/v1/chat/completions -H 'Content-Type: application/json' \
    -d '{"model":"'"$M"'","messages":[{"role":"user","content":"hi"}],"max_tokens":8}' >/dev/null
done
# (2) disconnect cancel: start long gen, drop after 2s -> log shows abort found, GPU drops
timeout 2 curl -s localhost:$PORT/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"'"$M"'","messages":[{"role":"user","content":"write a very long essay"}],"max_tokens":4000}' >/dev/null
# (3) regression: normal stream + non-stream still correct
curl -s localhost:$PORT/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"'"$M"'","messages":[{"role":"user","content":"2+2?"}],"max_tokens":10}'
```

Split out of #1428 (which now carries only the DeepSeek-V4 DSML tool-call changes) so this fix can be reviewed on its own.

### DIFF
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -440,17 +440,36 @@ async def generate_async(
         raise
     engine.core_mgr.add_request([seq])
 
-    while True:
-        item = await token_queue.get()
-        token_ids = item.get("token_ids") or []
-        if token_ids:
-            if first_token_at is None:
-                first_token_at = item.get("ts", time.time())
-            last_token_at = item.get("ts", time.time())
-            all_token_ids.extend(token_ids)
-        if item.get("finished", False):
-            finish_reason = item.get("finish_reason")
-            break
+    _finished_ok = False
+    try:
+        while True:
+            item = await token_queue.get()
+            token_ids = item.get("token_ids") or []
+            if token_ids:
+                if first_token_at is None:
+                    first_token_at = item.get("ts", time.time())
+                last_token_at = item.get("ts", time.time())
+                all_token_ids.extend(token_ids)
+            if item.get("finished", False):
+                finish_reason = item.get("finish_reason")
+                _finished_ok = True
+                break
+    finally:
+        # Two responsibilities, on EVERY exit path:
+        #   1) If we didn't finish (client disconnected / cancelled), tell the
+        #      engine to stop so the seq doesn't run to max_tokens and burn GPU.
+        #   2) Always drop the seq from io_processor.requests. The engine frees
+        #      its own KV on finish, but this dict is only cleaned up here for
+        #      non-stream requests -- without an unconditional pop, every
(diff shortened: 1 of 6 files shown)


yhl-amd:
Split out of #1428 so the client-disconnect / non-stream-leak fix can be reviewed on its own. #1428 now carries only the DeepSeek-V4 DSML tool-call changes.

PR #1449, opened 2026-07-03
[ci][atomesh] add ATOMesh nightly coverage for Kimi-K2.5 and MiniMax-M3

wanzhenchn:
## Summary

- Add ATOMesh nightly coverage for Kimi K2.5 and MiniMax M3, including TP4 single-node 1P1D, multi-node 2P1D DPA, and Eagle3 variants.
- Extend ATOMesh matrix/server submission logic to support flexible worker layouts, role-specific server args, Minimax quantization configs, Eagle3 options, and model path overrides.
- Improve workflow dispatch controls for selecting benchmark models, overriding benchmark/lm_eval concurrency, and configuring per-model paths across different Slurm clusters.
- Fix eval concurrency precedence so suite-level `eval_concurrency` overrides accuracy-level defaults, with benchmark concurrency as fallback.
- Standardize GSM8K result formatting to 4 decimal places in benchmark summaries.

## Workflow Dispatch Runs

https://github.com/ROCm/ATOM/actions/runs/28634140999


.github/benchmark/models_atomesh.yaml
```diff
@@ -103,3 +106,360 @@ models:
...
+            prefill:
+              HIP_VISIBLE_DEVICES: "0,1,2,3"
+            decode:
+              HIP_VISIBLE_DEVICES: "4,5,6,7"
+          benchmark:
+            wait_server_timeout: 1800
+          runner:
+            time_limit: "04:00:00"
+          run_eval: true
+          eval_concurrency: [64]
```
Jasen2201: gsm8k eval pls also add  256 conc

.github/benchmark/models_atomesh.yaml
```diff
@@ -103,3 +106,360 @@ models:
...
+            tp: 4
+          decode:
+            workers: 1
+            tp: 4
+          benchmark:
+            wait_server_timeout: 1800
+          runner:
+            time_limit: "06:00:00"
+          run_eval: true
+          eval_concurrency: [128]
```
Jasen2201: 2p1d pls also add 768 conc for gsm8k

.github/benchmark/models_atomesh.yaml
```diff
@@ -103,3 +106,360 @@ models:
...
+    server:
+      common_args:
+        trust_remote_code: true
+        kv_cache_dtype: fp8
+        block_size: 128
+        gpu_memory_utilization: 0.8
+        max_model_len: 32768
+        max_num_seqs: 256
+        max_num_batched_tokens: 32768
+        online_quant_config: '{"global_quant_config": "ptpc_fp8", "exclude_layer": ["lm_head", "model.embed_tokens", "vision_tower", "multi_modal_projector", "patch_merge_mlp", "*block_sparse_moe"]}'
```
Jasen2201, line 345: I think it's different for prefill and decode. cudagraph capture size args. Can we set this as a configurable option here?

Jasen2201 (dismissed):

.github/benchmark/models_atomesh.yaml
wanzhenchn: done

.github/benchmark/models_atomesh.yaml
wanzhenchn: done

PR #1465, opened 2026-07-04
perf: optimize model loading speed

ftyghome:
## Motivation

Loading large MoE checkpoints is slow on MI355X. Kimi-K2.5-MXFP4 stores each expert's
weight as a separate tensor, so the weight loader copies every expert into its slice of the
fused GPU parameter individually. Each copy moves very little data but still pays a fixed kernel-launch +
pageable-staging + GIL cost, and it's that per-copy overhead, not bandwidth, that dominates
load time. It shows up as a stall after the safetensors shard progress bar finishes, while
the loader thread pool drains the per-expert copies.

## Technical Details

Checkpoint iteration and the loader thread pool are unchanged. Only the way a fused MoE
parameter's experts reach the GPU changes:

- Each expert of a fused param is accumulated into a single pinned CPU staging buffer, and
  the buffer is flushed to the GPU with one large H2D copy once all experts have arrived
  (`atom/model_loader/loader.py`). This turns the hundreds of thousands of small copies into
  one copy per fused param.
- `FusedMoE.stage_expert_weight` writes an expert's shard into the staging buffer instead of
  the GPU param.

**This also resolves an issue when it takes too long to warm up the model, then NCCL times out and crashes.**

## Test Plan

Model: Kimi-K2.5-MXFP4
TP4 on MI355X (gfx950)

The metric is `load_model()` wall time on the slowest TP rank.

## Test Results

| Model | Before | After | Speedup | Reduction |
|---|---:|---:|---:|---:|
| Kimi-K2.5-MXFP4 | 208.1 s | 29.4 s | 7.08x | -85.9% |
| DeepSeek-R1-0528 | 535.2 s | 42.3 s | 12.66x | -92.1% |
| DeepSeek-V4-Pro | 822.2 s | 50.6 s | 16.26x | -93.9% |
| GLM-5-FP8 | 677.3 s | 48.3 s | 14.02x | -92.9% |


valarLip:
give some numbers for other models like dsr1/dsv4/glm5 ?

benenzhu:
This can reslove an issue that Kimi model load may randomly failed due to the slow ssds, and the broadcast ended timeouts.

Or https://github.com/ROCm/aiter/pull/4081 can fix this too by increasing the rccl process group timeout from 600s to 1200s.

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28670767395/job/85033302286#step:4:419
```python
Traceback (most recent call last):
  File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/app/ATOM/atom/model_engine/async_proc.py", line 130, in __init__
    self.runners = [runner_class(rank, *args, **kwargs)]
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/ATOM/atom/model_engine/model_runner.py", line 736, in __init__
    self.warmup_model()
  File "/app/ATOM/atom/model_engine/model_runner.py", line 1169, in warmup_model
    self.forward(dummy_batch)
  File "/opt/venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/app/ATOM/atom/model_engine/model_runner.py", line 2193, in forward
    fwd_output = self.postprocess(
                 ^^^^^^^^^^^^^^^^^
  File "/app/ATOM/atom/model_engine/model_runner.py", line 2112, in postprocess
    sampled_tokens = get_tp_group().broadcast(sampled_tokens, src=0)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/aiter-test/aiter/dist/parallel_state.py", line 1018, in broadcast
    torch.distributed.broadcast(
  File "/opt/venv/lib/python3.12/site-packages/torch/distributed/c10d_logger.py", line 83, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py", line 2907, in broadcast
    work = group.broadcast([tensor], opts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.distributed.DistBackendError: [2] is setting up NCCL communicator and retrieving ncclUniqueId from [0] via c10d key-value store by key '0', but store->get('0') got error: wait timeout after 600000ms, keys: /default_pg/0//3//cuda//0
```

valarLip:
## Review: perf: optimize model loading speed

Nice speedup, and the shape of the change is sound. Batching per-expert H2D copies into one pinned staging buffer per fused param is the right lever, `_copy_expert_shard` is good de-duplication (one copy path shared by `weight_loader` and the batched path so they can't drift), and switching the final drain from `concurrent.futures.wait()` to `future.result()` is a real improvement — it surfaces loader exceptions that `wait()` used to swallow.

A few things worth addressing before merge, roughly in priority order:

### 1. Pinned buffer is allocated + zeroed under the global lock (undercuts the speedup)
In `_stage_task`, entry creation runs inside `with staging_lock:`:
```python
entry = staging_map[pid] = {"staging": _make_staging(param), ...}
```
`_make_staging` does a full-param-size pinned `torch.empty(...)` + `.zero_()` — the single most expensive step — while holding a process-wide lock. All 16 workers serialize on it exactly during the allocation that should run in parallel, capping the batching win this PR is built to deliver. Suggest building the buffer outside the global lock (double-checked insert, or reuse the per-entry `lock`) and keeping only the dict insert under `staging_lock`.

### 2. `max_workers=16` is a hardcoded magic number
`main` used the default (`min(32, cpu+4)`). Hardcoding 16 silently caps parallelism on high-core MI355X hosts, and the sibling toggle `ATOM_LOADER_USE_THREADPOOL` is already a named env var. Please promote this to a named constant / `ATOM_LOADER_MAX_WORKERS` env and document it in `docs/environment_variables.md`.

### 3. Under-filled groups become silent partial loads
`loaded_weights_record.add(...)` fires at submit time. If `arrived < expected` at drain (e.g. `expected_batched_arrivals` returns `local_num_experts*2` but shared experts load elsewhere under `disable_fused_shared_loading`, so the routed param never reaches `expected`), the safety flush writes a partly-zero buffer with only a WARN — and the existing "NOT loaded from checkpoint" guard stays silent because the name is already recorded. The counting invariant (`expected` == exact non-`-1` arrivals) holds for the tested Kimi path but is untyped/fragile; a mismatch degrades to a silent partial load instead of a hard error. Consider an assert or a real error here.

### 4. Bias / per-Tensor scale are marked batchable but always fall back
`expected_batched_arrivals` reports `w13_bias`/`w2_bias` and per-Tensor `w13/w2_weight_scale` as batchable, but `stage_expert_weight`/`_copy_expert_shard` always return `False` for them. Each such param allocates a full pinned buffer on first arrival, then immediately pops it and falls back to `weight_loader`. No corruption (the fallback endpoint is correct), but it's wasted allocation and the "batchable" label is misleading — better to exclude these types in `expected_batched_arrivals`/`_param_is_batchable` so they never enter the staging path.

### 5. `_make_staging` zeroes the whole buffer unconditionally
In the normal all-experts-arrive case every byte is overwritten before flush, so `.zero_()` is a wasted full memory-write pass per param per layer for large fp4/fp8 tensors. It's only load-bearing for the under-fill safety path (#3) — defer or condition the zeroing on that case.

### 6. Docs not updated
`docs/model_support_guide.md` documents `load_model()`'s expert-loading flow and the `weight_loader` path, but not the new public `stage_expert_weight`/`expected_batched_arrivals` methods or the staging/flush mechanism. Please update it to match.

#1–2 bear directly on the PR's own performance goal and repo conventions; #3 is the correctness/robustness item worth a guard; #4–6 are cleanup/docs.

ftyghome:
Hi @valarLip,

Thanks for the review! I hope the commits I just pushed address these comments:

> 1. Pinned buffer is allocated + zeroed under the global lock (undercuts the speedup)

A helper function has been added, and pinning / zeroing can now be executed in parallel.

> 2. max_workers=16 is a hardcoded magic number

A new environment variable, `ATOM_LOADER_NUM_THREADS`, has been introduced and documented.

> 3. Under-filled groups become silent partial loads

I have changed the warning to a runtime assertion. The example in your comment does not seem likely to happen in practice. Disabling `disable_fused_shared_loading` should be fine, since the number of loading tasks still equals `expected` in this scenario.

> 4. Bias / per-Tensor scale are marked batchable but always fall back

The eligibility check logic has been refined, so they will no longer be marked as batchable.

> 5. _make_staging zeroes the whole buffer unconditionally

I think this is fine because model loading only happens once and this step is fast. Zeroing the buffer can also help prevent potential correctness issues.

> 6. Docs not updated

I have updated the docs and added performance measurements to this PR.

| Model | Before | After | Speedup | Reduction |
|---|---:|---:|---:|---:|
| Kimi-K2.5-MXFP4 | 208.1 s | 29.4 s | 7.08x | -85.9% |
| DeepSeek-R1-0528 | 535.2 s | 42.3 s | 12.66x | -92.1% |
| DeepSeek-V4-Pro | 822.2 s | 50.6 s | 16.26x | -93.9% |
| GLM-5-FP8 | 677.3 s | 48.3 s | 14.02x | -92.9% |

valarLip:
how is this happend? it only cost 30s for dsv4 pro for me

benenzhu:
> how is this happend? it only cost 30s for dsv4 pro for me

Maybe the Tensorwave SSD is a little slow?
This workflow uses **509s** to load the dsv4-pro.
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28693549865/job/85099471588#step:4:340
image

benenzhu:
Also Kimi have a problem that some rank loads maybe lagged.
Slowest rank's load time is about **12 mins** more than the fastest rank. 
Makes it easy to exceed the default 600s all_reduce timeout.
https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28670767395/job/85033302246#step:4:256
image
image

valarLip:
> > how is this happend? it only cost 30s for dsv4 pro for me
> 
> Maybe the Tensorwave SSD is a little slow? This workflow uses **509s** to load the dsv4-pro. https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28693549865/job/85099471588#step:4:340 image

i see.. tw node have some issue... our solution was disable_mmap, you can check this one https://github.com/ROCm/ATOM/actions/runs/28809636117/job/85434147979#step:7:112 however your way looks works better

benenzhu:
> > > how is this happend? it only cost 30s for dsv4 pro for me
> > 
> > 
> > Maybe the Tensorwave SSD is a little slow? This workflow uses **509s** to load the dsv4-pro. https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28693549865/job/85099471588#step:4:340 image
> 
> i see.. tw node have some issue... our solution was disable_mmap, you can check this one https://github.com/ROCm/ATOM/actions/runs/28809636117/job/85434147979#step:7:112 however your way looks works better

Thanks, it makes a lot easier. We will give it a try by setting `ATOM_DISABLE_MMAP=true`.

benenzhu:
> > > > how is this happend? it only cost 30s for dsv4 pro for me
> > > 
> > > 
> > > Maybe the Tensorwave SSD is a little slow? This workflow uses **509s** to load the dsv4-pro. https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28693549865/job/85099471588#step:4:340 image
> > 
> > 
> > i see.. tw node have some issue... our solution was disable_mmap, you can check this one https://github.com/ROCm/ATOM/actions/runs/28809636117/job/85434147979#step:7:112 however your way looks works better
> 
> Thanks, it makes a lot easier. We will give it a try by setting `ATOM_DISABLE_MMAP=true`.

Yeah disable_mmap works well, I submit a run, and it can reduce the load time from `12m` to `4m`.

https://github.com/SemiAnalysisAI/InferenceX/actions/runs/28881278580/job/85677479526

valarLip:
could you please fix the code style fail to make ci run

ftyghome:
I have updated the code and ruff should pass now.

valarLip:
Suggest also flipping the CI `disable-mmap` default to `false` as part of this PR.

Rationale: with the batched parallel loader introduced here, weight loading no longer needs the non-mmap path, so CI can keep mmap enabled by default instead of forcing `-e ATOM_DISABLE_MMAP=true` into every container.

Proposed change in `.github/actions/setup-gpu-container/action.yml`:

```diff
   disable-mmap:
     description: >-
-      When 'true' (default) inject `-e ATOM_DISABLE_MMAP=true`. Set 'false' for
-      callers whose container never set it (e.g. mmstar) to stay byte-for-byte
-      equivalent.
+      When 'true', inject `-e ATOM_DISABLE_MMAP=true` into the container.
+      Defaults to 'false' (mmap enabled) now that the batched parallel loader
+      makes the non-mmap load path unnecessary; set 'true' for callers that
+      still need mmap disabled.
     required: false
-    default: "true"
+    default: "false"
```

Blast radius (workflows using this action):
- `atom-test.yaml` and `atomesh-accuracy-validation.yaml` rely on the default → they will now run with **mmap enabled**.
- `atom-mmstar-ci.yaml` sets `disable-mmap: "false"` explicitly → **unchanged**.

Note: `atom/utils/envs.py` already defaults `ATOM_DISABLE_MMAP` to `false` in code; this only changes what CI injects into the container.

ftyghome:
> Suggest also flipping the CI `disable-mmap` default to `false` as part of this PR.

Thanks! I have your patch applied.

valarLip:
looks no help? https://github.com/ROCm/ATOM/actions/runs/28936754449/job/85864848300?pr=1465#step:22:291

ftyghome:
> looks no help? https://github.com/ROCm/ATOM/actions/runs/28936754449/job/85864848300?pr=1465#step:22:291

That's strange..

Unfortunately I don't have access to MI355 machines right now, maybe I can test it again once I regain access.

ftyghome:
Hi @valarLip ,

I investigated this a bit further, and it looks like this PR and `disable_mmap` are addressing different parts of the model loading pipeline.

Model loading mainly consists of two stages:

1. reading model weights from storage (I/O);
2. copying them from host memory to GPU (H2D copy).

This PR targets the second stage, while `disable_mmap` mainly helps with the first.

On TensorWave/CI machines, the model may stored on NFS (wekafs). When the model is loaded through `mmap` concurrently, the NFS connection can become contended and slows down the I/O stage. In that scenario, `disable_mmap` helps mitigate the loading bottleneck.

After the weights are in host memory, ATOM issues a large number of tiny H2D copies, which account for a significant portion of the loading time (as shown in this PR's description, which are all measured when model is on the local SSD). This PR batches those copies together, substantially reducing the H2D overhead.

I have just reverted the `disable_mmap` flip commit, and will appreciate if you can take another look at this.

Thanks!

valarLip:
sounds not make sense to me, h2d never take that long

ftyghome:
Hi @valarLip ,

I tested on MI355X again, and the results can be reproduced stably. The following command was used to launch the ATOM server:

```bash
AITER_BF16_FP8_MOE_BOUND=0 \
ATOM_MOE_GU_ITLV=1 \
ATOM_DISABLE_MMAP=0 \
ATOM_USE_TRITON_MXFP4_BMM=1 \
AITER_QUICK_REDUCE_QUANTIZATION=INT4 \
MORI_SHMEM_HEAP_SIZE=8G \
    --model deepseek-ai/DeepSeek-V4-Pro \
    --server-port 8000 \
    --kv_cache_dtype fp8 \
    --max-model-len 10240 \
    --trust-remote-code \
    -tp 4 \
    --no-enable_prefix_caching \
    --no-enable_chunked_prefill \
    --scheduler-delay-factor 10
```

Here are the detailed logs:

[upstream_log.txt](https://github.com/user-attachments/files/30034526/upstream_log.txt)

Model load takes ~880s (14m40s). The safetensors shard read finishes in ~8s, and the remaining ~14min is entirely the per-expert copy drain.

---

[optload_log.txt](https://github.com/user-attachments/files/30034524/optload_log.txt)

Model load takes ~48s, all four ranks finish within 3s of each other (04:46:02–04:46:05).

valarLip:
can you please help check why this never show any diff in our ci run, it's also tw nodes same as SA's

ftyghome:
> can you please help check why this never show any diff in our ci run

I think this is expected. This PR won't help reducing load time on slow I/O scenerios (e.g. the model is located on NFS drive and `disable_mmap` is on. In this case, I/O and H2D are overlapped). The performance improvments are measured when model is on the local nvme disk.

(The logs attached above are captured on TensorWave's machine, where model are placed on the RAID nvme device /dev/md0)

valarLip:
@gyohuangxin take a look this？

gyohuangxin:
> > can you please help check why this never show any diff in our ci run
> 
> I think this is expected. This PR won't help reducing load time on slow I/O scenerios (e.g. the model is located on NFS drive and `disable_mmap` is on. In this case, I/O and H2D are overlapped). The performance improvments are measured when model is on the local nvme disk.
> 
> (The logs attached above are captured on TensorWave's machine, where model are placed on the RAID nvme device /dev/md0)

Makes sense. Our CI loads models from NFS, so the bottleneck is likely storage I/O.

valarLip (approved):

valarLip:
oppps.. https://github.com/ROCm/ATOM/actions/runs/29600492073/job/87966982085#step:22:701

ftyghome:
> oppps.. https://github.com/ROCm/ATOM/actions/runs/29600492073/job/87966982085#step:22:701

Looks like it has been solved by commit d13bfb0

PR #1493, opened 2026-07-07
[fix OOM][Sparse MLA] chunked indexer and online topk to reduce the mem foorprint of logits tensor

### LINKED ISSUE #1376: GLM-5.2-FP8 OOM in sparse_attn_indexer (fp8_mqa_logits) under high concurrency + ~115k context
## Summary

On 8× MI355 (`gfx950`) with `rocm/atom-dev:latest` and the official GLM-5.2-FP8 serving recipe (TP8, bf16 KV cache), firing many concurrent requests with ~115k-token contexts causes a HIP OOM in the DSA sparse-attention indexer. All ModelRunner ranks die; `/health` still returns 200.

## How we found it

We were stress-testing GLM-5.2-FP8 with aiperf using long shared prefixes (~111k tokens) plus ~3.5k input tokens per request. Concurrency 8 with that shape completed. Stepping concurrency to 64 caused a reproducible crash after ~3–5 minutes of heavy prefill (GPU at 100%, then OOM). Attached `repro_glm52_oom.sh` is the minimal version of that load.

## Setup

Followed [ATOM GLM-5.2 Recipe](https://github.com/ROCm/ATOM/blob/main/recipes/GLM-5.md#glm-52-indexshare)

- Hardware: 8× MI355X
- Image: `rocm/atom-dev:latest`
- Model: `zai-org/GLM-5.2-FP8`
- Server command (inside container):

    python -m atom.entrypoints.openai_server --model zai-org/GLM-5.2-FP8 -tp 8 \
      --kv_cache_dtype bf16 --gpu-memory-utilization 0.8 --server-port 7777

- Client: aiperf 0.10.0 on host (Python 3.12 venv), `HF_HOME` pointing at cached model

## Repro

Uses [repro_glm52_oom.sh](https://github.com/user-attachments/files/29400064/repro_glm52_oom.sh)

1. Start a `rocm/atom-dev:latest` container with host networking and the HF model cache mounted (see GLM-5.2 recipe). Run the server command above inside it; wait for ready.
2. On the host: `pip install aiperf`, `export HF_HOME=...`
3. Run attached `repro_glm52_oom.sh`. Set `CONTAINER` to your Docker container name. Do not kill aiperf early; crash takes ~3–6 min.
4. Expect: `CRASH AS PREDICTED`, OOM lines in the server log, `All runners are shutdown`, `/health` still 200

## Stack trace (representative)

    atom/models/deepseek_v2.py", line 1219, in sparse_attn_indexer
        logits = fp8_mqa_logits(
    aiter/ops/triton/attention/fp8_mqa_logits.py", line 101, in fp8_mqa_logits
        logits = torch.full(
    torch.OutOfMemoryError: HIP out of memory. Tried to allocate ~49 GiB
    ...
    AsyncIOProcManager(ModelRunner): All runners are shutdown.

## Notes

- Same context shape at concurrency 8 does not crash.
- `--max-num-seqs 8` avoids the OOM but queues excess requests; not a real fix.
- Suspect `fp8_mqa_logits` needs bounded/chunked allocation rather than one large `torch.full`.

zejunchen-zejun:
fix: https://github.com/ROCm/ATOM/issues/1376

GLM5.2 FP8 TP4
|gsm8k|      3|flexible-extract|    20|exact_match|↑  |0.9469|±  |0.0062|
|     |       |strict-match    |    20|exact_match|↑  |0.9477|±  |0.0061|

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -1252,7 +1252,11 @@ def sparse_attn_indexer(
         # budget is disabled (0) or a single chunk fits, the loop runs exactly
         # once and matches the original single-shot behavior.
         budget_bytes = SPARSE_INDEXER_LOGITS_BUDGET_MB * 1024 * 1024
-        if budget_bytes > 0 and total_kv > 0 and budget_bytes // (total_kv * 4) < num_rows:
+        if (
+            budget_bytes > 0
+            and total_kv > 0
+            and budget_bytes // (total_kv * 4) < num_rows
+        ):
             # 4 bytes per fp32 logit; total_kv * 4 is one query row's footprint.
             # Round the budget-derived row count DOWN to keep the buffer within
             # budget: a multiple of 128 (aligned to the kernel's row tiling) in
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -127,7 +127,7 @@ ENABLE_DS_INPUT_RMSNORM_QUANT_FUSION = envs.ATOM_ENABLE_DS_INPUT_RMSNORM_QUANT_F
 ENABLE_DS_INDEXER_QK_ROPE_CACHE_FUSION = (
     envs.ATOM_ENABLE_DS_INDEXER_QK_ROPE_CACHE_FUSION
 )
-SPARSE_INDEXER_CHUNK_TOKENS = envs.ATOM_SPARSE_INDEXER_CHUNK_TOKENS
+SPARSE_INDEXER_LOGITS_BUDGET_MB = envs.ATOM_SPARSE_INDEXER_LOGITS_BUDGET_MB
 _FP8_DTYPES = tuple(
     dtype
     for dtype in (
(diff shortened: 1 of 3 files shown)


atom/models/deepseek_v2.py
```diff
@@ -1231,28 +1321,49 @@ def sparse_attn_indexer(
...
-            stride1=logits.stride(1),
-        )
+        # The dense logits buffer is [num_rows, total_kv] fp32. total_kv is the
+        # sum of all co-scheduled prefill contexts and is unbounded by
+        # max_num_batched_tokens, so a burst of long-context requests can push a
+        # single allocation to tens of GiB (#1376). Chunk along KV once total_kv
+        # exceeds the budget; short contexts keep the single-shot fast path.
+        chunk_tokens = SPARSE_INDEXER_CHUNK_TOKENS
+        if chunk_tokens > 0 and total_kv > chunk_tokens:
+            topk_indices_prefill.copy_(
```
ganyi1996ppo: better be inplace write instead of `copy_`?

zejunchen-zejun: make sense, will add out tensor

junhaha666 (approved):

valarLip:
give some perf number?

PR #1502, opened 2026-07-07
[Fix] fix GLM5.2 n-shot100 accuracy

yitingw1:
## Motivation

Fix GLM5.2 n-shot100 accuracy.

Scripts
```python
  --model_args model=zai-org/GLM-5.2-FP8,base_url=http://localhost:9677/v1/completions,num_concurrent=64,max_retries=3,tokenized_requests=False \
  --tasks gsm8k --num_fewshot 100 --limit 100
```

Before:

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|   100|exact_match|↑  | 0.36|±  |0.0482|
|     |       |strict-match    |   100|exact_match|↑  | 0.36|±  |0.0482|

After fix:

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|   100|exact_match|↑  | 0.98|±  |0.0141|
|     |       |strict-match    |   100|exact_match|↑  | 0.98|±  |0.0141|

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_v2.py
+++ b/atom/models/deepseek_v2.py
@@ -175,11 +175,19 @@ def _supports_fused_indexer_kernel_config(config: PretrainedConfig) -> bool:
 
 
 def _is_neox_rope_style(
-    config: PretrainedConfig, interleave_attr: str, default_interleave: bool
+    config: PretrainedConfig, interleave_attr: str, *, default_is_neox: bool
 ) -> bool:
-    interleave = getattr(config, interleave_attr, default_interleave)
+    """Resolve ``is_neox_style`` from a model's ``*_interleave`` rope config flag.
+
+    neox and interleaved (GPT-J) are the two mutually-exclusive rope layouts, so
+    an interleave flag of True means ``is_neox_style=False``. A missing or null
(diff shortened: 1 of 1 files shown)


atom/models/deepseek_v2.py
```diff
@@ -1861,7 +1861,9 @@ def __init__(
                 max_position=max_position_embeddings,
                 base=rope_theta,
                 rope_scaling=rope_scaling,
-                is_neox_style=True,
+                is_neox_style=_is_neox_rope_style(
+                    config, "indexer_rope_interleave", True
```
zejunchen-zejun: Hi, @yitingw1 

Should default value be False? For GLM5, it is ok because it has indexer_rope_interleave, while for DSV3.2, it doesn't have this attribute and the _is_neox_rope_style will return False
https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/config.json

valarLip (approved):

PR #1505, opened 2026-07-07
fix: dpa kv transfer error on spur

JiaoliangYu:
This patch fixes several issues in ATOM disaggregated inference with DPA:

  1. Preserve data_parallel_rank in OpenAI completion requests
     - Add data_parallel_rank to CompletionRequest.
     - Pass it through the non-streaming /v1/completions path.
     - Attach it to the sequence before engine scheduling.

  2. Honor pinned DP ranks in the ATOM engine
     - Route requests to the requested DP rank when data_parallel_rank is set.
     - Keep existing round-robin behavior when no rank is specified.
     - Validate invalid DP rank values.

  3. Align prefill and decode DP ranks in mesh
     - Select matching prefill@N and decode@N workers for ATOM PD requests.
     - Inject data_parallel_rank into worker requests.
     - Avoid prefill/decode rank mismatch during KV transfer.

  4. Fix KV-transfer metadata rank selection
     - Track selected prefill/decode DP ranks in the ATOM pair context.
     - Use the aligned prefill DP rank when enriching decode KV metadata.

  5. Improve ATOM DP worker discovery
     - Fall back to /kv_transfer_info when /server_info lacks DP metadata.
     - Fall back to /v1/models for model id discovery.
     - Allow mesh to expand ATOM endpoints into per-DP-rank workers.

  6. Wait for expanded DP workers during mesh startup
     - Wait until all configured ATOM endpoints are registered as DP workers.
     - Prevent routing before the full worker set is ready.

  7. Make Mooncake transfer completion safer
     - Return transfer success/failure from block and slot transfer paths.
     - Avoid sending write-done after a failed transfer.
     - Improve RDMA-local IP detection from IB devices/GIDs.

### DIFF
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -438,6 +438,7 @@ async def generate_async(
     sampling_params: SamplingParams,
     request_id: str,
     kv_transfer_params: Optional[Dict[str, Any]] = None,
+    data_parallel_rank: Optional[int] = None,
 ) -> AsyncGenerator[Dict[str, Any], None]:
     """Generate text asynchronously for non-streaming requests."""
     global engine, tokenizer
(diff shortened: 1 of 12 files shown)


Jasen2201 (approved):
LGTM

simondanielsson:
Nice work! When can we expect this to be merged? Thanks

JiaoliangYu:
> Nice work! When can we expect this to be merged? Thanks

ASAP  :smile:

ZhangLirong-amd (approved):

PR #1519, opened 2026-07-08
publish ATOMesh benchmark data with MI350X/MI355X hardware tag on dashboard

wanzhenchn:
- publish ATOMesh benchmark data with MI350X/MI355X hardware tag on dashboard
- allow dashboard publishing whenever successful benchmark artifacts exist instead of requiring every matrix case to pass.


.github/workflows/atomesh-benchmark.yaml
```diff
@@ -327,18 +331,38 @@ jobs:
           pattern: atomesh-model-benchmark-*
           merge-multiple: true
           path: atomesh-results
+        continue-on-error: true
```
gyohuangxin: Why set `continue-on-error: true` here, I think it's an important step for the results summry.

wanzhenchn: Good point. The artifact download is required for the summary job, so hiding failures here could mask real CI/artifact issues. 

I removed continue-on-error and kept the no-successful-result handling in the summary step via has_results=false.

gyohuangxin (approved):

PR #1527, opened 2026-07-09
Fix DeepSeek V4 fused shared expert mapping

cubezhang:
## Summary

Fix DeepSeek V4 expert weight mapping so fused shared expert weights are included in `FusedMoE.make_expert_params_mapping()` when the owning MoE layer is built in fused-shared mode.

The previous lookup only detected modules whose class name was exactly `FusedMoE`. In the vLLM plugin path this can miss the actual fused MoE module/wrapper, causing `num_fused_shared` to stay `0`. The generic loader then rewrites `ffn.shared_experts.*` to `ffn.experts.{n_routed_experts}.*`, but `get_expert_mapping()` does not include that final shared slot, so the shared expert tensors are silently dropped and model quality collapses.

## Validation

- Patched service started successfully in a clean container on port `8002`.
- Loader log no longer reports `ffn.shared_experts.*` tensors being silently dropped.
- Completion smoke test: `3 cars * 4 wheels` returns `12`.

- `lm_eval --tasks gsm8k --num_fewshot 5 --limit 20`: `0.9` exact match.
- Prior full validation on the same fix: full GSM8K 5-shot recovered to `flexible-extract=0.9477`, `strict-match=0.9484`.

### DIFF
--- a/atom/models/deepseek_v4.py
+++ b/atom/models/deepseek_v4.py
@@ -3054,9 +3054,21 @@ class DeepseekV4ForCausalLM(nn.Module):
         # mis-load expert weights -> garbage.
         num_fused_shared = 0
         for _m in self.model.modules():
-            if _m.__class__.__name__ == "FusedMoE":
+            if hasattr(_m, "num_fused_shared_experts"):
                 num_fused_shared = getattr(_m, "num_fused_shared_experts", 0)
                 break
+        if num_fused_shared == 0:
+            # Some plugin builds wrap/alias FusedMoE such that the exact class-name
+            # probe above misses it.  If the owning MoE layer was constructed in
+            # fused-shared mode, the loader will rewrite ffn.shared_experts.* to
+            # ffn.experts.{n_routed_experts}.*; include that final slot here so the
+            # generic expert mapping loads it into w13/w2 instead of dropping it.
+            for _m in self.model.modules():
+                if _m.__class__.__name__ == "MoE" and getattr(
+                    _m, "_fuse_shared_into_routed", False
+                ):
+                    num_fused_shared = getattr(self.args, "n_shared_experts", 0)
+                    break
         num_experts = self.args.n_routed_experts + num_fused_shared
         return FusedMoE.make_expert_params_mapping(
             ckpt_gate_proj_name="w1",


valarLip:
shared expert and routed expert has different data type for dsv4, how can we fuse?

cubezhang:
Good point. This patch is not intended to force fusion when the shared expert and routed experts have different dtypes.


My understanding is that this should still be guarded by the existing quant-config checks:


- `is_rocm_aiter_fusion_shared_expert_enabled_for_quant_config(...)`
- `MoE._fuse_shared_into_routed`
- `FusedMoE.num_fused_shared_experts`


For DSV4 configs where shared experts and routed experts have different dtypes, `_fuse_shared_into_routed` should be `False`, `num_fused_shared_experts` should stay `0`, and this patch should not add the shared expert slot.


The issue I observed was with `DeepSeek-V4-Flash-FP8`: the loader had already decided to rewrite `ffn.shared_experts.*` into `ffn.experts.{n_routed_experts}.*`, but `get_expert_mapping()` missed the fused shared slot because it only checked `__class__.__name__ == "FusedMoE"`. As a result, those rewritten shared expert tensors were dropped.


So the patch is only trying to make `get_expert_mapping()` consistent with the already-constructed MoE state. It should not enable mixed-dtype fusion.


If the intended behavior for this DSV4 config is actually standalone shared experts, then the bug is probably earlier in the loader/quant-config decision: `should_fuse_shared_expert_weight()` should return false and the loader should not rewrite `ffn.shared_experts.*` at all. I can update the PR in that direction if that is the preferred fix.




cube
***@***.***

yitingw1:
So for `DeepSeek-V4-Flash-FP8`, what are the datatypes of shared experts and routed experts?

cubezhang:
for DeepSeek-V4-Flash-FP8, both shared experts and routed experts are FP8 E4M3.






cube
***@***.***

yitingw1 (approved):
LGTM.

yitingw1:
Please add the same modification in deepseek_v4_mtp.py, too.
```python
diff --git a/atom/models/deepseek_v4_mtp.py b/atom/models/deepseek_v4_mtp.py
index 8f4582d9..29ccc4ac 100644
--- a/atom/models/deepseek_v4_mtp.py
+++ b/atom/models/deepseek_v4_mtp.py
@@ -253,9 +253,21 @@ class DeepseekV4MTP(nn.Module):
         """
         num_fused_shared = 0
         for m in self.model.modules():
-            if m.__class__.__name__ == "FusedMoE":
+            if hasattr(m, "num_fused_shared_experts"):
                 num_fused_shared = getattr(m, "num_fused_shared_experts", 0)
                 break
+        if num_fused_shared == 0:
+            # Some plugin builds wrap/alias FusedMoE such that the exact class-name
+            # probe above misses it. If the owning MoE layer was constructed in
+            # fused-shared mode, the loader rewrites ffn.shared_experts.* to
+            # ffn.experts.{n_routed_experts}.*; include that final slot here so the
+            # generic expert mapping loads it instead of dropping it.
+            for m in self.model.modules():
+                if m.__class__.__name__ == "MoE" and getattr(
+                    m, "_fuse_shared_into_routed", False
+                ):
+                    num_fused_shared = getattr(self.args, "n_shared_experts", 0)
+                    break
         return FusedMoE.make_expert_params_mapping(
             ckpt_gate_proj_name="w1",
             ckpt_down_proj_name="w2",
```
I've verified locally.

PR #1562, opened 2026-07-10
[Frontend] Abort engine request on client disconnect (free leaked KV)

yhl-amd:
## Summary
Non-streaming API handlers were not cancelled when the client hung up (Starlette only cancels `StreamingResponse`, not plain handlers), so the engine kept generating and the sequence's KV blocks leaked until it hit `max_tokens`. This adds a client-disconnect abort path.

## Changes
- **engine**: `EngineCoreMgr.abort_request` broadcasts an `abort_request` utility command; `_handle_abort_request` marks the seq `aborted`; the scheduler finishes it at the next step via the normal stop path (frees KV, emits a finished `RequestOutput`). Adds `Sequence.aborted`.
- **api_server**: `_run_nonstream_with_disconnect` runs `generate_async` in a task and polls `request.is_disconnected()`; on disconnect it cancels the task, whose teardown aborts + pops the request. Wired into `/v1/chat/completions` and `/v1/completions` (return HTTP 499). `generate_async` / `generate_async_multimodal` / `generate_async_fanout` / `cleanup_streaming_request` abort + pop on early exit.

## Test plan
- [ ] Disconnect a non-streaming client mid-generation → seq finishes at next step, KV freed (no leak until max_tokens).
- [ ] Normal (non-disconnect) requests unaffected.


yhl-amd:
ci:atom

atom/model_engine/sequence.py
```diff
@@ -86,6 +86,9 @@ def __init__(
         # out-of-window SWA blocks can be freed while compressed blocks persist).
         # Empty / unused for non-SWA models.
         self.swa_block_table = []
+        # Set True when the client disconnected; the scheduler finishes the seq
+        # at the next step via the normal stop path (frees KV, emits finished).
+        self.aborted = False
```
valarLip: SequenceStatus merged into this one?

yhl-amd: done

Issue #1507, opened 2026-07-08
[DeepSeek-V4] index_topk=1024 sparse indexer can't recall verbatim lines from ~18k+ prompts (breaks agentic cwd/env recall)


yhl-amd:
## Summary

When DeepSeek-V4-Pro is served with the fp8 sparse indexer at the default
`index_topk=1024`, the model **cannot reliably recall a verbatim line that is
present in its own prompt** once the prompt reaches a realistic agentic size
(~18k+ tokens with many tool definitions). Instead of reading the line back it
hallucinates a plausible-but-wrong value.

This is isolated to **recall** (no tools involved): the exact answer is literally
in the system prompt, the model is asked to copy it verbatim, and it still fails.

## Impact

Agentic clients (Claude-Code-style via `/v1/messages`, Codex via `/v1/responses`)
put the working directory / environment in the system prompt alongside ~20+ tool
definitions (~20–30k tokens total). The model then cannot recall its own `cwd`,
so it fabricates absolute paths (e.g. `/sgl-workspace/...`, `/home/<random>/...`),
runs shell commands against non-existent paths, loops, and either goes off-topic
or burns hundreds of K tokens without answering.

## Environment

- Model: `DeepSeek-V4-Pro`, fp8 sparse indexer (main default, no fp4 flag)
- Serve: `-tp 8 --kv_cache_dtype fp8 --hf-overrides '{"index_topk": 1024}'
  --enable_prefix_caching --enable_chunked_prefill --max-model-len 1048576
  --gpu-memory-utilization 0.7` (gfx950 ×8)
- ATOM: main `0c7d50f9` + PR #1423 (paged SWA) + #1443 + #1441 picked
- `temperature=0`, `stream=false`

## Reproduction (self-contained)

Builds a ~18k-token system prompt = 700 distinct tool-like lines with one unique
needle buried ~55% in, then asks the model to copy it verbatim (no tools):

```bash
python3 - <<'PY'
import json,urllib.request
needle="/opt/acme-widgets/build-42/workspace"
V=["Creates","Deletes","Updates","Queries","Renders","Compiles","Streams","Validates","Encrypts","Schedules","Indexes","Archives","Migrates","Profiles","Balances"]
N=["orders","invoices","clusters","shards","payloads","sessions","tenants","widgets","pipelines","caches","registries","snapshots","manifests","tokens","routes"]
[...]

PR #1563, opened 2026-07-10
[Frontend] DeepSeek-V4 native OpenAI/Anthropic/Responses API + DSML tool parser

yhl-amd:
## Summary
Serve DeepSeek-V4 (Pro/Flash) so **Claude Code** (Anthropic `/v1/messages`) and **Codex CLI** (OpenAI Responses `/v1/responses`) can talk to ATOM directly, with **DSML tool-call parsing** shared across `/v1/chat/completions`, `/v1/messages` and `/v1/responses`.

> Stacked on #1562 (uses the `abort_request` disconnect path). Review/merge that first; until then this PR's diff includes that commit.

## Changes
- **tool_parser**: DeepSeek-V4 DSML tool-call format (`<｜DSML｜invoke ...>`) with marker-less / self-closing / direct-JSON recovery, schema-driven type coercion and key aliases; streaming + non-streaming (alongside existing Qwen/GLM/MiniMax).
- **serving_chat**: reasoning-filter + tool-call streaming for `/v1/chat/completions` (reasoning_content deltas during think phase, tool_calls deltas).
- **serving_responses (new)** + `/v1/responses` endpoint: OpenAI Responses translation (input/tools ↔ chat, SSE emitter, reasoning/function_call items). Codex compat: inject a mandatory DSML tool-format instruction; normalize shell tool name/param aliases (`exec`/`shell_exec`/… → registered exec tool; `command`/`script` → `cmd`) so Codex tool calls execute instead of erroring.
- **/v1/messages**: pass the tool schema to the parser so tool args are typed correctly (fixes Claude Code "Invalid tool parameters"); streaming **UTF-8 incremental detokenization** (vLLM-style sliding window) so multi-byte chars (CJK, box-drawing) aren't split into `U+FFFD`.
- **reasoning / chat_encoders**: DeepSeek-V4 message encoding + reasoning tags.

## Test plan
- [ ] `claude-local` (Claude Code → `/v1/messages`): multi-tool agentic session, tools execute, no "Invalid tool parameters", CJK/box-drawing render cleanly.
- [ ] `codex-local` (Codex → `/v1/responses`): tool calls execute (no "unsupported call"/"missing field cmd"), multi-step task completes.
- [ ] `/v1/chat/completions` streaming: reasoning + tool_calls deltas.


yhl-amd:
Pushed 7 follow-up commits (append-only fast-forward) making the DSV4 native agentic path actually work end-to-end and hardening it:

**Makes tool-calling work / native agentic:**
- `[Bugfix] Wire batched stream-flush hook` — the batched stream-dispatch hook was left unwired (`_flush_stream_batch_fn = None`, never resolved/called), so streaming `/v1/messages` + `/v1/chat/completions` emitted only the synthetic role chunk then hung. Now resolved lazily after engine init.
- `[Bugfix] Inject DSML tool-format instruction on /v1/messages too`
- `[Frontend] Inject Claude Code cwd into Bash tool description on /v1/messages` — mitigates the long-context cwd recall cliff.

**Accuracy (DeepSeek-V4 #1507):**
- `[Bugfix] DeepSeek-V4: drop Hadamard on indexer Q to match un-rotated K (#1507)` — the lightning-indexer scored `(Hq)·k` instead of `q·k` (Q was Hadamard-rotated, the compressor's K was not), flattening sparse top-k so long-context needles couldn't be recalled. Apply RoPE-only to Q. Verified 13/13 needle recall on DSV4-Pro TP8 fp8 at default `index_topk=1024`, incl. ~40k-token prompts in the sparse regime. (Note: this is a model-path fix living in this frontend PR for convenience — happy to split it out to a standalone PR against #1507 if preferred.)

**Robustness (from a 10-angle review of this PR):**
- `[Bugfix] Abort leaked seq on streaming client disconnect (chat + responses)` — three stream generators cleaned up with `aborted=False`, so client disconnect never aborted the engine seq (ran to max_tokens, wasting KV/GPU). Restored the `aborted` tracking pattern from `serving_completion.py`.
- `[Bugfix] Guard Responses tool translation against malformed inputs` — null `function` field (AttributeError) + non-numeric `offset`/`limit` (ValueError mid-stream).
- `[Bugfix] Preserve cwd paths containing spaces in Codex <cwd> extraction`.

PR #1573, opened 2026-07-13
[Bugfix] DeepSeek-V4: fix sglang/vllm plugin SWA after paged-SWA migration (#1417/#1423)

### LINKED ISSUE #1417: [Bug] DeepSeek-V4: prefix caching corrupts output on cross-request hit (SWA window read from stale per-request ring)

yhl-amd:
## Summary

PR #1423 (\"content-addressed paged SWA\") migrated the **entire shared V4 SWA path** — the index kernels (`write_v4_paged_{decode,prefill}_indices`), `swa_write`, and the shared model `_attn_core` (which the plugins reuse) — from a per-request **ring** to a **paged** layout addressed by `swa_block_tables`/`block_size`. Its own commit message deferred *\"the sglang/vllm plugin-bridge SWA layout\"* as a follow-up, which was never done. So the plugins still passed the removed `state_slot_per_seq`/`cs` kwargs and crashed at server startup:

```
TypeError: write_v4_paged_decode_indices() got an unexpected keyword argument 'state_slot_per_seq'
```

during `init_cuda_graph_state`, failing the **Accuracy (DeepSeek-V4-Pro TP8)** sglang CI job. The plugins also still **write** SWA into a per-request ring (`swa_kv`, `state_slot`, `swa_cache_size`; no `swa_block_tables`/windowed pool), so a blind kwarg swap would make paged reads disagree with ring writes.

## Fix (plugin-only, no shared engine/kernel change)

Keep the ring pool but speak the new paged ABI via a **degenerate ring-emulating SWA block table**: every logical block of a request maps to the request's single ring slot (`swa_block_tables[bid, :] = state_slot`) and `block_size = cs` (`= swa_cache_size`). The paged offset then collapses to the exact ring offset:

```
swa_kv[swa_block_tables[bid, pos//bs]*bs + pos%bs] == swa_kv[slot*cs + pos%cs]
```

- **sglang/vllm bridges**: bind `swa_kv` as the flat `[pages, head_dim]` view and set `attn.swa_block_size = cs`; add `_build_swa_ring_block_tables` (broadcast fill, CUDA-graph-capture-safe via a persistent buffer in the graph-buffer classes); set `md.swa_block_tables` in every decode/prefill metadata builder; repoint the six index call sites to the paged ABI. The vllm persistent decode path keeps its unchanged ring kernel `write_v4_decode_indices_fused` (already emits `slot*cs+pos%cs`, consistent with the degenerate-table write).
- **sglang draft-extend wrapper**: `swa_kv` is now flat, so derive `cache_size` from `attn.swa_block_size` and pass `swa_block_tables`/`swa_block_size`.

This keeps the plugins on ring semantics (memory-bounded, correct without radix cache — the failing CI job runs `--disable-radix-cache`). The paged content-addressed #1417 fix stays engine-only; porting the plugins to a real windowed paged pool (to fix #1417 under prefix caching there) is a separate follow-up.

## Test plan

- [x] **Kernel equivalence**: degenerate-table paged decode kernel produces byte-identical SWA offsets to the pre-#1423 ring formula for spec=0 (cs=128) and spec=3 (cs=131).
- [x] **End-to-end**: replicated the failing CI job locally (`rocm/atom-dev:sglang-latest` + this ATOM, DeepSeek-V4-Pro TP8, `--disable-radix-cache --page-size 256`). Server reaches `Application startup complete` (no crash at CUDA-graph capture); **gsm8k 5-shot = 0.9515** flexible + strict (CI threshold 0.94), matching the engine's validated ~0.952.
- [ ] **Follow-up (not covered)**: MTP/spec (cs != 128) and the draft-extend graph are outside the failing CI (spec=0); the emulation math holds but each needs its own validation.


yhl-amd:
## Validation results

Environment: 8×MI355 (gfx950), single node. sglang via `rocm/atom-dev:sglang-latest`; vLLM via `rocm/atom-dev@sha256:4c9fb5e5…` (the d20260710 OOT image). All runs use this branch's ATOM.

### 1. SGLang accuracy — gsm8k (reproduces the failing CI job)

Config: `DeepSeek-V4-Pro TP8`, `--tensor-parallel-size 8 --kv-cache-dtype fp8_e4m3 --mem-fraction-static 0.9 --swa-full-tokens-ratio 0.1 --max-running-requests 256 --page-size 256 --disable-radix-cache --disable-shared-experts-fusion --tool-call-parser deepseekv4 --reasoning-parser deepseek-v4`; lm_eval gsm8k **5-shot, num_concurrent=8**.

| Metric | This PR | CI threshold |
|---|---|---|
| gsm8k flexible-extract | **0.9515** | ≥ 0.94 ✅ |
| gsm8k strict-match | **0.9515** | ≥ 0.94 ✅ |
| Completed | 320/320, exit 0 | |

Matches the engine-side #1423 self-test (~0.952) and the catalog baseline (0.9530 ± 0.0058). Server reaches `Application startup complete` — the original `TypeError: write_v4_paged_decode_indices() got an unexpected keyword argument 'state_slot_per_seq'` during CUDA-graph metadata build is gone.

### 2. Kernel-level equivalence — degenerate table vs the pre-#1423 ring formula

The ring-emulating `swa_block_tables[bid,:] = state_slot` + `block_size = cs` produces SWA offsets identical to the old ring `slot*cs + pos%cs`:

| Config | Result |
|---|---|
| spec=0 (cs=128) | byte-identical ✅ |
| spec=3 (cs=131) | byte-identical ✅ |

### 3. vLLM performance — this PR vs true-ring baseline (same machine / image / client)

Server (identical, via `atom_oot_test.sh`): `vllm serve … --async-scheduling --load-format fastsafetensors --compilation-config '{"cudagraph_mode":"FULL_AND_PIECEWISE"}' --kv-cache-dtype fp8 --tensor-parallel-size 8 --gpu-memory-utilization 0.9 --max-num-seqs 512 --tokenizer-mode deepseek_v4 --no-enable-prefix-caching`; env `AITER_BF16_FP8_MOE_BOUND=0 ATOM_MOE_GU_ITLV=1`.
Client (CI InferenceMax `benchmark_serving.py`): random 1024/1024, `--random-range-ratio 0.8 --num-prompts 320 --max-concurrency 32 --num-warmups 64 --request-rate inf --ignore-eos`.

| Metric (c32) | Baseline (baked true-ring `5a5d3da9`) | This PR (main+#1423+shim) | Δ |
|---|---|---|---|
| Output tok/s | 1108.78 | 1106.78 | **-0.18%** |
| Total tok/s | 2214.06 | 2210.06 | -0.18% |
| Mean TPOT | 27.82 ms | 27.90 ms | +0.3% |
| Median TPOT | 28.01 ms | 28.11 ms | +0.4% |
| Completed | 320/320 | 320/320 | ✅ |

Performance parity with true ring (within measurement noise) — expected, since the emulated SWA offsets are byte-identical.

### Note: pre-existing crash unrelated to this PR

`vllm bench serve` (the vLLM CLI client) crashes V4-Pro on mixed decode+chunked-prefill batches with `pa_sparse_prefill_opus_kernels.cu: kv_indptr_prefix length must be N+1`. This reproduces **identically on the true-ring baseline** (`5a5d3da9`, `081045f^`) — it is a pre-existing plugin/aiter chunked-prefill contract issue exposed by that specific client, **orthogonal to the SWA ring↔paged change**. The CI benchmark client (InferenceMax) does not trigger it; both this PR and the baseline complete cleanly with it (tables above).

valarLip (approved):

PR #1591, opened 2026-07-14
Fix DeepSeek-V4 SGLang index top-k metadata

junna2016:
## Motivation

The DeepSeek-V4 SGLang bridge previously hard-coded `index_topk=1024` when constructing attention metadata and CUDA graph buffers. However, the DeepSeek-V4-Flash-FP8 model uses:

```json
{
  "index_topk": 512
}
```

This caused an inconsistency between the ATOM indexer and the SGLang bridge:

- ATOM model/indexer: index_topk=512
- SGLang attention metadata: index_topk=1024
- SGLang CUDA graph buffers: allocated using index_topk=1024

The issue is difficult to detect with short prompts because DeepSeek-V4 CSA uses a compression ratio of 4:

```
effective_topk = min(sequence_length / 4, index_topk)
```

For a 1K prompt, the effective number of CSA entries is only about 256, so both 512 and 1024 produce the same result. The mismatch starts affecting execution after approximately 2K tokens.
At longer sequence lengths, the inconsistent value may cause incorrect sparse-attention metadata, incompatible graph buffer sizes, accuracy degradation, or invalid kernel inputs.

## Technical Details

This PR adds `_resolve_v4_index_topk()` to `atom/plugin/sglang/deepseek_v4_bridge.py`.
The resolver obtains `index_topk` from the following sources:

1. `model.args.index_topk`
2. `model.atom_config.hf_config.index_topk`
3. The cached value on the SGLang proxy pool
4. The active ATOM Hugging Face configuration
5. The DeepSeek-V4 default value as a fallback

The resolved value is cached on the proxy pool because the first eager SGLang metadata batch may be constructed before the model proxy cache views are bound.
The resolved value is now used consistently in:

- Eager attention metadata
- Decode CUDA graph metadata and buffers
- Verify CUDA graph metadata and buffers
- Proxy cache metadata

CUDA graph buffers are recreated when their existing `index_topk` does not match the active model configuration.
The proxy cache binding path also validates that each DeepSeek-V4 layer's indexer uses the same `index_topk` as the SGLang metadata. A mismatch now fails early with a descriptive error instead of propagating inconsistent metadata to sparse-attention kernels.

## Test Result

Flexible final-number matching results:

| Input length | Completed samples | Accuracy | Request errors |
|---|---|---|---|
| 3,072 | 1,319/1,319 | 92.65% | 0 |
| 12,288 | 1,319/1,319 | 86.43% | 0 |
| 16,384 | 1,319/1,319 | 85.29% | 0 |

### DIFF
--- a/atom/plugin/sglang/deepseek_v4_bridge.py
+++ b/atom/plugin/sglang/deepseek_v4_bridge.py
@@ -10,6 +10,58 @@ import torch
 ATOM_DEEPSEEK_V4_BLOCK_SIZE = 128
 
 
+def _resolve_v4_index_topk(model: Any = None, proxy_pool: Any = None) -> int:
+    """Resolve the indexer width from the active ATOM model configuration.
+
+    The first eager SGLang metadata batch is built before proxy cache views are
+    bound to the model, so it cannot rely on ``model._atom_v4_meta_params``.
+    ATOM's active config is available then and carries the same Hugging Face
+    ``index_topk`` value used to construct the indexer.
+    """
+    value = None
+    source = None
+    if model is not None:
+        value = getattr(getattr(model, "args", None), "index_topk", None)
+        source = "model.args.index_topk"
+        if value is None:
+            value = getattr(
+                getattr(getattr(model, "atom_config", None), "hf_config", None),
+                "index_topk",
+                None,
+            )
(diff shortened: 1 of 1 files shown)


yitingw1 (approved):
LGTM

ZhiweiYan-96 (approved):

PR #1593, opened 2026-07-14
[NO REVIEW][atom/atom-vllm] fix MTP accept-rate drop at con>=256 on sparse-MLA (DSA) decode

zejunchen-zejun:
Root cause: at large MTP batches the sparse-indexer decode path issues one big deepgemm_fp8_paged_mqa_logits + top_k_per_row_decode launch whose later rows can retain stale sparse-KV indices in the persistent buffer (rows beyond the CU wave capacity), so requests past the cliff attend to wrong KV and their drafts get rejected -> acceptance halves (~50% -> ~25% at con256, cliff at seq 128).

so when specify max num seqs to the value which is smaller than 128, the accept ratio restores to the normal value, while for default 512, most of requests has low draft token accept ratio. Here is the histogram illustration.
image

Fix:
- Chunk the sparse-indexer decode launches to <= get_num_sms()*wave_per_eu rows (new helper sparse_indexer_decode_rows_per_launch), so every row's logits/topk are actually computed. Applied to both native (deepseek_v2) and plugin paths.
- triton_gather_kv_indices_sparse: use token_to_seq_idxs as the authoritative row count and assert topk/sparse_kv_indptr are large enough instead of silently truncating (which left stale rows in the persistent sparse-KV index buffer).
- plugin: handle requires_padding by packing weights and unpacking topk indices.

## atom-vllm

| Configuration | Position 1 Acceptance Rate | Position 2 Acceptance Rate | Position 3 Acceptance Rate | Avg Draft Acceptance Rate |
| --- | ---: | ---: | ---: | ---: |
| without fix | 0.422 | 0.245 | 0.108 | 25.8% |
| with fix | 0.721 | 0.494 | 0.235 | 48.3% |

## atom

| Configuration | Acceptance Rate | Accepted 0 Tokens | Accepted 1 Token | Accepted 2 Tokens | Accepted 3 Tokens |
| --- | ---: | ---: | ---: | ---: | ---: |
| without fix | 26.87% | 52.30% | 23.90% | 14.70% | 9.10% |
| with fix | 53.53% | 17.64% | 27.65% | 31.20% | 23.51% |

### DIFF
--- a/atom/model_ops/attention_mla.py
+++ b/atom/model_ops/attention_mla.py
@@ -1628,17 +1628,19 @@ def triton_gather_kv_indices_sparse(
     assert topk_indices.shape[1] == NUM_TOPK_TOKENS
     assert NUM_TOPK_TOKENS % BLOCK_N == 0
 
-    # token_to_seq_idxs is the authoritative output row count. Do not silently
-    # truncate if the sparse indexer produced fewer top-k rows; that would leave
-    # stale values in the persistent sparse-KV index buffer.
-    num_tokens = token_to_seq_idxs.shape[0]
-    assert topk_indices.shape[0] >= num_tokens, (
-        "sparse indexer under-produced top-k rows: "
-        f"topk_rows={topk_indices.shape[0]} expected_rows={num_tokens}"
-    )
-    assert sparse_kv_indptr.shape[0] >= num_tokens + 1, (
-        "sparse_kv_indptr is shorter than sparse token metadata: "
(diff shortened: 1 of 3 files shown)


zejunchen-zejun:
这个bug的现象是GLM5.2并发256的时候，atom和atom-vllm的draft token接受率从正常的50%掉到25-30%
一个batch的256个request只有前128个request的接受率正常，后128个request的接受率几乎是0，于是综合接受率腰斩

zejunchen-zejun:
real fix should be in aiter side: https://github.com/ROCm/aiter/pull/4244

PR #1594, opened 2026-07-14
Add MoRIIO write-push KV transfer with DeepSeek-V4 and fabric (scale-up) backend

maning00:
## Summary

Adds a **write/push** transfer mode to the MoRIIO PD-disaggregation KV connector, and uses it to support **DeepSeek-V4** (hybrid sparse MLA) and the mori-io **fabric (UALink scale-up)** backend. Read/pull remains available but is no longer the default.

MoRIIO previously only supported read/pull (decode RDMA-reads prefill's KV). That cannot express DeepSeek-V4's KV layout (compressed blocks + sliding-window ring + FP8 indexer + per-request FP32 state), and cannot use the fabric backend, which moves KV GPU-to-GPU inside a UALink super-node without RDMA NICs.

## What's included

**Write/push mode**
- Consumer-initiated write: decode advertises its registered memory to prefill; prefill issues one-sided RDMA/fabric writes into decode's KV and notifies completion.
- Non-blocking — transfers run on a worker pool, so the engine step loop never stalls. Robust completion handling: per-request failure-wins, receive/orphan timeouts, and staging backpressure.
- `transfer_mode` defaults to `write`; `read` stays available via config.

**DeepSeek-V4 support**
- Registers KV via region descriptors (block / slot / staging) and transfers them in a single pass; geometry comes entirely from the attention backend (no hard-coded layer counts, ratios, or offsets).

**Fabric (UALink) backend**
- Opt-in via `ATOM_MORIIO_FABRIC=1` or `moriio_backend: "fabric"`; RDMA remains the default and is unchanged.
- Adds a backend-agnostic KV mem-pool hook to `KVConnectorFactory`: a backend can register a custom allocation pool (fabric needs fabric-exportable VMM memory), and `ModelRunner` wraps KV allocation with one generic call. Fabric symbols are imported lazily, so non-fabric users gain no new dependency.
- Under fabric each region is registered as a single descriptor and transfers are finalized via blocking status waits (fabric completion is callback-driven).

**Pre-existing fixes (needed for co-located / DP+TP runs)**
- Globally-unique mori-io engine key (role + pid + uuid) so co-located prefill/decode no longer collide.
- Correct TP size on every `get_port_offset` path, fixing ZMQ port collisions when `dp_size > 1` and `tp_size > 1`.

## Configuration

```jsonc
{"kv_role": "kv_producer" | "kv_consumer", "kv_connector": "moriio", "handshake_port": 6301}
```

- `transfer_mode`: `write` (default) | `read`
- Fabric: set `ATOM_MORIIO_FABRIC=1` (or `moriio_backend: "fabric"`) on both sides; GPUs must share a UALink vPOD.
- Write mode requires the router to set `do_remote_decode=true` on prefill requests (the atomesh `atom` backend does this).

## Testing

- Unit tests: protocol/schema, port-offset grid, region chunking, staging deferral, receive/orphan timeouts, and late-completion handling.
- End-to-end on 8×MI355X:
  - DeepSeek-V4-Flash over RDMA — full block+slot+staging path; GSM8K 5-shot = 96.5%.
  - DeepSeek-V3 (MLA) over RDMA — block-only path.
  - DeepSeek-V4 over fabric — single-node and cross-node within one UALink vPOD.

## Notes

- Read/pull does not support DeepSeek-V4 (V4 uses the region-based write path).
- Cross-TP (prefill_tp != decode_tp) writes are not yet supported and fail fast.

### DIFF
--- a/atom/kv_transfer/disaggregation/factory.py
+++ b/atom/kv_transfer/disaggregation/factory.py
@@ -52,6 +52,8 @@ class KVConnectorFactory:
         worker_class: str,
         scheduler_module: str,
         scheduler_class: str,
+        mem_pool_module: str | None = None,
+        mem_pool_func: str | None = None,
     ) -> None:
         """Register a KV connector backend.
 
@@ -61,12 +63,21 @@ class KVConnectorFactory:
             worker_class: Class name within *worker_module*.
             scheduler_module: Fully qualified module path for the scheduler connector.
             scheduler_class: Class name within *scheduler_module*.
+            mem_pool_module: Optional module path providing a custom KV
+                allocation pool for this backend (see
+                :meth:`get_kv_mem_pool_context`). Omit if the backend can
+                register/transfer ordinary torch memory (e.g. RDMA).
+            mem_pool_func: Name of a ``func(config) -> context-manager`` in
+                *mem_pool_module* returning a ``torch.cuda.use_mem_pool``
+                context (or nullcontext when not applicable).
         """
         cls._registry[name] = {
             "worker_module": worker_module,
             "worker_class": worker_class,
             "scheduler_module": scheduler_module,
             "scheduler_class": scheduler_class,
+            "mem_pool_module": mem_pool_module,
+            "mem_pool_func": mem_pool_func,
         }
 
     @classmethod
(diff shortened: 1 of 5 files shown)


ZhangLirong-amd:
@maning00  Hi, Do you have any performance detail compared with morio push and pull?

PR #1595, opened 2026-07-14
[atom-vllm] DeepSeek-V4: reconcile CUDA-graph padding in vLLM PIECEWISE attention

whx-sjtu:
## Summary

Fixes a crash that hangs the atom-vllm DeepSeek-V4 server mid gsm8k run:

```
[AITER] pa_sparse_prefill_opus_kernels.cu:60 kv_indptr_prefix length must be N+1
```

**Root cause.** Under `cudagraph_mode=FULL_AND_PIECEWISE`, `DeepseekV4Attention.forward` takes the **NARROW** split: `_attn_pre` (captured dense piece) → `v4_core_attention` op → `self._attn_core(...)` **directly**. It never goes through `forward_impl`. The vLLM plugin's padded-bucket reconciliation (introduced in #1573 / 724f08d) lived **only** in the `forward_impl` override — i.e. the WIDE/FULL path. So when a prefill/mixed batch small enough to fit under `max_cudagraph_capture_size` is padded up to a capture bucket (e.g. 342 real tokens → 352, visible as `M:352` GEMMs just before the abort), `_attn_core` gets a `T_pad`-row `q` while the sparse-attention metadata (`batch_id_per_token` / `kv_indptr_*`) is sized to the real token count. `sparse_attn_v4_paged_prefill` then aborts because `len(kv_indptr_prefix) != N+1`, killing `EngineCore` and hanging the server.

This is the path the launch config actually uses, so the earlier `forward_impl`-only fix never ran for it.

**Fix (plugin-only).** Add an `_attn_core` override in `DeepseekV4AttentionVllm` that mirrors the existing `forward_impl` reconciliation for the narrow path: for non-dummy, non-DECODE batches where `batch_id_per_token.shape[0] < x.size(0)`, clip every per-token input (`x, q, kv_pre, qr, qr_scale, positions, idx_q_fp8, idx_weights`) to the real token count, run the native core, then pad the output back to the bucket width (so the piecewise output buffer and the downstream graphed `_attn_post` keep the captured shape). `forward_impl` is kept for the WIDE path; the two are idempotent together. DECODE (metadata already padded to the bucket) and dummy runs are untouched. No native engine/kernel changes.

## Test plan
DeepSeek V4 Flash + TP4:
- [x] Accuracy healthy / uncorrupted: `flexible-extract = 0.9333`, `strict-match = 0.9340`.

### DIFF
--- a/atom/plugin/vllm/models/deepseek_v4.py
+++ b/atom/plugin/vllm/models/deepseek_v4.py
@@ -212,20 +212,35 @@ class DeepseekV4AttentionVllm(DeepseekV4AttentionBase):
 
     Under ``cudagraph_mode=FULL_AND_PIECEWISE`` vLLM captures/replays the dense
     regions of ATOM's torch.compiled graph at the padded bucket width, while the
-    ``deepseek_v4_attention`` op (a graph break, marked as a splitting op) runs
-    eagerly. So for a prefill/mixed batch whose bucket was captured, ``x`` /
-    ``positions`` arrive padded to ``T_pad``, but the sparse-attention metadata
-    is built for the *real* token count (the bridge's prefill path sets
-    ``batch_id_per_token`` to length == real tokens).
+    eager attention split op (a graph break) runs at that same padded width. So
+    for a prefill/mixed batch whose bucket was captured, ``x`` / ``positions``
+    (and every downstream per-token projection) arrive padded to ``T_pad``, but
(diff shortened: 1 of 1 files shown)


zejunchen-zejun (approved):

zejunchen-zejun:
atom-vllm V4 CI recovers

PR #1601, opened 2026-07-14
Fix(mxfp4): align activation quant rounding with Quark offline calibr…

thpereir:
…ation

MXFP4 activation quantization during inference used hardcoded HIP RoundUp rounding while Quark's offline weight calibration uses Even rounding.  The mismatch might degrade e2e accuracy: +1.82% GSM8K flexible-extract on Qwen3.5-35B-A3B-MXFP4 when switching to Even.

Changes
-------
atom/model_ops/linear.py
  - Add _resolve_mxfp4_act_round_mode(quant_config) which reads quantization_config.global_quant_config.input_tensors .scale_calculation_mode from the model checkpoint (Quark standard field) and maps it to the corresponding quant_mxfp4_hip round_mode int. Defaults to Even, which is Quark's calibration standard.
  - LinearBase.__init__ now resolves the round_mode once at load time and stores it as self.mxfp4_act_round_mode.
  - gemm_a4w4_quant accepts an act_round_mode parameter (default=2/Even) and uses functools.partial(quant_mxfp4_hip, round_mode=...) at all three activation quant call sites, replacing get_hip_quant(per_1x32) which always used RoundUp.
  - LinearBase.forward passes self.mxfp4_act_round_mode into gemm_a4w4_quant so the resolved mode is used at inference time.
  - Escape hatch: ATOM_ACT_QUANT_HIP_ROUNDUP=1 forces legacy RoundUp for debugging or comparison.

atom/quantization/quark/utils.py
  - quant_mxfp4_online_even: handle higher-rank weight tensors (e.g. fused MoE expert weights shaped [num_experts, N, K]) by reshaping to 2D before quant_mxfp4_hip (which requires 2D input) and restoring the leading dimensions on both packed-weight and scale outputs.

## Test Result

# Accuracy Results — GSM8K 3-shot, n=1319

| Model | Kernel | Rounding | flexible-extract | strict-match | Δ flex |
|-------|--------|----------|:----------------:|:------------:|:------:|
| Qwen3.5-35B-A3B-MXFP4 | HIP | RoundUp (before) | 0.8506 ± 0.0098 | 0.8378 ± 0.0102 | — |
| Qwen3.5-35B-A3B-MXFP4 | HIP | **Even (after)** | **0.8688 ± 0.0093** | **0.8537 ± 0.0097** | **+1.82%** |
| Kimi-K2.6-MXFP4 | HIP | RoundUp (before) | 0.8666 ± 0.0094 | 0.8567 ± 0.0097 | — |
| Kimi-K2.6-MXFP4 | HIP | Even (after) | 0.8673 ± 0.0093 | 0.8575 ± 0.0096 | +0.07% |
| GLM-5-MXFP4 | HIP | RoundUp (before) | 0.9371 ± 0.0067 | 0.9416 ± 0.0065 | — |
| GLM-5-MXFP4 | HIP | Even (after) | 0.9378 ± 0.0067 | 0.9416 ± 0.0065 | +0.07% |

Some extra results  MATH-500 and GPQA-Diamond

  ## Accuracy: activation-rounding fix (Even vs RoundUp)

  Two MXFP4 models, thinking mode enabled, greedy (temp 0), single run.
  **Fix ON = Even rounding** (default); **Fix OFF = RoundUp** (`ATOM_ACT_QUANT_HIP_ROUNDUP=1`).

  ### MATH-500 — `math_verify` (sympy), 500 Q, 0-shot, 32768 tok

  | Model | Even (fix ON) | RoundUp (fix OFF) | Δ (fix) |
  |-------|:-------------:|:-----------------:|:-------:|
  | Kimi-K2.6-MXFP4 | **0.972** ± 0.007 | 0.960 ± 0.009 | **+1.2** |
  | GLM-5.2-MXFP4   | **0.614** ± 0.022 | 0.580 ± 0.022 | **+3.4** |
  | Qwen3.5-35B-A3B-MXFP4 | **0.964** ± 0.008 | 0.954 ± 0.009 | **+1.0** |

  

  ### GPQA-Diamond — `exact_match`, 198 Q, 0-shot, 32768 tok

  | Model | Even (fix ON) | RoundUp (fix OFF) | Δ (fix) |
  |-------|:-------------:|:-----------------:|:-------:|
  | Kimi-K2.6-MXFP4 | 0.783 ± 0.029 | **0.803** ± 0.028 | −2.0 |
  | GLM-5.2-MXFP4   | **0.667** ± 0.034 | 0.657 ± 0.034 | +1.0 |
  | Qwen3.5-35B-A3B-MXFP4 | **0.778** ± 0.030 | 0.753 ± 0.031 | **+2.5** |

  

  ### Summary

  - On **MATH-500** (higher power, n=500) the Even-rounding fix improves **both** models — GLM +3.4 pts (~1.5σ), Kimi +1.2 pts.
  - On **GPQA-Diamond** (n=198) the fix is +1.0 for GLM and −2.0 for Kimi, but the Kimi result is within ~1σ and reverses on the larger MATH-500 set — i.e. small-sample noise, not a regression.
  - **Net: consistent with the fix's intent, no real regression on either model.**

  

  <details>
  <summary>Setup / caveats</summary>

  - Serving: ATOM openai_server, TP4, `--kv_cache_dtype fp8`, chat-completions endpoint with `--apply_chat_template` (thinking template applied server-side).
  - MATH-500 via ungated `HuggingFaceH4/MATH-500`; GPQA-Diamond via ungated mirror `fingertap/GPQA-Diamond` (official `Idavidrein/gpqa` needs an HF token).
  - MATH-500 headline metric is `math_verify` (sympy equivalence). Minerva's `exact_match` reads 0.000 for all cells — it keys on a rigid `\boxed{}` string that free-form thinking output doesn't match, so it's not meaningful here.
  - Single run, greedy; ± is the metric's stderr.

  </details>

## Submission Checklist

### DIFF
--- a/atom/model_ops/linear.py
+++ b/atom/model_ops/linear.py
@@ -20,6 +20,7 @@ from aiter import (
 # import torch.distributed as dist
 from aiter.dist.parallel_state import get_tp_group
 from aiter.jit.utils.torch_guard import torch_compile_guard
+from aiter.ops.quant import quant_mxfp4_hip
 from aiter.tuned_gemm import tgemm
 from aiter.utility import fp4_utils
 from torch import nn
@@ -45,6 +46,20 @@ from atom.utils import envs
 from atom.utils.decorators import mark_trace
 
 logger = logging.getLogger("atom")
+
+
+def _resolve_mxfp4_act_round_mode() -> int:
+    """Return the HIP quant_mxfp4_hip round_mode for activation quant.
+
+    Every other ATOM MXFP4 quant path (weights, MoE, fused SiLU/RMSNorm)
+    already produces Even scales, so we hardcode Even here for consistency.
+    Set ATOM_ACT_QUANT_HIP_ROUNDUP=1 to restore the legacy RoundUp path.
+    """
+    from aiter.utility.mx_types import MxScaleRoundModeInt
+
+    if envs.ATOM_ACT_QUANT_HIP_ROUNDUP:
+        return int(MxScaleRoundModeInt.RoundUp)
+    return int(MxScaleRoundModeInt.Even)
 
 
 def use_triton_gemm() -> bool:
@@ -116,6 +131,7 @@ def gemm_a4w4_quant_fake(
     params_dtype: torch.dtype,
     input_scale: torch.Tensor,
     output_size: int,
+    act_round_mode: int = 2,  # 2 = Even
 ) -> torch.Tensor:
     return torch.empty((*x.shape[:-1], weight.shape[0]), dtype=otype, device=x.device)
 
@@ -131,6 +147,7 @@ def gemm_a4w4_quant(
     params_dtype: torch.dtype,
     input_scale: torch.Tensor,
     output_size: int,
+    act_round_mode: int = 2,  # 2 = Even
 ) -> torch.Tensor:
     # Non-shuffle FP4 Triton path: keep x/weight/scale in the original MXFP4
     # layout and call the non-preshuffled gemm_afp4wfp4 kernel.
@@ -140,7 +157,7 @@ def gemm_a4w4_quant(
                 "ATOM_USE_FP4_NON_SHUFFLE_TRITON_GEMM=1 requires aiter.ops.triton.gemm_afp4wfp4"
             )
         if x_scale is None:
-            quant_func = get_hip_quant(QuantType.per_1x32)
+            quant_func = functools_partial(quant_mxfp4_hip, round_mode=act_round_mode)
             x, x_scale = quant_func(
                 x,
                 quant_dtype=params_dtype,
@@ -191,7 +208,7 @@ def gemm_a4w4_quant(
             device=x.device,
         )
         if x_scale is None:
-            quant_func = get_hip_quant(QuantType.per_1x32)
+            quant_func = functools_partial(quant_mxfp4_hip, round_mode=act_round_mode)
             x, x_scale = quant_func(
                 x,
                 quant_dtype=params_dtype,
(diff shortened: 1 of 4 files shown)


atom/model_ops/linear.py
```diff
@@ -42,6 +43,37 @@
 logger = logging.getLogger("atom")
 
 
+def _resolve_mxfp4_act_round_mode(quant_config=None) -> int:
```
lihaoyang-amd: How about moving this to the quark utils.py?

thpereir: @lihaoyang-amd after looking deeper at how the complete flow works, adding this dynamic setting might actually be worse than just hard coding `even` as the whole path already uses `even`.

I will drop this, set to `even` and keep a debug option to switch back to round-up if needed.

New push soon

atom/model_ops/linear.py
```diff
@@ -47,6 +48,22 @@
...
+
+    Every other ATOM MXFP4 quant path (weights, MoE, fused SiLU/RMSNorm)
+    already produces Even scales, so we hardcode Even here for consistency.
+    Set ATOM_ACT_QUANT_HIP_ROUNDUP=1 to restore the legacy RoundUp path.
+    """
+    import os
+
+    from aiter.utility.mx_types import MxScaleRoundModeInt
+
+    if os.environ.get("ATOM_ACT_QUANT_HIP_ROUNDUP", "0") == "1":
```
valarLip: switch to env.ATOM_xxxxx

tests/test_mxfp4_act_rounding.py
```diff
@@ -0,0 +1,344 @@
...
+        yield patches
+    finally:
+        for name, orig in saved.items():
+            if orig is None:
+                sys.modules.pop(name, None)
+            else:
+                sys.modules[name] = orig
+
+
+def _load_module(rel_path: str, module_name: str):
```
valarLip: why we need these stuff

valarLip (changes requested):

atom/model_ops/linear.py
thpereir: done

tests/test_mxfp4_act_rounding.py
thpereir: I had some issues with loading dependencies when running the tests... But I found a simpler way to run this now. Removed this method and updated the tests

PR #1605, opened 2026-07-15
[feat](gpt-oss): Eagle3 speculative decoding support for gpt-oss-120b

ProgMastermind:
## Summary
Adds Eagle3 speculative decoding support for `openai/gpt-oss-120b`.

## Changes
- `atom/models/gpt_oss.py`: capture final target layer `aux_hidden_state` and feed it to the Eagle3 draft.
- `atom/models/eagle3_llama.py`: pass `dtype` and `rope_scaling` to the draft RoPE; add `norm_before_fc` input RMSNorm before the Eagle3 `fc` projection; fix `fc_norm` chunking.
- `atom/spec_decode/eagle.py`: share target `embed_tokens`/`lm_head` with the Eagle3 draft when the draft checkpoint does not include them.
- `recipes/GPT-OSS.md`: add Eagle3 server example and accuracy test command.

## Verification
Server starts with:
```bash
  --model openai/gpt-oss-120b \
  --kv_cache_dtype fp8 \
  --gpu-memory-utilization 0.5 \
  --method eagle3 \
  --num-speculative-tokens 1 \
  --draft-model nvidia/gpt-oss-120b-Eagle3-throughput
```

- `python -m atom.benchmarks.benchmark_serving` (single stream, random ISL=9592 / OSL=1500, CONC=1): **158.5 tok/s** output throughput.
- `lm_eval` GSM8K 5-shot: **flexible-extract = 0.9030** (CI threshold >= 0.88).
- `lm_eval` GSM8K 3-shot: **flexible-extract = 0.9007** (CI threshold >= 0.88).

### DIFF
--- a/atom/models/eagle3_llama.py
+++ b/atom/models/eagle3_llama.py
@@ -114,6 +114,9 @@ class Eagle3LlamaAttention(nn.Module):
             max_position=max_position_embeddings,
             base=rope_theta,
             is_neox_style=True,
+            dtype=getattr(config, "dtype", None) or torch.bfloat16,
+            rope_scaling=getattr(config, "rope_parameters", None)
+            or getattr(config, "rope_scaling", None),
         )
 
         sliding_window = -1
@@ -334,6 +337,7 @@ class Eagle3LlamaModel(nn.Module):
         self.target_hidden_size = target_hidden_size
         self.num_aux_hidden_states = num_aux
         self.norm_output = getattr(config, "norm_output", False)
+        self.norm_before_fc = getattr(config, "norm_before_fc", False)
 
         # Independent embedding (vocab matches target model). The draft embed is
         # NOT shared with the (still TP-sharded) lm_head, so it can be replicated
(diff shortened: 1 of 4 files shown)


atom/models/eagle3_llama.py
```diff
@@ -412,13 +423,17 @@ def combine_hidden_states(self, aux_hidden_states) -> torch.Tensor:
                 )
             else:
                 fc_in = aux_hidden_states
+            if self.input_norm is not None:
```
valarLip, line 426: it‘s same thing like self.fc_norm?

ProgMastermind: Not the same , input_norm is one RMSNorm over the whole concatenated aux tensor, while fc_norm is a ModuleList that normalizes each aux chunk separately. The gpt-oss-120b checkpoint config has norm_before_fc=True, so the model expects the full [N, target_hidden_size * num_aux] tensor to be normalized before fc. fc_norm alone would normalize per chunk, which is a different operation, so input_norm is needed here.

PR #1640, opened 2026-07-20
[Feature] DeepSeek-V4 paged-SWA sparse checkpoint prefix-cache retention

yhl-amd:
## Summary

DeepSeek-V4's paged-SWA prefill only writes each chunk's trailing `window`
tokens to the content-addressed SWA cache, so cross-request prefix hits whose
reuse boundary sits mid-sequence are truncated by `bounded_hit()` (the SWA tail
is gone). On agentic branch/replay workloads this collapses the actual server
prefix-cache hit rate.

**Measured on identical BF16 / DeepSeek-V4-Flash / TP8 / c=48 / same AgentX trace:**
ATOM default reuse is dominated by SWA-tail loss while vLLM reaches ~97.5% on the
same box and budget. Root cause: ATOM stores SWA tails densely (write-all) in a
small, isolated pool, so live-window churn overwrites a tail before a branch
reuses it. vLLM keeps SWA tails **sparsely** in a shared pool
(`VLLM_PREFIX_CACHE_RETENTION_INTERVAL`, vllm-project/vllm#43447).

This PR adds opt-in **sparse checkpoint SWA-tail retention** (default off — dense
behavior byte-identical):

- `ATOM_SWA_FULL_RETAIN=1` writes/materializes the full chunk so mid-sequence
  tails exist to reuse; `swa_pool.py`/`model_runner.py`/`block_manager.py` size
  the SWA pool from a budget fraction (not a 1:1 mirror, which starves the
  compressed index).
- `ATOM_SWA_RETENTION_INTERVAL` (tokens; 32768 mirrors vLLM) makes `publish_hash`
  **pin** the SWA tail only at segment + prompt boundaries — an extra ref so
  `free_out_of_window`/eviction skip it, LRU-capped at `ATOM_SWA_CHECKPOINT_FRAC`
  of the pool. ATOM needs the explicit pin (vs vLLM's mask) because its SWA pool
  is small and isolated, not a shared pool. `bounded_hit` is unchanged and stays
  correct: it truncates a hit to the nearest present checkpoint, so a boundary
  whose tail was evicted is safely shortened, never mis-served.
- CacheStats instrumentation splits `Compressed-hit` vs `Lost-to-SWA-gate` to
  attribute reuse loss.

## Results (Flash BF16, TP8, c=48, mem0.9, same trace, warmed interval)

| metric | default dense (f=0.2) | **sparse 32k (this PR)** | vLLM |
|---|---:|---:|---:|
| Actual prefix hit | 5.6% | **96.45%** | 97.5% |
| Compressed-hit | 96.95% | 96.60% | ~97.5% |
| Lost-to-SWA-gate | 91.35% | **0.16%** | ~0 |

SWA-gate loss collapses 91% → ~0.2%; actual hit tracks compressed-hit within
~0.2 pt, matching vLLM. Stable at c=48 through 700+ requests, no `No free SWA
blocks`, no OOM.

## Test plan
- [x] Unit test: checkpoints survive window-free + churn; non-checkpoints
      reclaimed; LRU eviction bounded.
- [x] Flash BF16 c=48: hit 5.6% → 96.45%, gate loss 91% → 0.16%, no crash.
- [x] Default off (interval 0) byte-identical to prior window-only behavior.
- [ ] DeepSeek-V4-Pro BF16 c=48 sparse-on vs default-off end-to-end throughput.
- [ ] fp8 KV (needs an aiter build with the fused-decode `swa_block_tables`
      kwarg; smoke tests used bf16 to isolate the logic).


atom/model_engine/swa_pool.py
valarLip, line 1: let's rename this one as "state cache pool" later

valarLip (approved):

PR #1654, opened 2026-07-21
[DSV4] remove kv_last_len

amd-ruitang3:
## Summary
Remove the redundant `kv_last_page_lens` from the V4 page_size=1 asm decode
path. aiter's `mla_decode_fwd_v4_nm` derives kv_seq_len from the token-level
`kv_indptr` and ignores the value, so building/forwarding it is dead work.

## Changes
- deepseek_v4_attn.py: stop building/staging the `v4_kv_last_page_lens`
  CpuGpuBuffer (removes a per-fwd slice + H2D copy in both decode metadata paths).
- paged_decode.py: drop `kv_last_page_lens` from the aiter call; simplify the
  asm-decode return path (remove the `out_16_nosplit` / logits branch, return
  `output` directly).
- deepseek_v4.py: drop the `kv_last_page_lens` argument at the call site.

## Test plan
Verified on this branch (with the current aiter build), DeepSeek-V4-Pro fp8, tp8:
- gsm8k accuracy — all pass: base 0.9538, MTP 0.9530 (accept 64.4%),
  TBO+DPA conc1000 0.9454 (thresholds 0.94 / 0.94 / 0.93).
- perf conc256 1k/1k: 9292 tok/s total, no regression.
- `black --check` + `ruff check` clean.

### DIFF
--- a/atom/model_ops/v4_kernels/paged_decode.py
+++ b/atom/model_ops/v4_kernels/paged_decode.py
@@ -916,7 +916,7 @@ def _sparse_attn_v4_paged_decode_asm(
     q_packed_in: torch.Tensor,
     q_rope_in: torch.Tensor,
     qo_indptr: torch.Tensor,
-    kv_last_page_lens: torch.Tensor,
+    kv_last_page_lens: torch.Tensor | None = None,  # unused on v4 nm (page_size=1)
     num_kv_splits: int | None = None,
 ) -> torch.Tensor:
     """Native 2buff fp8 V4 decode via the aiter assembly kernel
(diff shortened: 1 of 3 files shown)


atom/model_ops/v4_kernels/paged_decode.py
```diff
@@ -916,7 +916,7 @@ def _sparse_attn_v4_paged_decode_asm(
     q_packed_in: torch.Tensor,
     q_rope_in: torch.Tensor,
     qo_indptr: torch.Tensor,
-    kv_last_page_lens: torch.Tensor,
+    kv_last_page_lens: torch.Tensor | None = None,  # unused on v4 nm (page_size=1)
```
valarLip, line 919: why still have this one

valarLip (approved):

PR #1660, opened 2026-07-21
[CI] harden aiter-whl download (digest-mismatch fallback) + raise fd limit for ATOM editable build

sunway513:
## What

Harden two intermittent, PR-agnostic failures in the `atom-sglang-test` accuracy jobs (`.github/workflows/atom-sglang-test.yaml`):

1. **`Download aiter wheel` flakiness.** `actions/download-artifact@v8` intermittently fails the ~600 MB `aiter-whl` transfer with a `digest-mismatch` after its 5 internal retries (observed on `DeepSeek-R1-FP4-V2 TP8 MTP3`, ~37 min then hard fail). The step is now `continue-on-error`, followed by a resilient fallback that re-fetches the same run-scoped `aiter-whl` artifact via the REST API with 5 backoff retries. A flaky artifact transfer no longer red-lights the accuracy job.

2. **ATOM editable build fd exhaustion.** `pip install -e .` (which builds the Rust extension via cargo) fails with `cargo build ... Too many open files (os error 24)` when the runner's default `nofile` soft limit is low. Raise it in the same shell before the build.

## Why

Both failures are unrelated to any model/PR change and reproduce on `main`:
- `unified_kv_rope` / cudagraph and download failures on in-flight PRs (e.g. #477) trace to these infra issues, not the PR content.
- The cargo fd failure hit `main`'s `Install ATOM and dependencies` step directly.

## Not covered here (needs infra / image owner)

The `DeepSeek-V4-Pro` accuracy job separately fails with `'DeepseekV4Attention' object has no attribute 'unified_kv_rope'` during SGLang cudagraph capture. `unified_kv_rope` is set in `atom/model_ops/attentions/deepseek_v4_attn.py` and read in `atom/models/deepseek_v4.py` — both consistent on current `main`, so this is a **stale/mismatched ATOM in the SGLang base image**. The fix is to rebuild the SGLang base image from current `main` and pin the base-image + aiter-wheel to a coherent set (rather than "latest main"). Tracked alongside ROCm/aiter#4304.

## Test plan

Re-run the `atom-sglang-test` accuracy matrix; the aiter-whl download should survive a digest-mismatch via the fallback, and the ATOM editable build should no longer hit the fd limit. (CI validation needs a maintainer run.)

### DIFF
--- a/.github/actions/atom-bench-container/action.yml
+++ b/.github/actions/atom-bench-container/action.yml
@@ -59,8 +59,8 @@ runs:
         MODEL_PROGRESS_INTERVAL: "60"
       run: |
         set -euo pipefail
-        if [ ! -d "/models" ]; then
-          echo "/models directory not found, skipping model download"
+        if [ -z "${MODEL_HOST_ROOT:-}" ]; then
+          echo "Model cache directory not found, skipping model download"
           exit 0
         fi
         model_dir="/models/${{ inputs.model-path }}"
--- a/.github/actions/setup-gpu-container/action.yml
+++ b/.github/actions/setup-gpu-container/action.yml
@@ -87,11 +87,20 @@ runs:
           DEVICE_FLAG="--device /dev/dri"
         fi
 
-        if [ -d "/models" ]; then
-          MODEL_MOUNT="-v /models:/models"
+        MODEL_HOST_ROOT=""
+        for d in /shared_nfs/huggingface_models /models /data/models; do
+          if [ -d "$d" ]; then
+            MODEL_HOST_ROOT="$d"
+            break
+          fi
+        done
+
+        MODEL_MOUNT=""
+        if [ -n "$MODEL_HOST_ROOT" ]; then
+          MODEL_MOUNT="-v ${MODEL_HOST_ROOT}:/models"
+          echo "MODEL_HOST_ROOT=${MODEL_HOST_ROOT}" >> "$GITHUB_ENV"
         else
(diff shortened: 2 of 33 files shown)


sunway513:
@okakarpa please review before merge. This hardens two PR-agnostic CI failures in `atom-sglang-test` (both reproduce on main and other PRs, see the linked analysis):
1. **aiter-whl download** — `download-artifact@v8` digest-mismatch on the ~600MB wheel after its 5 internal retries; added `continue-on-error` + a run-scoped API re-download with backoff (curl+python3, since gh CLI isn't on the self-hosted GPU runner).
2. **ATOM editable build** — `cargo ... Too many open files (os error 24)`; raise `nofile` ulimit before `pip install -e .`.

Static checks are green (Black/Ruff/actionlint/unit). The GPU accuracy jobs are path-skipped on this workflow-only PR, so the download-fallback/ulimit are exercised only on a full accuracy re-run. Cross-PR evidence + root causes: https://amd.atlassian.net/wiki/spaces/~pensun/pages/1809672378

gyohuangxin:
@sunway513 Thanks for the enhancement. This is very helpful. We already have a unified script for handling aiter wheel downloads, and I am optimizing it in this commit: https://github.com/ROCm/ATOM/pull/1658/changes/0fd8cb741ffae6807096425e5969bf6ff70dead3. I think I can adopt your optimization in that script after this lands, so we only need to maintain the logic in one place.

gyohuangxin (approved):

gyohuangxin (approved):

PR #1665, opened 2026-07-22
Drop stale GLM5.2 AITER JIT artifacts

qichu-yun:
## Motivation
Fix GLM-5.2 SGLang startup failures caused by stale AITER JIT artifacts baked into the image.

```
File "/app/aiter-test/aiter/ops/moe_op.py", line 597, in ck_moe_stage2_fwd
    ck_moe_stage2(...)
RuntimeError: Unsupported block_m value for moe heuristic dispatch: 16
Exception: Capture cuda graph failed: ...
```

## Technical Details
This PR removes the stale GLM-5.2 MoE b16 AITER JIT .so and its build directory during image build for both atom_sglang and atom_image. This forces AITER to regenerate the kernel from the current source at runtime, avoiding Unsupported block_m value for moe heuristic dispatch: 16.

## Submission Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -270,6 +270,12 @@ RUN echo "========== [SGLANG-ATOM 4/6] Restore base image Triton ==========" && 
     "${VENV_PYTHON}" -c "import triton; print(f'triton.__version__ = {triton.__version__}')" && \
     "${VENV_PYTHON}" -m pip show triton
 
+RUN echo "========== [SGLANG-ATOM] Drop stale GLM-5.2 AITER JIT artifacts ==========" && \
+    if [ -d /app/aiter-test/aiter/jit ]; then \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -270,6 +270,12 @@ RUN echo "========== [SGLANG-ATOM 4/6] Restore base image Triton ==========" &&
     "${VENV_PYTHON}" -c "import triton; print(f'triton.__version__ = {triton.__version__}')" && \
     "${VENV_PYTHON}" -m pip show triton
 
+RUN echo "========== [SGLANG-ATOM] Drop stale GLM-5.2 AITER JIT artifacts ==========" && \
```
valarLip, line 273: should fix in aiter instead of here,,

PR #1669, opened 2026-07-22
[fea](req): change transformers version to 5.12.1

PerryZhang01:
## Motivation

This PR updates ATOM’s Transformers requirement to 5.12.1. It excludes only aiperf’s Transformers dependency to prevent it from pulling the latest version, while SGLang remains pinned to 5.2.0 until upgraded.

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -579,6 +579,7 @@ _CONFIG_REGISTRY: dict[str, str] = {
     # through as extra config attrs and are read in DeepseekV4Args.from_hf_config.
     "glm_moe_dsa": "deepseek_v3",  # GLM 5.0 MoE, structure similar to DeepSeek v3.2
     "kimi_k2": "deepseek_v3",
+    "qwen3_next": "qwen3_next",
 }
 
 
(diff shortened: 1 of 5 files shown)


zhuyuhua-v (dismissed):

atom/config.py
valarLip, line 675: why this one not help for full_attention_interval..

PerryZhang01: That loop only runs for model types in _CONFIG_REGISTRY. qwen3_next is not registered there, so it follows the generic AutoConfig.from_pretrained() path and never reaches that re-injection logic.

In Transformers 5.12.1, Qwen3NextConfig.__post_init__() pops full_attention_interval to build layer_types without retaining it as an attribute, while ATOM still reads hf.full_attention_interval. Therefore we need to restore it explicitly after the generic loading path. The added test covers this case.

PerryZhang01: changed

valarLip (approved):

valarLip (approved):

PR #1681, opened 2026-07-23
[DSV4] perf(tbo): overlap pure-TP all_reduce on TBO + Delay TP

ZhangLirong-amd:
## Motivation

```
AITER_BF16_FP8_MOE_BOUND=0 ATOM_MOE_GU_ITLV=1 python -m atom.entrypoints.openai_server   --model=/mnt/m2m_nobackup/lirzhang/DeepSeek-V4-Pro/   --tensor-parallel-size=8  --max-num-batched-tokens=16384   --kv-cache-dtype=fp8   --gpu-memory-utilization=0.85   --enable-tbo
```
For each 16k prefill, time 567ms -> 495ms, about 15% improve for each prefill

image

## Submission Checklist

### DIFF
--- a/atom/model_ops/linear.py
+++ b/atom/model_ops/linear.py
@@ -924,7 +924,10 @@ class LinearBase(nn.Module):
                 if self.bias is not None:
                     y += self.bias
         if self.tp_dim == 1 and self.tp_size > 1 and self.reduce_results:
-            y = get_tp_group().all_reduce(y, ca_fp8_quant=False)
+            if getattr(self, "tbo_aware", False):
+                y = torch.ops.aiter.tbo_all_reduce(y)
+            else:
+                y = get_tp_group().all_reduce(y, ca_fp8_quant=False)
         return y
 
 
@@ -1818,6 +1821,7 @@ class RowParallelLinear(LinearBase):
         bias: bool = False,
         quant_config: Optional[QuantizationConfig] = None,
         reduce_results: bool = True,
+        tbo_aware: bool = False,
         source_quant_dtype: torch.dtype = None,
         prefix: str = "",
         **kwargs,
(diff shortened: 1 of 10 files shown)


atom/model_ops/linear.py
```diff
@@ -924,7 +924,10 @@ def forward(
                 if self.bias is not None:
                     y += self.bias
         if self.tp_dim == 1 and self.tp_size > 1 and self.reduce_results:
-            y = get_tp_group().all_reduce(y, ca_fp8_quant=False)
+            if getattr(self, "tbo_aware", False):
+                y = torch.ops.aiter.tbo_all_reduce(y)
+            else:
+                y = get_tp_group().all_reduce(y, ca_fp8_quant=False)
```
valarLip: let’s have a ATOM AR layer to deal with, so we don't have to add these "if..else.." for every model/op with AR

valarLip: i would like something like  from ato,.model_ops.communication_op import tensor_model_parallel_all_reduce

atom/utils/tbo/ubatching.py
```diff
@@ -305,6 +305,47 @@ def sync_dp_metadata(
 _CURRENT_CONTEXTS: list["TBOContext | None"] = []
 _NUM_UBATCHES: int = 2
 
+_TP_AR_ORDER_EVENT: "torch.cuda.Event | None" = None
+
+
+def tbo_get_tp_ar_order_event() -> "torch.cuda.Event | None":
```
valarLip: never used?

atom/model_ops/linear.py
ZhangLirong-amd: good idea

valarLip (approved):

PR #1699, opened 2026-07-25
feat(dp_sticky) : add new dp_sticky policy for dp-aware routing.

Yuechguo:
## Motivation

Data-parallel workers need stable session affinity so requests in the same conversation continue to reach the worker holding the session state. At the same time, inactive sessions should not remain pinned indefinitely when load distribution changes.

## Technical Details
Add the dp_sticky routing policy, selected through policy: { type: dp_sticky }.
Route requests with the same X-Session-ID to the same healthy worker.
Select the healthy worker with the lowest in-flight load for new sessions, missing session IDs, or unhealthy assignments.
Track each assignment’s last-access time.
Reassign sessions idle for more than five minutes to the current lowest-load healthy worker.

## Test Result
start atom server
```
HIP_VISIBLE_DEVICES=0,1 \
nohup python3 -m atom.entrypoints.openai_server \
    --model /data/models/Qwen3.5-27B-FP8 \
    --host 0.0.0.0 \
    --server-port 8001 \
    --tensor-parallel-size 1 \
    --data-parallel-size 2 \
    --trust-remote-code \
    --kv_cache_dtype fp8 \
    --gpu-memory-utilization 0.90 \
    > atom_dp2.log 2>&1 &
```
start atomesh
```
nohup /data/mesh-all/ATOM/atom/mesh/target/release/atomesh launch \
    --host 0.0.0.0 \
    --port 8000 \
    --worker-urls http://127.0.0.1:8001 \
    --policy dp_sticky \
    --dp-aware \
    --backend atom \
    --log-level debug \
    > atomesh_dp_sticky_8000.log 2>&1 &
```
start requests
```
HOLD_SESSION_ID="$(cat /proc/sys/kernel/random/uuid)"
SECOND_SESSION_ID="$(cat /proc/sys/kernel/random/uuid)"

echo "hold session: ${HOLD_SESSION_ID}"
echo "new session: ${SECOND_SESSION_ID}"

curl -sS http://127.0.0.1:8000/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -H "X-Session-ID: ${HOLD_SESSION_ID}" \
    -d "{
        \"model\": \"/data/models/Qwen3.5-27B-FP8\",
        \"messages\": [{\"role\": \"user\", \"content\": \"Write exactly 1000 words about data parallel routing.\"}],
        \"max_tokens\": 1024,
        \"stream\": false
    }" >/dev/null &

sleep 0.5

curl -sS http://127.0.0.1:8000/v1/chat/completions \
    -H 'Content-Type: application/json' \
    -H "X-Session-ID: ${SECOND_SESSION_ID}" \
    -d "{
        \"model\": \"/data/models/Qwen3.5-27B-FP8\",
        \"messages\": [{\"role\": \"user\", \"content\": \"Hello\"}],
        \"max_tokens\": 32,
        \"stream\": false
    }" >/dev/null

wait
```
result
image

## Usage for Aiperf
```
atomesh launch \
  --worker-urls http://worker-0:8000 http://worker-1:8000 \
  --policy dp_sticky \
  --dp-aware
```

```
export AIPERF_HTTP_X_SESSION_ID_FROM_CORRELATION_ID=true
aiperf ...
```


atom/mesh/src/policies/dp_sticky.rs
```diff
@@ -0,0 +1,277 @@
...
+struct WorkerIdentity {
+    url: String,
+    dp_rank: Option<usize>,
+}
+
+#[derive(Debug)]
+pub struct DpStickyPolicy {
+    /// Session ID -> worker identity. The DP rank distinguishes logical workers
+    /// that share a common endpoint URL.
+    assignments: DashMap<String, StickyAssignment>,
```
JiaoliangYu, line 39: Do used session IDs need to be aged out?

atom/mesh/src/policies/dp_sticky.rs
```diff
@@ -0,0 +1,277 @@
...
+                if let Some(index) = Self::healthy_worker_for_identity(workers, &assignment.worker)
+                {
+                    return Some(index);
+                }
+            }
+        }
+
+        // The assigned worker was removed or became unhealthy. The next
+        // selection establishes a replacement affinity.
+        self.assignments.remove(session_id);
```
JiaoliangYu: Do we need to make the remove and update operations atomic?

Yuechguo: It is an atomic operation. Dashmap in Rust uses a sharded lock mechanism, and remove is an atomic operation. Moreover, the entry API covers both Occupied and Vacant cases, and both of these operations are atomic.

atom/mesh/src/policies/dp_sticky.rs
Yuechguo: It is indeed necessary to age out, but the current design does not have a background mechanism to scan and perform age-out. This feature may overlap with the background management mechanism of the worker later on, so I don't have a definitive plan yet. However, since this is currently for testing the agentic dataset for SA, the number of session IDs is limited, so it shouldn't be a major issue for now.

JiaoliangYu (dismissed):
lgtm

JiaoliangYu (dismissed):
lgtm

gyohuangxin (approved):

PR #1725, opened 2026-07-29
[atom]fix lm_cache write back

zhuyuhua-v:
## Motivation
LMCache load-back restored KV data into allocated GPU blocks and advanced `seq.num_cached_tokens`, but the restored blocks never passed through `BlockManager.hash_blocks()`.
As a result, the current request could consume the restored KV, but later requests could not discover it through `BlockManager.can_allocate()`. Frequently reused prefixes were therefore loaded repeatedly from CPU instead of becoming subsequent GPU cache hits.

Hence, this PR will:
Publish LMCache-restored KV blocks into ATOM's GPU prefix-cache index after all TP workers complete the load successfully.
Preserve canonical hash mappings when concurrent requests restore the same prefix.
Track the exact LMCache load range and emit `OFFLOAD-PROMOTE` diagnostics.

## Technical Details
- Record the actual `[HBM prefix, LMCache prefix)` load range when emitting LMCache load metadata.
- After TP-wide load completion, publish complete block-aligned restored ranges through `BlockManager.publish_loaded_prefix()`.
- Keep an existing canonical `hash_to_block_id` mapping when duplicate physical blocks contain the same prefix.
- Annotate duplicate blocks with the correct chained hash so newly computed suffix blocks retain the correct parent hash.
- Skip promotion on failed loads and clear transient load-range state.
- Emit `[OFFLOAD-PROMOTE]` logs with the restored range and indexed token count.

## Test Plan
GLM5.2 fp4 tp4 agentic perf test

## Test Result
image

GPU+LMCache scenario gpu cache hit raise from 17.7% to 90.28 in conc=2, 60.68% to 91.49% in conc=8


atom/kv_transfer/offload/connector.py
```diff
@@ -757,6 +757,11 @@ def build_connector_meta(self) -> LMCacheOffloadMetadata:
                 self._clear_pending_load(sid)
                 continue
             # num_cached after load = max(HBM, offload); never drop below HBM.
+            # Persist the exact physical load range on the sequence. The
+            # scheduler uses it after all TP workers report success to publish
+            # the restored blocks into BlockManager's GPU prefix index.
+            seq.offload_load_start_tokens = hbm
+            seq.offload_load_end_tokens = lmc
```
yhl-amd: nowhere use it

zhuyuhua-v: removed

yhl-amd (dismissed):
LGTM

valarLip (dismissed):

valarLip (approved):

PR #1727, opened 2026-07-29
[fix] bunch of fixes to pass Kimi-K3 KVV 

gbyu-amd:
# Kimi-K3 KVV compliance fixes + tool/reasoning parser refactor

This PR is driven by the Kimi-K3 KVV (Kimi Vendor Verifier) test suite. While
fixing the failed cases found in the test, some code refactor has also been done
with the tool/reasoning parser.

Here are the details:

## Part 1 — KVV compliance fixes

Behavioral fixes that close the gaps the KVV suites flagged
(`tool_call_json_schema`, `k3_features`, `prompt_tokens`).

**`atom/entrypoints/openai/serving_chat.py`**
- `_normalize_finish_reason()`: maps the raw K3 stop-token id (e.g.
  `finish_reason="stop_163586"`) back to the OpenAI-standard
  `"stop"` / `"tool_calls"` / `"length"`. Previously the raw token id leaked into
  the response and masked otherwise-correct `k3_features` results.
- Threads `starts_thinking` into the streaming `ReasoningFilter` so streamed
  responses split `reasoning_content` correctly when the chat template opens the
  reasoning channel in the prompt.

**`atom/entrypoints/openai/api_server.py`**
- `_resolve_thinking()`: resolves K3 thinking on/off + effort for the chat
  template. `thinking.type == "disabled"` (or `reasoning_effort == "none"`)
  disables thinking; effort precedence is
  `thinking.effort` > `reasoning_effort` > template default.
- Forwards `thinking` / `thinking_effort` / `response_format` / `tool_choice`
  into the chat-template kwargs so K3 actually honors them (fixes the
  `response_format` injection gap behind the `prompt_tokens` / `response_format`
  cases).
- `_validate_chat_request()` (+ `_validate_one_tool` / `_validate_tool_list`):
  general request-shape validation — `tool_choice` enum, tool-name / uniqueness,
  and `response_format` shape — returned as HTTP 400 instead of an opaque 500.
- Seeds the streaming filter via `prompt_starts_in_reasoning()` so the server
  knows when the rendered prompt already opened the reasoning channel.

**`atom/entrypoints/openai/protocol.py`**
- Adds the request fields these fixes rely on: `thinking`, `response_format`,
  `reasoning_effort`.

> Note: an earlier param-immutability lock (`KIMI_PARAM_LOCK`) was intentionally
> **removed** — enforcing Kimi's parameter immutability is not a general
> serving-framework concern, so those `tests/params` cases fail by design.
> Request *shape* validation is kept.

## Part 2 — Tool parser refactor (`tool_parser.py` → `tool_parser/` package)

The old monolithic `tool_parser.py` (466 lines) mixed three unrelated
on-the-wire formats. Split into a package where each model dialect owns its file:

- **`tool_parser/base.py`** — shared primitives: `ToolCall`, id generation,
  param type-coercion, and the streaming `emit_call_events` helper.
- **`tool_parser/kimi_k3_parser.py`** — K3 XTML channel format
  (`<|open|>call tool="…"|>`). **This is the core KVV fix**: the old code had
  *no* K3 tool-call parser, so `tool_calls` was always empty. Adds non-stream
  `parse_kimi_k3` + streaming `K3StreamHandler`. Also fixes argument coercion to
  **preserve significant whitespace in string values** — `_k3_coerce` previously
  stripped every value, which corrupted string arguments whose leading/trailing
  whitespace is meaningful (e.g. an enum of whitespace strings,
  `TestEnforcerCases:38`); stripping is now applied only to numeric/bool/JSON
  coercion. This took `tool_call_json_schema` from 406/408 to 408/408.
- **`tool_parser/kimi_k2_parser.py`** — K2 special-token format (moved out of
  the monolith; its tokens now live local to the file).
- **`tool_parser/qwen3_parser.py`** — Qwen3 XML format (moved out).
- **`tool_parser/__init__.py`** — the two dispatchers (`parse_tool_calls`,
  `ToolCallStreamParser`); detection order K3 → Qwen → K2. Adding a model = one
  new module + one line.

## Part 3 — Reasoning parser refactor (dialect-driven engine)

`reasoning.py` had grown a thicket of per-model `if` branches. Refactored into a
data-driven engine:

- **`reasoning_dialects.py`** (new) — model-specific marker knowledge as data:
  a `ReasoningDialect` per family (`kimi_k3` XTML channels; `think_tag` for the
  `<think>…</think>` family covering DeepSeek / Qwen3 / K2 / MiniMax). Adding a
  model = add one dialect entry.
- **`reasoning.py`** — now dialect-agnostic. `separate_reasoning()` and the
  streaming `ReasoningFilter` iterate `DIALECTS`; **no per-model conditions
  remain**. The refactor is behavior-preserving (verified 286/286 differential
  match — 16/16 non-stream + 270/270 streaming — against the pre-refactor code).

Following vLLM's convention (each parser owns its token strings; no shared token
module), the K3 channel tokens are declared inline in the two files that use them
(`reasoning_dialects.py`, `kimi_k3_parser.py`), accepting minor duplication in
exchange for self-contained parsers.

## Files changed

```
 atom/entrypoints/openai/api_server.py              | 251 ++++++++++-
 atom/entrypoints/openai/protocol.py                |  14 +-
 atom/entrypoints/openai/reasoning.py               | 210 ++++++----
 atom/entrypoints/openai/reasoning_dialects.py      | 130 ++++++   (new)
 atom/entrypoints/openai/serving_chat.py            |  58 ++-
 atom/entrypoints/openai/tool_parser.py             | 466 ------------  (removed)
 atom/entrypoints/openai/tool_parser/__init__.py    | 116 +++++   (new)
 atom/entrypoints/openai/tool_parser/base.py        | 117 ++++++   (new)
 atom/entrypoints/openai/tool_parser/kimi_k2_parser.py | 172 ++++   (new)
 atom/entrypoints/openai/tool_parser/kimi_k3_parser.py | 132 ++++   (new)
 atom/entrypoints/openai/tool_parser/qwen3_parser.py   | 158 ++++   (new)
```

## Test results

Verified against a live Kimi-K3 deployment (MI355 × 8, tp=8, fp8 KV,
`max-num-batched-tokens=16384`) using the KVV suite from
`Kimi-Vendor-Verifier/tests`.

| Suite | Raw result | Effective result | Notes |
|-------|-----------|------------------|-------|
| `tool_call_json_schema` | **408/408** | 408/408 | run with `--reruns 3`; whitespace fix took `TestEnforcerCases:38` from a stable fail (406/408 pre-fix) to pass, no regressions |
| `k3_features` | 49/63 | ~52/63 pass, 2 skew, 9 skip | 3 dynamic-tool fails are model stochasticity (pass on `--reruns`); 2 fails are the +3 skew below |
| `prompt_tokens` | 0/44 | **0 real failures** | all 44 fail by a constant **+3** token skew — see below |
| `params` | 5/18 | by design | the 13 fails are all `test_wrong_param_rejected`; param-immutability enforcement was intentionally dropped (see note above) |
| `kimi_toolcall` † | 0.7345 acc | 0.7345 (±0.0417) | end-to-end tool-calling on 113 thinking samples, `temperature=1.0`; 113/113 completed, no errors |

† `kimi_toolcall` is **deprecated in the latest KVV repo** — the current
`Kimi-Vendor-Verifier` no longer ships this inspect-ai eval (`eval.py` only
exposes `ocrbench` / `mmmu` / `aime2025`), and tool-call verification is now the
`tests/tool_call_json_schema/` pytest suite above. It was run here from the
legacy harness purely as a supplementary end-to-end cross-check.

Reasoning-refactor equivalence: **286/286** differential cases match the
pre-refactor implementation (16/16 non-stream `separate_reasoning`,
270/270 streaming `ReasoningFilter`).

### On `prompt_tokens`: the 44 "failures" are not real defects

Every one of the 44 `prompt_tokens` cases fails by **exactly +3 tokens**
(server reports N+3, groundtruth expects N). This is a **counting-convention
difference, not a bug**, and it is not fixable without diverging from standard
serving semantics.

The 3 tokens are precisely the **assistant generation stub
`<|open|>think<|sep|>`** — the thinking-channel opener that the chat template
appends when `add_generation_prompt=True`. Verified on every case:
`apply_chat_template(..., add_generation_prompt=True)` ends in exactly
`['<|open|>', 'think', '<|sep|>']`, and `rendered_tokens - 3 == expected`
(e.g. 102→99, 94→91, 113→110).

- **This server (and vLLM / OpenAI-compatible servers in general)** count the
  generation-prompt stub in `prompt_tokens`, because those tokens *are* fed to
  the model as input before the first generated token.
- **The KVV groundtruth** uses Kimi's internal `chatt` /
  `MessageTemplateEncoderV4` convention, which treats `<|open|>think<|sep|>` as
  the *start of generation* and excludes it. The KVV README states this
  explicitly: `expected_prompt_tokens = len(golden prompt ids) - 3`,
  "subtracting the assistant generation stub (`[open]think[sep]`, 3 tokens)".

The skew is a **uniform +3 across all 44 cases** (never variable), which is
itself evidence that the server matches the groundtruth on everything else —
including the subtle rule that each assistant *history* message carries an extra
empty think block (6 tokens). If any of that were mishandled, multi-turn cases
would show a different offset; they do not. The single residual difference is
the trailing generation stub.

> If strict KVV `prompt_tokens` parity is required, it can be added as an opt-in
> mode that subtracts the trailing `<|open|>think<|sep|>` from the reported
> count. This is **not** enabled by default, because it would make
> `prompt_tokens` deviate from vLLM/OpenAI semantics for every non-KVV client.

The same +3 also accounts for the 2 `k3_features` `response_format` failures
(those assertions likewise check `prompt_tokens`).

### DIFF
--- a/atom/entrypoints/openai/reasoning_dialects.py
+++ b/atom/entrypoints/openai/reasoning_dialects.py
@@ -1,0 +1,147 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Reasoning-channel dialects (model-specific data for the general engine).
+
+The engine in ``reasoning.py`` is dialect-agnostic: it iterates ``DIALECTS`` to
+detect/split the reasoning channel. All model-specific marker knowledge lives
+here. Adding a model = add one ``ReasoningDialect`` entry (and its ``split`` for
+whole-response separation).
+
+Two dialects today, named by format rather than model:
+  - inline ``<think>...</think>`` (DeepSeek-R1, Qwen3, Kimi-K2, MiniMax, ...).
+    The opening tag may be emitted in the output or injected by the template.
+  - structured channel format: one stream split into named channels (think /
+    response / tools), each wrapped in framing tokens. The same concept as
+    OpenAI Harmony's analysis/final/commentary channels (gpt-oss). The opening
+    tag is template-injected, so the output begins *inside* the reasoning
+    channel. Different channel-format models use different framing tokens; the
+    entry below carries Kimi-K3's (``<|open|>think<|sep|>`` ...), and another
+    such model would add its own entry with its own tokens.
+"""
+
+import re
+from collections.abc import Callable
+from dataclasses import dataclass
+
+# Structured-channel format tokens (``<|open|>SECTION<|sep|>`` ... framing).
+# Named by the format concept, not the model: channel formats are a cross-model
+# pattern (e.g. gpt-oss/Harmony uses the same idea with different framing tokens).
+# The token *values* below are Kimi-K3's — a different channel-format model would
+# declare its own values. Declared locally so this module is self-contained
+# (each parser owns the token strings it uses); the tool-call parser keeps its
+# own copies of the subset it needs.
+CHANNEL_THINK_START = "<|open|>think<|sep|>"
+CHANNEL_THINK_END = "<|close|>think<|sep|>"
+CHANNEL_RESPONSE_START = "<|open|>response<|sep|>"
+CHANNEL_RESPONSE_END = "<|close|>response<|sep|>"
+CHANNEL_MESSAGE_END = "<|close|>message<|sep|>"
+CHANNEL_END_OF_MSG = "<|end_of_msg|>"
+CHANNEL_TOOLS_START = "<|open|>tools<|sep|>"
+CHANNEL_CALL_PREFIX = '<|open|>call tool="'
+
+# Result of splitting a full response: (reasoning_content or None, content).
+SplitResult = tuple[str | None, str]
+
+
+@dataclass(frozen=True)
+class ReasoningDialect:
+    """How one model family delimits its reasoning channel.
+
+    - ``prompt_open_marker``: what a rendered prompt ends with when the template
+      has already opened the reasoning channel (output then begins in reasoning).
+    - ``output_open_marker``: the marker the model *emits* to open reasoning
+      mid-stream (``<think>``); ``None`` when the template injects it instead.
+    - ``think_end_marker``: the marker that ends the reasoning channel.
+    - ``split``: whole-response separator returning ``SplitResult`` or ``None``
+      if this dialect does not apply to the text.
+    """
+
+    prompt_open_marker: str
+    output_open_marker: str | None
+    think_end_marker: str
+    split: Callable[[str], SplitResult | None]
+
+
+# --- Structured-channel dialect ---
+
+
+def _strip_channel_response_markers(text: str) -> str:
+    # Preserve tool-call sections: they follow <|close|>response<|sep|> (an
+    # empty response channel), so truncating at CHANNEL_RESPONSE_END would drop
+    # the whole tools block. Leave it intact for parse_tool_calls to handle.
+    if CHANNEL_TOOLS_START in text or CHANNEL_CALL_PREFIX in text:
+        return text
+
+    text = text.removeprefix(CHANNEL_RESPONSE_START)
+
+    for marker in (CHANNEL_RESPONSE_END, CHANNEL_MESSAGE_END, CHANNEL_END_OF_MSG):
+        if marker in text:
+            text = text.partition(marker)[0]
+    return text.strip()
+
+
+def _split_channel(text: str) -> SplitResult | None:
+    combined = CHANNEL_THINK_END + CHANNEL_RESPONSE_START
+    if combined in text:
+        reasoning, _, content = text.partition(combined)
+        return (reasoning.strip() or None, _strip_channel_response_markers(content))
+    if CHANNEL_RESPONSE_START in text:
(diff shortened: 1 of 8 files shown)


atom/entrypoints/openai/reasoning_dialects.py
```diff
@@ -0,0 +1,147 @@
...
+# model. Channel-format models differ in their framing tokens, so each gets its
+# own DIALECTS entry; the entry below carries Kimi-K3's token values.
+#
+# Detection/priority order: structured-channel dialects before inline-tag ones,
+# so a specific channel marker is tried before the generic <think> tag.
+# separate_reasoning() returns the first dialect whose split() matches. A dialect
+# is identified by its markers/split behavior, not a label.
+DIALECTS: tuple[ReasoningDialect, ...] = (
+    # Structured channel format — Kimi-K3 token values (see CHANNEL_* above)
+    ReasoningDialect(
```
gbyu-amd, line 138: @valarLip  remove hardcoded KIMI_K3 already.

valarLip (approved):

PR #1682, opened 2026-07-23
Enable index_share_for_mtp_iteration in native MTP EagleProposer.

fanxingran:
Wire GLM-5.2 draft index_share_for_mtp_iteration like vLLM: step 0 runs the MTP indexer, steps 1+ reuse sparse_kv_indices_buffer via skip_topk and compact_topk_indices. Gated on method=mtp, DSA index_topk, and the config flag so other draft backends are unchanged.

## Motivation

GLM-5.2-MXFP4 sets `index_share_for_mtp_iteration: true` in `config.json`. Within one MTP draft iteration, step 0 should run the DSA indexer once; steps 1+ should reuse the same top-k indices from `sparse_kv_indices_buffer` instead of re-scoring the full KV cache.

vLLM already implements this in the MTP proposer. ATOM native `openai_server` did **not**, so each of the 3 draft steps ran a full indexer pass. This PR **aligns ATOM draft behavior with vLLM and the model config** (correctness / parity), not only a micro-optimization.

**Scope:** draft propose only. Target verify is unchanged (long-ISL gap vs vLLM is mostly verify + acceptance; separate follow-up).

**Goals:**

- Honor `index_share_for_mtp_iteration` for GLM-5.2 MTP (n=3)
- Reduce redundant draft indexer work (3× → 1× per propose)
- No weight, algorithm, or verify-path changes; other draft backends unchanged

---

### Changed files

| File | Change |
|------|--------|
| `atom/models/deepseek_mtp.py` | Add `set_skip_topk` / `compact_topk_indices` on `DeepSeekMultiTokenPredictor`; delegate from `DeepSeekMTP` |
| `atom/spec_decode/eagle.py` | Gate + drive skip/compact in `EagleProposer.propose()` |
| `tests/test_mtp_index_share.py` | Unit tests for helpers and gate logic |

### `deepseek_mtp.py`

- `DeepSeekMultiTokenPredictor.set_skip_topk(skip)` — sets `self_attn.skip_topk` on MTP layers with an indexer
- `DeepSeekMultiTokenPredictor.compact_topk_indices(slot_ids)` — copies top-k rows from `last_token_indices` into the front of `sparse_kv_indices_buffer`
- `DeepSeekMTP` delegates both methods

### `eagle.py`

In `__init__`, detect `_share_mtp_indices` and log when enabled.

In `propose()` loop:

- Step 0, before forward: `set_skip_topk(False)`
[...]

PR #1742, opened 2026-07-30
fix(mtp): recycle post-final-norm hidden between MTP draft steps

fanxingran:
Draft step 0 gets the target's post-final-norm hidden, but steps 1+ were fed the pre-final-norm block output, skipping shared_head.norm. RMSNorm's per-channel weight is not idempotent, so the error compounded down the draft chain and acceptance collapsed at later steps; output stayed correct because the target verifies every token. Chain the post-final-norm hidden instead -- already materialized for the LM head, so it is free. Matches vLLM and SGLang.

A/B on 54f4d00 (GLM-5.2 MXFP4, n=3, TP4): +7~9% tok/s, accept_len 3.04 -> 3.46, GSM8K unchanged.

## Motivation

MTP draft step 0 is fed the target's post-final-norm hidden, but steps 1+ were fed the MTP block's
pre-final-norm output, skipping `shared_head.norm` — an input that layer never saw during training.
RMSNorm's divide-by-RMS is idempotent, but its per-channel weight is not (GLM-5.2 spans roughly
`[-0.07, 1.81]`), leaving a ~15% relative L2 deviation that compounds down the draft chain. Output
stays correct because the target verifies every drafted token, so the only symptom was a suppressed
acceptance rate at later draft positions.

vLLM carried the same bug and fixed it in vllm#45895. #1682 ported the `index_share` half of that
PR; this is the other half. SGLang's `deepseek_nextn` has always done it this way.

## Technical Details

- `compute_draft_token(..., return_normed=True)` also returns the post-final-norm hidden. That
  tensor is already materialized for the LM head, so this is free — no extra kernel, no extra
  memory traffic.
- `EagleProposer` chains it into steps 1+ in place of the raw block output.

Gated on `returns_postnorm_hidden`, so only DeepSeek MTP drafts change behaviour — `deepseek_v3`,
`deepseek_v32`, `glm_moe_dsa` per `_MTP_TYPE_MAP`. EAGLE3 and the other MTP families are untouched,
matching vLLM's `DeepSeekMTPModel` gate. No new kernels, no new config; prefill and verification
are unchanged.

## Test Plan

In-place A/B on the parent commit (54f4d00): same box, same container, patch applied then reverted,
so the only delta is this diff. GLM-5.2-MXFP4, TP4, `--kv_cache_dtype fp8`, `--method mtp
--num-speculative-tokens 3`, cudagraph FULL.

- Throughput: `benchmark_serving`, `dataset=random`, `ignore_eos`, 160 prompts, OSL 1024, CONC 16,
  ISL ∈ {1024, 8192, 10240}.
- Accuracy: GSM8K, full 1319 questions, 5-shot, `temperature=0`.

## Test Result

| ISL | out tok/s | TPOT mean | accept_len |
|---|---|---|---|
| 1024 | 1726 → **1873** (+8.5%) | 8.77 → **7.96 ms** | 3.044 → **3.452** |
| 8192 | 1253 → **1342** (+7.1%) | 11.80 → **10.99 ms** | 3.073 → **3.465** |
| 10240 | 1147 → **1230** (+7.2%) | 12.87 → **11.86 ms** | 3.082 → **3.476** |

GSM8K unchanged at 96.06% (1267/1319), against 96.06–96.29% spread across baseline reruns. Median
TTFT is flat, as expected since prefill is untouched.

On natural text the gain is larger (+16.8% tokens/forward) and the per-position breakdown shows the
compounding signature directly: pos0 87.7% → 87.7%, pos1 68.3% → **80.6%**, pos2 36.3% → **80.8%**.
Step 0 is unaffected because it always received the correct input.

## Submission Checklist

### DIFF
--- a/atom/models/deepseek_mtp.py
+++ b/atom/models/deepseek_mtp.py
@@ -194,7 +194,8 @@ class DeepSeekMultiTokenPredictor(nn.Module):
         self,
         hidden_states: torch.Tensor,
         spec_step_idx: int = 0,
-    ) -> torch.Tensor:
+        return_normed: bool = False,
+    ) -> torch.Tensor | tuple[torch.Tensor, torch.Tensor]:
         """Greedy draft token via distributed argmax over the TP-sharded vocab —
         avoids all-gathering the full [N, vocab] logits every draft step.
 
@@ -202,11 +203,19 @@ class DeepSeekMultiTokenPredictor(nn.Module):
         rank's logit shard to (max_val, global_idx) and all-gathers only [N, 2]
         instead of the O(vocab) logits. Token-identical to
         compute_logits(...).argmax(-1).
+
+        ``return_normed`` also hands back the post-final-norm hidden the next
+        draft step must consume. shared_head.norm is the MTP layer's counterpart
+        of the backbone's final norm, and its per-channel weight is not
+        idempotent, so recycling the pre-norm hidden feeds the draft an
+        off-distribution input and costs acceptance. It is already materialized
+        here, so returning it is free.
         """
         current_step_idx = spec_step_idx % self.num_mtp_layers
         mtp_layer = self.layers[str(self.mtp_start_layer_idx + current_step_idx)]
         normed = mtp_layer.shared_head(hidden_states)
-        return mtp_layer.shared_head.head.compute_argmax_token(normed)
+        token = mtp_layer.shared_head.head.compute_argmax_token(normed)
+        return (token, normed) if return_normed else token
 
     def set_skip_topk(self, skip: bool) -> None:
         """Toggle ``skip_topk`` on MTP sparse-attention layers.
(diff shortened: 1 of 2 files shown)


fanxingran:
@junhaha666 @whx-sjtu review please when you get a chance — ROCm/ATOM#1742.
MTP draft steps 1+ were fed the pre-final-norm hidden, skipping
`shared_head.norm`. RMSNorm's per-channel weight isn't idempotent, so the error
compounded down the draft chain and acceptance collapsed at later steps; output
stayed correct since the target verifies every token. Now chains the
post-final-norm hidden, already materialized for the LM head, so it's free.
A/B on GLM-5.2 MXFP4, TP4, n=3: +7~9% output tok/s, accept_len
3.04 → 3.46, GSM8K flat at 96.06%. Per-position: pos0 flat, pos1 68→81%,
pos2 36→81%.
2 files, +56/-6, no new kernels or config.  Thanks！

valarLip:
thanks your contribution, however can we get this PR more clean

fanxingran:
> thanks your contribution, however can we get this PR more clean

@valarLip thanks for your review! Trimmed. Diff is down from +56/-6 to +33/-6: dropped the
dead `apply_final_norm()` path, swapped the `hasattr` probe for an explicit
`returns_postnorm_hidden` flag, and flattened the branch in `propose()`. Logic
unchanged; description shortened too.

yitingw1 (approved):
LGTM

PR #1756, opened 2026-07-31
[Kimi-K3] assign 0 to padded slot instead of directly return with the initialized value

zejunchen-zejun:


zejunchen-zejun:
close as it theoratically doesn't really make sense, the pad row will not affect the acc value because all computations are row isolated

PR #1759, opened 2026-07-31
support mamba prefix cache

ganyi1996ppo:
# Prefix caching for mamba-like models (Qwen3-Next / Qwen3.5 / Kimi KDA)

Adds prefix caching support for linear-attention models — Qwen3-Next,
Qwen3.5, and Kimi K3. On Kimi K3 this is **2.4× better TTFT p50 and +35%
output throughput** on a shared-prefix workload, with no accuracy change.

ATOM has not been able to reuse a prefix on these models. The recurrent state
of a linear-attention layer is a fold over every token, so a KV prefix hit
that is not clamped to a published state checkpoint resumes the recurrence
mid-stream — **wrong output, silently, at full cache speed**. Rather than risk
that, `BlockManager.can_allocate` declines *every* KV hit for such a model.

Making those hits safe needs machinery that did not exist: periodic
checkpoints of the recurrent state, published on a grid the KV layer can
address, and a hit clamped down to the deepest one available. That is what
this PR adds, for both linear-attention families ATOM supports.

One real defect comes along with it. `enable_prefix_caching` defaults to on
(`config.py:1313`), so a Qwen3-Next / Qwen3.5 / Kimi K3 user today gets a flag
that reads as enabled while doing nothing — no reuse, no error, no log line
saying why. After this PR the flag either works or explains itself.

## Structure

Two layers, and they are worth reviewing separately — the commits are already
split this way:

**1. Generic SSM state cache** (`b0fdefa85`) — model-independent. The pool, the
clamp, the scheduler/block-manager integration, and the checkpoint save/restore
path for the GDN (gated delta-net) kernels that Qwen3-Next and Qwen3.5 use.

| component | what it does |
|---|---|
| `state_cache.py` (new, 610 lines) | the pool: checkpoint entries, LRU with lazy eviction, `bounded_hit` clamp, stats |
| `block_manager.py` | the gate — declines a KV hit that no checkpoint can back, clamps the rest |
| `sequence.py` | per-seq checkpoint bookkeeping: `state_load_slot`, `state_save_slot/pos`, `block_hashes`, `fork_hit_blocks`, `has_recurrent_state` |
| `gdn_attn.py`, `attention_gdn.py` | `apply_state_cache_loads` / `_checkpoint_targets`; checkpoints live in the same tensor as runtime slots so save/load is a plain slice copy |
| `fla_ops/state_checkpoint.py` (new) | the checkpoint write kernel |
| `fla_ops/chunk.py`, `chunk_delta_h.py` | paged variant that hands back per-chunk states, so an interior checkpoint costs no extra forward |
| `config.py`, `arg_utils.py`, `model_runner.py`, `scheduler.py` | gating, sizing, stats logging |

**2. Kimi K3 / KDA wiring** (`4ce9149ba`, `8d2923721`) — the second family. Kimi
uses KDA rather than GDN, so it needs its own paged kernel; everything above it
is shared.

| component | what it does |
|---|---|
| `fla_ops/chunk_kda.py` (new) | `chunk_kda_paged` — same math as `_run_kda`, verified bit-exact against `fla.ops.kda.chunk_kda`, but pool-indexed in-kernel so states come back for checkpoint slicing |
| `models/kimi_k3.py` | `_run_kda_paged`, `gather_kda_initial_state`, checkpoint write at the KDA call site |
| `config.py` auto-enable | `kimi_linear` in `LINEAR_ATTENTION_MODEL_TYPES` |
| `docs/ssm_state_cache_design.md` (new) | design doc |

The two families reach the same machinery by different routes:
`ModelRunner.is_qwen_next()` → GDN, `is_kimi_linear()` → KDA. A test pins that
every member of `LINEAR_ATTENTION_MODEL_TYPES` is matched by one of them, so a
future linear model cannot be added to one place and forgotten in the other.

## Approach

**The state cache is derived, not configured.** It is a precondition of prefix
caching on these models, not a companion knob, so pairing two flags was the
bug. `Config.__post_init__` enables it when all three hold:

1. `enable_prefix_caching` is on,
2. the model type is in `LINEAR_ATTENTION_MODEL_TYPES`,
3. `--block-size` divides the 64-token kernel chunk grid (16, 32, or 64).

There is no `--enable_ssm_state_cache` flag. `--no-enable_prefix_caching`
turns off both.

**An incompatible block size warns and declines; it never raises.** A command
line that works today keeps working, with exactly the (correct,
hit-declining) behaviour it has now:

```
Linear-attention model (kimi_linear) with prefix caching, but --block-size 128
does not divide the kernel chunk grid (64), so the SSM state cache cannot be
enabled and every prefix hit will be declined (correct, but no reuse). Use
--block-size 16, 32, or 64 to get it.
```

Why 64: it is the only grid on which split-and-replay of the recurrence is
bit-exact. A block size above it floors `blocks_per_ckpt` to 0 (nothing is
ever checkpointed); one that does not divide it puts checkpoints off the KV
block boundaries that hit length is expressed in. Two requirements, one check.

**`LINEAR_ATTENTION_MODEL_TYPES` is deliberately narrower than
`InputOutputProcessor._per_req_cache_model_types()`.** That set also covers
per-request state which is not a recurrence and cannot be checkpointed —
notably `deepseek_v4`, which implements none of this machinery and forces
`kv_cache_block_size = 256`, a size this feature's divide-64 rule rejects.
Auto-enabling on `has_per_req_cache` would have caught it. Every member of the
new set must also appear in the older one, or its sequences get no state slot
at all; a test pins the containment in both directions.

**Sizing is a share of what is actually left.** `--ssm_state_cache_ratio`
(default **0.3**) applies to the KV capacity remaining *after* the per-request
runtime state tensor, not to the budget before it — the old denominator
included memory the checkpoints could never have had.

This is a **retention** budget, not a concurrency one. Eviction is lazy, so a
published checkpoint stays hittable after its sequence ends, and the deep slots
are where cross-request reuse lives. Judge the size against `evict` in the
stats line, not against `max_num_seqs`.

## Results — Kimi K3, TP8, MI355X

Model: Minimax-m3-xiaobing-0727, `--block-size 64 --max-num-seqs 64
--max-model-len 16384 --kv_cache_dtype fp8`.

### Throughput and latency

Length-matched A/B, same server, 128 prompts × 2 passes, concurrency 16,
294,912 total input tokens in each arm:

| arm | TTFT mean | TTFT p50 | out tok/s | duration | state hits |
|---|---|---|---|---|---|
| shared 2048-token prefix | **1342 ms** | **788 ms** | **421** | 38.9 s | 128/128 |
| no shared prefix (control) | 2215 ms | 1859 ms | 313 | 52.0 s | **0/256** |

**2.4× better TTFT p50, +35% output throughput.**

The load-bearing number is `0/256`: across both control passes the cache never
hit once. The hits in the treatment arm are the shared prefix being reused —
not an internal self-match, not coincidence.

### Hit rate

| pass | hits / 128 reqs | writes | evict | saved_tok |
|---|---|---|---|---|
| cold (populates) | 119 (93.0%) | 117 | 129 | 243,712 |
| warm (can hit) | **128 (100%)** | 15 | 0 | 262,144 |

**Exactly 2048.0 tokens saved per hit** — the full prefix, nothing lost to
checkpoint-grid rounding. The cold pass already hitting 93% is expected: at
concurrency 16 the first wave publishes checkpoints and the rest of the pass
hits them.

KV side agrees: 82.0% cached/total in the shared-prefix window vs 22.5% in the
control.

### Accuracy — unchanged

GSM8K, cache-on bracketed by two cache-off controls run either side of it:

| run | state cache | exact_match (strict) |
|---|---|---|
| control 1 | off | 0.9530 |
| **treatment** | **on** | **0.9507** |
| control 2 | off | 0.9484 |

The cache-on result sits *between* the two controls, which differ from each
other by more than either differs from it. No detectable change — the expected
result, since this is a gating and sizing change, not a math change. Two
controls rather than one because a single A/B proves nothing at this noise
floor.

### Sizing on this config

```
SSM state cache: 291 checkpoint slots (15.22GB, 29.7% of the 51.34GB left
after runtime state) sized by --ssm_state_cache_ratio 0.300; granularity=64
```

The ratio binds, not a cap. An earlier `2*max_num_seqs` cap clamped this to
128 slots and evicted 91% of what it wrote; it was removed because it bounded
the *write* path while retention — the *read* path — is what reuse depends on.

### Gating paths, each verified on a real server

| launch | evidence | outcome |
|---|---|---|
| `--block-size 64 --enable_prefix_caching` | auto-enable line + sizing line | cache on |
| `--block-size 128 --enable_prefix_caching` | decline warning, no sizing line | warns, serves, declines hits — **no crash** |
| `--no-enable_prefix_caching` | no sizing line | both off |

### Coverage gap, stated plainly

**Only the KDA / Kimi K3 path was measured end-to-end.** The GDN path
(`qwen3_next`, `qwen3_5_text`, `qwen3_5_moe_text`) is in the auto-enable set,
shares every layer above the kernel, and its checkpoint kernels are unit-tested
— but no Qwen server benchmark or accuracy run is in this PR. If a reviewer
wants that before merge, it is a fair ask; the alternative is to narrow the
auto-enable set to `kimi_linear` and add the others when measured.

## Reviewer notes

**Do not read `Engine kwargs:` to confirm the cache is up.** That dump is the
`EngineArgs` fields and the flag is not one of them any more. The load-bearing
signal is the sizing line from `get_num_blocks`, emitted only when the cache
is enabled — its absence is the real negative.

**The logged `hit_rate=` understates the truth and should not be read as a
health metric.** It reported 4.6% while the per-request rate was 100%.
`lookups` counts *probes*, not requests: `bounded_hit` (`state_cache.py:394`)
scans checkpoint boundaries deepest-first with early exit, so a hit costs ~1
lookup while a long miss costs one per boundary. `saved_tok` and
hits-per-request are the honest measures. Worth a follow-up to either rename
the field or count it per-request.

**`has_recurrent_state` is deliberately not derived from `num_cached_tokens`.**
That counter tracks the paged-KV prefix hit and is set at admission, before any
forward runs — while the freshly popped per-request slot still holds the
*previous* tenant's state, because `deallocate` returns slots with contents
intact. Deriving it would resume the recurrence from a foreign sequence. This
is the subtlest correctness point in the PR.

**`all_reduce(MIN)` on the slot count is load-bearing.** The remaining-KV
budget varies per rank (measured on Kimi TP8: ranks computed 51/51/52/52/53,
all agreeing on 51 after the reduce). Ranks must not disagree on tensor shape.

**`max(1, ...)` on the slot count is load-bearing.** `StateCachePool.enabled`
is `num_slots > 0`, so flooring to 0 would turn every method into a silent
no-op rather than an error.

**Two fla/Triton fixes ride along and are not cosmetic:**
- `chunk_kda.py` applies `l2norm_fwd` to q and k, matching what
  `ChunkKDAFunction.forward` does before `chunk_kda_fwd`. Without it every
  prefill under the state cache is wrong by ~27% relative — silently.
- `state_checkpoint.py` passes real strides (`stride_x_t`, `stride_cd_s/d/l`)
  because `conv_state` is a transposed view and `x` is a column slice of the
  fused in-projection.

**New env var** `ATOM_SSM_STATE_CACHE_LOG_INTERVAL` (default 100). The
counters are cumulative, so the last line is the whole-run figure — but only
for runs that are a multiple of the interval. A 1319-request eval at the
default last reports at 1300 and drops the tail; set 1 when the exact rate for
a fixed-size eval matters.

## Reproducing

```bash
# serve
python -m atom.entrypoints.openai_server --model <kimi-k3> -tp 8 \
  --block-size 64 --enable_prefix_caching --kv_cache_dtype fp8 \
  --max-num-seqs 64 --max-model-len 16384 --max-num-batched-tokens 16384 \
  --gpu-memory-utilization 0.93 --trust-remote-code --server-port 8014

# confirm the cache is up (absence of this line is the real negative)
grep "SSM state cache: .* checkpoint slots" server.log

# treatment: shared 2048-token prefix
python -m atom.benchmarks.benchmark_serving --backend openai --port 8014 \
  --endpoint /v1/completions --model <m> --tokenizer <m> --trust-remote-code \
  --dataset-name random --random-prefix-len 2048 --random-input-len 256 \
  --random-output-len 128 --random-range-ratio 1.0 \
  --num-prompts 128 --max-concurrency 16 --request-rate inf \
  --num-warmups 0 --seed 0

# control: same token count, no sharing
#   --random-prefix-len 0 --random-input-len 2304
```

`--num-warmups 0` is deliberate: the built-in warmup sends the first prompt of
the set, which would seed the shared prefix and let the cold pass hit too,
collapsing the distinction between the passes.

## Limitations

- **Only Kimi K3 / KDA was measured.** See the coverage gap above.
- **The `--no-enable_prefix_caching` server half was not run.** What is
  measured is cache-on with vs without a shared prefix. A structurally
  cache-absent server would additionally rule out the shared prefix helping
  through some path other than the state cache.
- **The single-server cold→warm delta (1503→1342 ms) should not be quoted.**
  The cold pass also pays autotune and graph capture, which happens with the
  cache off too. The control pair is what carries the result.
- **`evict` ran level with `writes` on the cold pass** (129 vs 117). It cost
  nothing here — warm still hit 128/128 — but a longer reuse distance would
  need a higher ratio. The knob exists for this.
- **GSM8K cannot measure this feature's benefit.** lm_eval re-samples its
  5-shot exemplars per question (1203 distinct first exemplars across 1319
  docs), capping the measured hit rate near 1% however well the cache works.
  It is an accuracy guard here, not a performance one.


zejunchen-zejun:
Hi @ganyi1996ppo 

Could you add `ATOM_SSM_STATE_CACHE_LOG_INTERVAL` and `--ssm_state_cache_ratio` into the K3 recipe?

PR #1770, opened 2026-08-03
Add Periodic Engine Status Log

yitingw1:
## Motivation

ATOM server mode has no periodic scheduler-state log (running/waiting reqs, KV usage, prefix-cache hit rate, throughput) — unlike the vllm-atom plugin, which reuses vLLM's `LoggingStatLogger`. This PR adds that line (vLLM format, every 10s, default on, opt-out via `--no-enable-log-stats`):

```bash
Engine 000: Avg prompt throughput: 97.4 tokens/s, Avg generation throughput: 71.0 tokens/s, Running: 8 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 4.4%
```

Along the way it merges the three separate stats trackers into one `EngineStats` (`Scheduler.engine_stats`) and extends the throughput line to the P/D-disaggregated schedulers.

## Design

`SpecStats` / `CacheStats` / `ThroughputStats` are merged into a single flat `EngineStats`. Merging is purely structural — each **section keeps its own enable gate and its own pace**; a disabled section's `update_*` is a no-op, so call sites drop their guards.

| Section | Gate | Pace | Log line |
|---------|------|------|----------|
| spec (MTP) | `use_spec` | every `log_interval * mtp_k` draft tokens | `[MTP Stats]` |
| cache | `enable_prefix_caching` | every 100 requests | `[Cache Stats]` |
| throughput | `enable_log_stats` | every 10s wall-clock | `Engine NNN: ...` |

## Test

```bash
python -m pytest tests/test_scheduler.py tests/test_state_checkpoint.py \
                 tests/test_prefill_scheduler.py tests/test_disagg_modes.py -q
# 214 passed
```
Verified curl `mtp_status`, `cache_status`, `metrics` still efficient.

## Performance

Since `enable_log_stats` is **on by default**, its overhead lands on every deployment, so a real A/B comparison was run:
8k1k (in=8192, out=1024, range-ratio=0.8), c64 (128 prompts / concurrency 64), default-on vs `--no-enable-log-stats`. Each config was run **twice**: pass 1 `on→off`, pass 2 **reversed** `off→on`, to cancel out the ordering bias where whichever group runs first absorbs aiter JIT compilation and GPU cold-start cost. All 12 runs completed 128/128 requests successfully.

**Total Token throughput (tok/s)**

| Model | Config | Pass 1 | Reversed pass | Mean | Intra-group spread |
|------|------|--------|--------|------|----------|
| DeepSeek-R1 | ON | 10560.11 | 10693.50 | 10626.81 | 1.26% |
| | OFF | 10615.09 | 10664.13 | 10639.61 | 0.46% |
| GLM-5.2 | ON | 4553.65 | 4635.50 | 4594.58 | 1.80% |
| | OFF | 4640.15 | 4655.53 | 4647.84 | 0.33% |
| DeepSeek-V4-Flash-Base | ON | 17718.59 | 18059.92 | 17889.26 | 1.93% |
| | OFF | 17928.93 | 18032.47 | 17980.70 | 0.58% |

*Intra-group spread* = `(max - min) / min` over the two runs of the **same** config —
pure measurement jitter, and the bar the ON-vs-OFF deltas below have to clear.

**ON vs OFF delta**

| Model | Mean ratio | Best-run ratio | TPOT mean ratio |
|------|--------|------------|-------------|
| DeepSeek-R1 | -0.12% | **+0.28%** (ON faster) | +0.03% |
| GLM-5.2 | -1.15% | **-0.43%** | +1.33% |
| DeepSeek-V4-Flash-Base | -0.51% | **+0.15%** (ON faster) | +0.15% |

## Known Limitations

1. **P/D throughput not yet live-verified** — unit-tested only; the aggregated path is live-verified.


valarLip:
Reviewed at `a858fa994`. The unification is a real improvement and the refactor itself is clean — a normalized diff of `EngineStats` against the deleted `SpecStats`/`CacheStats` shows only additions, `__slots__` is complete, the `_cache_log_line` argument order matches both call sites, and the added per-step cost is 0.07% of the host budget (measured, see #9 below), so this is not a throughput regression. Findings below are ordered by consequence.

---

## 1. `PrefillEngineCore` swaps the scheduler but never rewires the utility handler

`engine_core.py:181` constructs `EngineUtilityHandler` with the base `Scheduler`. `PrefillEngineCore` (class at `:797`) then replaces `self.scheduler` at `:839`:

```python
# Replace the base Scheduler created by EngineCore.__init__ with
# PrefillScheduler, which has no BlockManager and only schedules ...
self.scheduler = PrefillScheduler(config, disagg_cu_shm_name=...)
```

`DecodeEngineCore` explicitly fixes this at `:1111` (`self.utility_handler.scheduler = self.scheduler`, "wire the real one in for MTP stats"). `PrefillEngineCore` does not — `:1111` is the only rewire in the file.

So on a P/D prefill process, `push_metrics()` fires every 5s and publishes `engine_utility.py:347`'s `self.scheduler.engine_stats` from the **orphaned base Scheduler**: mtp/cache all zeros, `kv_blocks_*` plausible-looking, while the live `PrefillScheduler.engine_stats` is the object actually being fed at `scheduler.py:2713`. The new module docstring at `engine_stats.py:8-9` asserts the opposite — "all three schedulers own one, and engine_utility reads it".

Worth flagging that the naive fix then crashes: `PrefillScheduler` (`:2600`) is not a `Scheduler` subclass, has no `get_request_counts`, and its `block_manager` is None, so `collect_metrics` would raise `AttributeError` inside the busy loop.

## 2. The window tick is hand-placed at six return points instead of the one choke point

`engine_core.py:342`'s `result = self.scheduler.schedule()` in `_process_engine_step_inner` is the single place every scheduler goes through, and `total_tokens_num_prefill` is already a field on the returned `ScheduledBatch`. Instead the tick lives at `scheduler.py:957`, `:1199`, `:1368`, `:2925`, `:2959`, `:2976`, plus `:2258` and the two heartbeats.

That makes the 10s cadence a whole-program invariant maintained by hand inside a ~700-line `schedule()`. A seventh early return — the `if not self.running and not self.waiting` shape at `:958` is exactly the kind that gets added — silently stalls the status line for as long as that branch fires, and nothing fails. The log just goes quiet, which is indistinguishable from an idle engine.

Findings 3 and 4 are this failure mode already realized twice, before merge.

## 3. The PP head busy loop has the same idle shape but no heartbeat

`pp_engine_core.py:74`:

```python
if self._in_flight or not self.scheduler.is_finished(): self._pp_head_step()
elif self.has_pending_kv_work(): self._advance_idle_kv_transfer()
```

byte-for-byte the shape that got `heartbeat_throughput(now)` at `engine_core.py:316` and `:706` — and `grep -rn heartbeat_throughput atom/` returns exactly those two call sites, nothing in `pp_engine_core.py`.

With `pipeline_parallel_size > 1`, server idle 60s then a 30000-token burst: elapsed ~70s instead of ~10s, so the line prints ~430 tok/s for a burst that ran at ~3000. This is precisely what `test_idle_does_not_leave_the_window_start_stale` asserts against; it passes because it drives `EngineStats` directly and never touches the PP loop. Note `_head_busy_loop` has no `now` variable at all, so the fix needs a `time.monotonic()` too. (`_downstream_busy_loop` does not schedule and does not need it.)

## 4. Both busy loops `continue` past the heartbeat on the RL-offload path

`engine_core.py:311` `if self._is_idle_rl_weights_offloaded(): continue`, and the DP twin at `:701` `if global_offloaded: time.sleep(0.01); continue`, both sit **above** the `else:` arm that holds `heartbeat_throughput`. `_throughput_last_log_time` has exactly three writers and all three are reachable only via `schedule()` or `heartbeat_throughput()`, both bypassed here.

An RL trainer that offloads for 20 minutes resumes and the first `schedule()` records 8000 prompt tokens with elapsed ~1200s: `Avg prompt throughput: 6.7 tokens/s`.

## 5. The new counter is overridden; the old one feeding `/metrics` is not

`Scheduler._status_counts` (new, `:608`) and `Scheduler.get_request_counts` (existing, `:2527`) have identical bodies — `return len(self.running), len(self.waiting)`. `DecodeScheduler` (`:2770`) overrides only `_status_counts` (`:2788`), while `engine_utility.collect_metrics:343` calls `get_request_counts`.

On a P/D decode engine holding 200 requests in `prefill_waiting`, the new log line correctly says `Running: 200` while `/metrics` and the gh-pages dashboard simultaneously report `requests_running: 0, requests_waiting: 0`. The override's own docstring names this exact behaviour a bug ("reports an idle engine while it is at full load") and leaves it in place one method away. `DecodeScheduler` already overrides `is_finished`, `has_requests` and `get_num_unfinished_requests` for these same two queues — the queue inventory is now encoded four times and can drift.

## 6. `num_new_generation_tokens` counts the forward output, not what the client receives

`scheduler.py:2006` does `num_new_token = len(token_ids); num_new_generation_tokens += num_new_token`. But `:2149`'s `new_tokens = new_tokens[:keep]` drops accepted-draft tokens past EOS/stop (the code's own comment says these otherwise leak into the response), and `:2075` prepends one with `new_tokens = [injected_t0] + list(new_tokens)`. `new_tokens` is what reaches `RequestOutput` at `:2181`, and `:2214` computes `total_generation_tokens` from the post-truncation `num_tokens`.

On MTP-3 with short answers (GSM8K-shaped), the 10s line reports up to `mtp_k` extra tokens per finishing request — exceeding both what the benchmark client receives and the `/metrics` counter produced by the same `postprocess` call.

## 7. `prefix_cache_hit_rate`: structurally 0.0% on decode, and lifetime-cumulative on a windowed line

`engine_stats.py:391`, two separate issues:

**(a)** `update_cache` is reachable only from `Scheduler._schedule_prefill_seq` (`:1672`); `DecodeScheduler.schedule()` is a decode-only override that never gets there. But `cache_enabled = config.enable_prefix_caching` (default True), so the n/a branch is never taken and a P/D decode engine logs `Prefix cache hit rate: 0.0%` every 10s forever while prefix caching works fine on the prefill side. Same for the first N seconds of the aggregated `Scheduler` (`total_full_tokens == 0` → 0.0%, not n/a).

**(b)** `cache_hit_rate` is `total_cached / total_full` since process start, while throughput / Running / Waiting / KV on the same line are windowed. vLLM's `LoggingStatLogger` uses a sliding window precisely so the number tracks the current workload. An 8-hour run whose first hour hit 80% and whose next seven reuse nothing still prints ~71% and barely moves.

## 8. Two mutations that stay green

Verified by mutation, not by reading:

- Monkeypatch `Scheduler._record_throughput` to drop its `num_prompt_tokens` argument, simulating loss of the `scheduler.py:1199` kwarg → **4226 passed, zero new failures**. "Avg prompt throughput", this PR's headline number on the non-P/D path, can become permanently 0.0 with CI green. The only prompt-token assertion in the PR (`tests/test_prefill_scheduler.py:211`) is on `PrefillScheduler` — the path the PR description itself lists as not yet live-verified.
- Monkeypatch `_record_throughput` to a no-op → `test_decode_schedule_records_throughput_when_log_stats_on` **still passes**; it asserts only `schedule() is None`, `batch is not None`, `seq.id in seqs`. All three `DecodeScheduler` tick sites (`:2925`, `:2959`, `:2976`) can be deleted and the test named after them stays green — which matters because `DecodeScheduler.schedule()` deliberately does not call `super().schedule()`.

Related: `tests/conftest.py:60` defaults `MockConfig.enable_log_stats` to False while production defaults it to True, so `_record_throughput` returns at its first line across ~150 legacy scheduler/PP/state-checkpoint tests and only the handful of new tests passing `enable_log_stats=True` reach `maybe_log_throughput`. A future subclass whose `_status_counts` override touches an attribute assigned after `super().__init__()` — `DecodeScheduler` already has that shape, `prefill_done` at `:2814` vs `super().__init__` at `:2808` — would be green across the whole suite and fail only on a real server. Flipping the default to True and letting individual tests opt out would match the production shape.

## 9. `_kv_usage` is not a scan, and the cheap gate is on the wrong path

`_record_throughput`'s docstring (`scheduler.py:624`) calls `_kv_usage()` a "scan", while `_kv_usage`'s own docstring nine lines above says "cheap (no traversal of seq tables)". `BlockPool.num_used` is `len(self._used)` (`block_pool.py:97`) — two attribute reads and a divide.

This matters because the comment is load-bearing misinformation: the whole added per-step path (`_status_counts` + `_kv_usage` + the extra `time.monotonic()`) measures **0.209 μs = 0.07%** of the ~0.3 ms host budget on this box. Left as written it will send the next person optimizing the fwd→fwd host path hunting an O(seqs) scan that does not exist.

Separately, `window_expired`'s own docstring says it exists "so the busy loop's idle heartbeat can skip the rest of `_record_throughput`" — the cheap gate got applied to the rare path and skipped on the 1 kHz one. Adding it to `_record_throughput` cuts the added path to 0.093 μs and makes the two agree.

## 10. Smaller items

- **`tests/test_scheduler.py:152` is flaky by construction.** `throughput_log_interval_s=0.05` paired with a real `time.sleep(0.06)` and then `7700 tokens / rate > 10_000` tolerates only ~0.77 s of total wall time — a 20% margin, the tightest real-clock margin in the suite. A GC pause or a contended runner produces a red build whose message ("window still spans the idle stretch") points at a nonexistent regression. It is also the only `time.sleep` in either new test file; the two sibling tests use the looser `time.monotonic() - _throughput_last_log_time < 1.0` form.
- **`--no-enable-log-stats` silences only the throughput section** (`arg_utils.py:431`). The flag reaches only `EngineStats.throughput_enabled` (`engine_stats.py:89`); `[MTP Stats]`, `[Cache Stats]` and `[Cache Pool]` are gated by `use_spec` / `enable_prefix_caching` and keep firing. An operator hitting log flooding reaches for vLLM's `--disable-log-stats` semantics and still gets the three-line burst every 100 requests.
- **The 10s cadence has no knob, and its guard is stripped under `-O`.** `throughput_log_interval_s=10.0` is a keyword default (`:85`) never plumbed from Config or the CLI, while comparable observability knobs in the same area are env-sourced (`config.py:1533`, `:1568`). The bare `assert` at `:128` is the only thing between `interval <= 0` and `self.num_prompt_tokens / elapsed` at `:387`; under `-O` the engine dies with `ZeroDivisionError` inside the scheduler loop. `if elapsed <= 0: return` plus a real `ValueError` in `__init__` covers it.
- **Offline mode gets the line too, despite the "(Server Mode)" title.** `enable_log_stats` lives on Config with default True and `EngineArgs._get_engine_kwargs` is a blanket `fields()` passthrough, so `python -m atom.examples.simple_inference` and every embedded `LLM(...)` — including `start_simple_inference.sh` and `run_gsm8k_eval.sh` runs — now emit it every 10s.
- **Docs were not swept.** Stale `SpecStats`: `docs/eagle3_speculative_decoding.md:141` (class and file both wrong now), `docs/scheduling_kv_cache_guide.md:597`, and `.claude/commands/debug-guide.md:99` and `:144` — the last two are on the path CLAUDE.md mandates for server/GPU problems, so someone chasing a low MTP acceptance rate greps `SpecStats` and gets zero hits. Missing: no row for `atom/model_engine/engine_stats.py` in the architecture guide's source-file table; no `enable_log_stats` row in `configuration_guide.md:44` or `:357`, where `enable_prefix_caching` has both; no `--enable-log-stats` in the `serving_benchmarking_guide.md` EngineArgs table.
- **`ThroughputStats` never existed.** Both the module and class docstrings say `EngineStats` "replaced the former SpecStats / CacheStats / ThroughputStats". `git log -S ThroughputStats --all -- atom/` returns only this PR's own three commits (`2f63b7c43`, `0897b2b04`, `4b2c306a1`) and a repo-wide grep returns nothing — only `SpecStats` and `CacheStats` were ever merged; the throughput section is new code. The same false claim is pasted in two places.

valarLip (approved):

PR #1789, opened 2026-08-04
[vLLM-ATOM] Support DSpark speculative decoding for Kimi-K3

whx-sjtu:
## Summary

Enables DSpark block-drafting speculative decoding for Kimi-K3 under the ATOM vLLM plugin backend, with the standalone [Inferact/Kimi-K3-DSpark](https://huggingface.co/Inferact/Kimi-K3-DSpark) draft checkpoint. Validated at TP8 on MI355 with full CUDA graphs and asynchronous scheduling.

The draft is a standalone MLA model rather than an in-checkpoint MTP head, which is what most of this diff is about:

- **Causal block drafting.** vLLM's DSpark speculator drafts non-causally, so every block position attends to the whole block. ATOM's `mla_decode_fwd` masks tail-aligned-causally whatever it is asked for, so that upper triangle would have to be merged in by log-sum-exp. `spec_decode_patch.py` drafts causally instead for the `K3DSparkModel` architecture only, matching ATOM's native DSpark — one drafting semantics across both engines and no merge to build. This is a scope decision, not a kernel limitation: the persistent decode kernel does return a usable LSE on gfx950, which is what #1746 just put DCP's cross-rank merge on, so non-causal is a buildable follow-up. No other model's drafting is touched.
- **Standalone draft plumbing.** `SpeculativeConfig` rewrites every non-Qwen3 DSpark draft into a DeepSeek-V4 one, which makes the Kimi-K3 draft config unresolvable; the patch keeps its own architecture. `model_wrapper.py` generalizes the eagle3-specific standalone-draft handling to cover DSpark, and stops exposing `get_mtp_target_hidden_states` on non-MTP models, which MRV2 probes with a bare `hasattr`.
- **Separate bf16 draft KV pool.** The draft gets its own KV cache group, pinned to bf16 regardless of the target's `--kv-cache-dtype fp8`.
- **Persistent MLA work-descriptor sizing.** The descriptors were sized for a 1-token decode but filled at the real decode width. They are now sized for the worst case over every width `split_decodes_and_prefills` can route to `_build_decode`, i.e. `1..reorder_batch_threshold`. The sizes are not monotonic in query length, so this takes a per-slot maximum rather than the value at the largest width. Costs no extra memory: the descriptor shapes saturate well below the threshold.
- **Padded decode slots.** A full decode graph replays at its captured request count, so vLLM pads the batch with slots carrying no query rows and no sequence. Those empty slots left the persistent kernel waiting on partials nothing produces; they now get one page of KV and their share of query rows.
- **DFlash slot-mapping units.** `BlockTables` stores block ids in kernel blocks, but the DFlash speculator converts them back to slots with the manager's block size. Those agree for vLLM's own backends; ATOM's MLA pages by the token, so the draft wrote past its pool. Wrapped for the `set_attn` call only, and only when the two sizes actually differ.

## Validation

Full 1319-example GSM8K, 5-shot, over the served endpoint at 64 concurrency, TP8, fp8 KV, `FULL_AND_PIECEWISE` CUDA Graph, async scheduling on:

| | exact_match |
|---|---|
| flexible-extract | 0.9545 ± 0.0057 |
| strict-match | 0.9538 ± 0.0058 |

The target alone scores 0.9553 ± 0.0057 under the same harness and concurrency (see the [Kimi-K3 recipe](recipes/atom_vllm/Kimi-K3.md)), so drafting is accuracy neutral as expected for verify-based speculative decoding.

Acceptance over that run: mean accepted length **3.777** tokens per forward, acceptance rate **39.7%**, over 36,119 verify forwards. Per-position acceptance rate is `0.923 / 0.774 / 0.602 / 0.442 / 0.034 / 0.002 / 0.000` — causal drafting means the last three of the seven block positions contribute about 1% of accepted tokens between them, which is the cost of not needing the LSE merge.

Also verified that prompts of every length from 1 to 20 tokens serve correctly. Prompts of 10 to 15 tokens are the interesting ones: `parallel_drafting` widens `reorder_batch_threshold` to `1 + 2 * num_speculative_tokens = 15`, so their prefill is classified as a decode and reaches `_build_decode`. An accuracy sweep never covers that window.

Rebased onto `main` over #1746 (DCP persistent PA), which touched the same two attention files. The work-descriptor sizing now uses that PR's `persistent_num_heads` inside this PR's per-query-length worst-case loop, and its DCP-on-gfx942 persistent opt-out runs before this PR's capacity assertion. Every number above was measured against the rebased tree, and the length sweep and smoke were re-run on it too. Acceptance held at 3.77-3.79 across four separate runs spanning the pre- and post-#1746 trees.

## Test plan

- [x] Full GSM8K 1319 examples over `vllm serve`, 64 concurrency, fresh server
- [x] Prompt length sweep 1..20 tokens against the served endpoint
- [x] Deterministic smoke from the recipe (`Question: What is 17 + 25? Answer:` returns ` 42`)
- [x] Torch profile at 8k input / 100 output / 16 concurrency with stacks, both models visible under their own MLA kernels (`aiter::mla_a8w8_*` target, `aiter::mla_a16w16_*` draft)
- [ ] Reviewer: confirm no regression for DeepSeek-V4 MTP and MiniMax-M3 EAGLE3, which share the patched speculative-decoding paths

### DIFF
--- a/atom/plugin/vllm/attention/layer_mla.py
+++ b/atom/plugin/vllm/attention/layer_mla.py
@@ -946,6 +946,12 @@ class AttentionForVllmMLA(MLAAttention, AttentionLayerBase):
         q = q[:num_actual_toks, ...]
         k_c_normed = k_c_normed[:num_actual_toks, ...]
         k_pe = k_pe[:num_actual_toks, ...]
+        # Model runner V2 sizes slot_mapping to the CUDA-graph token width and
+        # marks the tail with PAD_SLOT_ID, but leaves num_actual_tokens unpadded
+        # on the piecewise path; V1 never pads it. The cache kernels launch one
+        # block per slot and index q/kv by that block, so slot_mapping must not
+        # outrun the tensors sliced above. Same trim layer_mha already applies.
+        slot_mapping = attn_metadata.slot_mapping[:num_actual_toks]
 
         decode_q = q[:num_decode_tokens]
         prefill_q = q[num_decode_tokens:]
@@ -973,7 +979,7 @@ class AttentionForVllmMLA(MLAAttention, AttentionLayerBase):
                     k_c_normed,
                     self.rotary_emb_cos_sin_cache,
                     self.rotary_emb.is_neox_style,
-                    attn_metadata.slot_mapping,
+                    slot_mapping,
                     kv_cache,
                     self.kv_cache_dtype,
                     self._k_scale,
@@ -985,7 +991,7 @@ class AttentionForVllmMLA(MLAAttention, AttentionLayerBase):
                         k_c_normed,
                         k_pe.squeeze(1),
                         kv_cache,
-                        attn_metadata.slot_mapping.flatten(),
+                        slot_mapping.flatten(),
                         kv_cache_dtype=self.kv_cache_dtype,
                         scale=self._k_scale,
                     )
@@ -1066,7 +1072,7 @@ class AttentionForVllmMLA(MLAAttention, AttentionLayerBase):
                         self.kv_lora_rank + self.qk_rope_head_dim,
                     ),
                     decode_q,
-                    attn_metadata.slot_mapping,
+                    slot_mapping,
                     self._k_scale,
                     self._q_scale,
                     positions,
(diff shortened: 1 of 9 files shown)


whx-sjtu:
Superseded by #1899, which brings this to the vLLM 0.26.1 line (targets `vllm0.26.1`, the #1839 upgrade branch) with the DSpark optimizations from #1876 already in place. Closing in favour of that one.

PR #1752, opened 2026-07-30
[k3] enable dual stream for shared expert and some fusions

gbyu-amd:
## Motivation

Depends on: https://github.com/ROCm/aiter/pull/4467 and https://github.com/ROCm/aiter/pull/4435

This pr:
- Added dual stream for shared/routed expert overlapping
- Added ptpc fp8 quant for attn and shared expert gemms, fuse the introduced per-token quant accordingly

Accuracy:
```bash
local-completions ({'model': '/workspace/shared/data/amd_int/models/Kimi-K3', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 64, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 5, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9568|±  |0.0056|
|     |       |strict-match    |     5|exact_match|↑  |0.9568|±  |0.0056|
```

Dual stream overlapping:
image

E2E perf:
**Main**:
| Input Tokens | Output Tokens | Max Concurrency | Num Prompts | Median TTFT (ms) | Median TPOT (ms) | Output Token Throughput | Total Token Throughput |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 8192 | 1024 | 4  | 20  | 557.01 | 26.60 | 142.59 | 1270.64 |
| 8192 | 1024 | 8  | 40  | 548.81 | 29.98 | 248.98 | 2238.21 |
| 8192 | 1024 | 16 | 80  | 565.70 | 35.91 | 417.48 | 3713.71 |
| 8192 | 1024 | 32 | 160 | 596.72 | 47.92 | 624.52 | 5636.14 |
| 8192 | 1024 | 64 | 320 | 799.17 | 72.08 | 839.41 | 7506.28 |

**This pr**:
| Input Tokens | Output Tokens | Max Concurrency | Num Prompts | Median TTFT (ms) | Median TPOT (ms) | Output Token Throughput | Total Token Throughput |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 8192 | 1024 | 4  | 20  | 512.26 | 25.66 | 147.14 | 1311.19 |
[...]

PR #1790, opened 2026-08-04
fix(plugin): gate dual-stream MoE by runtime graph mode

wuhuikx:
## Problem

In vLLM plugin mode, CUDA/HIP graph capture is owned by vLLM and the concrete runtime mode is selected per forward (`FULL`, `PIECEWISE`, or `NONE`). ATOM's MoE dispatcher instead checked its static compilation config.

For `FULL_AND_PIECEWISE`, the plugin maps vLLM's compile mode to ATOM level 3, which initializes the ATOM-side static graph mode as `PIECEWISE`. As a result, decode forwards that vLLM actually dispatches as `FULL` are incorrectly forced onto the single-stream MoE path. This was observed while validating Kimi-K3 dual-stream shared/routed expert overlap in ROCm/ATOM#1752.

## Solution

- Resolve the concrete graph mode from the active frontend's forward context.
- Normalize frontend enums by name instead of relying on matching numeric values.
- Fall back to ATOM's native forward context when vLLM context is unavailable.
- Disable dual-stream MoE only for an actual `PIECEWISE` forward; preserve it for eager/`NONE` and whole-model `FULL` capture.
- Add focused tests for enum translation, frontend fallback, runtime dispatch, token threshold, TBO, and the existing `_use_dual_stream` gate.

## Validation

- [x] `python -m pytest -q tests/test_module_dispatch_ops.py tests/plugin/test_plugin_config_translation.py tests/plugin/test_vllm_kimi_k3.py` — 22 passed
- [x] Kimi-K3 TP8 with vLLM `FULL_DECODE_ONLY` and `FULL_AND_PIECEWISE`
- [x] Profiler confirmed decode dual-stream graph execution on all TP ranks; with the native-matched online quant configuration, measured stream overlap was 99.1%–100%

### DIFF
--- a/atom/model_ops/module_dispatch_ops.py
+++ b/atom/model_ops/module_dispatch_ops.py
@@ -25,9 +25,10 @@ Currently registered:
 
 import torch
 
-from atom.config import get_current_atom_config
+from atom.config import CUDAGraphMode, get_current_atom_config
 from atom.utils import envs
 from atom.utils.custom_register import direct_register_custom_op
+from atom.utils.forward_context import get_current_cudagraph_runtime_mode
 
 # ---------------------------------------------------------------------------
 # Dual-stream MoE dispatch (V2 / V3.2 / V4)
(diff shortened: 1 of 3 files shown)


wuhuikx:
Closing to resubmit the change from @XiaobingSuper.

PR #1793, opened 2026-08-04
fix(deepseek_v4): gfx942 dtype-dialect and MXFP4 swizzle fixes for DeepSeek-V4-Flash

AMD-melliott:
## Motivation

DeepSeek-V4-Flash-0731 (`DeepseekV4ForCausalLM`) uses FP8 for its dense/attention
layers and FP4/MXFP4 for its routed MoE experts. Bringing it up on gfx942 (MI300X)
surfaced three issues, all with the same root cause: gfx942's FP8/FP4 hardware dialects
differ from gfx950/gfx1250, and a few code paths only handled the latter.

## Technical Details

1. **`wo_a` FP8→BF16 dequant silently skipped on gfx942**
 `DeepseekV4Attention.process_weights_after_loading()` dequantizes the
 attention output projection (`wo_a`) from FP8 to BF16 after weights load. 
 The dtype check matched `torch.float8_e4m3fn` (the OCP dialect used on gfx950). 
 On gfx942, AITER's FP8 tensors are `torch.float8_e4m3fnuz` (same bit
 layout, different exponent bias), so the check silently did nothing,
 leaving `wo_a` in FP8 at forward time and crashing CUDA graph warmup with a
 dtype mismatch.

 Already fixed upstream on `main`. No regression test existed, so this PR
 adds one (`tests/test_deepseek_v4_wo_a_dequant.py`) covering both dialects
 plus idempotency (a second call must be a no-op).

2. **Same dialect bug in dummy-weight init**
 `atom/model_loader/loader.py`'s `initialize_dummy_weights()` only
 recognized `(float8_e4m3fn, float8_e5m2)` as FP8 dtypes. On gfx942, real
 FP8 params are `float8_e4m3fnuz`, so they fell into the FP4-packed
 byte-fill branch instead, producing dummy weights with the wrong
 magnitude. This only affects the `--load_dummy` profiling path, no real
 checkpoint is affected. Fixed by adding `float8_e4m3fnuz` to the check.

3. **MXFP4 MoE scale-shuffle crash on gfx942**
 `atom/model_ops/fused_moe_triton.py`'s `_swizzle_mxfp4()` called AITER's
 `shuffle_scale_moe()` unconditionally for MXFP4 routed-expert weights.
 That function only supports the preshuffled FP4 tensor-core scale layout
 used on gfx950/gfx1250; on any other arch it raises `UnboundLocalError`
 (its sibling function, `_shuffle_scale_tile`, has the same arch guard but
 ends in a clean `ValueError`, this one is just missing it). gfx942 has no
 such layout to shuffle for, and the Triton MXFP4 GEMM kernels already have
 a working fallback for unshuffled scales. Fixed by gating both call sites
 on `get_arch() in {"gfx950", "gfx1250"}` and routing gfx942 through the
 existing fallback.

4. **Test hygiene**
 The new test for (3) originally used `pytest.importorskip("aiter")` alone
 to skip on the non-GPU CI gate. That's not reliable: `tests/test_pp.py`
 and `tests/test_pd_pp.py` install a fake `sys.modules["aiter"]` stub with
 no teardown, which can leak into later test files in the same pytest run.
 Hardened the check to verify the specific symbol the test needs
 (`ActivationType`), so a leaked stub causes a clean skip instead of a failure.

## Testing Notes (gfx94x)

- Set `ATOM_USE_TRITON_MOE=1` explicitly when launching. `Mxfp4MoEMethod`
selects the Triton MoE path automatically on gfx94x, but doesn't set this
env flag itself, and `_swizzle_mxfp4` checks the flag directly. Leave it at
`=1` (or unset, since it's the default). Explicitly setting it to `0`
crashes on the first forward pass, since gfx942 has no native aiter FP4 MoE
kernel support to fall back to.
- Set `TENSILE_SOLUTION_SELECTION_METHOD=2` when testing under sustained
concurrent load. hipBLASLt/Tensile can pick a GlobalSplitU (split-K) GEMM
kernel on gfx942 whose completion-flag polling loop livelocks under load,
hanging the engine. This appears to be an upstream hipBLASLt issue, unrelated
to the fixes in this PR. This env var switches hipBLASLt to its Stream-K
kernel selection strategy, which avoids the lock scenario.

## Test Plan

- `black .` / `ruff check .` clean (ruff scoped to this PR).
- CI's exact non-GPU unit test command
(`python -m pytest tests/ --ignore=tests/plugin --ignore=tests/entrypoints/test_openai_server.py`):
- CI's exact non-GPU unit test command
(`python -m pytest tests/ --ignore=tests/plugin --ignore=tests/entrypoints/test_openai_server.py`):
**836 passed, 32 skipped, 0 failed.**
- End-to-end serving on 8x MI300X with a real checkpoint (not `--load_dummy`): 
weight load -> MoE post-load hooks -> `torch.compile` -> CUDA graph capture 
all completed cleanly, followed by completion and chat requests returning coherent text. 

## Test Result

```bash
python -m atom.entrypoints.openai_server --model deepseek-ai/DeepSeek-V4-Flash-0731
  -tp 8 --kv_cache_dtype fp8 --gpu-memory-utilization 0.9
```

falls back to bf16 KV cache on gfx942 (pre-existing, correct behavio)r. fp8 2-buffer KV cache 
is gfx950/gfx1250-only.

`gsm8k` via `lm_eval` (`num_fewshot=3`, `num_concurrent=65` - matching CI's
own command shape; no CI reference exists for this exact config, so these
parameters are stated explicitly rather than implying direct comparability):

| Tasks | Version | Filter | n-shot | Metric | Value | Stderr |
|-------|--------:|--------|-------:|--------|------:|-------:|
| gsm8k | 3 | flexible-extract | 3 | exact_match | 0.8681 | ±0.0093 |
| | | strict-match | 3 | exact_match | 0.8681 | ±0.0093 |

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


atom/model_ops/fused_moe_triton.py
```diff
@@ -73,14 +84,16 @@ def _swizzle_mxfp4(
     w2_triton_layout = w2.transpose(-2, -1)
     w2_scale_triton_layout = w2_scale.transpose(-2, -1)
 
-    if N_1 % 32 == 0 and K_1 % (32 * 8) == 0:
+    can_swizzle = get_arch() in _MXFP4_SWIZZLE_CAPABLE_ARCHS
+
+    if can_swizzle and N_1 % 32 == 0 and K_1 % (32 * 8) == 0:
         w1_scale_triton_layout, w1_swizzle_layout = shuffle_scale_moe(
```
valarLip, line 77: thanks, however... this "can_swizzle = get_arch() in _MXFP4_SWIZZLE_CAPABLE_ARCHS" should land in shuffle_scale_moe, intead of framework

yitingw1: I've removed check for shuffle_scale_moe in ATOM and added it in aiter https://github.com/ROCm/aiter/pull/4588

PR #1805, opened 2026-08-05
[kimi-k3] DSpark speculative decoding on vLLM 0.26

whx-sjtu:
Ports Kimi-K3 DSpark speculative decoding (#1789, written against vLLM 0.25) onto
this branch's vLLM 0.26. Re-derived on 0.26's surfaces rather than carried over
verbatim: three of the original patches exist only to work around gaps 0.26
closed, so they are gone rather than ported.

## What changed relative to #1789

**Causality comes from config, not a monkeypatch.** 0.26 resolves a draft's
causality through `dflash_has_any_non_causal()`, and `dflash_config.causal`
overrides every layer at once. Injecting that at `SpeculativeConfig.__post_init__`
reaches all three consumers — backend selection, the draft's `use_non_causal`,
and the per-group causality the speculator hands its metadata builders — so the
0.25 patch that reached into speculator attributes is dropped.

**`pin_kv_cache_dtype()` is deleted.** 0.26 added
`speculative_config.kv_cache_dtype`, which `load_dspark_model()` folds into the
cache config the draft is built under. Since ATOM's MLA layer already reads its
dtype from `cache_config.cache_dtype`, setting that field is enough and pinning
the dtype after construction is redundant. `align_page_size_to_engine_dtype`
stays: the draft still caches bf16 against an fp8 target, so its block size
still has to scale so the two page sizes match.

**No `K3DSparkConfig` shim.** 0.26 ships its own, so the registry points
straight at ATOM's model.

## Validation

Full GSM8K 5-shot, TP8, fp8 KV, block size 128, 64 concurrent,
`FULL_AND_PIECEWISE` (FULL 48/48 and PIECEWISE 51/51 captured), freshly started
server:

| | this branch (0.26) | #1789 (0.25) |
|---|---|---|
| gsm8k flexible / strict | 0.9515 / 0.9507 | 0.9530 / 0.9530 |
| acceptance rate | 39.4% | 39.7% |
| mean accepted length | 3.758 | 3.776 |

Accuracy differs by 0.0015 against a stderr of 0.0059, and acceptance is within
0.3pp, over 36,145 drafts.

Per-position acceptance is 91.96 / 76.74 / 59.80 / 43.86 / 3.28 / 0.14 / 0.00%.
As a conditional that is 92 / 83 / 78 / 73%, then a step down to 7 / 4 / 2%. The
tail matches #1789's documented behaviour for causal drafting, and the aggregate
numbers reproduce it, so it is inherited rather than introduced here. The shape
still looks more like a block-width limit than a smooth information decay —
worth a separate look, out of scope for this port.

## Bugs found while validating

Collecting the target's auxiliary hidden states by swapping the tapped layers'
`forward` for a closure that recorded into a module-level list split the body's
forward across a Dynamo graph break, and ATOM's compile backend accepts exactly
one graph — startup died in `profile_run` with `VllmBackend can only be called
once`. They are now collected inline in the layer loop into a local list, as
every other ATOM and vLLM target that feeds a drafter does.

Two silent correctness bugs came out with it. The tap recorded the stream
entering the layer it sat on, but vLLM converts a draft's `target_layer_ids`
into residual-stream indices by adding one before handing them over, so every
captured tensor was a layer early. And it summed only the deferred FFN output,
missing the shared-expert output an MoE layer defers separately, leaving the aux
tensor short a term on every MoE layer. Neither crashes; both just feed the
draft the wrong tensor.

## Not included

- No recipe doc yet — happy to add `recipes/atom_vllm/DSpark.md` with the
  numbers above.
- Serving throughput is not measured here; only accuracy and acceptance.

### DIFF
--- a/atom/plugin/vllm/attention/layer_mla.py
+++ b/atom/plugin/vllm/attention/layer_mla.py
@@ -843,6 +843,11 @@ class AttentionForVllmMLA(MLAAttention, AttentionLayerBase):
             q_scale=self._q_scale,
             kv_scale=self._k_scale,
             return_lse=return_lse,
+            # A bidirectional draft block needs every position to see the whole
+            # block. The asm kernel picks its .co by this flag and the work
+            # descriptors were planned for it, so it has to match what the
+            # metadata builder planned.
+            causal=attn_metadata.decode.causal,
         )
         if do_fold:
             o = o.view(ori_total_s, ori_nhead, -1)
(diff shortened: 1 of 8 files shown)


ZhangLirong-amd (approved):

whx-sjtu:
Superseded by #1899, which brings this to the vLLM 0.26.1 line (targets `vllm0.26.1`, the #1839 upgrade branch) with the DSpark optimizations from #1876 already in place. Closing in favour of that one.

PR #1811, opened 2026-08-06
fix(glm-dsa): enable stable top-k for tensor parallel

yhl-amd:
## Summary

- enable AITER stable per-row top-k only for GLM-5.2 sparse attention when the tensor-parallel world size is greater than one
- carry the stable_topk policy through native ATOM, vLLM, SGLang, and RTP-LLM sparse-indexer custom ops for both prefill and decode
- keep GLM-5.2 TP=1 and non-GLM models on their existing non-stable paths; vLLM prefill continues to use its existing operator when stable top-k is disabled
- recognize the GLM-5.2 MTP draft through index_share_for_mtp_iteration after its model_type is remapped to deepseek_mtp

## Dependency

Depends on https://github.com/ROCm/aiter/pull/4415, specifically the explicit stable parameter introduced by commit 34ecbcbf4.

## Policy

| Model | TP world size | top-k behavior |
| --- | ---: | --- |
| GLM-5.2 target or MTP draft | > 1 | pass stable=True |
| GLM-5.2 target or MTP draft | 1 | existing default path |
| Other models | any | existing default path |

## Validation

- Black check on all four changed files
- Python bytecode compilation for all four changed files
- custom-op import/schema validation for native, vLLM, SGLang, and RTP-LLM
- AITER PR #4415 stable top-k test suite: deterministic, ascending, and reference-matching prefill/decode results across short/long rows and tie-heavy inputs
- live GLM-5.2 MXFP4 TP=8 service validation, including a successful 4,617-token prompt that exercises sparse-indexer prefill above the 2,048-token top-k threshold


yhl-amd:
CI note: the Offline inference smoke test fails in the Docker Login step because fork-triggered workflows receive empty DOCKER_USERNAME/DOCKER_PASSWORD secrets (the log reports `username is empty`). The job does not reach checkout, build, or inference. Black, Ruff, catalog validation, non-GPU unit tests, Read the Docs, and all three heavy-CI gate checks pass.

valarLip (approved):

PR #1820, opened 2026-08-06
[ATOM][Kimi-K3] enable KDA prefill on aiter

Liang-jianhao97:
## Motivation

Kimi-K3's KDA prefill recurrence is hard-wired to `fla.ops.kda.chunk_kda`.
aiter ships an equivalent Triton implementation.

## Technical Details

- Rename `transpose_state_layout` -> `state_v_first`, fla's spelling since
  0.5.1. This raises the floor to fla >= 0.5.1; the release image installs
  0.5.2 (`docker/atom_release.dockerfile:200`) and the recipe pins 0.5.1
  (`recipes/atom_vllm/Kimi-K3.md:17`).
  Worth noting the direction of the risk: `chunk_kda` ends in `**kwargs`, so on
  a pre-0.5.1 fla the new name is swallowed without error and the state comes
  back K-first instead of matching `mamba_v_cache` -- a silent wrong answer,
  not an exception. The old name has the same exposure once fla drops its
  deprecation shim.

## Test Plan

- Kimi-K3 accuracy test

## Test Result

- Kimi-K3 accuracy passed.

## Submission Checklist

### DIFF
--- a/atom/models/kimi_k3.py
+++ b/atom/models/kimi_k3.py
@@ -958,36 +958,35 @@ class KimiKDAAttention(nn.Module):
         cu_seqlens: torch.Tensor | None,
         output_final_state: bool,
     ):
-        from fla.ops.kda import chunk_kda
+        from aiter.ops.triton.kimi_delta_attn import chunk_kimi_delta_attn
 
-        kwargs = {
-            "q": q,
-            "k": k,
-            "v": v,
-            "g": g,
-            # Keep beta in fp32: fla computes b = sigmoid(beta) in-kernel with
-            # use_beta_sigmoid_in_kernel, and triton's sigmoid follows the input
-            # dtype -- a bf16 beta yields a bf16 write strength, which erodes the
-            # delta-rule state update across the 71 KDA layers (measured gsm8k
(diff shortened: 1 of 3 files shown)


atom/models/kimi_k3.py
```diff
@@ -980,14 +979,30 @@ def _run_kda(
...
-        # and preserves the same chunk-KDA forward semantics.
-        return chunk_kda(**kwargs, disable_recompute=True)
+        if envs.ATOM_USE_AITER_KDA:
+            from aiter.ops.triton.kimi_delta_attn import chunk_kimi_delta_attn
+
+            return chunk_kimi_delta_attn(**kwargs)
+
+        from fla.ops.kda import chunk_kda
+
+        return chunk_kda(**kwargs)
```
whx-sjtu: Is it necessary to still keep fla fall back path here? Maybe we can just make aiter path as default and remove dependency of fla.

Liang-jianhao97: done

whx-sjtu (approved):

zejunchen-zejun (approved):

valarLip (approved):

PR #1824, opened 2026-08-07
[ATOM+SGL]ci(sglang): fallback to recent nightly benchmark image

zhangxinyuanliuhengyu:
## Summary

- Update scheduled SGLang Benchmark image selection to avoid blocking only on the current-day nightly Docker tag.
- Prefer the same-day `sglang-v{version}-nightly_YYYYMMDD` image, but fall back to a same-version nightly tag from today/yesterday when the current-day image is missing.
- Keep fallback bounded to recent same-version images so benchmark data does not silently use stale Docker images.

## Why

The scheduled SGLang Benchmark can currently skip the whole benchmark matrix if the same-day SGLang Docker release fails or is delayed. Since this benchmark does not need to be hard-blocked by a missing same-day tag, using a recent same-version nightly image is safer than losing the entire run.

## Test plan

- `git diff --check`
- Reviewed the updated scheduled image wait logic in `.github/workflows/atom-sglang-benchmark.yaml`

## Notes

- The fallback is only for scheduled `main` runs.
- `workflow_dispatch` can still override the image explicitly via `docker_image`.

### DIFF
--- a/.github/workflows/atom-sglang-benchmark.yaml
+++ b/.github/workflows/atom-sglang-benchmark.yaml
@@ -136,7 +136,7 @@ jobs:
           import re
           import sys
           import time
-          from datetime import datetime
+          from datetime import datetime, timedelta
           from pathlib import Path
           from urllib.error import HTTPError, URLError
           from urllib.parse import quote, urlencode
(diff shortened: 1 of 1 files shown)


zhuyuhua-v (approved):

zhuyuhua-v:
check last 2 days' image since benchmark change to bi-day schedule.

PR #1835, opened 2026-08-07
fix(scheduler): unblock decode behind a long chunked prefill

hippothewild:
## Motivation

A single long prompt stops every other request in the engine until it finishes.

`Scheduler.schedule()` returns as soon as it has scheduled any prefill, and the decode
scheduling block sits below that return. This is harmless for short prompts, which
occupy one tick. A chunked prefill, however, consumes the entire
`max_num_batched_tokens` budget on every tick it runs, so a 600K-token prompt keeps a
prefill in the batch for all ~37 of its ticks and no decode runs for their whole
duration.

Measured on a single MI355X x8 node (Kimi-K3 + DSpark `mtp_k=7`, TP=8, DP=1, PP=1,
`max_num_seqs=64`, `max_num_batched_tokens=16384`, chunked prefill enabled,
`kv_cache_dtype=fp8`), a 101-token request arriving 2 s into a 600,104-token prefill
waited **82.4 s** for its first token. With this change the same request is answered in
**1.9 s**, and the long prompt still finishes within 0.6 s of its original time.

Timeline: upstream blocks the short request for the whole prefill; the patched build answers it while the prefill continues

<sub>Schematic of the two scenarios. Exact A/B numbers are under Test Result.</sub>

Two further defects surfaced while fixing this and are included here, because they are
not independently landable. Fixing the scheduling order makes defect 3 reachable:
decode-priority can preempt a mid-chunk prefill, and the Phase 1 early return is
precisely what makes that path unreachable at `dp_size == 1` today. Splitting this into
separate PRs would merge a change that activates a latent state-corruption bug before
the fix for it.

### 1. Decode never runs on a tick that carries prefill

`atom/model_engine/scheduler.py`

The fix does not mix prefill and decode in one batch — the model runner does not support
that today. It changes the ordering instead: when decode work is ready, schedule a
decode-only batch first and let the prefill resume on a later tick. Batch shapes are
unchanged, so the runner needs no changes.

Both directions are bounded in time, and the second bound matters more than it appears.
An unbounded decode-priority produces the inverse failure: a busy decode batch keeps
decode work ready on every tick, so Phase 1 and Phase 2 — the only paths that admit new
requests, resume chunked prefills, and promote remote-KV consumers — are never reached.
An unbounded variant measured 36.4 s to first token for a 91-token request against a
0.65 s idle baseline. `_prefill_admission_starving()` yields the tick back once a queued
prefill has waited past the bound.

Symmetrically, a large prefill resume steps aside for a short queued prompt, but only
once and only for a bounded interval, so a stream of short prompts cannot freeze a long
one.

Three bounds are read from the environment to keep the diff small:

| variable | default | effect |
| --- | --- | --- |
| `ATOM_PREFILL_STARVE_MS` | 2000 | decode-priority yields to a prefill queued longer than this |
| `ATOM_LARGE_PREFILL_DEFER_TOKENS` | 30000 | a resume with more remaining tokens than this may be deferred |
| `ATOM_LARGE_PREFILL_DEFER_MS` | 2000 | maximum time such a resume is deferred |

### 2. The MTP status queue is consumed on every step but filled only on some

`atom/model_engine/model_runner.py`

`prepare_input_ids()` pops `pending_mtp_status_copies` unconditionally. The queue is
filled from `postprocess`, which the pure-middle-chunk / non-last-PP early return in
`forward` skips. A chunked prefill therefore discards the previous decode's
rejected/bonus counts, and the next decode step reads an empty queue:

```
TypeError: 'NoneType' object is not subscriptable
  File "atom/model_engine/scheduler.py", line 1817, in postprocess
    num_rejected = int(fwd_output.num_rejected[idx])
```

Any prompt above roughly 50K tokens killed the engine core within seconds. The fix gates
the consumer on the same predicate that gates the producer, `batch.produces_output()`.

Guarding the consumer in `Scheduler.postprocess` instead is worse than the crash. That
does not skip the step — it routes real deferred tokens into the non-speculative branch,
which appends tokens rather than overwriting the MTP placeholders. The placeholders then
stay in the context permanently, `num_tokens` inflates, and a stale `seq.num_rejected`
shifts both the next step's write offset and `preempt()`'s strip arithmetic. The crash
becomes silent context corruption.

### 3. `preempt()` truncates a chunked prefill's prompt

`atom/model_engine/scheduler.py`

`postprocess` appends the `mtp_k - num_rejected` speculative placeholders only to
sequences that are RUNNING and not mid-chunked-prefill, but `preempt()` strips
`mtp_k + seq.num_rejected` from every sequence it touches. Preempting a sequence that is
still prefilling deletes real prompt tokens.

`num_prompt_tokens` is assigned once at construction, so the sequence can never satisfy
`num_cached_tokens >= num_prompt_tokens` again. It stays `is_partial_prefill` forever on
every re-admission and pins its KV blocks, and because Phase 1 `break`s rather than
`continue`s, every later chunked prefill queues behind it. Nothing is logged — the
client simply hangs.

## Test Plan

Four regression tests are added. All four fail on `main` without this change and pass
with it.
They drive the real `Scheduler` on CPU through the existing `tests/conftest.py`
fixtures, so they need no GPU.

- `tests/test_scheduler_preempt_partial_prefill.py` — a preempted partial prefill keeps
  every prompt token and never drops below `num_prompt_tokens`.
- `tests/test_scheduler_hol.py` — decode runs ahead of a freshly queued prefill;
  decode-priority yields once an arrival ages past `_HOL_STARVE_MS`; a large resume is
  deferred for a short prompt and then released after `_HOL_DEFER_MS`.

End-to-end runs used the node configuration described under Motivation. The baseline is
the identical container image with `scheduler.py` and `model_runner.py` reverted to
`9acb0843`, so nothing else — AITER revision, kernels, model weights, engine flags —
differs between the two sides. Each build was driven with the same three probes: one 600K
prompt with short arrivals spread across its prefill, a c=16 throughput sweep, and a
three-way run combining sustained decode load, a 600K prefill, and short arrivals.

## Test Result

```
tests/test_scheduler_hol.py ...                                          [ 75%]
tests/test_scheduler_preempt_partial_prefill.py .                        [100%]
4 passed in 0.07s
```

Both builds are the same image with only `scheduler.py` and `model_runner.py` differing,
so the comparison isolates this change. Numbers below are the second 600K run on each
build — the first long prefill after engine start is warm-up dominated (95 s vs 84 s) and
is discarded on both sides.

**One 600K-token prompt, with N short requests fired together 10 s into its prefill.**
Median time to first token across the N:

| short requests in flight | upstream `9acb0843` | with this PR |
| --- | --- | --- |
| 1 | 74.5 s | **2.5 s** |
| 4 | 74.1 s | **3.1 s** |
| 16 | 74.3 s | **2.9 s** |

Upstream is flat at ~74 s because the number of short requests is irrelevant: none of them
runs until the prefill is done, so each one just waits out the ~75 s still remaining. With
this change the level drops ~25x and stays flat as concurrency grows — the short requests
are being served alongside the prefill rather than after it.

The prompt that causes the stall is unaffected: it completes in 85.6–86.0 s upstream and
86.1–86.3 s with this PR, across the same three runs.

Arrival time does not change the picture either. Staggering four arrivals at t = 2 / 20 /
45 / 70 s instead of firing them together gives 82.4 / 64.4 / 39.4 / 14.4 s upstream —
four different numbers that are all the same absolute deadline, the moment the prefill
ends — against 1.9 / 3.8 / 3.0 / 5.6 s with this PR.

**Sustained load, c=16, three rounds after a discarded warm-up:**

| | upstream `9acb0843` | with this PR |
| --- | --- | --- |
| total output | 606.9 / 694.2 tok/s | 650.4 / 660.1 tok/s |
| per request | 48.3 / 47.1 tok/s | 47.7 / 48.2 tok/s |

Run-to-run spread is wider than the difference between the two builds, so throughput is
unchanged.

**Crash reproduction.** Driving upstream with eight sustained decode streams plus a 600K
prefill killed the engine core, which is defect 2:

```
[atom 11:02:01] run_engine: exception: 'NoneType' object is not subscriptable
Traceback (most recent call last):
  File "/app/ATOM/atom/model_engine/engine_core.py", line 215, in run_engine
    engine.busy_loop()
  File "/app/ATOM/atom/model_engine/engine_core.py", line 245, in busy_loop
    self._process_engine_step()
  File "/app/ATOM/atom/model_engine/engine_core.py", line 258, in _process_engine_step
    return self._process_engine_step_inner()
  File "/app/ATOM/atom/model_engine/engine_core.py", line 314, in _process_engine_step_inner
    finished_seqs = self.scheduler.postprocess(
  File "/app/ATOM/atom/model_engine/scheduler.py", line 1817, in postprocess
    num_rejected = int(fwd_output.num_rejected[idx])
TypeError: 'NoneType' object is not subscriptable
```

Every in-flight request then hung and the run had to be abandoned. The same workload on
the patched build completed: the 600K prompt returned its first token in 113.9 s despite
the decode load, the three short arrivals in 2.8 / 3.7 / 3.9 s, no request starved, and
the log contains no traceback.

Style gates: `black --check .` is clean across the tree, and `ruff check .` reports 1585
findings both with and without this change — no new ones.

Under `.github/scripts/run_unit_tests.sh`, the command Pre Checkin actually runs:

| | passed | failed |
| --- | --- | --- |
| `main` | 756 | 0 |
| with this PR | 760 | 0 |

The four added tests are the whole difference; nothing else moved.

(A bare `pytest tests/` — without the script's `--ignore=tests/plugin` — breaks these and
several existing scheduler tests alike with `ImportError: cannot import name
'get_kvconnector' from 'atom.utils.forward_context'`. That is the module-stub leakage the
runner script already documents as its reason for excluding `tests/plugin`, and it
affects `main` identically, so I left it alone.)

Two points I would rather a maintainer decide:

- the three bounds should probably become `Config` fields
  (`prefill_starve_ms`, `large_prefill_defer_tokens`, `large_prefill_defer_ms`) with CLI
  flags rather than environment variables.
- `_schedule_decode_batch()` is now shared by `schedule()` and
  `_try_schedule_decode_priority()`. Splitting it differently, or folding the
  decode-priority decision into `schedule()` itself, is a reasonable alternative.

## Submission Checklist

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -387,7 +387,24 @@ class tokenIDProcessor:
         ]
         self.input_ids.copy_to_gpu(total_tokens_prefill)
 
-        self.prev_rejected_num, self.prev_bonus_num = self.recv_mtp_status_async()
+        # `pending_mtp_status_copies` is filled in `postprocess`, which the
+        # pure-middle-chunk / non-last-PP early return in `forward` skips. Popping it
+        # on every step drains it: a chunked prefill that produces no token discards
+        # the status belonging to the previous decode, and the next decode step then
+        # reads an empty queue and gets (None, None). Consume it under the same
+        # predicate that fills it so producer and consumer stay in lockstep.
+        if batch.produces_output():
+            self.prev_rejected_num, self.prev_bonus_num = self.recv_mtp_status_async()
+            if (
+                self.is_deferred_out
+                and self.prev_rejected_num is None
+                and batch.total_seqs_num_decode > 0
+                and not batch.is_dummy_run
+            ):
+                logger.warning(
+                    "MTP status queue was empty on a token-producing decode step; "
+                    "speculative bookkeeping for this step is unreliable."
+                )
 
         # TODO: remove this when we support mixed prefill and decode in one batch
         if total_reqs_prefill > 0:
--- a/atom/model_engine/scheduler.py
+++ b/atom/model_engine/scheduler.py
@@ -18,6 +18,7 @@ This module provides:
 from __future__ import annotations
 
 import logging
+import os
 import struct
 import threading
 import time
@@ -26,6 +27,15 @@ from collections import deque
 import numpy as np
 
 from atom.config import Config
+
+# --- head-of-line scheduling bounds (see Scheduler.schedule) -----------------
+# Upper bound on how long decode-priority may hold prefill admission back.
+_HOL_STARVE_MS = float(os.environ.get("ATOM_PREFILL_STARVE_MS", "2000"))
+# A partial prefill with more than this many tokens left counts as "large" and may
+# be deferred so that short waiting prefills can be admitted first.
+_HOL_LARGE_TOKENS = int(os.environ.get("ATOM_LARGE_PREFILL_DEFER_TOKENS", "30000"))
+# Upper bound on how long one large prefill resume may be deferred.
+_HOL_DEFER_MS = float(os.environ.get("ATOM_LARGE_PREFILL_DEFER_MS", "2000"))
 from atom.kv_transfer.disaggregation import KVConnectorOutput
 from atom.model_engine.block_manager import BlockManager
 from atom.model_engine.request import RequestOutput
@@ -970,6 +980,21 @@ class Scheduler:
         if not self.running and not self.waiting:
             return None
 
+        # Decode-priority: a chunked prefill occupies the whole token budget for
+        # every tick it runs, so scheduling it ahead of ready decodes stalls every
+        # running sequence for the length of the prompt.  Run decode first and let
+        # the prefill resume on a later tick -- but only while prefill work has not
+        # been held back past `_HOL_STARVE_MS`, otherwise a steady decode load would
+        # block new admissions forever.
+        if (
+            self.running
+            and self._has_decode_ready_seqs()
+            and not self._prefill_admission_starving()
+        ):
+            decode_result = self._try_schedule_decode_priority()
+            if decode_result is not None:
+                return decode_result
+
         # ---- Phase 1: resume partial prefills from running ----
         # Gated by `delayer_allows` so cross-DP alignment still holds when one
         # rank is mid-chunked-prefill: a delayer veto skips both Phase 1 and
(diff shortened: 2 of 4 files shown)


yitingw1 (changes requested):
Blocking: the decode-priority early return is a per-rank, wall-clock-timed decision that bypasses PrefillDelayer's cross-DP all-reduce lockstep.

- At dp_size > 1, ranks can land on different phases in the same tick (A decodes, B prefills) — exactly what PrefillDelayer exists to prevent.
- e2e results and reasoning here only cover DP=1.

Suggest gating decode-priority on dp_size == 1, or confirming/adding DP>1 coverage before merge.

PR #1841, opened 2026-08-10
(fix)Fix missing MTP metadata in interleaved agentic prefill and decode

zhuyuhua-v:
## Motivation
Long-running agentic workloads can interleave chunked prefill with deferred MTP decode output. In this case, prefill outputs legitimately contain no speculative rejection metadata, and an interleaved prefill may consume the queued D2H status before the previous MTP decode output is processed.

The scheduler previously assumed that every output had num_rejected, num_bonus, and draft_token_ids whenever speculative decoding was enabled. Indexing these optional arrays caused EngineCore to crash with:
image

## Technical Details

 - Propagate is_prev_prefill from ModelRunner to ScheduledBatchOutput.
 - Record the configured speculative decoding method in the scheduler.
 - Handle missing speculative metadata according to the output type:
    - For prefill output, set rejected and bonus token counts to zero because prefill does not perform speculative verification.
    - For deferred plain-MTP decode output, reconstruct the fixed-width MTP status from the number of returned tokens:
       - num_bonus = num_new_tokens - 1
       - num_rejected = mtp_k - num_bonus
 - Restrict reconstruction to plain MTP outputs with a valid token count.
 - Require num_rejected and num_bonus to be either both present or both absent.
 - Clear stale draft-token state when no draft metadata is returned.
 - Preserve integer normalization for metadata returned as NumPy scalar values.

## Test Result

With this fix, the AgentX MTP3 matrix at concurrency 1, 2, 4, 8, and 10 completed the full profiling duration with zero request errors and passed the AIPerf metric-coverage checks.
image

## Submission Checklist

### DIFF
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -3210,6 +3210,9 @@ class ModelRunner:
             is_deferred_out=self.tokenID_processor.is_deferred_out,
             num_rejected=prev_rejected_num,
             num_bonus=prev_bonus_num,
+            is_prev_prefill=(
+                prev_batch is not None and prev_batch.total_seqs_num_prefill > 0
+            ),
             logprobs=logprobs_map,
             dspark_ell=dspark_ell,
         )
--- a/atom/model_engine/scheduler.py
+++ b/atom/model_engine/scheduler.py
@@ -1817,8 +1817,7 @@ class Scheduler:
                 # that case (it previously crashed long AgentX MTP runs).
                 if fwd_output.num_rejected is None or fwd_output.num_bonus is None:
                     if not (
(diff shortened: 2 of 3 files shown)


yitingw1 (commented):
For DSpark/eagle3, raise RuntimeError will actually be triggered — trading a silent crash for an explicit one, not a real fix.
Suggestion: state explicitly in the PR description that this patch only covers plain MTP, and the same class of crash on DSpark/eagle3 still needs fix in [#1835](https://github.com/ROCm/ATOM/pull/1835).

zhuyuhua-v:
close as https://github.com/ROCm/ATOM/pull/1771 fixed the issue

PR #1874, opened 2026-08-12
feat(kv-cache): add V4 checkpoint capacity headroom

yhl-amd:
## What this does

Adds the `STATE_CKPT_EXTRA_ENTRIES=N` environment variable for native ATOM DeepSeek-V4, allowing the STATE pool to reserve checkpoint capacity above the `max_num_seqs` live-request floor.

With the variable set, V4 receives `max_num_seqs + N` state groups. The x268 benchmark configuration is:

```bash
STATE_CKPT_EXTRA_ENTRIES=268
```

With `max_num_seqs=96`, that produces `364` STATE groups.

This is intentionally an environment-only temporary control. It does not add a Config field, EngineArgs field, or CLI flag.

## Why

With only the live-request floor, reaching `max_num_seqs` consumes every STATE group and drains retained checkpoints. Explicit headroom lets deployments trade PAGE capacity for checkpoint retention without increasing scheduler concurrency or CUDA-graph capture sizes.

## Behavior and tradeoffs

- Unset or empty values resolve to `0`, preserving existing sizing and behavior.
- The V4 STATE slot declaration directly passes `int(envs.STATE_CKPT_EXTRA_ENTRIES)` as flat extra entries.
- The reservation is static and reduces the PAGE byte remainder.
- Extra groups remain reclaimable for admission; they are not a protected checkpoint partition.
- The reservation remains allocated when checkpointing or prefix caching is disabled.
- Set a valid integer value before starting the server/workers.
- Separately launched ranks or PD endpoints must use the same value.
- This control is native-ATOM DeepSeek-V4 only. GDN and the vLLM/SGLang plugin pools are unchanged.

## Implementation

- Registers `STATE_CKPT_EXTRA_ENTRIES` through the centralized lazy `atom.utils.envs` mechanism.
- Reads it directly in the native DeepSeek-V4 STATE slot declaration.
- Accounts for the extra entries in pool sizing and the PAGE memory remainder.
- Skips a checkpoint copy when that content hash is already indexed.
- Keeps newly selected copy destinations off the free list until the current pending loop is complete, then releases them together.
- Adds CPU-safe coverage for environment parsing, the absence of Config/EngineArgs/CLI surfaces, pool arithmetic, V4-only wiring, checkpoint headroom, and existing-checkpoint reuse.

## Verification

- Focused CPU tests: `218 passed, 47 skipped`
- `tests/test_state_checkpoint.py`: `97 passed`
- Randomized checkpoint-copy state regression: `1500 passed`
- Black: all changed Python files pass
- Ruff: all changed Python files pass
- `git diff --check`: clean
- GitHub Offline inference smoke test: passed
- Direct 8x MI355X/gfx950, TP8, BF16, FULL CUDAGraph server smoke:
  - unset: `96` STATE groups, about `2.40 GiB/GPU`
  - x268: `364` STATE groups, about `9.10 GiB/GPU`
  - both started, served a streaming chat request with HTTP 200, had zero fault signatures, and shut down cleanly
- Forced-capacity two-request A/B with `max_num_seqs=1`, checkpoint interval `256`, and the same 305-token prompt:
  - unset: STATE groups `1`, second-request cached tokens `0`, kept/dropped `0/2`, lost-to-checkpoint `41.97%`
  - x268: STATE groups `269`, second-request cached tokens `256`, kept/dropped `1/0`, lost-to-checkpoint `0%`

Full GPU model accuracy and x268 long-run testing on the current PR source remain follow-up validation.


atom/model_engine/state_pool.py
```diff
@@ -537,20 +537,28 @@ def _commit_pending(self) -> None:
         """
         if not self._checkpoint_pending:
             return
+        pending: dict[int, int] = {}
         for seq in self._checkpoint_pending:
             h, seq.pending_checkpoint = seq.pending_checkpoint, -1
             src = seq.per_req_cache_group
             if h == -1 or src < 0:
```
valarLip, line 544: 1

atom/model_engine/model_runner.py
```diff
@@ -618,6 +618,19 @@ def __init__(self, rank: int, config: Config):
         self.use_mla = self.is_deepseek_mla()
         self.use_gdn = self.is_qwen_next()
         self.use_v4 = self.is_deepseek_v4()
+        try:
+            self.state_ckpt_extra_entries = (
+                envs.STATE_CKPT_EXTRA_ENTRIES if self.use_v4 else 0
```
valarLip: 1

atom/model_ops/attentions/deepseek_v4_attn.py
```diff
@@ -900,7 +900,12 @@ def sub_pool_specs(self) -> list[SubPoolSpec]:
         row_bytes = self.plane_row_bytes()
         return [
             page_pool(geo.block_bytes(row_bytes) + self._indexer_block_bytes()),
-            state_pool(STATE_SLOT_CLASS, geo.slot_bytes(row_bytes), entries_per_req=1),
+            state_pool(
+                STATE_SLOT_CLASS,
+                geo.slot_bytes(row_bytes),
+                entries_per_req=1,
+                extra_entries=int(envs.STATE_CKPT_EXTRA_ENTRIES),
```
valarLip: move envs.STATE_CKPT_EXTRA_ENTRIES into state_pool, so all the other model can also use this one

valarLip (approved):

PR #1882, opened 2026-08-13
Enhance GLM-5.2 with DP attention, prefix cache optimizations, and logging

Jasen2201:
## Motivation

This PR extends GLM-5.2 pipeline-parallel (PP) + data-parallel (DP) attention serving with 9 interconnected improvements spanning KV transfer, prefix caching, speculative decoding, load balancing, and observability.

## Summary

- **PP-aware KV offload aggregation** — new ZMQ channel + aggregator so the head scheduler knows when all PP stages finish loading/saving
- **PD incremental KV transfer** — decode reuses cached prefix blocks; only the delta tail crosses RDMA
- **PP shared-indexer top-k carry** — DSA IndexShare sparse indices cross PP boundaries via IntermediateTensors
- **MTP on PP prefill nodes** — draft model fills KV on every chunk (not just the final sampling chunk)
- **Sparse MLA bounds hardening** — workspace size check prevents illegal memory access under DP metadata skew
- **LMCache CPU budget PP scaling** — budget proportional to local layer count, not full model
- **Deferred prefix hash publishing** — hashes from non-output PP chunks wait for sampled tokens before publishing

### 1. PP-Aware KV Offload Aggregation

Under pipeline parallelism each PP stage holds different model layers. KV offload operations (load/save to LMCache) complete independently per stage, but the head scheduler must know when *all* stages are done for a given request before it can advance that sequence.

A new `PPKVAggregator` class collects per-stage completion reports and emits a terminal event only when all `pp_size` stages have reported in. A third ZMQ channel (`kv_status`) carries these reports from downstream stages back to the head.

```mermaid
flowchart TD
    subgraph "PP Stage 0 (Head)"
        S0_CONN["Offload Connector\n(local layers)"] -->|"finished_loading\nfinished_saving"| S0_POLL["_poll_kv_transfer_progress()"]
        ZMQ_RECV["ZMQ kv_status PULL"] -->|"downstream reports"| S0_POLL
        S0_POLL -->|"non-offload\n(mooncake P/D)"| SCHED["Scheduler\n(direct)"]
        S0_POLL -->|"offload events"| AGG["PPKVAggregator"]
        AGG -->|"all pp_size stages\nreported for req"| SCHED
    end

    subgraph "PP Stage 1"
        S1_CONN["Offload Connector"] -->|"finished_loading"| S1_SEND["send_kv_status()"]
        S1_SEND -->|"ZMQ PUSH"| ZMQ_RECV
    end

    subgraph "PP Stage N-1"
        SN_CONN["Offload Connector"] -->|"finished_saving"| SN_SEND["send_kv_status()"]
        SN_SEND -->|"ZMQ PUSH"| ZMQ_RECV
    end
```

**Aggregation logic:** `PPKVAggregator` tracks `{ReqId: set[pp_rank]}` for loading/saving/failed. When a request's stage set reaches `pp_size`, it's emitted as globally done.

```mermaid
flowchart LR
    IN["ingest(event)"] --> CHK{event type?}
    CHK -->|finished_loading| LOAD["_loading[req_id]\n.add(pp_rank)"]
    CHK -->|finished_saving| SAVE["_saving[req_id]\n.add(pp_rank)"]
    CHK -->|failed_loading| FAIL["_failed_loading[req_id]\n.add(pp_rank)"]
    LOAD --> LEN1{"len == pp_size?"}
    SAVE --> LEN2{"len == pp_size?"}
    FAIL --> LEN3{"len == pp_size?"}
    LEN1 -->|yes| EMIT["emit globally_done"]
    LEN2 -->|yes| EMIT
    LEN3 -->|yes| EMIT
    LEN1 -->|no| WAIT["accumulate"]
    LEN2 -->|no| WAIT
    LEN3 -->|no| WAIT
```

**Connector-only batches:** When no new requests arrive but parked sequences need their offload loads kicked off, `_dispatch_connector_only_batch()` sends a metadata-only batch (no model forward) so the connector starts prefetching.

---

### 2. PD Incremental KV Transfer

In prefill-decode disaggregation, the decode node now registers received prompt blocks into its local prefix cache. On subsequent requests sharing the same prefix, only the **delta blocks** (the ones not already cached on decode) cross RDMA.

```mermaid
sequenceDiagram
    participant P as Prefill Node
    participant D as Decode Node

    Note over D: Request A arrives (new prefix)
    P->>D: Transfer ALL blocks [0..N]
    Note over D: Register blocks [0..N] in prefix cache

    Note over D: Request B arrives (shares prefix with A)
    D->>D: Prefix cache lookup → M blocks hit
    D->>P: num_computed_blocks = M
    Note over P: Compute only delta blocks [M..N']
    P->>D: Transfer ONLY blocks [M..N']
    Note over D: ~79% transfer avoided on multi-turn
```

**Implementation path:**

```mermaid
flowchart LR
    subgraph "Send Side (Prefill)"
        SEQ["Sequence"] -->|"num_cached_tokens\n÷ block_size"| NCB["num_computed_blocks"]
        NCB --> META["KVTransferMeta\n.num_computed_blocks"]
    end

    subgraph "Receive Side (Decode)"
        META --> OFF["off = meta.num_computed_blocks"]
        OFF --> SLICE["blocks = blocks[off:]"]
        SLICE --> ALLOC["Allocate only\ndelta blocks"]
        ALLOC --> RECV["RDMA receive\ndelta only"]
    end
```

---

### 3. PP Shared-Indexer Top-K Carry

GLM-5.2's DSA (Dense-Sparse-Attention) uses IndexShare groups where "shared" layers reuse a prior "full" layer's sparse top-k indices. When a PP boundary splits a shared group, the receiving rank has no "full" layer to source indices from. This fix carries `sparse_kv_indices` across the PP boundary via `IntermediateTensors`.

```mermaid
flowchart LR
    subgraph "PP Stage K"
        L_FULL["Layer i (full)\ncomputes top-k indices"] --> BUF["_sparse_kv_indices_gpu"]
        L_SHARED1["Layer i+1 (shared)\nreuses indices"] -.->|reads| BUF
        BUF --> SEND["IntermediateTensors\n+ sparse_kv_indices"]
    end

    SEND -->|"PP transport\n(pp_comm.py)"| RECV

    subgraph "PP Stage K+1"
        RECV["Receive IntermediateTensors"] --> BUF2["_sparse_kv_indices_gpu\n(write incoming)"]
        L_SHARED2["Layer i+2 (shared)\nreuses indices"] -.->|reads| BUF2
        L_SHARED3["Layer i+3 (shared)\nreuses indices"] -.->|reads| BUF2
    end
```

Detection: `_setup_pp_shared_indexer()` inspects `indexer_types` and `index_topk_pattern` to set `_pp_send_needs_sparse` / `_pp_recv_needs_sparse` flags per rank. Zero-cost no-op when no boundary crosses a shared group.

---

### 4. MTP on PP Prefill Nodes

Enables MTP (Multi-Token Prediction / speculative drafting) on disaggregated prefill nodes running under PP. Previously, MTP was gated incorrectly and skipped on non-last PP ranks. The key insight: the draft model must fill its KV cache on *every* prefill chunk so it has context for proposal when the last chunk arrives.

```mermaid
flowchart TD
    FWD["run_model()\ntarget forward"] --> CHK{"Last PP rank\n& last chunk?"}
    CHK -->|"yes"| SAMPLE["Sample tokens\n+ MTP propose"]
    CHK -->|"no (middle chunk\nor non-last rank)"| PRECOMP["drafter.precompute_context_kv()"]
    PRECOMP --> EAGLE["EagleProposer\n.precompute_context_kv()"]
    EAGLE --> ANCHOR{"Any anchor == -1?\n(sampling pending)"}
    ANCHOR -->|"yes"| SKIP["Skip — no anchor yet"]
    ANCHOR -->|"no"| FILL["Draft model forward\nover context tokens\n→ KV cache filled"]
    FILL --> READY["Draft KV ready\nfor all chunks"]
    READY --> PROPOSE["MTP propose()\nwith full context"]
    PROPOSE --> ACCEPT["~71% acceptance rate"]
```

---

---

### 5. Sparse MLA Index Bounds Hardening

`_sparse_index_workspace()` now validates that the workspace tensor has sufficient capacity for the requested `total_out` slots. Under DP attention, metadata skew between ranks could leave the workspace undersized → HIP illegal-access fault. Now raises a descriptive `RuntimeError` instead of silently over-reading.

---

### 6. Prefix Cache & Block Pool Observability

- **Per-request prefix cache logging** (toggle: `ATOM_LOG_PREFIX_CACHE_PER_REQ`) — on each sequence admission, logs hit/miss count, eviction delta, and block pool occupancy.
- **Block pool occupancy counters** — the block manager tracks allocated vs. free blocks, eviction rate, and hash-chain health.

---

### 7. LMCache CPU Budget PP Scaling

Under PP, each stage holds only a fraction of the model's layers. The CPU budget now scales proportionally:

```
scaled = configured × pp_size × local_layers ÷ total_layers
```

`kv_shape` also uses `num_layers` (local) not `total_layers` so only local layers are cached.

---

### 8. Deferred Prefix Hash Publishing for PP

Under PP, non-output batches (middle prefill chunks) must not publish their prefix hashes immediately — the sampled tokens haven't arrived yet. Publishing with wrong/missing tokens corrupts the hash chain.

```mermaid
sequenceDiagram
    participant FWD as Forward Pass
    participant PEND as _pending_prefix_hash
    participant SCHED as Scheduler

    Note over FWD: Chunk 1 (non-output, middle)
    FWD->>PEND: append (batch, seqs)
    Note over PEND: Deferred — no tokens yet

    Note over FWD: Chunk 2 (non-output, middle)
    FWD->>PEND: append (batch, seqs)

    Note over FWD: Chunk 3 (output, final)
    FWD->>SCHED: Sampled tokens arrive
    SCHED->>SCHED: _flush_pending_prefix_hashes()
    Note over SCHED: register_prefill_hashes() for chunks 1, 2, 3 in order
    PEND-->>SCHED: Pending batches drained
```
image
<html>
<body>
<html><head>
            
            <style>
              table { border-collapse: collapse; width: 100%; }
              th, td { border: 1px solid #ccc; padding: 8px; text-align: left; }
            </style>
          </head>
          <body>
Run | Weights | Cache | Parallel | all-gather | Conc | req/s | tput/GPU | ratio/B300 | input/GPU | output/GPU | input tok/s | output tok/s | TTFT p50 (ms) | ITL p50 (ms) | p90 intvty tok/s/user | realized prefix hit | theoretical hit | prompt tokens | generated tokens | errors/requests | Effective Concurrency
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
glm52_pp4pd_dpa_aiperf_1h_20260814_0416 | MXFP4 | fp8 | P:PP4×TP1 / D:DPA dp4×tp1 | n/a (tp=1) | 32 | 0.82 | 14481 | — | 14382 | 99.6 | 115053 | 797 | 7153 | 16.11 | 87.99 | 96.87% | 97.41% | 418794457 | 2900856 | 0/3002 | 19.46

        
      </body></html>
</body>
</html>

## Test Plan

| Test File | Covers |
|-----------|--------|
| `test_pp.py` | Prefix-hash publishing with parked sequences |
| `test_pd_pp.py` | PP head/downstream metadata for request-less batches |
| `test_lmcache_offload_connector.py` | CPU budget scaling across PP stages |
| `test_block_manager.py` | Hash-chain backfill and received-prefix registration |

## Test Result

- **Hardware:** MI355X
- **GSM8K accuracy:** 0.93–0.94
- **DP rank distribution:** Even across all workers
- **MTP acceptance:** ~71% on all four decode ranks


atom/kv_transfer/offload/connector.py
```diff
@@ -104,12 +104,22 @@ def register_kv_caches(
...
 
         tp = get_tp_group()
-        rank, world = tp.rank_in_group, tp.world_size
+        pp_rank = getattr(
+            getattr(self._config, "parallel_config", None),
+            "pipeline_parallel_rank",
+            0,
+        )
+        pp_size = int(getattr(self._config, "pipeline_parallel_size", 1) or 1)
+        rank = pp_rank * tp.world_size + tp.rank_in_group
```
yhl-amd: [P1] Isolate LMCache lookup identity per DP engine

With `--enable-dp-attention`, `EngineCoreManager` turns TP4 into four DP EngineCores with TP=1. This calculation therefore assigns every DP process `rank=0/world=1`. `build_lmcache_metadata` also fixes `engine_id` to `atom-offload` and `lookup_server_worker_ids` to `[0]`, so all DP processes bind/connect the same rank-0 IPC endpoint. Only one lookup server can own it; the other schedulers query that DP rank, making their local caches invisible and allowing false hits when rank 0 has data that the current rank lacks.

Please include DP identity in the lookup namespace (for example, one `engine_id` per DP group) and compute matching worker/scheduler rank/world inside that group's PP × TP[/PCP] cache shards. DP replicas should not be aggregated as lockstep cache shards. A multi-DP lookup-path test would catch this.

atom/model_engine/pp_engine_core.py
```diff
@@ -150,16 +157,96 @@ def _pp_head_step(self):
...
+            kvoutput = KVConnectorOutput()
+
+        # Non-offload fields go directly to scheduler.
+        non_offload = KVConnectorOutput(
+            finished_sending=kvoutput.finished_sending,
+            finished_recving=kvoutput.finished_recving,
+            failed_recving=kvoutput.failed_recving,
+        )
+        if not non_offload.is_empty():
+            self.scheduler._update_from_kv_xfer_finished(non_offload)
```
yhl-amd, line 261: [P1] Gate `finished_sending` on all PP-stage LMCache saves

For `multi(mooncake producer + lmcache_offload)`, stage 0 gets the global Mooncake release, but its `MultiConnector` pairs that send only with stage 0's LMCache save. Forwarding `finished_sending` here lets `Scheduler` deallocate/reuse the shared block IDs while downstream stages may still be D2H-saving them, so new forwards can overwrite KV being persisted. Conversely, downstream `MultiConnector` instances never receive a local Mooncake send completion and retain their `finished_saving` forever, so `PPKVAggregator` cannot reach global save completion.

Please move this pairing to a PP-wide coordinator: only forward `finished_sending` after the global Mooncake send **and** all PP-stage LMCache saves are done; downstream save completions must not wait for a local send that only stage 0 receives.

atom/model_engine/pp_engine_core.py
```diff
@@ -86,7 +90,10 @@ def _pp_head_step(self):
             if result is None:
                 break
             scheduled_batch, seqs = result
-            if scheduled_batch is None or len(scheduled_batch.req_ids) == 0:
+            if scheduled_batch is None:
+                break
+            if len(scheduled_batch.req_ids) == 0:
+                self._dispatch_connector_only_batch(scheduled_batch)
```
yhl-amd, line 104: [P1] Keep advancing connector work after scheduler queues become empty

This connector-only dispatch is reached only from `_pp_head_step`, but the outer loop stops calling it once the last postprocess leaves no waiting/running requests and `_in_flight` is empty. Final asynchronous LMCache saves and their `finished_saving` messages can therefore remain unpolled, leaving deferred blocks/save state pinned until an unrelated request arrives (and potentially through shutdown).

Please include deferred/connector/aggregator work in the liveness condition (for example, `has_pending_kv_work()`) and keep idle dispatch + polling active until metadata, in-flight saves, status aggregation, and deferred frees are drained.

atom/kv_transfer/disaggregation/pp_kv_aggregator.py
```diff
@@ -0,0 +1,68 @@
...
+            self._loading.pop(rid, None)
+            self._failed_loading.pop(rid, None)
+        for rid in done_saving:
+            self._saving.pop(rid, None)
+
+        return KVConnectorOutput(
+            finished_loading=done_loading,
+            failed_loading=failed,
+            finished_saving=done_saving,
+        )
```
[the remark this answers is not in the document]
yhl-amd: [P1] There is a more immediate correctness race than state resurrection: `failed_loading` is emitted as soon as any stage fails, before `success_ranks | failed_ranks` covers `pp_size`. The scheduler can then fall back to prefill while other stages are still asynchronously retrieving into the same logical block IDs, so the retrieve and the new forward can concurrently write KV.

A tombstone alone does not fix this. Please wait until every PP stage has reached a load terminal state, then emit failure if any stage failed. Any terminal tracking should be scoped to that load generation so the fallback path can still perform a legitimate save for the same request.

atom/model_engine/model_runner.py
```diff
@@ -3127,7 +3222,10 @@ def postprocess(
             num_reject_tokens = self.drafter.mtp_k - num_bonus_tokens
             next_token_locs = num_bonus_tokens
 
-        if get_tp_group().world_size > 1 and self.tokenID_processor.is_deferred_out:
+        # Drafter input must agree across TP ranks.
+        if get_tp_group().world_size > 1 and (
+            self.tokenID_processor.is_deferred_out or hasattr(self, "drafter")
```
ZhangLirong-amd, line 3244: why we need add `hasattr(self, "drafter")` here?

Jasen2201: PP nodes have is_deferred_out=False, but when they carry an MTP drafter they still need sampled_tokens broadcast across TP ranks to keep draft proposals deterministic — so the condition is widened to also trigger on hasattr(self, "drafter").

atom/kv_transfer/offload/connector.py
Jasen2201: now , DP-attention and PP>1 are not compatible , I have push a commit  
if config.enable_dp_attention and config.pipeline_parallel_size > 1:  
    raise ValueError

atom/model_engine/pp_engine_core.py
Jasen2201: The gap is real but self-healing — any new request re-enters the loop and drains the pending saves.

atom/model_engine/scheduler.py
```diff
@@ -656,20 +675,28 @@ def __init__(
...
-        # on: `--method mtp` on a checkpoint carrying `dspark_block_size` still
-        # resolves to DSpark, so the method name alone would be wrong here.
+        # EAGLE/MTP needs the successor token; DSpark does not.
         self.drafter_needs_next_token = self.use_spec and not (
             config.speculative_config.use_dspark()
         )
+        # True when this engine both drafts and verifies; False under PP
+        # (which only drafts for handoff to the decode node).
+        pp_size = getattr(config, "pipeline_parallel_size", 1)
+        self.spec_decode_local = self.use_spec and pp_size == 1
```
ZhangLirong-amd, line 685: pp not supported?

Jasen2201: PP prefill node：MTP drafter creat  draft token，but no need do  verify（verify is on decode）

atom/model_engine/pp_engine_core.py
Jasen2201: have fixed with new commit

atom/kv_transfer/offload/connector.py
yhl-amd: Thanks, but this guard does not address the reported case. The collision occurs in the supported `pp=1` DPA configuration: `EngineCoreManager` converts TP4 into four DP EngineCores with `tp=1`, and each LMCache worker still computes `rank=0/world=1` while sharing `engine_id="atom-offload"`. Rejecting DPA only when `pp>1` leaves those four DP replicas binding/querying the same rank-0 lookup endpoint.

This needs a replica-specific scheduler/worker namespace (for example, an `engine_id` containing `data_parallel_rank`) with rank/world scoped consistently inside that DP replica's cache-sharing group. The original P1 therefore remains.

atom/model_engine/pp_engine_core.py
yhl-amd: Re-reviewed `482f919d`. This partially fixes the original race for requests that actually perform both a Mooncake send and an LMCache save: downstream stages now pass save completion through, and the head waits for the PP save quorum.

However, it introduces a new P1 liveness issue. Once `_pp_kv_aggregator` has been initialized, lines 239-241 hold **every** later `finished_sending`, even when that request had no LMCache `save_spec` (for example, a short or already-persisted prompt). Such a send-only request will never produce `finished_saving`, so it remains in `_held_sending` forever and its producer blocks are never freed.

Please track per request whether a PP-wide save is actually pending; send-only completions must pass through immediately. Also, no tests were added for this barrier, and the latest targeted multi-connector suite currently has 4 failures because the test helper does not initialize the new `_pp_is_head` field.

atom/model_engine/pp_engine_core.py
yhl-amd: A future unrelated request eventually draining the state does not resolve the liveness issue. In a single-request/offline workload, during a long idle period, or during shutdown, the final save/status and deferred blocks can remain pending indefinitely because the outer loop still stops when `waiting/running` and `_in_flight` are empty.

The loop needs to treat connector/aggregator/deferred-free work as unfinished and continue idle dispatch + polling until it is drained. This P1 remains unchanged in `482f919d`.

atom/kv_transfer/disaggregation/pp_kv_aggregator.py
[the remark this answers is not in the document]
yhl-amd: Rechecked the latest head `482f919d`; this code is unchanged. `PPKVAggregator(2).ingest(0, failed_loading={"r"})` still emits `failed_loading={"r"}` immediately, before rank 1 reaches any terminal state. A later success from rank 1 then leaves stale partial loading state.

The scheduler can therefore still start fallback prefill while another stage is retrieving into the same block IDs. Please wait for `finished_ranks | failed_ranks` to cover all `pp_size` stages, then emit failure if any stage failed. The load-failure P1 remains.

yhl-amd:
LMCache re-review after the discussions were marked resolved: the PR head is still `482f919d` and no follow-up commit has landed, so the following P1s remain in the current code:

- DPA with supported `pp=1` still gives every DP replica the same LMCache rank/world and shared `engine_id="atom-offload"`: https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153191
- After `_pp_kv_aggregator` has been initialized, a later send-only request is still put in `_held_sending` and can never receive the `finished_saving` event required to release it: https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153584
- The PP head loop still stops without draining final connector, aggregator, and deferred-free work: https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153991
- `PPKVAggregator` still emits `failed_loading` after the first failing stage instead of waiting until all PP stages have reached a load terminal state: https://github.com/ROCm/ATOM/pull/1882#discussion_r3796154404

The non-GPU CI is also still red with the four `_pp_is_head` `AttributeError` failures introduced by this change (`4 failed, 1209 passed, 109 skipped`). Please reopen/address these discussions and add the corresponding regression coverage before merge.

Jasen2201:
> LMCache re-review after the discussions were marked resolved: the PR head is still `482f919d` and no follow-up commit has landed, so the following P1s remain in the current code:
> 
> * DPA with supported `pp=1` still gives every DP replica the same LMCache rank/world and shared `engine_id="atom-offload"`: [Enhance GLM-5.2 with DP attention, prefix cache optimizations, and logging #1882 (comment)](https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153191)
> * After `_pp_kv_aggregator` has been initialized, a later send-only request is still put in `_held_sending` and can never receive the `finished_saving` event required to release it: [Enhance GLM-5.2 with DP attention, prefix cache optimizations, and logging #1882 (comment)](https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153584)
> * The PP head loop still stops without draining final connector, aggregator, and deferred-free work: [Enhance GLM-5.2 with DP attention, prefix cache optimizations, and logging #1882 (comment)](https://github.com/ROCm/ATOM/pull/1882#discussion_r3796153991)
> * `PPKVAggregator` still emits `failed_loading` after the first failing stage instead of waiting until all PP stages have reached a load terminal state: [Enhance GLM-5.2 with DP attention, prefix cache optimizations, and logging #1882 (comment)](https://github.com/ROCm/ATOM/pull/1882#discussion_r3796154404)
> 
> The non-GPU CI is also still red with the four `_pp_is_head` `AttributeError` failures introduced by this change (`4 failed, 1209 passed, 109 skipped`). Please reopen/address these discussions and add the corresponding regression coverage before merge.

① Head holds an unpaired send forever 
▎ Fixed in 7140e226. Once _pp_kv_aggregator exists it never goes back to None, so every later finished_sending landed in _held_sending and only released on an aggregated finished_saving. A send-only request — any prompt shorter than LMCache's 256-token chunk, or one already persisted — waits for an event that never comes, and its blocks are never freed. Now only sends that arrive paired with a local save wait for the PP-wide quorum; the rest pass straight through, as the non-PP engine core already does.

② CI red — _pp_is_head

▎ Also fixed in 7140e226 (the test helper builds instances with __new__ and never got the new field). be8ce0da adds a test that constructs MultiConnector for real, so it can't drift again.

③ PP head loop doesn't drain

▎ Fixed in c8be0bc8. Not PP-specific — a request leaves running in the same pass that parks it in deferred_free_blocks, so is_finished() reads idle while its send/save is in flight. All three loops now OR in one predicate, EngineCore.has_pending_kv_work(); the PP head extends it via super() with _held_sending and the aggregator tallies. Covered by tests/test_kv_drain_liveness.py.

④ PPKVAggregator early failed_loading

▎ Fixed in 08cf96a4. Now mirrors the TP aggregator: terminal only when done | failed >= pp_size. Reporting early woke the request to recompute into blocks the other stages were still loading into, and dropped the tally that suppresses their late reports. Three tests in tests/test_pp_kv_status.py.

yhl-amd (dismissed):
LGTM

yhl-amd (approved):

PR #1886, opened 2026-08-13
fix(dpa+mesh) stabilize atomesh routing sticky and scheduling

wanzhenchn:
### Summary

- Preserve mesh-selected DP ranks across streaming and non-streaming request paths to maintain sticky affinity.
- Add role-specific DP rendezvous ports to prevent co-located prefill/decode conflicts on single node.
- Recover missing deferred-output placeholders before scheduler token replacement.
- Add coverage for DP sticky routing and rendezvous environment overrides.

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -881,6 +881,12 @@ class ParallelConfig:
             self.data_parallel_rank = envs.ATOM_DP_RANK
         if envs.is_set("ATOM_DP_RANK_LOCAL"):
             self.data_parallel_rank_local = envs.ATOM_DP_RANK_LOCAL
+        if envs.is_set("ATOM_DP_MASTER_IP"):
+            self.data_parallel_master_ip = envs.ATOM_DP_MASTER_IP
+        if envs.is_set("ATOM_DP_MASTER_PORT"):
+            self.data_parallel_master_port = envs.ATOM_DP_MASTER_PORT
+        if envs.is_set("ATOM_DP_BASE_PORT"):
+            self.data_parallel_base_port = envs.ATOM_DP_BASE_PORT
 
 
 _DSPARK_DEFAULT_MAX_BLOCK = 16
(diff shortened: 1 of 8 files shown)


junyyang-amd (dismissed):

atom/config.py
```diff
@@ -881,6 +881,12 @@ def __post_init__(self) -> None:
             self.data_parallel_rank = envs.ATOM_DP_RANK
         if envs.is_set("ATOM_DP_RANK_LOCAL"):
             self.data_parallel_rank_local = envs.ATOM_DP_RANK_LOCAL
+        if envs.is_set("ATOM_DP_MASTER_IP"):
```
Yuechguo, line 884: What are these new environment variables used for?

atom/model_engine/scheduler.py
```diff
@@ -2086,9 +2086,21 @@ def postprocess(
                     ell_r = fwd_output.dspark_ell.get(seq.id)
                     if ell_r is not None:
                         seq.dspark_next_ell = int(ell_r)
+                required_placeholders = num_placeholder + offset
```
Yuechguo, line 2089: In what cases would this part of the code be used?

atom/config.py
wanzhenchn: These variables allow Atomesh to configure separate DPA distributed communication endpoints for prefill and decode:

- `ATOM_DP_MASTER_IP / ATOM_DP_MASTER_PORT` configure DP coordination.
- `ATOM_DP_BASE_PORT` configures the model-runner distributed process group.

This **prevents port conflicts when prefill and decode run on the same nod**e. The overrides apply only when explicitly set, so existing defaults remain unchanged.

atom/model_engine/scheduler.py
wanzhenchn: This handles an issue observed during GLM-5.2 TP4-DPA warmup in agentic scenario: 

Image

- an aborted request was rescheduled while chunked prefill/deferred output was active. When its delayed output arrived, `output_tokens` contained fewer placeholders than expected, causing `IndexError: list assignment index out of range` and eventually cascading failures across all DPA ranks.

The code restores the missing temporary slots before writing the real tokens. The inserted EOS values are immediately overwritten and never returned to the user.

Yuechguo (dismissed):
LGTM

JiaoliangYu (approved):

PR #1890, opened 2026-08-13
Fuse block-banking cat into attn_res kernel

yanxuer-999:
## Motivation

 Route Kimi-K3 attn_res through aiter's fused kernel by default so the model gets fused per-token FP8 output and block-banking, and remove the old env-based routing switch. 

## Technical Details

apply_attn_res now dispatches to aiter's attn_res_gate via four kimi_k3_attn_res_gate* custom ops (plain / close_block / quant / quant+close_block), while ATOM's local Triton kernel stays registered only as an A/B baseline; AttnRes.forward passes out_quant_dtype from fused out_norm and can return ((fp8, scale), prefix_out) for downstream Linears. 
## Test Plan

 Validated with real AttnRes/RMSNorm/ReplicatedLinear under eager and torch.compile(backend="inductor"), plus a runtime dispatch probe confirming aiter is called and the local kernel is not.. 

## Test Result

Inductor correctness is fixed (no uninitialized outputs), the layer loop compiles to a single graph, and cudagraph replay keeps per-call overhead negligible on the production path.. 

## Submission Checklist


valarLip (dismissed):

zejunchen-zejun:
aiter PR: https://github.com/ROCm/aiter/pull/4726 merged

PR #1891, opened 2026-08-13
[ATOM SGL] fix ci error

ZhiweiYan-96:
## Bring the sglang plugin up to the current V4 and GDN metadata APIs

V4's window/compress addressing now lives in UnifiedPoolGeometry, and GDN carries a request's recurrent state as a read slot plus a write slot (both introduced in #1771). The sglang bridge still builds metadata for the earlier shape of both, so two CI jobs crash at decode graph capture:

- **DeepSeek-V4-Flash TP8** — write_v4_paged_decode_indices() got an unexpected keyword argument 'cache_size'
- **Qwen3.5-35B-A3B-FP8 TP2** — `'NoneType' object is not subscriptable` on `non_spec_state_indices_in_tensor`

### Changes

Everything is under atom/plugin/sglang/; no core file is touched.

- **GDN**: the bridge supplied only the write-side state indices, while attention_gdn.py reads the read side unconditionally. It now supplies both, giving the read side its own storage as that field's contract requires (decode replays a CUDAGraph with the pointer baked in at capture).
- **V4**: ports the bridge's index-builder calls to the geometry ABI. The shared builders address a compressed entry as phys * envelope_rows + row, with no base term, which fits a pool whose layer view already begins at its compressed region. This bridge keeps a private plane per layer, so the plane is reordered to `[compressed ++ windows]` to match that convention, and the windows carry the offset through PerLayerPoolGeometry.ring_start. That geometry is plugin-local and only reads WindowParams from core.

The vLLM bridge needs the same update; it is not covered here.

### DIFF
--- a/atom/plugin/sglang/attention_backend/attention_gdn.py
+++ b/atom/plugin/sglang/attention_backend/attention_gdn.py
@@ -29,6 +29,33 @@ from atom.utils.forward_context import (
 logger = logging.getLogger(__name__)
 
 
+def _state_indices_in(linear_backend: Any, idx: torch.Tensor) -> torch.Tensor:
+    """The slots this forward READS its incoming recurrent state from.
+
+    SGLang hands one mamba slot per request and this bridge never forks a
+    request's state, so the read slots hold the same values as the write ones.
+    They still get their own storage, per
+    ``GDNAttentionMetadata.non_spec_state_indices_in_tensor``: decode replays a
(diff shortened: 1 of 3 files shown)


ZhiweiYan-96:
Results for Dsv4 Flash TP8:
image

Results for Qwen3.5 35B TP2

image

Results for Deepseek v4 Pro MTP3

image

PR #1898, opened 2026-08-14
fix(scheduler) Include CPU-offloaded prefix hits in request usage

wanzhenchn:
## Background

In agentic scenario, AIPerf obtains cache-hit tokens from:

`usage.prompt_tokens_details.cached_tokens`

In ATOM, the value follows this path:

`seq.prefix_cache_hit_tokens`  
→ `RequestOutput.num_cached_tokens`  
→ `prompt_tokens_details.cached_tokens`  
→ `total_usage_prompt_cache_read_tokens`

Previously, when LMCache restored a longer prefix from CPU/NVMe, **the scheduler updated only** `seq.num_cached_tokens`. It did not update `seq.prefix_cache_hit_tokens`.

**ATOM therefore skipped the CPU-restored tokens correctly during prefill**, but AIPerf reported only the initial GPU HBM cache hit.

## Before This Change

Assume one request has:

- GPU HBM hit boundary: `gpu_hit`
- LMCache-loaded cumulative boundary: `loaded`

The scheduler updated:

```python
seq.num_cached_tokens = loaded
```

However, the reported value remained:

```text
cached_tokens = gpu_hit
```

Therefore:

```text
Reported cache hit = GPU hit only
Missing CPU hit = loaded - gpu_hit
```

Example:

```text
GPU HBM hit boundary:      40
LMCache loaded boundary:   70
Actual combined cache hit: 70
AIPerf reported hit:       40
Missing CPU hit:           30
```

The previous benchmark script also calculated cache statistics from periodic scheduler log messages. This approach had several limitations:

- Warmup requests could be included.
- Cancelled or unrelated requests could be included.
- Statistics were logged only every 100 requests, so trailing requests could be missed.
- Multiple DPA ranks could write interleaved output into the same log file.

## After This Change

After a successful LMCache load, the scheduler synchronizes the reported cache-hit boundary:

```python
seq.prefix_cache_hit_tokens = max(seq.prefix_cache_hit_tokens, loaded)
```

The reported per-request cache hit becomes:

```text
cached_tokens = max(existing_or_inherited_hit, loaded)
```

`loaded` is the cumulative cache boundary after CPU offload loading. It already includes the GPU-cached prefix, so the two values must not be added directly.

Example:

```text
GPU HBM hit boundary:    40
LMCache loaded boundary: 70
Additional CPU hit:      70 - 40 = 30
Reported cached tokens:  70
```

Using `max()` also prevents a smaller local value from overwriting a larger cache-hit value inherited from an upstream prefill node.

When LMCache is disabled, this code path is not executed, so existing GPU prefix-cache behavior remains unchanged.

## ATOMesh Benchmark Metric Changes

The benchmark script now reads AIPerf's profiling-only aggregate counters directly:

```text
cache_hit_tokens = total_usage_prompt_cache_read_tokens
cache_total_tokens = total_usage_prompt_tokens
cache_hit_rate = cache_hit_tokens / cache_total_tokens
```

These counters cover successful profiling requests and exclude AIPerf internal warmup and requests cancelled during grace-period draining.

The calculation changes from:

```text
Before:
AIPerf cache-hit tokens = Σ GPU/inherited cache-hit tokens
```

to:

```text
After:
AIPerf cache-hit tokens =
    Σ max(GPU/inherited hit, LMCache loaded boundary)
```

## Validation

- `tests/test_scheduler.py`: 62 tests passed
- No scheduler lint errors
- A follow-up AIPerf benchmark should confirm that CPU-restored prefix tokens are included in `total_usage_prompt_cache_read_tokens`

### DIFF
--- a/.github/scripts/atomesh/pd_server_atom.sh
+++ b/.github/scripts/atomesh/pd_server_atom.sh
@@ -742,41 +742,6 @@ ensure_aiperf() {
   "${AIPERF_VENV}/bin/aiperf" --version
 }
 
-# Cumulative prefix-cache token counters of every prefill server on this run,
-# read out of their logs: ATOM has no metrics endpoint, so the scheduler's
-# periodic "[Cache Stats]" line is the only place these numbers surface.
-# Prints "<cached> <total>" summed across prefill ranks, or returns non-zero when
-# no rank has printed a line yet -- callers must treat that as "unknown", never
-# as a zero hit rate.
-prefill_cache_hit_totals() {
-  local log cached total line
-  local sum_cached=0 sum_total=0 found=0
-  shopt -s nullglob
-  local logs=("${RUNTIME_LOG_DIR}"/prefill-rank-*.log)
-  shopt -u nullglob
-  for log in "${logs[@]}"; do
-    # The cumulative line is "[Cache Stats         ]" and the per-interval one is
-    # "[Cache Stats Interval]". Both carry "Cached/Total:", so the space run after
-    # "Stats" is what separates them -- matching on "Cached/Total:" alone would
-    # pick up whichever of the two happened to be printed last.
-    # The trailing "|| true" is load-bearing under `set -e -o pipefail`: a log with
-    # no such line yet is the normal early state, not an error.
-    line="$(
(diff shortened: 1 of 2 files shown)


yhl-amd (approved):
LGTM

Yuechguo (approved):
LGTM

PR #1903, opened 2026-08-15
[Gluon] [GFX12] add gluon mxfp8 gemm support

k50112113:


atom/model_ops/linear.py
```diff
@@ -270,6 +290,16 @@ def gemm_a8w8_blockscale_preshuffle_impl(
     dtype: torch.dtype = torch.bfloat16,
     prefix: str = "",
 ) -> torch.Tensor:
+    if use_gluon_mxfp8_gemm():
+        return gemm_afp8wfp8_preshuffle(
```
valarLip, line 294: can we add this into aiter like a16w16 gemm do... so vllm/sglang can also using muti backend gemm and we can always choose best one across different impl

k50112113: that's a good point... we have to plan that, we can discuss

k50112113:
I am doing AITER side wiring, closing this PR

PR #1911, opened 2026-08-16
perf(mla): take the decode's KV-split budget from the machine, not a hardcoded 16

zejunchen-zejun:
The persistent aiter MLA metadata build asks the kernel for a global KV-split budget, and we quoted a hardcoded 16. aiter derives that number itself when asked to: `max_split_per_batch < 0` means `num_splits = num_clusters` (`csrc/kernels/mla/metadata/v1_2_device.cuh:894`), i.e. cut the KV walk into as many parts as the machine has clusters. That is what ATOM's MHA path has always passed and what aiter defaults to.

`num_splits` is the divisor the planner uses for `payload = ceil(sum_blocks / num_splits) + 16 blocks`, the amount of KV work handed to each CU part. Quoting 16 therefore only binds while `16 * batch_size < num_clusters` — it was never a tuning choice, it was a small-batch throttle, and it left a batch-1 decode on 16 of a gfx950's 256 CUs. A batch-1 decode has no parallelism other than that KV walk: MLA's `num_heads_k` is 1, so the head dimension contributes none, and decode has no query dimension to speak of.

**The change is to stop overriding aiter.** All five call sites now pass `-1`: the shared persistent decode metadata, the sparse-MTP set, the TBO per-ubatch build, and the two sparse-prefill builds. On the prefill path the cap was inert to begin with — sparse prefill treats every query token as its own batch, so `num_batches` is the chunk's token count and `16 * num_batches` already saturates `num_clusters` for any chunk of 16 tokens or more.

**Buffer sizing is byte-for-byte unchanged, deliberately.** `get_mla_metadata_info_v1`'s fast-mode bound, `min(bs + max_splits - 1, (max_splits - 1) * 2) * tiles`, is already the worst case for `num_splits == num_clusters`: the planner walks `num_cu` CU parts and each emits at most one partial before yielding, so the `reduce_partial_map` entry count is bounded by `num_cu + tile_cnt` regardless of this parameter. aiter #3855 says the same — its capped expression `tile_cnt + per_tile_cap` is a conservative upper bound kept `>=` the fast-mode one, not a fix for an under-allocation. Nothing here changes what `mla_decode_fwd` allocates for its fp32 `logits`.

### Effect

`num_splits = min(num_clusters=256, cap * bs)` on MI355X at occupancy 1:

| bs | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---|---|---|---|---|---|
| before (cap 16) | 16 | 32 | 64 | 128 | 256 | 256 |
| after (aiter's -1) | 256 | 256 | 256 | 256 | 256 | 256 |

Bit-identical at bs >= 16; the change is confined to the batch sizes that cannot fill the machine. Short contexts are not over-split either — the planner's `payload` floor of 16 blocks means a 2k-token context still yields ~7 work items instead of 5, while a 180k one goes from 15 to 187.

### Measurement

Kimi-K3 MXFP4, 8x MI355X, tp8, DSpark N=2, InferenceX AgentX trace replay at concurrency 1. The budget the kernel picks there is 256, the same number the measurement forced (16 heads, 256 CUs, occupancy 1):

```
mla_a8w8_qh16_qseqlen4_gqaratio16_v3_ps   416.7us -> 48.0us per call
paired kn_mla_reduce_v1                     7.8us -> 41.0us per call
aggregate ITL                              12.68ms -> 9.76ms   (1.30x)
replay of a fixed 168-request set          3,358s -> 2,813s    (1.19x)
```

Output sequence lengths matched the unpatched run on every request. The reduce gets more partials to combine, so the win narrows as batch_size grows and the throttle stops binding anyway.

### Notes for review

- On by default for the native MLA persistent path (DSv3/V4/GLM/Kimi), so it needs `ci:atom` accuracy green before merge: the split count changes the fp32 reduce order.
- Only concurrency 1 is measured. bs=2..8 also changes (bs=4: 64 -> 256 splits) and is worth one replay pass before merge, even though the partial count cannot grow past `num_clusters` at any batch size.
- The one prefill case this does change is a chunk of fewer than 16 query tokens (a chunked-prefill tail, or few owned queries per rank under DCP). Those are small enough that the 16-block payload floor keeps the work-item count essentially where it was.
- `atom/plugin/vllm/attention/metadata.py:1177` still hardcodes the same 16 on the vLLM-plugin persistent decode. Same ceiling, same one-line fix; left for a follow-up PR so this one stays single-file.
- Unrelated and pre-existing, noticed while reading the planner: the kernel recomputes its cluster count from each call's actual q-len, and occupancy is a point condition on `num_heads * q_len == 64` rather than monotonic in q_len. A runtime q-len below the sizing q-len can therefore land on occupancy 2 while the buffers were sized at occupancy 1 (e.g. num_heads=32, num_spec_tokens=2, `AITER_ENABLE_EXPERIMENTAL` on). That is independent of this parameter — worth a separate look, not addressed here.

### Changed since the earlier revisions

The first revision exposed the budget as `ATOM_MLA_MAX_SPLIT_PER_BATCH` (plus an `ATOM_MLA_KV_GRANULARITY` escape hatch) defaulting to the old 16; the second derived the cluster count host-side via `get_mla_decode_fwd_max_splits` and fed it to both the sizing and the runtime call. Both are gone. aiter derives the same number from the device it is running on, so the framework has nothing to contribute here, and the sizing side never needed a matching value. Any `ATOM_MLA_MAX_SPLIT_PER_BATCH` export in a launch script is a no-op and can be dropped.

### DIFF
--- a/atom/model_ops/attentions/aiter_mla.py
+++ b/atom/model_ops/attentions/aiter_mla.py
@@ -51,14 +51,10 @@ try:
 except (TypeError, ValueError):
     _MLA_META_SUPPORTS_MAX_SPLIT = False
 
-# Cap on the decode's KV-split budget: aiter cuts the KV walk into
+# Cap on the KV-split budget: aiter cuts the KV walk into
 # `min(num_clusters, cap * batch_size)` parts, and a negative cap means uncapped
 # -- as many parts as the machine has clusters (v1_2_device.cuh:894).
 _MLA_SPLIT_BUDGET_AUTO = -1
-
-# Prefill caps it: it already has query-dimension parallelism, so extra KV splits
-# there only buy a bigger reduce.
-_MLA_PREFILL_MAX_SPLIT_PER_BATCH = 16
 
 
 def _mla_seg_meta_kwargs() -> dict:
@@ -1050,7 +1046,7 @@ class AiterMLAMetadataBuilder(CommonAttentionBuilder):
                 max_seqlen_qo=1,
                 uni_seqlen_qo=1,
                 fast_mode=1,
-                max_split_per_batch=_MLA_PREFILL_MAX_SPLIT_PER_BATCH,
+                max_split_per_batch=_MLA_SPLIT_BUDGET_AUTO,
             )
             attn_metadata.sparse_prefill_work_meta_data = var[
                 "sparse_prefill_work_meta_data"
@@ -1431,7 +1427,7 @@ class AiterMLAMetadataBuilder(CommonAttentionBuilder):
             max_seqlen_qo=1,
             uni_seqlen_qo=1,
             fast_mode=1,
-            max_split_per_batch=_MLA_PREFILL_MAX_SPLIT_PER_BATCH,
+            max_split_per_batch=_MLA_SPLIT_BUDGET_AUTO,
         )
         attn_metadata.sparse_prefill_work_meta_data = var[
             "sparse_prefill_work_meta_data"
--- a/atom/model_ops/attentions/aiter_mla.py
+++ b/atom/model_ops/attentions/aiter_mla.py
@@ -51,31 +51,14 @@ try:
 except (TypeError, ValueError):
     _MLA_META_SUPPORTS_MAX_SPLIT = False
 
-# Let aiter size the decode's KV-split budget. `max_split_per_batch < 0` means
-# `num_splits = num_clusters` (csrc/kernels/mla/metadata/v1_2_device.cuh:894),
-# i.e. cut the KV walk into as many parts as the machine has clusters, which is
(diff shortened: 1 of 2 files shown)


zejunchen-zejun:
image
image

zejunchen-zejun:
For dspark model mla, aiter::mla_a8w8_qh16_qseqlen4_gqaratio16_msk0_v3_ps
image

zejunchen-zejun:
aiter PR for using a cost model to choose the optimal split size: https://github.com/ROCm/aiter/pull/4899

zejunchen-zejun:
Hi @valarLip @Zzz9990 

Could we merge this PR? The AITER PR has been merged just now.

Thank you

valarLip (approved):

PR #1919, opened 2026-08-17
[Kimi-K3] Integrate aiter fused KDA decode kernel

mengfei-jiang:
- When  batch <= 64, use aiter's fused_kda_decode (conv1d + recurrence + gated RMSNorm in one Triton kernel)
- Fused output (bf16) goes directly to o_proj which handles quant internally
- Falls back to original 3-kernel path for batch > 64 or spec decode

## Motivation

Single Triton kernel fusing three KDA decode operations:

  - causal_conv1d_update (Q + K + V)
  - delta-rule recurrence with KDA lower-bounded sigmoid gate
  - gated RMSNorm output normalization

## Submission Checklist


atom/models/kimi_k3.py
```diff
@@ -926,6 +926,8 @@ def __init__(
             prefix=f"{prefix}.o_proj",
         )
 
+        self._use_fused_kda_decode = envs.ATOM_ENABLE_FUSED_KDA_DECODE_TRITON
```
ganyi1996ppo: I think we can remove this flag, make it as default path

ganyi1996ppo (dismissed):

zejunchen-zejun (dismissed):

zejunchen-zejun:
aiter PR https://github.com/ROCm/aiter/pull/4712

ganyi1996ppo (dismissed):

zejunchen-zejun (dismissed):

zejunchen-zejun:
Thanks for putting this together. E2E performance numbers are still being measured and I'll post them once the runs settle.

In the meantime, there is a scoping problem worth resolving before this lands: **the fused path does not support DSpark speculative decoding or ReplaySSM, and those two are exactly what low-concurrency Kimi-K3 serving runs with.** At con 1 / 2 / 4 / 8 we always enable DSpark, and ReplaySSM defaults on whenever speculative decoding is on. That is the same regime where a decode-side fusion has the most to gain — so as written, the optimization is unavailable precisely where it would pay off.

Details below, in case they help scope the follow-up.

---

## 1. DSpark: the integration point is never reached

`GDNAttentionMetadataBuilder` branches on a **server-level** flag, not a per-batch one (`atom/model_ops/attentions/gdn_attn.py`):

```python
if not self.use_spec_decode or is_prefill:
    ...                        # non-spec path
else:
    num_spec_decodes = num_decodes
    num_prefills = 0
    num_decodes = 0            # all-or-nothing
```

with `self.use_spec_decode = self.num_spec > 0` and `self.num_spec = model_runner.drafter.mtp_k`.

So once a draft model is attached, **every** decode forward takes the else-branch and `num_decodes` is forced to 0 — there is no mixed batch. This PR hooks into

```python
elif kda_metadata.num_decodes > 0:      # atom/models/kimi_k3.py
```

which is therefore never entered under DSpark.

I verified this empirically rather than by reading alone. I added a counter on both sides of the new branch and ran the server for 900 s with `--method dspark --num-speculative-tokens 3` and the fused path **enabled**:

```
[FUSEDKDA] occurrences = 0
```

Not one line — neither the fused counter nor the fallback counter, since both live inside that `elif`. For contrast, with DSpark off the same build reports `fallback=…` when the fusion is disabled and `fused=…` when it is enabled.

The draft model is not a second entry point either: `atom/models/kimi_k3_dspark.py` only builds `K3DSparkMLAAttention` layers, so the DSpark draft has no KDA layers at all.

## 2. What the kernel would need for spec decode

The good news is that the hard structural piece is already there — `fused_conv_recurrent_norm_kernel` loops `for i_t in range(seq_T)`, which is exactly the `1 + num_spec` verify window. What is missing is the state bookkeeping.

**(a) Per-token snapshot slots (2D index tensor).** Spec decode passes `spec_state_indices_tensor` shaped `[batch, num_spec+1]` — one snapshot slot per token. The kernel takes a 1D `[B]` tensor, computes `p_h` once before the loop, and writes back to that same slot every iteration:

```python
state_idx = tl.load(ssm_state_indices_ptr + i_n)      # 1D, single slot
p_h = ssm_state_ptr + state_idx * stride_ssm_slot + ...
for i_t in range(seq_T):
    ...
    tl.store(p_h, b_h)                                 # same slot each token
```

**(b) Resume offset.** The incoming state must be read from `slot[num_accepted_tokens - 1]`, not slot 0. The kernel has no `num_accepted_tokens` argument.

(a) and (b) are small, and there is already a reference implementation inside aiter — `_triton_kernels/gated_delta_rule/decode/fused_rearrange_sigmoid_gdr.py`:

```python
i_t = tl.load(num_accepted_tokens + i_n).to(tl.int64) - 1 if IS_SPEC_DECODING else 0
state_idx = tl.load(ssm_state_indices + i_n * stride_indices_seq + i_t * stride_indices_tok)
...
for i_t in range(T):
    ...
    final_state_idx = tl.load(ssm_state_indices + i_n * stride_indices_seq
                              + i_t * stride_indices_tok)
    if final_state_idx >= 0:
        tl.store(...)
```

Adding `num_accepted_tokens`, an `IS_SPEC_DECODING` constexpr, `stride_indices_seq` / `stride_indices_tok`, and moving the write-slot lookup inside the loop is on the order of 15 lines.

**(c) conv_state rewind — this one is a rewrite, not a new parameter.** The conv buffer is sized with the speculation window (`GDNAttentionMetadataBuilder._state_shape`):

```python
conv_state_shape = (conv_kernel_size - 1 + num_spec, conv_dim // tp_world_size)
```

With `num_spec = 7` that is `state_len = 4 - 1 + 7 = 10`, **not 3**. The fused kernel hardcodes a `W-1` window: it reads `j = 0 … W-2` and appends the new input at `W-2`.

The actual semantics in `atom/model_ops/mamba_ops/causal_conv1d.py` are a right-aligned append with an acceptance-driven rewind:

```
VAL = state_len - seqlen                            # new tokens land at VAL … state_len-1
conv_state_token_offset = num_accepted_tokens - 1   # read-window start, spec mode
source shift = 1 if IS_SPEC_DECODING else seqlen
```

as its own comment describes:

```
Before forward:    [history1, history2, ..., historyM]
After forward:     [history2, ..., historyM, draft1, ..., draftN]
After accepting 2: [history3, ..., historyM, draft1, draft2]
```

The fused kernel has no notion of `state_len`, no acceptance offset, and a compile-time-constant shift, so this part needs reimplementing.

**(d) Two index tensors, not one.** The kernel currently uses a single `state_idx` for both `conv_state` and `ssm_state`. On the spec path conv uses `spec_state_indices[:, 0]` (one slot per sequence) while the recurrence uses the full 2D array; these have to become separate arguments.

## 3. ReplaySSM: the kernel does not emit the replay records

When ReplaySSM is active the decode path does not call `fused_sigmoid_gating_delta_rule_update` at all — it calls a different op:

```python
def replayssm_sigmoid_gating_delta_rule(
    q, k, v, a, b, A_log, dt_bias,
    ckpt,                        # one checkpoint per request, not per token
    buf_k, buf_u, buf_g,         # per-token record streams the kernel must WRITE
    write_pos, slot_idx,         # ring-buffer bookkeeping
    cu_seqlens, max_query_len, o=None, ...
)
```

The design trades per-token state snapshots for one checkpoint plus a replayable `(k, u, g)` record — the per-token states are rebuilt from those records on demand. `fused_conv_recurrent_norm_kernel` writes none of the three streams, so a request whose drafts are rejected has nothing to replay from.

This one should be cheap to add: the expensive intermediate is already in registers. `u = beta · (v − h·k)` is exactly the value held right after

```python
b_v = b_v - b_dot
b_v = b_v * b_beta        # this is u
```

so it is three more output pointers plus the ring-buffer index arithmetic — roughly 20–30 lines, and materially easier than the conv_state rewind in §2(c). The extra traffic is small: `3 × head_dim` values per (token, head) against the `head_dim²` state the kernel already moves, on the order of 2%.

### One correctness item to fix regardless of merge order

The new early-return is inserted **before** the ReplaySSM check:

```python
elif kda_metadata.num_decodes > 0:
    decode_state_indices = state_indices[:num_actual_tokens]
    #  <-- fused path returns here
    q, k, v = causal_conv1d_update(...)
    ...
    if getattr(kda_metadata, "replayssm", False):
        replayssm_sigmoid_gating_delta_rule(...)     # never reached
    else:
        fused_sigmoid_gating_delta_rule_update(...)
```

If the ReplaySSM change lands first, this PR as written will **silently disable ReplaySSM for every batch of ≤ 64 tokens** — the record buffers are never written, and a later rollback replays stale `(k, u, g)` with no error raised. Worth guarding with `if not getattr(kda_metadata, "replayssm", False)` (or moving the insertion point below that branch) even before full support exists, especially since ReplaySSM defaults on with speculative decoding:

```python
self.replayssm = self.use_spec_decode if replayssm_override is None else replayssm_override
```

## 4. Reachability summary

| DSpark | ReplaySSM | fused path taken? |
|---|---|---|
| on | on (default) | no — `num_decodes == 0` |
| on | off (explicit) | no — `num_decodes == 0` |
| off | on (explicit) | no — ReplaySSM branch, **but silently bypassed as written** |
| off | off | yes — the only combination |

zejunchen-zejun:
Here is the perf number w/ and w/o fusion
image
image

valarLip (approved):

PR #1953, opened 2026-08-19
feat(lmcache): add GLM-5.2 multiprocess offload

yhl-amd:
## Summary

- Add the `lmcache_mp` connector (with `LMCacheMPConnector` as an alias) for native ATOM GLM-5.2, transferring the 576-element MLA latent cache and padded 144-byte DSA index cache as two physical LMCache groups over one shared paged-block ID space.
- Keep an independent LMCache KV rank and object key for every TP worker, use a layout-specific namespace, enforce chunk/block geometry, and reuse the existing dense offload scheduling policy for chunk-aligned lookup, store, and retrieve.
- Preserve LMCache device futures and CUDA events until transfers are provably terminal, with explicit lookup-lock cleanup and fail-closed handling so late GPU writes cannot race KV-block reuse or recomputation.
- Install the pinned LMCache v0.5.3 ROCm wheel in ATOM images, rebuild only the HIP `c_ops` extension against the image's PyTorch ABI, and backport the scheduler-heartbeat fix from LMCache#3828.
- Add a GLM-5.2 deployment recipe covering server startup, connector options, cold/warm validation, capacity planning, timeout behavior, and failure semantics.

## Validation

- 312 relevant unit tests passed.
- Ruff, Black, and `git diff --check` passed.
- `docker buildx build --check` passes for `docker/atom_release.dockerfile`; `docker/Dockerfile` reports only its pre-existing undefined `$PYTHONPATH` warning.
- TP8 ROCm cold/store/warm/load validation completed successfully: every GPU reached 8,448 cumulative stored and retrieved tokens. Server-failure testing also exercised degraded mode and worker cache re-registration after recovery.

## Known limitations

- Native ATOM GLM-5.2 (`glm_moe_dsa`) only; exactly one LMCache MP server and TP-only deployment. PP, DCP, PCP, DP, and DP Attention are rejected.
- Supports LMCache `auto` and `lmcache_driven` transfer modes; `engine_driven` cannot represent the two physical cache groups.
- Lookup admission is synchronous, and a control-plane call can block the scheduler for up to the configured MQ timeout.
- There is no engine-level shutdown callback yet; LMCache registrations are released when worker processes exit.
- A permanent server failure after GPU transfer work is already in flight is fail-stop for the affected request and requires an ATOM engine restart for safe recovery.


yhl-amd:
## CxS long-context benchmark results (GLM-5.2 TP8)

I ran the official `LMCache/LMBenchmark` `real-multi-round-qa` workload against this PR and compared the native `lmcache_mp` connector with no offload and ATOM's legacy in-process `lmcache_offload` connector.

### Configuration

- Model: `zai-org/GLM-5.2-FP8`, native ATOM, TP8
- Workload: C=2, S=2; 4 sessions; 5 turns/session; 20 requests total
- Input: Gutenberg novels truncated to 8192 tokens with the GLM-5.2 tokenizer
- Output: 64 tokens/request, temperature 0, fixed document-selection seed
- Server: 16k max context, 16k max batched tokens, 4 max sequences
- KV: FP8, 16-token ATOM blocks, prefix cache disabled
- LMCache: CPU L1, 256-token chunks, `OFFLOAD_MIN_LOAD_TOKENS=8192`

The upstream CxS client only reads `delta.content`. GLM-5.2 streams thinking tokens through `delta.reasoning_content`, so I made a local benchmark-only compatibility change to include `reasoning_content` in TTFT measurement and subsequent conversation history. Without it, TTFT/TPOT are incorrectly reported as zero.

### Results

| Mode | Wall time (s) | Req/s | Total tok/s | Mean TTFT (s) | P95 TTFT (s) | Warm mean TTFT (s) | Warm P95 TTFT (s) | Mean latency (s) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| No cache | 15.2664 | 1.3101 | 10,991.46 | 0.7678 | 1.0077 | 0.7812 | 1.0084 | 1.5264 |
| `lmcache_mp` | **13.0566** | **1.5318** | **12,851.58** | 0.5388 | 1.0423 | **0.4338** | **0.7832** | **1.3055** |
| Legacy `lmcache_offload` | 14.0865 | 1.4198 | 11,912.13 | **0.5384** | **0.8197** | 0.4857 | 0.8070 | 1.4084 |

All modes completed 20/20 requests. No-cache and legacy processed 166,520 prompt tokens; MP processed 166,518 (a two-token difference caused by nondeterministic FP8 MoE greedy output entering later turns). All modes generated 1,280 completion tokens.

### Relative performance

- MP vs no cache: **+16.92%** total token throughput, **-14.47%** wall time, **-44.47%** warm mean TTFT, and **-22.33%** warm P95 TTFT.
- MP vs legacy LMCache: **+7.89%** total token throughput, **-7.31%** wall time, **-10.69%** warm mean TTFT, and **-2.95%** warm P95 TTFT.

MP cold TTFT was 0.9588s versus 0.7140s for no-cache and 0.7494s for legacy, which is why MP's all-request P95 is slightly higher even though its cache-hit turns are faster.

### MP transfer validation

- 8/8 TP ranks registered successfully
- 800 L1 objects, 11,501,568,000 bytes used
- Every GPU reported 25,600 stored tokens and 106,496 retrieved tokens
- Operation/token counts matched across all eight GPUs
- No transfer error or traceback was observed

This is one small CxS point rather than a capacity sweep. A follow-up should sweep C/S and repeat each point at least three times, but this run confirms that the connector is exercising real MP store/retrieve and improves warm multi-turn performance over both baselines.

yhl-amd:
Added an interactive architecture walkthrough for this PR (Chinese text, with source links):

**[Open the GLM-5.2 MP offload walkthrough](https://yhl-amd.github.io/ATOM/glm52-mp-offload/)**

The diagrams show the ATOM worker / independent LMCache server process boundary, control messages versus GPU IPC data access, and the store/retrieve sequence. They also explain how MLA and indexer caches use separate copy layouts while their data for the same token chunk is stored together in one LMCache memory object.

Checked against PR head `00ceb94`. Layer counts, block IDs, and token counts in the diagrams are illustrative.

[HTML source](https://github.com/yhl-amd/ATOM/blob/1a7bb5c9a0b20ecb6f77f13a5f03de71f83bf701/glm52-mp-offload/index.html)

valarLip (approved):

PR #1973, opened 2026-08-20
Qwen38 accuracy

PerryZhang01:
## Motivation

This PR supports qwen3.8-27B and 2.4T on vllm-atom backend and  improves the accuracy of qwen3.8 on atom backend.

## vllm-atom 侧模型支持

- **qwen-2.4T 架构注册** 
  - `Qwen3_5MoeForCausalLM` → `Qwen3_5MoeForCausalLMVllm`
- **新增 wrapper 类**
  - 基类 `ATOMMoEForCausalLM` + `IsHybrid`
  - 补齐 GDN 的 mamba state dtype / shape / copy func，vLLM 才能给线性注意力分配 state cache
- **FULL CUDA Graph 元数据修复** 
  - 原实现用 GPU 布尔 mask 做高级索引 ——  改为 CPU 侧校验 padding 落在尾部，然后直接切 `[:real_num_decodes]`
  - padding 不在尾部 / decode 行不是单 token / block-table 形状变化 → 抛 `RuntimeError` 点名坏掉的假设
  - 取代原先静默读错 state slot 的行为

 ## 原生 ATOM 后端

- **thinking 控制名映射（会导致精度降低）**
  - ATOM 的 `thinking` / `thinking_effort` 镜像到 HF 模板认的 `enable_thinking` / `reasoning_effort`
  - 不映射时 effort 静默回落到模板默认值，跑出来的 baseline 会降低
- **`min_tokens` **
  - `temperature=1.0` 下首个 decode step 就可能采到 EOS，返回空回复，GSM8K 直接判错
- **mask 实现** 
  - 未达 `min_tokens` 时把 EOS、`stop_token_ids`、单 token stop string 的 logits 置 `-inf`
  - 在 logits 上做而不是丢弃已采样的 EOS —— 把 EOS 喂回模型会开启新 message，改变请求的分布
  - 多 token stop 序列不受影响，仍由 scheduler 在看到完整后缀后终止

## Test Plan

参考recipe.

###  ATOM backend

- **Qwen3.8-27B**（TP1 / KV fp8）：**0.9727**
- **Qwen3.8-2.4T-A95B-Quark-MXFP4**（TP8 / KV bf16）：**0.9833**

###  vLLM-ATOM backend

- **Qwen3.8-27B**（TP1）：**0.9757**
- **Qwen3.8-2.4T-A95B-Quark-MXFP4**（TP8 + EP8）：**0.9795**


JiaoliangYu (dismissed):

valarLip:
Reviewed at `7387efaf5`. Moving stop-string matching off the scheduler and onto text is the right direction — the old token-prefix matcher could not express a stop string that does not align to a token boundary. Findings below are ordered by consequence; I verified the load-bearing ones against the branch.

---

# 1. The `min_tokens` floor can be crossed on all three paths

This is the PR's headline feature, so taking the three together:

**(a) `scheduler.py:433` — `num_completion_tokens` under-counts placeholders by `mtp_k + num_rejected`.**

It subtracts only `seq.num_placeholder_tokens`, but a running sequence carries more than that. `postprocess` computes the real length as `seq.num_tokens - num_placeholder_width - num_rejected` (`:2335`) and then appends `num_placeholder - num_rejected` fresh placeholders (`:2517`), so physical-minus-real is `2*mtp_k + 1` (deferred) while `num_placeholder_tokens` is only `mtp_k + 1 - rej`.

Traced for `mtp_k=1`, deferred, prompt=4: after one real token the batch reports 2 completion tokens where the truth is 1. Both `apply_min_tokens_mask_with_spec_decode` and the bonus-row mask consume this, so `remaining = clip(min_tokens - completed, 0, None)` hits 0 while the request is still below its floor and EOS stops being masked. `test_min_tokens_metadata_discounts_placeholder_tokens` only exercises `mtp_k=0`, so it passes.

**(b) `scheduler.py:2361` — `below_min_tokens` is computed from the post-step total but gates a stop positioned at `stop_at_idx`.**

The two are equivalent only when a step yields exactly one token. With `min_tokens=5`, 0 completion tokens so far, and one MTP step accepting 5 tokens with EOS at index 0: `num_tokens - num_prompt_tokens == 5`, so `below_min_tokens` is False, the EOS branch fires with `stop_at_idx=0`, the truncation at `:2416` drops 4 tokens, and `:2495` commits with `leave_reason="eos"`. The request finishes with **1 completion token against a floor of 5**. The same shape applies to a P/D `injected_t0`, which the block's own comment names as the reason the guard exists. `test_min_tokens_holds_off_eos` feeds one token per step and cannot catch it. The comparison needs to be against the completion length *at* `stop_at_idx`.

**(c) `llm_engine.py:640` — a stop string ends a request below its floor, and it is the one stop condition neither guard covers.**

`min_tokens` appears nowhere in `_wrap_for_stop_strings` or in `check_stop_strings`. With `{"min_tokens": 50, "stop": ["."]}` the sampler masks EOS and the scheduler refuses to finish, but the wrapper calls `check_stop_strings` from the first delta and aborts on the first period — roughly 2 tokens against a floor of 50. vLLM has exactly this guard (`vllm/v1/engine/detokenizer.py`: `if self.stop and self.num_output_tokens() > self.min_tokens:`). This also makes the sampler docstring false where it says "a floor that is still in force simply keeps that check from running" — nothing keeps it from running. No test covers `min_tokens` together with `stop`.

# 2. An empty stop string truncates every completion to zero

`check_stop_strings` guards only `if not new_char_count or not stop:` — never the individual entries. For `stop_str = ""`, `output_text.find("", start)` returns `start` and `end = stop_index + 0`, so `best_end` is tiny; with the default `include_in_output=False` the wrapper sets `truncate_to=0`, `finished=True`, `finish_reason="stop_sequence"` and aborts on the very first token. The client gets an empty completion.

`SamplingParams.__post_init__` validates `stop_token_ids` but does nothing to `stop_strings`, and `api_server.py:1573/1801` passes `request.stop` straight through. The old code was immune because `tokenizer.encode("")` yields no tokens and the `if stop_tokens:` filter dropped the entry — so this is a regression, and an empty entry is a common artifact of templating an unset variable into a stop list. A `if not stop_str: continue` in the loop closes it.

# 3. Mutating `finished = True` in place discards the engine's real terminal output

`engine_core_mgr.py:566-585` does `callback(request_output)` and then tests `if request_output.finished: self._seq_id_to_callback.pop(seq_id, None)` — on the very object the wrapper just mutated, several engine steps before the abort lands.

The scheduler's genuine terminal chunk (`scheduler.py:2463-2481`, which calls `kv_connector.request_finished(seq)` and sets `kv_transfer_params_output`) then finds `callback is None` and is dropped. **On a P/D disaggregated deployment, a stop-string request never returns its KV transfer params.**

Two secondary effects worth knowing about while fixing it: the wrapper's `state["stopped"]` / `elif not request_output.finished: return` defence is unreachable dead code on this path — the manager's early pop is what actually provides it; and if `abort_request` throws (the wrapper swallows it at `:665`) nothing re-aborts, because SSE `cleanup_stream` runs with `aborted=False` after seeing a finished chunk, leaving the sequence to run to `max_tokens` on the GPU with no consumer.

# 4. Stop strings are inert on `generate_multimodal` and the rollout engine

```
llm_engine.py:253  def generate(
llm_engine.py:270      stream_callback=lambda _request_output: None,   <- added so the wrapper has something to wrap
llm_engine.py:284  def generate_multimodal(                            <- not added
```

`atom/rollout/async_engine.py:137` likewise calls `add_request` without it. `stream_callback_iter` falls to `itertools.repeat(None)`, no callback is registered — and since the engine core no longer decides stop strings at all after this PR, a `stop=[...]` request on those two paths runs to EOS or `max_tokens` with the stop string left in the output. Before this PR the scheduler matched for every path equally.

# 5. Three consumers never learned about `stop_truncate_to`, and streaming cannot retract

- **`api_server.py:1048` — `generate_async_fanout` (non-streaming `n>1`).** `make_callback._cb` copies only `token_ids`/`finished`/`finish_reason`/`ts`, and the builder at `:1126` does a plain `tokenizer.decode(per_tokens[i], ...)` with no cut. `n=2, stream=false, stop=["\n\n"]` returns each choice containing the stop string plus every token emitted before the abort landed, while the identical `n=1` request truncates correctly. `RequestOutput.stop_truncate_to`'s docstring claims it is "honoured by every consumer that builds text".
- **`atom_standalone_service.py:624` — atomesh got it only on `SingleRequestState`.** `FanoutRequestState._build_outputs` and `StreamRequestState.record` (`:660`) both decode without a cut, so on the same server an `n=1` non-streaming request truncates while an `n>1` or streaming one does not. The streaming case is the harder of the two: it keeps no accumulated text, so a whole-completion character offset has no frame to be applied in without new state.
- **`streaming_dispatch.py:242` — a stop string straddling two deltas has already been partly sent.** With `stop=["</answer>"]`, step N's delta ends in `</ans` and is delivered (one delivery per engine step); step N+1 completes `wer>`, and `check_stop_strings` deliberately searches back `len(stop_str)-1` characters and returns the match **start**, which lies inside the already-sent delta. `emitted_before` then exceeds `truncate_to`, `max(0, truncate_to - emitted_before)` is 0, and the current delta is emptied — leaving the client's concatenated stream ending in `</ans`. `merge_chunk` cannot repair it: it only appends and never re-applies a cut. The non-streaming path for the same request cuts correctly, so the two delivery modes disagree.

# 6. Within one step, EOS wins unconditionally over an earlier client stop token

`scheduler.py:2387`: the EOS test is the `if` and `elif stop_ids and any(t in stop_ids for t in token_ids):` is only the `elif`. With `stop_token_ids=[S]` and a spec-decode step yielding `[S, x, EOS]`, `stop_at_idx` becomes 2 instead of 0; the truncation at `:2416` (`stop_at_idx < num_new_token - 1`) then trims nothing, so `x` and `EOS` are both emitted past the client's stop token and `finish_reason` reads `eos` rather than the stop. The correct behaviour is the minimum of the two indices. Structurally the old code had the same shape, but it was harmless there because the set held only EOS aliases — with client-named ids the ordering is now semantically meaningful.

# 7. `tests/test_sampler.py:35` can abort the whole CI suite

`saved` is captured *after* the first import attempt has already failed. A failed import removes its own `sys.modules` entry, so `saved["atom.model_ops.sampler"]` is None and the `finally` loop executes `sys.modules.pop("atom.model_ops.sampler")` on the module it just successfully stub-built.

That only happens on the non-GPU runner — the one place the stub path runs — and CI runs the whole suite in one process, so any later module doing `import atom.model_ops.sampler` (anything pulling in `atom.model_engine.model_runner`) misses the cache and re-executes the module body with the aiter stubs already torn down. That is a collection-time ImportError, which in this repo aborts all ~4100 tests rather than failing one. It also leaves two distinct module objects under one name. Fix: capture `saved` before the first import attempt, or do not restore that key.

# 8. Two recipes document things that are not there

- **`recipes/atom_vllm/Qwen3.8.md:111`** — the "Architecture Notes" section describes GDN metadata compaction, a `[:real_num_decodes]` slice and RuntimeError guards. `atom/plugin/vllm/gdn_backend.py` is not in this PR's diff, and `AtomGDNAttentionMetadataBuilder` in the tree is a docstring-only subclass whose docstring says the opposite: *"A prior post-build compaction pass rewrote that metadata using `num_actual_tokens` ... and shrunk `num_decodes`, which corrupted GDN ssm_state on Qwen3.5 block-FP8 under cudagraph replay."* There is no `build()` override, no slicing, no `real_num_decodes`, no RuntimeError. **The removal of that pass was the fix** — so the doc promises a loud failure where the real one is silent, and anyone acting on the compaction bullet would re-add the pass that is on record as corrupting ssm_state. The PR body's "FULL CUDA Graph metadata fix" is likewise unbacked by any hunk.
- **`recipes/Qwen3.8.md:133`** — both documented invocations pass a Hugging Face repo id (`Qwen/Qwen3.8-27B`, `amd/Qwen3.8-2.4T-A95B-Quark-MXFP4`), but `scripts/test_qwen38_gsm8k.sh:89-90` runs `[[ -d "${MODEL_PATH}" ]] || die` then `[[ -f "${MODEL_PATH}/config.json" ]] || die`, and `:116` does `cd -- "${MODEL_PATH}"`. There is no `snapshot_download` path, so both commands exit 2 with `模型目录不存在: Qwen/Qwen3.8-27B` before anything starts. The recipe's own `openai_server` command and the CI JSON entries both use repo ids, so the script is the only consumer that cannot accept one.

# 9. `discard_stop_state` has no caller, so `_stop_string_hits` grows for the process lifetime

`git grep -n discard_stop_state atom/ tests/` returns exactly one hit — its own definition at `llm_engine.py:864`.

`_stop_string_hits[req_id]` is written by the wrapper (`:661`) and removed only in `postprocess` (`:824`) and that uncalled method. `postprocess` runs only from the offline `generate`/`generate_multimodal` loops and `atom/rollout/async_engine.py`; every OpenAI-server and atomesh teardown path instead does `engine.io_processor.requests.pop(seq.id, None)` (`cleanup_stream`, and the `finally` blocks in `generate_async` / `generate_async_multimodal` / the fan-out). So a serving process keeps a permanent `{int: int}` entry per stop-string hit, keyed by a monotonically increasing seq id, so it never self-limits. The method's docstring describes exactly the leak it fails to prevent, which will make a future leak report read as already-fixed.

# 10. `Qwen3_5MoeForCausalLMVllm` cannot load fused-expert checkpoints

`plugin/vllm/models/qwen3_5.py:122` inherits `ATOMModelBase.load_weights`, which passes no `load_fused_expert_weights_fn`. `loading_core.py:243-245` picks `detect_fused_expert_format` and `get_fused_expert_mapping` off the model via `getattr` but takes the fn only from the explicit parameter, and `weight_dispatch.py:169-176` requires all three — so `_dispatch_fused_expert` returns False and the stacked `experts.gate_up_proj` / `experts.down_proj` tensors fall through to the generic path.

Every other consumer of this architecture wires it explicitly: the multimodal sibling in this same file overrides `load_weights` to delegate to `self.model.load_weights` with `load_fused_expert_weights_fn=self.load_fused_expert_weights` (`:356`), as do the sglang (`:489`) and rtpllm (`:812`) plugins. `atom.models.qwen3_5.Qwen3_5MoeForCausalLM` defines the method (`models/qwen3_5.py:587`); it is simply never reached. The per-expert MXFP4 Quark checkpoint in the CI matrix does not use the fused layout, so nightly will not catch it.

---

## Smaller items

- `llm_engine.postprocess:818` applies a `skip_special_tokens=True` offset to a `skip_special_tokens=False` decode.
- `stop_token_ids` and `include_stop_str_in_output` are unreachable from every HTTP entrypoint, while the Rust mesh router already forwards `stop_token_ids`.
- `min_tokens` is missing from the Anthropic path, and is non-nullable, so `"min_tokens": null` 422s.
- `docs/scheduling_kv_cache_guide.md:515/632/659` still documents the deleted `stop_token_sequences`, and the recipe's `min_tokens` bullets describe "single-token stop strings" the code does not mask.
- The stop-string abort now runs per-request on the engine output thread, contradicting `_send_control`'s documented "none of them runs per request".
- `IncrementalDetokenizer.text` now accumulates for *every* streaming request — O(n²) attribute concat, in the same file whose `merge_chunk` documents the opposite discipline.
- Three `Sequence` attributes are now dead, and `tests/plugin` is excluded from CI so the new plugin test never runs.

## Checked and dismissed

Recording these so they do not get re-raised: the "double finish" theory is **not** a bug — `engine_core_mgr.py:583` pops the callback on the flag the wrapper just mutated, so no later output reaches the client (the real consequence is finding 3 instead). The CI JSON and workflow entries validate clean against the repo's own schema gate, runner labels and all 41 workflow choice strings. `--fewshot_as_multiturn 0` is fine — lm_eval 0.4.11 declares it with a bool converter, not `store_true`. Bulk-vs-incremental decode drift is real only with `clean_up_tokenization_spaces=True`, which none of the target models enable.

PR #1974, opened 2026-08-20
perf(dsv4): optimize FP8 and FP4 indexer prefill

yhl-amd:
## Summary

This PR optimizes both DeepSeek-V4 indexer-prefill cache formats:

- **FP8:** use the paged-cache scorer as the built-in real-prefill path. It reads the FP8 page cache directly, so prefill no longer gathers all sequences into one dense packed K buffer or remaps global top-k indices back to each sequence.
- **FP4:** replace the old row-count threshold table with a kernel-derived wave-task model. It selects K parallelism from device coverage and per-wave serial depth, and can launch the same work as independent one-wave CTAs for ragged/multi-sequence prefill.
- Build and reuse FP8 expanded page tables and FP4 schedules once per effective forward/ubatch. PCP, TBO, and logits-budget chunking rebuild matching metadata when needed.
- Keep the legacy FP8 scorer only for the pre-pool model warmup dry run. FP4 decode remains on its fixed 256x4 CUDAGraph schedule.

Real FP8 prefill is therefore paged by default and has no backend environment switch.

## FP8 paged prefill

The legacy FP8 path gathered all co-scheduled sequences into a packed K tensor. Its logits width was `sum(K_i)`, whereas the paged path is sequence-local with width `max(K_i)`. The paged path also removes the gather and global-to-local top-k remap.

The metadata builder expands the sequence block table to a contiguous per-query table once per effective forward and reuses it across CSA layers. Logits use a 256-column-aligned physical row stride because the AITER preshuffle kernel may issue a full tail-chunk store, while `visible_end_gpu` retains the exact logical causal bound.

Synthetic FP8 indexer-pipeline benchmark on MI355X, `Q=8192`, 2 GiB logits budget:

| Shape | Legacy / paged calls | Legacy | Paged | Speedup |
| --- | ---: | ---: | ---: | ---: |
| Ragged bs=4, final contexts 133120/264192/526336/690176 | 7 / 3 | 28.79 ms | 11.98 ms | 2.40x |
| bs=1, final context 696320 | 3 / 3 | 24.95 ms | 20.88 ms | 1.20x |

The ragged expanded page table is 84.25 MiB and takes 0.03 ms median to build. The full ragged comparison had five near-tie swaps across 8,388,608 top-k entries; set overlap was 99.99994%. The committed GPU tests use wide score margins and require exact top-k sets.

## FP4 source-driven scheduling

The production FlyDSL kernel has `MFMA_N=16`. Its source computes:

```text
N_TILES_PER_WARP = (BLOCK_K / 16) / num_warps
```

Both supported launch shapes therefore assign exactly four MFMA N tiles, or 64 K rows, to each wave:

```text
256 K rows / 4 waves = 64 rows per wave
 64 K rows / 1 wave  = 64 rows per wave
```

The kernel has no LDS, scratch, cross-wave reduction, barrier, or cross-wave data dependency. The two shapes execute the same MFMA/KV work for complete tiles. The one-wave form changes only hardware scheduling granularity and reduces causal-tail rounding from 256 to 64 rows. It also compiles with slightly fewer VGPRs (88 versus 92).

The new selector therefore reasons in **64-row wave tasks**, not arbitrary CTAs per row. Its device term is derived from gfx950 topology: `256 CUs * 32 resident waves/CU = 8192` resident wave slots, with one half-refill for ragged causal retirement (`12288` tasks). Its locality term uses the kernel payload (`64 B` packed FP4 + `4 B` scale per K token) and the reported `4 MiB` L2 capacity; it still requires at least three independent KV streams before paying the 4x CTA/descriptor cost of the fine launch.

1. request enough wave tasks to cover the MI355X and prevent excessively long per-wave dependency chains;
2. cap parallelism before each wave gets too little serial K work to amortize Q/scale/weight loads and descriptor/dispatch cost;
3. quantize the budget to a four-wave coarse CTA so 256x4 and 64x1 represent the same useful work;
4. use independent 64x1 workgroups only where causal-tail waste and multi-sequence load balance can repay the extra CTA/descriptor cost;
5. retain 256x4 for long single-sequence prefill, where Q already fills the device and KV locality dominates.

For example, at about 16K K context the model chooses 20/16/12/8 wave tasks per row as Q grows, instead of keeping 20 wave tasks through Q=1536 and then abruptly falling to 8. For a long single sequence at Q=8192, one 256x4 CTA per row is selected: measured latency was 0.6915 ms, while two CTAs/eight wave tasks per row was slower at 0.7013 ms. This is the expected source-model result: row parallelism is already sufficient, so extra K splitting adds scheduling cost.

The full-batch metadata path and budget-chunked fallback carry the same `block_k`, `num_warps`, and wave budget. The scheduler still caps the grid by useful `(row, K-chunk)` work, so no row can be dropped and short contexts do not launch unnecessary CTAs.

### FP4 measurements

Eight-GPU kernel measurements comparing the formula with the old fixed 20-wave small-Q policy:

| Q | Selected waves/row | Mean latency delta |
| ---: | ---: | ---: |
| 768 | 16 | -0.079% |
| 1024 | 12 | -0.721% |
| 1280 | 12 | -0.728% |
| 1536 | 8 | -0.561% |

The formula deliberately keeps Q=768 at 16 waves even though one isolated sweep favored 12; there is no per-shape special case.

DSV4-Pro TP8/FP4, 64K shared prefix + 2K suffix, 80 requests, concurrency 8, 16 output tokens. Repeated optimized runs after the formula change:

| Seed | Baseline throughput | Optimized throughput | Baseline / optimized TTFT p50 |
| ---: | ---: | ---: | ---: |
| 201 | 9.23 req/s | 9.66 req/s | 592.83 / 556.50 ms |
| 203 | 9.00 req/s | 9.58 req/s | 619.95 / 568.53 ms |

An earlier three-seed source-driven run measured +5.09% throughput, -5.45% TTFT p50, -5.45% TTFT p90, and -3.76% E2EL p50 versus baseline. A first formula run showed system-level p90 outliers on two seeds; those outliers did not reproduce in the isolated reruns above.

## Full GSM8K MTP-3 accuracy

The final pushed implementation was validated on 8 x MI355X with DeepSeek-V4-Pro, TP8, FP8 KV cache, MTP-3, 3-shot GSM8K, `local-completions`, concurrency 65, and all 1319 samples.

| Metric | FP8 paged | FP4 source model |
| --- | ---: | ---: |
| flexible exact match | 0.949204 +/- 0.006048 | 0.947688 +/- 0.006133 |
| strict exact match | 0.949962 +/- 0.006005 | 0.948446 +/- 0.006091 |
| MTP acceptance rate | 0.661963 | 0.662920 |
| average tokens / forward | 2.985888 | 2.988760 |

Both arms completed 1319/1319 and both server error scans were empty. The FP8/FP4 score difference is 0.152 percentage points, much smaller than the approximately 0.61 percentage-point standard error.

## Validation

- `py_compile`: pass
- `git diff --check`: pass
- FP4/metadata/transfer CPU contracts: **60 passed**
- FP4 GPU differential: **1 passed**, with random FP4 Q/K bytes and scales, BF16 weights, non-contiguous physical pages, nonzero local starts, and 63/64/65, 127/128/129, 255/256/257 boundaries
- 64x1 and 256x4 valid logits are bit-exact (`rtol=0`, `atol=0`)
- FP8 default-on TP8/MTP smoke: pass, expected completion returned, MTP endpoint healthy, server error scan empty
- FP4 and FP8 full GSM8K MTP-3: pass, 1319/1319 each, empty error scans

The experimental grouped-wave, dummy-CTA early-exit, overlap-prefetch, causal-tail early-exit, and tiled split-major prototypes are not part of this PR. No AITER source change is required: the existing generic FlyDSL kernel already supports both equivalent launch shapes.


yhl-amd:
@JiaoliangYu This combined FP8 paged-indexer + FP4 grid-tuning PR is ready for review. The PR body now includes the kernel/E2E data and the full 1319-sample DSV4-Pro MTP-3 GSM8K validation at d552377c: FP8 paged 0.951478 / 66.006% acceptance and FP4 tuned 0.953753 / 66.334% acceptance, both above the 0.94 / 0.622 gates with empty server error scans. I also enabled ci:atom; the official native accuracy matrix is running now.

yhl-amd:
Follow-up pushed in c54d89e4: removed `_env_choice` and `ATOM_V4_FP8_INDEXER_PREFILL_BACKEND` entirely. Real FP8 indexer prefill now always uses the paged path; only the pre-pool model warmup dry run retains the legacy scorer. Validation: 72 CPU contract/integration tests passed, FP8 paged GPU differential 4/4 passed, Black passed, and a no-backend-env TP8/MTP-3 DSV4-Pro real-prefill smoke passed with an empty server error scan.

atom/model_ops/v4_kernels/fp4_mqa_schedule.py
```diff
@@ -0,0 +1,51 @@
...
+        raise ValueError(f"max_seq_len must be non-negative, got {max_seq_len}")
+    if block_k <= 0:
+        raise ValueError(f"block_k must be positive, got {block_k}")
+    if min_parallel_unit_num <= 0:
+        raise ValueError(
+            "min_parallel_unit_num must be positive, got " f"{min_parallel_unit_num}"
+        )
+
+    chunks_per_row = max(1, (max_seq_len + block_k - 1) // block_k)
+    if num_rows <= 1536:
```
yhl-amd: 用公式

PR #1986, opened 2026-08-21
perf(dsv4): tune FP4 prefill MQA grid

yhl-amd:
## Summary

- add a shape-specific persistent-grid selector for the DeepSeek-V4 FP4 prefill MQA logits kernel
- use more K splits for small/medium prefill Q while avoiding dummy CTAs beyond the useful row/chunk pairs
- apply the same policy to both the precomputed full-batch schedule and the budget-chunked fallback
- keep the fixed 512-CTA decode/CUDAGraph schedule unchanged

This is stacked on #1974 because it changes the paged-prefill path introduced there. The base should be changed to `main` after #1974 merges.

## Motivation

`compute_prefill_schedule` launches exactly `parallel_unit_num` CTAs. The previous prefill policy, `max(512, Q)`, preserves row coverage but folds too many context chunks onto each CTA for small Q. Conversely, an unrestricted larger grid can launch dummy CTAs after all useful `(row, K-chunk)` pairs are assigned.

The selector uses five splits per row through Q=1536, two for intermediate Q, and one for a full 16K-row prefill with at most 64 compressed-context chunks. It always covers every query row and caps the grid by the number of useful row/chunk pairs.

## Validation

Unit tests:

```text
python3 -m pytest -q \
  tests/test_fp4_mqa_prefill_schedule.py \
  tests/test_v4_indexer_paged_metadata.py \
  tests/test_deepseek_v4_transfer_regions.py

40 passed
```

Kernel-only A/B on MI355X/gfx950 used alternating launch order, 10-20 warmups, 50-100 samples per arm, schedule construction outside the timed region, and bit-exact checks on representative output cells:

- all 20 aggregate-Q shapes observed in the DSV4-Pro 64K-prefix + 2K-suffix serving logs improved on one GPU; frequency-weighted latency delta: `-3.83%`
- the five most frequent shapes were repeated on all eight GPUs; all 40/40 GPU/shape cells improved
- eight-GPU median deltas: Q512 `-11.02%`, Q768 `-4.83%`, Q1536 `-3.36%`, Q3584 `-2.17%`, Q5376 `-3.49%`

DSV4-Pro TP8/FP4 end-to-end results remain effectively flat. Across four same-seed paired 64K-prefix + 2K-suffix runs, request-throughput delta had median `-0.53%`, mean `+0.88%`, and range `[-1.30%, +5.87%]`; p50 end-to-end latency delta had median `-0.03%`. The FP4 MQA kernel accounts for only about `0.03%` of total benchmark wall time at this workload, so its microsecond-scale gain is below end-to-end noise.

One caveat remains: optimized p99 TTFT was `+3.94%` to `+7.11%` slower and p99 end-to-end latency was `+3.31%` to `+5.60%` slower across those four pairs. The kernel-only median and p90 both improved, so this does not reproduce in the changed kernel itself, but the serving-tail signal needs follow-up with reversed arm order and/or a targeted trace.

### DIFF
--- a/atom/model_ops/attentions/deepseek_v4_attn.py
+++ b/atom/model_ops/attentions/deepseek_v4_attn.py
@@ -101,6 +101,7 @@ from atom.model_ops.v4_kernels import (
     FP4_MQA_BLOCK_K,
     FP4_MQA_PARALLEL_UNIT_NUM,
     fp4_indexer_enabled,
+    fp4_mqa_prefill_parallel_unit_num,
     hca_compress_paged_offsets,
     write_v4_paged_decode_indices,
     write_v4_paged_prefill_indices,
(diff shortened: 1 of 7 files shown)


yhl-amd:
The FP4 grid commit is now included in #1974 so the FP8 and FP4 indexer-prefill optimizations are reviewed together. GitHub automatically marked this stacked PR merged when its base branch advanced to the same commit.

PR #1994, opened 2026-08-24
fix: fix PD prefix cache for dcp

Phi-C:
## Motivation

When dcp is enabled, replace block_size with hash_block_size under some pd conditions
## Submission Checklist

### DIFF
--- a/atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py
+++ b/atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py
@@ -386,7 +386,17 @@ class MooncakeConnectorScheduler(KVConnectorSchedulerBase):
             # prefix cache. Per-request state (including the SWA ring slot) is
             # not covered by a block-only delta, so it takes a full transfer.
             num_computed_blocks = 0
-            if not seq.has_per_req_cache and self.hash_block_size > 0:
+            remote_hash_block_size = params.get("hash_block_size")
+            if remote_hash_block_size != self.hash_block_size:
+                logger.warning(
+                    "PD incremental transfer disabled for req %s: producer "
+                    "hash_block_size=%r, consumer hash_block_size=%d; "
+                    "falling back to full transfer",
+                    seq.id,
+                    remote_hash_block_size,
+                    self.hash_block_size,
+                )
+            elif not seq.has_per_req_cache and self.hash_block_size > 0:
                 num_computed_blocks = seq.num_cached_tokens // self.hash_block_size
             params["num_computed_blocks"] = num_computed_blocks
             logger.info(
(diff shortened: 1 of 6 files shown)


atom/model_engine/block_manager.py
```diff
@@ -1095,7 +1095,13 @@ def register_received_prefix(self, seq: Sequence) -> int:
                         f"seq={seq.id} block={block_id} indexed={indexed_block_id}"
                     )
                 block.update(h, token_ids)
-        return num_full
+
+        seq.num_hashed_tokens = max(seq.num_hashed_tokens, num_full * hbs)
```
yitingw1, line 1099: `record_remote_store` reports `self.block_size` (physical) as `BlockStored.block_size`, but token_ids/block_hashes are grouped per hash_block_size (register_received_prefix's _hash_block_tokens) — **under DCP these differ**, so external KV-event consumers slicing token_ids by block_size to align with block_hashes will get it wrong; same pattern at the other two _make_block_stored call sites (block_manager.py:614, :1040), so worth fixing consistently rather than just here.

atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py
```diff
@@ -378,15 +379,15 @@ def update_state_after_alloc(self, seq: Sequence) -> None:
...
             params["do_remote_prefill"] = False
             params["local_slot_index"] = slot_index
             # PD incremental: skip leading blocks already in the decode node's
             # prefix cache. Per-request state (including the SWA ring slot) is
             # not covered by a block-only delta, so it takes a full transfer.
             num_computed_blocks = 0
-            if not seq.has_per_req_cache and self.block_size > 0:
-                num_computed_blocks = seq.num_cached_tokens // self.block_size
+            if not seq.has_per_req_cache and self.hash_block_size > 0:
+                num_computed_blocks = seq.num_cached_tokens // self.hash_block_size
```
yitingw1, line 400: The handshake exchanges tp_size/remote_pp_size but never hash_block_size, so nothing catches a P/D pair with mismatched decode_context_parallel_size; since num_computed_blocks is computed on one side and used as a block-table index on the other, a mismatch here silently misaligns the slice exactly like the bug this PR fixes — recommend exchanging hash_block_size at handshake and falling back to a full transfer (off=0) with a warning when it doesn't match.

yitingw1 (dismissed):

yitingw1:
Please add tests in tests/test_block_manager.py::TestRegisterReceivedPrefix for this modification.

atom/model_engine/block_manager.py
Phi-C: Good catch, thanks. I've updated it accordingly.

atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py
Phi-C: For heterogeneous PD, If prefill and decode have different hash_block_size, just falling back to full transfer may also lead to problems. But we can add this warning and fallback path here as you suggested for now.

ZhangLirong-amd (approved):

PR #1995, opened 2026-08-24
[ATOM] add persistent path for DCP GLM5.2

zhuyuhua-v:
## Summary

This PR improves native GLM-5.2 sparse MLA serving under Decode Context Parallelism (DCP) in the following areas:

- Adds an experimental replicated index-cache layout that removes cross-rank index top-k communication.
- Supports replicated index-cache transfer through LMCache and selected Mooncake/PD topologies.
- Optimizes the AG+RS DCP output merge by removing full-tensor layout copies and an unused global-LSE write.

The main MLA KV cache remains DCP-sharded. Replicating the index cache is opt-in through:

```bash
export ATOM_DCP_REPLICATE_INDEX_CACHE=1
```

## Motivation

GLM-5.2 uses sparse MLA with IndexShare. The previous DCP path had two major limitations:

1. Sparse compact indices vary by full IndexShare layer, so persistent MLA metadata built before the layer-specific top-k was invalid. The path therefore fell back to non-persistent attention. For FP8 Q/KV with a gathered GQA ratio of 64, AITER only supports the persistent kernel, forcing a 64-to-128 head padding fallback.

2. The DSA index cache was sharded using the same DCP layout as the main MLA KV cache. This required:
   - prefill index-K all-gather and de-interleave;
   - per-rank decode top-k;
   - candidate packing and all-gather;
   - a second top-k over the gathered candidates.

The replicated-index mode trades additional index-cache memory for local global-top-k computation and removes that index communication pipeline.

### Persistent sparse MLA under DCP

- Rebuilds `get_mla_metadata_v1` work/reduce metadata after each full IndexShare layer has produced its compact rank-local indices.
- Reuses the previous full layer's compact indices, indptr, and persistent work plan on shared IndexShare layers.
- Supports the same per-layer rebuild in sparse prefill, where each query token is represented as a virtual `q_len=1` row.
- Keeps unsupported paths on the existing non-persistent fallback.
- Routes GLM-5.2 TP8/DCP8 and TP4/DCP4 FP8 attention to the native persistent GQA64 path instead of padding 64 heads to 128.
- Preserves CUDA graph replay by reusing preallocated work, reduce, indptr, count, and compact-index buffers.
- Keeps MTP/speculative sparse DCP out of this persistent path.

### Experimental replicated index cache

- Adds `ATOM_DCP_REPLICATE_INDEX_CACHE`, disabled by default.
- Limits the replicated allocation to GLM-5.2 `glm_moe_dsa`, DCP > 1, no speculative decoding, and MLA page size 1.
- Allocates index-cache rows only for the 21 `"full"` IndexShare layers; the 57 `"shared"` layers continue reusing the preceding full layer's result.
- Expands each index-cache page from `block_size` rank-local tokens to `block_size × dcp_size` global tokens.
- Adds a separate `index_slot_mapping` while leaving the existing MLA `slot_mapping` owner-only and DCP-sharded.
- Makes every DCP rank write the complete index K cache.
- Uses global context lengths and the expanded page size for local decode scoring.
- Computes one stable global top-k locally on every rank.
- Skips the decode candidate exchange, candidate packing, index all-gather, and second top-k.
- Skips the prefill index-K all-gather and de-interleave.
- Retains the existing ownership filter and order-preserving compaction so each rank attends only to the selected tokens stored in its DCP-sharded MLA KV cache.
- Handles dummy warmup, padded CUDA graph batches, TBO ubatches, cached prefill, and prefix-cache block lifetimes.

The selected top-k set must be deterministic across ranks, but the 2,048 indices do not need to be sorted by token position. The ownership filter preserves the incoming relative order and emits a compact, hole-free local index region.

### DCP output-merge optimization

For the AG+RS backend:

- Adds a correction kernel that writes directly to rank-major ReduceScatter input layout.
- Removes the full-tensor `movedim(...).contiguous()` copy before ReduceScatter.
- Makes ReduceScatter return contiguous `[batch, local_heads, value_dim]` output directly, removing the post-collective layout copy.
- Avoids allocating and writing global LSE in the production merge path because the caller does not consume it.
- Preserves NaN/+inf LSE sanitization and zeroes empty-rank contributions before reduction.

This does not modify the AITER MLA kernel.

### LMCache and PD compatibility

- Extends the raw-byte cache payload to include:
  - all local MLA KV rows;
  - the 21 compact, DCP-expanded full-index rows.
- Bumps the LMCache PAGE namespace to version 3.
- Adds the replicated-index flag, effective index-cache dtype, and exact `indexer_types` schedule to the namespace fingerprint.
- Prevents replicated and sharded layouts, or different IndexShare schedules, from reusing incompatible LMCache objects.
- Floors full-prompt LMCache hits to complete chunks after reserving the final token for logits recomputation. For example, an 8,192-token full hit now loads 7,936 tokens and recomputes the final 256-token chunk instead of attempting an invalid 8,191-token load and falling back to full prefill.
- Validates the expanded physical index-page width before exposing transfer regions.

Supported replicated-index transfer configurations are:

- standalone `lmcache_offload`;
- standalone Mooncake producer or consumer;
- `multi` containing exactly Mooncake producer plus LMCache offload.

Other transfer topologies fail fast.

### Documentation and tests

- Extends the context-parallel guide with:
  - sparse persistent metadata rebuild behavior;
  - replicated index-cache layout and runtime flow;
  - memory/capacity trade-offs;
  - LMCache and PD layout requirements;
  - validation and performance results.
- Adds regression coverage for:
  - DCP LSE merge correctness;
  - empty-rank/NaN handling;
  - direct ReduceScatter layout;
  - LMCache PAGE namespace separation;
  - full-prompt chunk alignment;
  - DCP virtual-block behavior;
  - compact/expanded index-cache byte accounting.

## Memory trade-off

For GLM-5.2 with 78 MLA layers, 21 full index layers, block size 16, FP8 index width 144, and DCP8:

- Sharded index bytes per scheduler block:
  `78 × 16 × 144 = 179,712`
- Replicated index bytes per scheduler block:
  `21 × 16 × 8 × 144 = 387,072`
- Total MLA+index bytes per block increase from 898,560 to 1,105,920, or approximately 23.1%.
- A fixed KV memory budget therefore holds approximately 18.8% fewer blocks.

The replicated layout is intentionally experimental and disabled by default because this capacity cost is workload-dependent.

### Unit and regression tests

Targeted DCP and LMCache suites completed successfully, including merge correctness, sparse filtering, persistent metadata, top-k, virtual blocks, namespace isolation, and LMCache connector behavior.

A combined targeted run completed with:

```text
396 passed
```

### TP8/DCP8 runtime

Validated on MI355/gfx950 with GLM-5.2-MXFP4, FP8 KV, and no MTP:

- CUDA graph capture completed.
- Short decode completed.
- 8K-context prefill/decode completed.
- Concurrency-16 serving completed without traceback or HIP errors.
- Replicated index-cache startup reported 21 full IndexShare rows on all eight ranks.

### Accuracy

Persistent sparse DCP without replicated index cache was validated on both TP8/DCP8 and TP4/DCP4 with the full 1,319-sample GSM8K 5-shot set:

```text
flexible-extract: 0.9689
strict-match:     0.9666
```

Using the same local chat evaluator for a replicated-versus-sharded TP8/DCP8 comparison:

```text
replicated index: flexible 0.9651 / strict 0.9621
sharded index:    flexible 0.9697 / strict 0.9697
```

The difference was approximately 0.7/1.1 combined standard errors and was not statistically significant in that run.

A 20-sample LMCache restore smoke completed 160/160 worker loads without fallback, traceback, or HIP error and retained flexible-extract accuracy at 0.95.

### LMCache restore

For an 8,192-token TP8/DCP8 prompt:

- every rank stored 67.5 MiB of MLA+replicated-index payload;
- every rank restored the aligned 7,936-token prefix with `status=ok`;
- the remaining 256 tokens were recomputed;
- TTFT dropped from 5.87 seconds cold to 0.46 seconds on restore.

### Performance

On an 8,192-input/128-output concurrency-16 benchmark:

```text
output throughput: 172.34 -> 179.77 token/s  (+4.3%)
mean TPOT:          53.14 -> 49.30 ms       (-7.2%)
median ITL:         23.15 -> 19.43 ms       (-16.1%)
```

The AG+RS correction/layout microbenchmark at H=64, D=512, DCP8 measured:

```text
batch 1:    15.5 -> 13.0 us
batch 16:  164.4 -> 13.2 us
8192 rows: 866.1 -> 417.2 us
```

Replicated-index performance is workload-dependent. In the long-context Agentic C16 profile it was slower than the sharded layout:

```text
TP8/DCP8 per-GPU throughput: 5,446.64 -> 4,945.76 token/s
TP4/DCP4 per-GPU throughput: 9,930.62 -> 9,227.99 token/s
```

This is why the feature remains opt-in.

## Known limitations

- Replicated index cache is native ATOM only.
- Sparse DCP + MTP/speculative decoding is not supported.
- Requires MLA page size 1.
- PCP, PP > 1, and RapidServe are not supported in replicated-index mode.
- Only the explicitly validated LMCache/Mooncake transfer topologies are accepted.
- The main MLA KV cache remains DCP-sharded, so query/output DCP communication is still required.
- Replicated index cache reduces available KV capacity.
- Long-context quality requires continued validation. A recent unique-prefix DCP4 replicated-index needle run retrieved 5/6 needles, while its pure TP4 control retrieved 6/6. The feature should remain experimental until this is fully characterized.

### DIFF
--- a/atom/kv_transfer/offload/config.py
+++ b/atom/kv_transfer/offload/config.py
@@ -23,10 +23,12 @@ from typing import Any
 
 import torch
 
-# Version 2 adds explicit DSV4 KV/index dimensions and DCP virtual-block byte
-# mapping to the PAGE identity. Keeping it separate prevents version-1 objects
-# from being reused after the layout correction.
-PAGE_LAYOUT_VERSION = 2
+from atom.utils import envs
+
+# Version 3 adds the GLM replicated-index layout and IndexShare schedule to the
+# PAGE identity. Its compact layer axis and DCP-expanded index pages are byte-
+# incompatible with the sharded-index v2 layout.
+PAGE_LAYOUT_VERSION = 3
 _PAGE_FINGERPRINT_BYTES = 16
 _OFFLOAD_LAYOUT_ALIASES = {
     "hybrid": "hybrid",
@@ -49,10 +51,12 @@ _HF_PAGE_FIELDS = (
     "qk_rope_head_dim",
     "compress_ratios",
     "indexer_dtype",
+    "indexer_types",
 )
 _HF_INTEGER_GEOMETRY_FIELDS = frozenset(_HF_PAGE_FIELDS) - {
     "compress_ratios",
     "indexer_dtype",
+    "indexer_types",
 }
 
 logger = logging.getLogger("atom")
@@ -198,6 +202,12 @@ def build_page_namespace(
                 else getattr(config, "decode_context_parallel_size", 1)
             ),
             minimum=1,
+        ),
+        "replicated_index_cache": bool(
+            envs.ATOM_DCP_REPLICATE_INDEX_CACHE
+            and getattr(hf, "model_type", None) == "glm_moe_dsa"
+            and (getattr(config, "decode_context_parallel_size", 1) or 1) > 1
+            and getattr(config, "speculative_config", None) is None
         ),
         "hf_geometry": _stable_hf_geometry(hf),
         "speculative_config": _stable_config_value(
--- a/atom/model_ops/attentions/aiter_mla.py
+++ b/atom/model_ops/attentions/aiter_mla.py
@@ -17,6 +17,7 @@ from aiter import (
 
 from atom.distributed.dcp_utils import (
     dcp_persistent_supported,
+    dcp_replicated_index_cache_enabled,
     get_dcp_rank,
     get_dcp_world_size,
 )
@@ -171,6 +172,57 @@ class AiterMLAMetadataBuilder(CommonAttentionBuilder):
 
         self.dcp_world_size = get_dcp_world_size()
         self.dcp_rank = get_dcp_rank()
+        self.replicate_index_cache = dcp_replicated_index_cache_enabled(config)
+        if envs.ATOM_DCP_REPLICATE_INDEX_CACHE:
+            unsupported = []
+            if getattr(hf_config, "model_type", None) != "glm_moe_dsa":
+                unsupported.append("model_type must be glm_moe_dsa")
+            if self.dcp_world_size <= 1:
+                unsupported.append("decode context parallel size must be > 1")
+            if config.speculative_config is not None:
+                unsupported.append("MTP/speculative decoding is not supported")
+            if self.block_size != 1:
+                unsupported.append("MLA page size must be 1")
+            if get_pcp_world_size() > 1:
+                unsupported.append("PCP is not supported")
+            if getattr(config, "pipeline_parallel_size", 1) > 1:
+                unsupported.append("pipeline parallelism is not supported")
+            if getattr(config, "kv_transfer_config", None):
+                unsupported.append("KV transfer/LMCache/PD is not supported")
+            if getattr(config, "enable_rapidserve", False):
+                unsupported.append("RapidServe disaggregation is not supported")
+            if unsupported:
+                raise ValueError(
+                    "ATOM_DCP_REPLICATE_INDEX_CACHE=1 is unsupported: "
+                    + "; ".join(unsupported)
+                )
+
+        self.full_index_layer_ids: tuple[int, ...] = ()
+        self.index_layer_to_cache_row: dict[int, int] = {}
+        if self.replicate_index_cache:
+            indexer_types = getattr(hf_config, "indexer_types", None)
+            if not indexer_types or len(indexer_types) != hf_config.num_hidden_layers:
+                raise ValueError(
+                    "Replicated GLM index cache requires one indexer_types entry "
+                    "per target layer"
+                )
+            self.full_index_layer_ids = tuple(
+                layer_id
+                for layer_id, indexer_type in enumerate(indexer_types)
+                if indexer_type == "full"
+            )
+            if not self.full_index_layer_ids:
+                raise ValueError("Replicated GLM index cache found no full IndexShare layers")
+            self.index_layer_to_cache_row = {
+                layer_id: row
+                for row, layer_id in enumerate(self.full_index_layer_ids)
+            }
+            logger.info(
+                "Replicating GLM index cache for %d full IndexShare layers "
+                "across %d DCP ranks",
+                len(self.full_index_layer_ids),
+                self.dcp_world_size,
+            )
 
         # DCP decode all-gathers Q on the head dim, so the head count reaching
         # mla_decode_fwd (and thus the persistent decode metadata) is the padded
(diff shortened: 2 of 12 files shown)


docs/context_parallel_guide.md
```diff
@@ -520,30 +520,124 @@ hit it.
...
-the module pads to and the width the persistent descriptors are planned for
-cannot disagree.
-
-**GLM-5.2 is the model that lands on it**: 64 query heads at `-tp 8` is 8 per
-rank, so `-dcp 8` gathers exactly 64 — sparse, hence non-persistent, hence padded
-to 128. Half the kernel's heads are padding in that configuration; making it 64
-means giving sparse + DCP a persistent path first.
-
-**Validated** (ATOM server, 8×MI355 gfx950, `-tp 8 -dcp 8`, fp8 KV, full 1319
+### Sparse DCP persistent attention and gqa=64
```
yitingw1, line 523: Please be more concise for user guide. For example, no need validation results.

atom/model_ops/attentions/aiter_mla.py
```diff
@@ -1764,6 +1980,13 @@ def prepare_decode(self, batch: ScheduledBatch, bs: int):
         var["positions"].np[:sum_scheduled_tokens] = positions
         var["context_lens"].np[:scheduled_bs] = context_lens
         var["context_lens"].np[scheduled_bs:bs] = 0
+        if self.replicate_index_cache:
+            var["index_slot_mapping"].np[:bs] = 0
+            if not batch.is_dummy_run:
+                var["index_slot_mapping"].np[:scheduled_bs] = [
+                    self._replicated_index_slot(block_table, int(seq_len) - 1)
+                    for block_table, seq_len in zip(block_tables, context_lens)
+                ]
```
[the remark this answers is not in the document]
yitingw1: Please fix this bug and show GSM8K 20-shot results.

yitingw1 (changes requested):

atom/model_ops/attentions/aiter_mla.py
```diff
@@ -332,6 +466,14 @@ def __init__(self, model_runner):
         mla_metadata["kv_last_page_lens"].cpu.fill_(1)
         mla_metadata["kv_last_page_lens"].copy_to_gpu()
         if self.is_sparse:
+            if self.replicate_index_cache:
+                mla_metadata["index_slot_mapping"] = CpuGpuBuffer(
+                    self.max_num_batched_tokens, **i64_kwargs
+                )
+                # CUDAGraph capture may execute cache writes before the first
+                # scheduled batch. Point those dummy writes at a valid slot.
+                mla_metadata["index_slot_mapping"].np.fill(0)
```
yitingw1, line 475: Here too.

atom/kv_transfer/offload/config.py
```diff
@@ -200,6 +204,12 @@ def build_page_namespace(
             ),
             minimum=1,
         ),
+        "replicated_index_cache": bool(
```
yitingw1, line 207: Maybe we can reuse `dcp_replicated_index_cache_enabled()` form dcp_utils.py?

PR #1996, opened 2026-08-24
[atom+sgl]fix(sglang): align DeepSeek-V4 bridge with geometry/dest_rows API

zhangxinyuanliuhengyu:
## Summary

Fix SGLang DeepSeek-V4 bridge for the post-#1771 `geometry` / `dest_rows` API, and use per-layer proxy pool geometry so decode row addressing matches SGLang's independent cache planes (not ATOM's interleaved `UnifiedPoolGeometry`).

Also:
- Register `DeepseekV4ForCausalLMNextN` / `deepseek_v4_mtp` so MTP draft keeps the V4 metadata path
- MI308 catalog: FP8 weights with BF16 KV (`gfx942` has no DSV4 FP8 KV decode kernels)

## Test plan

- [x] `ci:sglang` DeepSeek-V4 Flash / Pro / MTP on MI355


ZhiweiYan-96:
will this PR introduce any layout change for kv cache pool in sglang? The former design is that we use the same kv pool layout as atom, while implementing the minimum API requirement of SGLang w/o any layout change.

zhangxinyuanliuhengyu:
@ZhiweiYan-96 No change to the SGLang KV-pool architecture or allocator.

The proxy is still independent per-layer planes (the original SGLang-minimum design), not native ATOM’s interleaved `UnifiedPoolGeometry`. Allocator, page size, and the SWA API are unchanged; arena size is unchanged.

What this PR does is speak the post-#1771 `geometry` / `dest_rows` kernel contract against those existing per-layer views. Feeding ATOM’s interleaved row math into SGLang’s planes mis-addressed `dest_rows` and collapsed GSM8K on MI355 TP8. `_PerLayerPoolGeometry` is only an addressing adapter: `ring_start = num_blocks * block_rows(ratio)`, window contiguous after that layer’s own compressed rows.

The only physical tweak is the **within-layer** unified order: `[compressed | SWA]` so `unified[ring_start]` is the real window (it used to be SWA-first). That is not adopting ATOM’s interleaved layout.

ZhiweiYan-96 (dismissed):

zhuyuhua-v (dismissed):

zhuyuhua-v:
### P0-1 — `csa_n_committed_per_token` is staged without padding to `t_pad`

In `build_atom_v4_decode_graph_metadata_from_sglang`:

```python
if total:
    visible_np = visible_csa(pos_np).astype(np.int32)   # len == total
else:
    visible_np = np.zeros(0, dtype=np.int32)
md.csa_n_committed_per_token = bufs.stage(
    bufs.csa_n_committed_per_token, visible_np, t_pad   # n == t_pad
)
```

`stage()` does `buf.np[:n] = arr_np[:n]`. When `t_pad > total` (`t_pad = max(bs, total)`), the RHS is `(total,)` and the LHS is `(t_pad,)` — a hard `ValueError: could not broadcast`. Every neighbouring line in this same function handles `t_pad > total` explicitly (`batch_pad = np.full(t_pad, -1); batch_pad[:total] = batch_np`, and `out[total+1:] = out[total]` in `indptr()`), so the branch is clearly considered reachable.

The vLLM bridge writes it this way:

```python
visible_np = np.zeros(T_pad, dtype=np.int32)
if total:
    visible_np[:total] = visible_csa(pos_np).astype(np.int32)
md.csa_n_committed_per_token = bufs.stage(bufs.csa_n_committed_per_token, visible_np)
```

Zero-padding is also required for correctness even if the lengths happen to match: the only consumer is `deepgemm_fp8_paged_mqa_logits(..., attn_md.csa_n_committed_per_token[:total_tokens], attn_md.block_tables_per_token[:total_tokens], ...)`, which reads the whole `[:total_tokens]` range and **does not gate on the `batch_id` sentinel**. Leaving stale counts in the pad tail, combined with `safe_batch_ids.clamp_min(0)` pointing the pad rows at batch 0's block table, means scoring pad rows against another request's blocks.

Please mirror the vLLM form.

### P0-2 — `_bind_v4_state_slots` aliases `state_slot_in` to `state_slot_out`

```python
def _bind_v4_state_slots(md) -> None:
    slots = md.state_slot_mapping
    md.state_slot_in = slots
    md.state_slot_out = slots
```

`AttentionMetaData_DSV4.state_slot_in` documents the opposite:

> Equal in value to `state_slot_out` except on the one forward carrying a state fork, but **always its OWN buffer**: the decode path replays a CUDAGraph with this pointer baked in at capture, so **aliasing it to `state_slot_out` when nothing forked would make the distinction unreplayable the moment one does**.

The vLLM bridge honours this: two separate `CpuGpuBuffer`s (`self.state_slot_in = i32(S)` / `self.state_slot_out = i32(S)`) on the graph path, and an explicit `md.state_slot_in = md.state_slot_out.clone()` on the eager path — that `.clone()` is deliberate.

Here all three fields are one tensor, on the path that gets captured into a CUDA graph, and `maybe_compressors_async(x, plan, state_slot_in, state_slot_out, block_tables)` bakes both pointers in. Any future state fork (MTP rejection rollback, DSpark) becomes silently unrepresentable rather than failing loudly.

Suggest: add `state_slot_in` / `state_slot_out` buffers to `_V4SGLangDecodeGraphBuffers` and `_V4SGLangVerifyGraphBuffers` and stage the same `slot_arr` into both.

### P1-3 — `_geometry_serves_ratio` is always `True`; the new gating is dead code and removes an upstream guard

```python
@property
def classes(self) -> tuple[int, ...]:
    return (DENSE_RATIO, CSA_RATIO, HCA_RATIO)
```

Hard-coded — it never looks at `proxy_pool.stage_ratios`. So `has_csa` / `has_hca` are always `True`, and the commit *"fix(sglang): gate decode index buffers by pool geometry"* gates nothing.

Worse, `served_window_params(geometry)` in `paged_decode_indices.py` is explicit about why it is a dict:

> The classes that are ON index `served` directly, on purpose. A caller asking for rows of a class no layer belongs to then gets a **KeyError** here instead of a plausible row written from borrowed parameters — and unlike an `assert`, that holds under `python -O`.

`UnifiedPoolGeometry.classes` is `dict[int, ClassLayout]` and raises on an absent class. Replacing it with an always-full tuple removes that guard: a DSV4-Flash or NextN-draft config that lacks a class will now silently get a plausible-looking wrong row instead of failing fast.

Suggest deriving `classes` from `proxy_pool.stage_ratios` (`tuple(sorted(set(stage_ratios)))`) and adding a unit test asserting `_geometry_serves_ratio` returns `False` for an absent class.

### P1-4 — `envelope_rows` now means three different things

| Site | Value | Consumer |
|---|---|---|
| `md.envelope_rows = _CSA_ENVELOPE_ROWS` | `256 // 4 = 64` | `csa_translate_pack`: `row = block * envelope_rows + slot_in_block` |
| `_PerLayerPoolGeometry.envelope_rows` | `256 // 128 = 2` | `write_v4_decode_hca_compress_tail`, `write_v4_paged_prefill_indices` |
| `UnifiedPoolGeometry.envelope_rows` | sum across all layers | native ATOM |

Both values are individually correct for SGLang's per-layer planes, but carrying two contracts on one attribute name is fragile — the vLLM bridge literally contains `md.envelope_rows = int(meta_params.geometry.envelope_rows)`, so anyone "unifying" the two will break this silently. The `AttentionMetaData_DSV4.envelope_rows` docstring ("Rows one V4 block takes across every layer of the pool") is now false for this bridge.

The class already has `block_rows(ratio)`. Please say what you mean:

```python
md.envelope_rows = geometry.block_rows(CSA_RATIO)
...
envelope_rows=geometry.block_rows(HCA_RATIO),
```

`_CSA_ENVELOPE_ROWS` can then go away. Also note that `test_proxy_metadata_uses_per_layer_csa_block_stride` currently *pins* the ambiguity (`metadata.envelope_rows == BLOCK//4` **and** `geometry.envelope_rows == BLOCK//128`) rather than removing it — it should assert the two `block_rows` calls instead.

### P1-5 — `_PerLayerPoolGeometry` is an unchecked duck type

Both `write_v4_paged_decode_indices` and `write_v4_paged_prefill_indices` annotate `geometry: UnifiedPoolGeometry`, but the object passed here:

- exposes `classes` as `tuple[int, ...]` where the original is `dict[int, ClassLayout]` (`in` happens to work; `.values()` or `[ratio]` would not);
- has no `physical_slot`, `compress_index`, `layer_base_row`, `with_capacity`, `slot_rows`, …

The `v4_pool_geometry` module docstring calls itself the *"Sole owner of the arithmetic: allocation, view binding, the index kernels and their references all take their offsets from here rather than each re-deriving them."* This PR adds a parallel implementation. Two ways out:

1. Declare a `PoolGeometry` Protocol in `v4_pool_geometry.py` (`classes` / `envelope_rows` / `block_rows` / `window_params`), type the kernels against it, and let both implementations be checked in one place; or
2. Better: build a real `UnifiedPoolGeometry(ratios=[ratio], num_blocks=N, num_slots=S, ring_slots=cs, block_size=256, arena_rows=0)` per compress class. A single-layer pool's `envelope_rows` *is* the per-layer `block_size // ratio`, and you get `window_params` / `physical_slot` / `compress_index` for free. SGLang's `[blocks | slots]` layout already grows the same way the native plane does.

### P1-6 — Hard-coded constants, inconsistent with the ones this PR introduces

- `_V4_SWA_DEST_RATIOS = (0, 4, 128)` → `(DENSE_RATIO, CSA_RATIO, HCA_RATIO)` (the `classes` property right above already uses the named constants)
- `hca_rows_per_block=ATOM_DEEPSEEK_V4_BLOCK_SIZE // 128` in both the verify and prefill paths → `geometry.block_rows(HCA_RATIO)`
- `_geometry_serves_ratio(geometry, 4)` / `(geometry, 128)` → `CSA_RATIO` / `HCA_RATIO`
- `_slice_views` now uses `ratio in (CSA_RATIO, HCA_RATIO)` in the bf16 branch but still `ratio in (4, 128)` in the fp8 branch — two styles in one function

### P1-7 — `block_tables_per_token` / `csa_n_committed_per_token` are built unconditionally

`block_tables_per_token`'s docstring says "**FP8 decode only**; None under the FP4 indexer", and both fields are only consumed on the fp8 decode scoring path. Building them unconditionally costs, per decode step: a host-side `visible_csa(pos_np)`, a `clamp_min(0).long()` (two temporaries), and an `index_select` — all of which is pure waste under the BF16 KV configuration this same PR switches MI308 to.

The vLLM bridge keeps both inside the fp8 branch. Suggest `if proxy_pool.use_fp8_kv:`.

### P1-8 — `batch_id_per_q_token_cpu` has a different length from its GPU twin

On the decode-graph path, `md.batch_id_per_q_token = stage(batch_pad, t_pad)` is length `t_pad` with `-1` sentinels, while `md.batch_id_per_q_token_cpu = batch_np` is length `total` with none. A CPU/GPU pair under the same name with different lengths *and* different semantics is a trap for any downstream `len(...)` or `[:T]`. Either store `batch_pad`, or drop the field for now — I don't see a consumer.

zhangxinyuanliuhengyu:
Thanks @zhuyuhua-v — went through the eight notes against the current SGLang bridge and vLLM/native contracts. Summary of what landed in 3873f5c2:

**Addressed**
- **P0-1** decode-graph `csa_n_committed_per_token` is now `zeros(t_pad)` with the real region filled in `[:total]`, same pad contract as vLLM (idle/graph tail no longer `stage()`-crashes, and the pad is 0 so MQA logits do not score batch 0).
- **P0-2** decode/verify graph buffers now have distinct `state_slot_in` / `state_slot_out` addresses; eager path clones. Values are still identical today (SGLang does not fork state), but HIP graph replay can keep two pointers.
- **P1-3** `_PerLayerPoolGeometry.classes` is derived from `stage_ratios`; missing CSA is `False` and `window_params(4)` is `KeyError`. Covered by `test_proxy_geometry_omits_absent_stage_ratios`.
- **P1-4** `md.envelope_rows` is `geometry.block_rows(CSA_RATIO)`; HCA helpers take `block_rows(HCA_RATIO)`. Page size is 256 but ATOM `BLOCK_SIZE` is 128, so CSA/HCA strides are 32/1 not 64/2.
- **P1-6** dest-row / plan / slice ratios use `DENSE_RATIO`/`CSA_RATIO`/`HCA_RATIO`.
- **P1-8** dropped unused `batch_id_per_q_token_cpu` (CPU/GPU length mismatch; no remaining plugin reader).

**Not changing in this PR**
- **P1-5** duck-typed `_PerLayerPoolGeometry` stays. Native `UnifiedPoolGeometry` is interleaved slot-major; SGLang proxy is independent per-layer `[compressed | SWA]`. Building `UnifiedPoolGeometry(ratios=[ratio], ...)` would reintroduce the dest_rows mis-address that collapsed MI355 TP8 GSM8K. A Protocol is a reasonable follow-up; not swapping the adapter here.
- **P1-7** do not gate `csa_n_committed_per_token` / `block_tables_per_token` on `use_fp8_kv`. The consumer is the FP8 **indexer** (`deepgemm_fp8_paged_mqa_logits`), not KV dtype. MI308 BF16 KV still runs the FP8 indexer. Current vLLM also fills those on every decode and only puts `qo_indptr` behind the fp8 branch.

Local check: geometry tests 30 passed in `rocm/atom-dev:sglang-*`; Flash-FP8 TP8 smoke on :30111 returned 200 with generated tokens. CPU unit suite 4865 passed; 3 failures are `hipHostMalloc` with `HIP_VISIBLE_DEVICES=""` (no device), unrelated.

PR #2000, opened 2026-08-24
fix(kv-transfer): synchronize P/D sends with offload save completions

Jasen2201:
## Submission Checklist

### DIFF
--- a/atom/kv_transfer/disaggregation/__init__.py
+++ b/atom/kv_transfer/disaggregation/__init__.py
@@ -21,7 +21,6 @@ from atom.kv_transfer.disaggregation.types import (
     LoadOperationId,
     ReqMeta,
     SaveOperationId,
-    completion_req_key,
 )
 
 __all__ = [
@@ -35,5 +34,4 @@ __all__ = [
     "LoadOperationId",
     "ReqMeta",
     "SaveOperationId",
-    "completion_req_key",
 ]
--- a/atom/kv_transfer/disaggregation/__init__.py
+++ b/atom/kv_transfer/disaggregation/__init__.py
@@ -21,6 +21,7 @@ from atom.kv_transfer.disaggregation.types import (
     LoadOperationId,
     ReqMeta,
     SaveOperationId,
+    completion_req_key,
 )
 
 __all__ = [
@@ -34,4 +35,5 @@ __all__ = [
     "LoadOperationId",
     "ReqMeta",
     "SaveOperationId",
+    "completion_req_key",
 ]
--- a/atom/kv_transfer/disaggregation/multi/multi_connector.py
+++ b/atom/kv_transfer/disaggregation/multi/multi_connector.py
@@ -63,7 +63,11 @@ from atom.kv_transfer.disaggregation.base import (
     KVConnectorBase,
     KVConnectorSchedulerBase,
 )
-from atom.kv_transfer.disaggregation.types import ConnectorMetadata, KVConnectorOutput
+from atom.kv_transfer.disaggregation.types import (
+    ConnectorMetadata,
+    KVConnectorOutput,
+    completion_req_key,
+)
 
 logger = logging.getLogger("atom")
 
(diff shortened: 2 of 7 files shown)


yhl-amd (commented):
I found two correctness gaps that should be addressed before merge:

1. **PP can release the send after only the first save generation reaches quorum.**

   The worker now correctly waits for all locally known `SaveOperationId` values, but the PP head stores only one `_held_sending[request_key]`. In `_ingest_and_release()`, the first globally-complete save operation for that request pops the held send.

   Repro with the existing `test_pp_kv_status.py` fixtures:
   - stage 0 emits `finished_sending={9}` together with save generations 2 and 3;
   - stage 1 reports generation 2 in poll 1 and generation 3 in poll 2;
   - after poll 1, `released_sending == {9}` and `_held_sending == {}`, even though generation 3 has not reached PP quorum.

   That can free the producer KV blocks while a downstream stage is still saving from them. The PP head needs to retain the exact pending save-operation set associated with each held send and release only after that set is empty. Please add the staggered multi-generation PP regression test above.

2. **`MultiConnector.get_finished()` drops `connector_completions`.**

   DSV4 reports SLOT/sidecar save success or failure through `KVConnectorOutput.connector_completions`, but the MultiConnector merge only collects recv/send/load/save fields. A sidecar completion entering a sub-connector therefore produces an empty MultiConnector output. The new scheduler-side `process_completions()` fan-out cannot repair this because it never receives the event.

   Consequently, `_sidecar_save_inflight` is not cleared, the sidecar commit/failure state is not applied, and `has_pending_work()` can remain true indefinitely. This is especially relevant now that this PR explicitly treats `slot_save_spec` as a paired save. Please preserve and aggregate connector-owned completions through MultiConnector (and through the PP path where applicable), with a regression test covering DSV4 sidecar completion delivery.

The request/operation ID normalization and worker-side per-generation tracking are otherwise in the right direction, and the current focused tests pass, but they do not cover these two completion paths.

Jasen2201:
> I found two correctness gaps that should be addressed before merge:
> 
> 1. **PP can release the send after only the first save generation reaches quorum.**
>    The worker now correctly waits for all locally known `SaveOperationId` values, but the PP head stores only one `_held_sending[request_key]`. In `_ingest_and_release()`, the first globally-complete save operation for that request pops the held send.
>    Repro with the existing `test_pp_kv_status.py` fixtures:
>    
>    * stage 0 emits `finished_sending={9}` together with save generations 2 and 3;
>    * stage 1 reports generation 2 in poll 1 and generation 3 in poll 2;
>    * after poll 1, `released_sending == {9}` and `_held_sending == {}`, even though generation 3 has not reached PP quorum.
>    
>    That can free the producer KV blocks while a downstream stage is still saving from them. The PP head needs to retain the exact pending save-operation set associated with each held send and release only after that set is empty. Please add the staggered multi-generation PP regression test above.
> 2. **`MultiConnector.get_finished()` drops `connector_completions`.**
>    DSV4 reports SLOT/sidecar save success or failure through `KVConnectorOutput.connector_completions`, but the MultiConnector merge only collects recv/send/load/save fields. A sidecar completion entering a sub-connector therefore produces an empty MultiConnector output. The new scheduler-side `process_completions()` fan-out cannot repair this because it never receives the event.
>    Consequently, `_sidecar_save_inflight` is not cleared, the sidecar commit/failure state is not applied, and `has_pending_work()` can remain true indefinitely. This is especially relevant now that this PR explicitly treats `slot_save_spec` as a paired save. Please preserve and aggregate connector-owned completions through MultiConnector (and through the PP path where applicable), with a regression test covering DSV4 sidecar completion delivery.
> 
> The request/operation ID normalization and worker-side per-generation tracking are otherwise in the right direction, and the current focused tests pass, but they do not cover these two completion paths.

Thanks your suggestion. Both confirmed and fixed.

1. PP send released after only the first save generation

The hold now stores the exact pending set — (send_id, {SaveOperationId...}) — and releases only once every generation has reached PP-wide quorum. The set is complete when installed: MultiConnector flushes the send together with all accumulated generations, and mooncake sends only after the request's last chunk.

2. get_finished() drops connector_completions

Confirmed — the merge unions five of six fields. Now unioned alongside the others; both return paths build the same output, so pass-through and pairing are covered at once. Completions are a connector-owned channel, not a block-lifetime signal, so the pairing path passes them through rather than holding them with the send.

yhl-amd (approved):
LGTM

valarLip (approved):

PR #2004, opened 2026-08-24
[fix](dspark): fix kimi k3 dspark

PerryZhang01:
## Motivation

Kimi-K3 with DSpark crashes during PIECEWISE cudagraph capture: HIP error: operation not permitted when stream is capturing. This PR  marks the KDA mixer with @eager_break_during_capture so it runs eagerly instead of being recorded into the graph.

### DIFF
--- a/.github/scripts/atom_oot_test.sh
+++ b/.github/scripts/atom_oot_test.sh
@@ -289,8 +289,8 @@ PY
     --compilation-config '{"cudagraph_mode": "FULL_AND_PIECEWISE"}' \
     --trust-remote-code \
     --kv-cache-dtype fp8 \
-    "${extra_arg_array[@]}" \
     --no-enable-prefix-caching \
+    "${extra_arg_array[@]}" \
     > "${VLLM_LOG_FILE}" 2>&1 &
   echo $! > "${VLLM_PID_FILE}"
   echo "Server PID: $(cat "${VLLM_PID_FILE}")"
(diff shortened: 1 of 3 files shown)


whx-sjtu (dismissed):

atom/models/kimi_k3.py
```diff
@@ -1048,9 +1054,30 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
-            hidden_states, hidden_states_scale, self.layer_name
+        output = torch.empty(
+            hidden_states.shape, dtype=torch.bfloat16, device=hidden_states.device
         )
+        torch.ops.aiter.kda_attention_with_output(
+            hidden_states, hidden_states_scale, self.layer_name, output
+        )
+        return output
+
+    @_eager_break_during_capture
```
ZhangLirong-amd: why we need this vllm api, can we use @mark_spliting_op or @piecewise_core ?

PerryZhang01: This is vllm's implementation for PIECEWISE KDA, we will move it into plugin.

PerryZhang01: mark_spliting_op just break the fx graph, is_current_stream_capturing() is still True on KDA. piecewise_core also give the decorated function its own cudagraph. if graph capture in kda, it will crash because vllm metadata num_spec will be zero in dummy_run and native atom has no piecewise for prefill.

valarLip (approved):

Issue #14, opened 2025-11-27
More details about requirements in README.md


AlexHe99:
The Requirements in README.md may need more details to cover the question, 

1. Dose ATOM support any rocm/pytorch version within docker image?
2. Could it run without docker but native rocm+PyTorch environment?
3. Does ATOM support Radeon GPU and APU? What's the compatibility matrices of the GPU?

PR #2007, opened 2026-08-24
feat(pp) support dynamic chunked pipeline parallel

wanzhenchn:
# Dynamic Chunked Pipeline Parallelism

## Background
feat(pp) support dynamic chunked pipeline parallel
Motivated by [SGLang Dynamic Chunking](https://www.lmsys.org/blog/2026-01-15-chunked-pipeline/) and [vLLM Ascend Dynamic CPP](https://docs.vllm.ai/projects/ascend/en/latest/user_guide/feature_guide/dynamic_chunk_pipeline_parallel.html).

With fixed-size prefill chunks, attention cost grows with the cached prefix, so later chunks take longer than earlier ones. Pipeline stages fall out of sync and the pipeline develops bubbles. **Dynamic chunking shrinks later chunks so that every chunk costs about the same**.

### Latency model

A chunk of `x` new tokens after a cached prefix of `L` tokens is modeled as:

$$t(L, x) = c + \gamma L + b x + a (2Lx + x^2)$$

- `L` is the cached-prefix length, `x` is the number of new tokens in the current chunk, and `t(L, x)` is the predicted chunk latency.
- `c` is the fixed per-forward overhead, paid even when `L = 0`.
- `gamma * L` captures prefix-dependent work such as rebuilding cached-prefix state in MLA models. It is kept separate from `c` rather than folded into one constant because the two scale differently with `L`: `gamma * L` vanishes at `L = 0` and grows with the cached prefix, so collapsing them would erase exactly the prefix-dependent behavior this feature acts on.
- `b * x` captures work that grows linearly with the new-token count.
- `a * (2 * L * x + x^2)` models self-attention work: `2 * L * x` for the new chunk against its cached prefix, `x^2` for interactions within the new chunk.

`a`, `b`, `c`, and `gamma` are learned automatically from measured forward latency, so no offline coefficients or calibration flags are required.

### Solving for the next chunk

Every chunk of a request should cost what the first one did. The first chunk runs at `L = 0`, so its cost is `c + b*x0 + a*x0^2`. Since `c` is paid by every forward it cancels from both sides, and the budget a later chunk has to match is

$$b x_0 + a x_0^2 - \gamma L$$

- `gamma * L` is subtracted because it is a floor the chunk owes before any of its own tokens are attended to. Equalizing only the attention-area terms would leave every chunk paying that floor on top of an already-equal budget, making later chunks both too large to be equal-latency and too numerous.
- The scheduler solves the resulting quadratic for the `x` that spends the remaining budget, then applies the serving constraints: the smoothing factor, the minimum chunk size, an upper bound of the initial chunk size and the batch-token budget, and alignment to `max(KV block size, 64)`.
- When the prefix rebuild alone exhausts the budget, no chunk size can match the first chunk; ATOM stops shrinking and falls back to fixed-size chunking.

Equal-cost chunks keep pipeline stages in sync, which is what removes the bubbles and shortens TTFT for long requests.
## Performance

Kimi-K2.5-MXFP4, prefill PP4×TP1 on MI355X, PD disaggregation, OSL=128. 

- Input length 128k
image

- Input length 256k
image

Each cell is **throughput change / mean TTFT improvement**; positive means dynamic chunking wins.

### Baseline: tuned fixed chunk size

Fixed and dynamic each take their best result across 8K/16K/32K:

| Workload | conc=1 | conc=2 | conc=4 |
|---|---:|---:|---:|
| 128K fixed-length | **+2.9% / +3.3%** | +0.6% / +0.5% | -9.6% / -10.7% |
| 128K variable-length | **+4.3% / +4.9%** | -0.3% / -0.5% | +3.3% / +2.4% |
| 256K fixed-length | -3.6% / -4.1% | -0.2% / -0.2% | -0.2% / -0.2% |
| 256K variable-length | -3.0% / -3.4% | +0.8% / +0.4% | ≈0% / ≈0% |

Against a fully tuned fixed chunk size the gains are small and workload dependent; at 256K, fixed 8K is still the best configuration in most cases.

### Baseline: default `--max-num-batched-tokens=16384`

| Workload | conc=1 | conc=2 | conc=4 |
|---|---:|---:|---:|
| 128K fixed-length | **+10.7% / +11.3%** | -0.3% / +1.0% | +0.5% / +0.6% |
| 128K variable-length | **+9.7% / +10.4%** | -1.1% / -1.2% | -0.7% / -0.7% |
| 256K fixed-length | -0.8% / -0.9% | +0.1% / +0.2% | +0.1% / +0.1% |
| 256K variable-length | -3.3% / -3.7% | ≈0% / ≈0% | -0.1% / -0.1% |

At the default 16K budget, dynamic chunking mainly helps 128K at conc=1, worth roughly 10% on both throughput and TTFT. At 256K or higher concurrency it is essentially flat.

### Baseline: `--max-num-batched-tokens=32768` (common in agentic serving)

| Workload | conc=1 | conc=2 | conc=4 |
|---|---:|---:|---:|
| 128K fixed-length | **+43.7% / +34.3%** | -0.1% / -0.7% | -15.5% / -19.1% |
| 128K variable-length | **+39.4% / +32.0%** | -4.6% / -6.6% | +3.3% / +2.4% |
| 256K fixed-length | **+19.7% / +17.7%** | +3.8% / +3.9% | +3.1% / +3.2% |
| 256K variable-length | **+17.1% / +15.6%** | -0.4% / -0.8% | +0.1% / +0.1% |

The large 32K initial chunk benefits most, especially at long context and low concurrency. Gains at higher concurrency are not consistent, and 128K fixed-length at conc=4 still regresses noticeably.

Overall, **dynamic chunking reduces how sensitive a large initial chunk is to tuning at long context and low concurrency**, rather than beating a fully tuned fixed chunk size everywhere.

## Correctness

GSM8K, 3-shot, 1319 samples, with a 512-token budget to force multi-chunk prefills:

- 1107 requests actually took the multi-chunk path.
- Both fixed and dynamic completed 1319/1319 with no timeouts, crashes, or KV handoff failures.
- Flexible exact match: **92.72% → 93.56%**.
- Strict exact match: **92.57% → 93.48%**.

The accuracy difference is within run-to-run noise, so dynamic chunking does not regress accuracy.


atom/model_engine/model_runner.py
```diff
@@ -1250,6 +1270,180 @@ def warmup_model(self):
...
+
+        alignment = max(self.block_size, 64)
+        max_chunk = min(
+            self.config.max_num_batched_tokens,
+            self.config.max_model_len,
+            self.config.num_kvcache_blocks * self.block_size,
+        )
+        max_chunk -= max_chunk % alignment
+        if max_chunk < 8 * alignment:
+            raise ValueError(
```
yitingw1: Maybe better to use `try... except` like line1367 `fit_chunk_overhead` function?

atom/model_engine/dynamic_chunking.py
```diff
@@ -0,0 +1,470 @@
...
+    def _is_due(self) -> bool:
+        if len(self._timings) < MIN_CALIBRATION_SHAPES:
+            return False
+        if len({prefix for _, prefix in self._timings}) < MIN_CALIBRATION_PREFIXES:
+            return False
+        if len({chunk for chunk, _ in self._timings}) < MIN_CALIBRATION_CHUNK_SIZES:
+            return False
+        # A fit is attempted once per timing the last attempt did not see, so a
+        # failure costs one retry per new measurement rather than one per poll.
+        return self._since_fit > 0
```
yitingw1, line 199: This calibration sweep has no give-up mechanism: _calibrating only flips False inside install_chunk_latency_model(), which is only reached once a fit clears ChunkLatencyCalibrator.fit()'s quality gates. If those never pass on real hardware (timing jitter, MoE routing noise, EPLB rebalancing, background KV offload), the sweep never ends — every other request stays permanently chunked at max_num_batched_tokens // 4 for the life of the process, with no way to recover short of restarting without --enable-dynamic-chunking. Suggest capping the number of failed fit attempts and giving up to fixed chunking with a warning once exceeded.

wanzhenchn: Thanks for catching this. You’re right: repeated samples can keep triggering failed fits indefinitely while the scheduler remains in calibration mode. I’ll cap the number of failed fit attempts; once exceeded, calibration will stop, a warning will be logged, and the scheduler will fall back to fixed chunking.

atom/model_engine/model_runner.py
wanzhenchn: I agree it should fail gracefully like `fit_chunk_overhead`: I’ll log a warning and return an error so the server can continue with fixed chunking instead of failing startup.

yitingw1 (dismissed):

Yuechguo (dismissed):

yitingw1 (dismissed):

Yuechguo (dismissed):

Yuechguo (dismissed):

Jasen2201 (dismissed):
LGTM

atom/model_engine/model_runner.py
```diff
@@ -638,6 +652,15 @@ def __init__(self, rank: int, config: Config):
         self.tp_world_size = config.tp_world_size
         self.rank = rank
         self.label = f"Model Runner{rank}/{self.tp_world_size}"
+        # Dynamic chunking calibration state.
+        self._chunk_calibrator: ChunkLatencyCalibrator | None = None
+        self._chunk_sample_pending: deque[
```
valarLip, line 657: move all changes into atom/model_engine/dynamic_chunking.py?

valarLip (changes requested):

atom/model_engine/model_runner.py
wanzhenchn: It's a good suggestion.

valarLip:
Reviewed at `34eb44e16` (branch `zwan/feat-dynamic-ck`).

**Provenance:** findings marked **[verified]** are ones where I read the code at the cited line myself, and for #1 reproduced the failure. Findings marked **[reported]** came from a finder pass whose mechanism is traced through the diff but where I did not open the far end of the call chain — treat those as needing one more look before acting.

Cleared, so nobody re-chases them: `config.enable_dynamic_chunking` is mutated at `engine_core.py:154-176` *before* `Scheduler(config)` at `:199`, so there is no stale-flag problem; the `(req_ids, out)` tuple payload has exactly one producer and one consumer; `ChunkSizePredictor.from_coefficients` positional order matches the tuple `take_fit()` builds; and the new RPCs cannot hit the "worker RPC returning None deadlocks" footgun, because `outputs_queue` is fed only by rank 0 and rank 0 returns a dict on every branch.

---

# 1. [verified] `--enable-dynamic-chunking` cannot start: `_dummy_batch` assigns a list where an `array("i")` is required

```python
# dynamic_chunking.py:708
seq.block_table = list(range(math.ceil(chunk_size / block_size)))
```

`sequence.py:34` exists for exactly this, and says why:

> An `array("i")` rather than a list because **every forward marshals these into the int32 `block_tables` buffer**, where a list costs one CPython int unboxing per block (~17k per step at 50 seqs x 100k ctx) and an array is a memcpy.

`ScheduledBatch.__init__` (`scheduler.py:340`) collects `seq.block_table` into `self.block_tables` → `prepare_block_tables` → `pack_rows`, which does `flat[base:base+n] = row` on `memoryview(dst).cast("B").cast("i")`. Reproduced against that exact shape:

```
list         TypeError: a bytes-like object is required, not 'list'
array('i')   OK
```

The `TypeError` is raised inside `AsyncIOProc.busy_loop`, which has no handler, so the worker dies and EngineCore blocks in `outputs_queue.get()` until `monitor_procs` injects `SystemExit`. Startup fails on every PP stage, on the first profiling forward.

This is invisible everywhere else because `warmup_model` leaves `block_table` empty and the `if seq.block_table` filter skips it. Nothing in the test suite constructs `DynamicChunkingWorker` or `_dummy_batch` (see #14), which is why it ships undetected.

Fix: `new_block_table(range(...))`.

# 2. [verified] Same function reads `hf_config.vocab_size` instead of the runner's `hf_text_config`

```python
# dynamic_chunking.py:703
int(runner.config.hf_config.vocab_size)
```

This is the only such access in `atom/model_engine/`. `ModelRunner` already holds `self.hf_text_config = get_hf_text_config(hf_config)` (`model_runner.py:641`), and `get_hf_text_config` (`utils/__init__.py:178-188`) exists precisely because multimodal configs nest the text fields under `config.text_config`.

On such a model `profile()` raises `AttributeError` instead of degrading to `{"error": ...}` — breaking the error-handling contract this file follows everywhere else (see the `MIN_PROFILE_SAMPLES` and fit-failure paths, both of which return `{"error": ...}` rather than raising) — and takes the worker down the same way as #1.

# 3. [verified] The PP completion handshake is added unconditionally, changing the default PP path

```
$ grep -c "enable_dynamic_chunking" atom/model_engine/pp_engine_core.py
0
```

Previously `if not needs_output: popleft; release; continue` retired middle-chunk batches with no message, so `_in_flight` drained to empty every step and `while len(self._in_flight) < self.pp_size` let the head run ahead. Now every middle chunk costs a pickled `(tuple(req_ids), None)` ZMQ round trip from the last stage and holds an `_in_flight` slot until it arrives.

Concretely, on an existing PP chunked-prefill workload with no dynamic chunking at all — pp_size=4, 128K prompt, 16384-token chunks — eight middle chunks that used to retire instantly now each need a full pipeline traversal before the head may launch the next.

Knock-on: `if launched == 0: flush_pp_send` (`:126`, commented "when idle") now fires on **every** step once the head saturates at `pp_size`, adding a blocking all-TP RPC plus a 1 ms `_PP_HEAD_IDLE_POLL_MS` sleep to a step that is not idle.

This back-pressure may well be the right thing. But it needs its own A/B number, and as written it **confounds this PR's own regression rows** (128K fixed-length conc=4: −15.5% / −19.1%) — those numbers cannot distinguish "dynamic chunking is slower here" from "the unconditional handshake is slower here."

# 4. [verified] `predict()` returns `None` for two opposite conditions, and the scheduler reads it as "no cap"

Producer:

```python
raw = self.equal_latency_chunk(history_len, base_chunk_size)
if not math.isfinite(raw) or raw <= 0:
    return None                      # "no equal-latency chunk exists"
...
return aligned if aligned >= alignment else None
```

Consumer:

```python
# scheduler.py:2148-2151
chunk = min(
    num_new_tokens,
    budget_remaining,
    dynamic_limit if dynamic_limit is not None else num_new_tokens,   # None == no cap
)
```

`equal_latency_chunk` returns `nan` when `prefix_coeff * history_len >= target_latency(base)`. So a 200K prompt shrinks monotonically toward `min_chunk_size`, and at the prefix length where `gamma*L` crosses the budget the very next chunk **jumps back to the full `max_num_batched_tokens`** — producing the single worst pipeline bubble of the request, exactly where the chunk should be smallest.

The docstring at `:627` says "the caller should stop shrinking"; the caller does the opposite. `None` here needs to mean "clamp to `min_chunk_size`".

# 5. [verified] The two new numeric validations run outside the feature guard

```python
assert 1 <= self.pipeline_parallel_size
if not 0.0 <= self.dynamic_chunking_smooth_factor <= 1.0:
    raise ValueError("dynamic_chunking_smooth_factor must be in [0, 1]")
if self.dynamic_chunking_min_chunk_size <= 0:
    raise ValueError("dynamic_chunking_min_chunk_size must be positive")
if self.enable_dynamic_chunking:          # <- the guard starts here
    ...
```

`--dynamic-chunking-smooth-factor` defaults to `envs.ATOM_DYNAMIC_CHUNKING_SMOOTH_FACTOR` (`arg_utils.py:101`), so a stale `ATOM_DYNAMIC_CHUNKING_SMOOTH_FACTOR=1.5` left in a shared launch script hard-fails a plain `-tp 8` server with no PP and the feature off. A non-numeric value crashes even earlier, inside the envs lambda's `float()`.

Related inconsistency in the same block: `enable_dynamic_chunking` with `pipeline_parallel_size == 1` **raises**, whereas every other failure path in this feature — profiling failure, calibration give-up, useless model — degrades gracefully to fixed chunking.

# 6. [verified] The two calibration sweep arms can be equal, and there is no give-up path for "never enough diversity"

```python
max(self.max_num_batched_tokens // CALIBRATION_SWEEP_RATIO,
    self.dynamic_chunking_min_chunk_size)
```

returns exactly `max_num_batched_tokens` whenever `min_chunk_size >= max_num_batched_tokens` — e.g. the default `min=4096` with `--max-num-batched-tokens 4096`, or `min=16384` with the 16384 default. Both arms are then identical, `num_chunk_sizes` never reaches `MIN_CALIBRATION_CHUNK_SIZES=2`, `_is_due()` is False forever, `maybe_fit()` returns `None` forever.

`gave_up` only counts *rejected* fits (`_failed_fits`), never "never enough diversity", so it never fires. The engine ZMQ-polls `take_dynamic_chunking_fit` every 32 steps for the life of the process, and every sole-prefill request is permanently served in chunks that were only ever meant to be temporary. The same permanent limbo occurs if the workload is simply never sole-prefill, so zero samples reach `add()`.

# 7. [verified] The model is gated at the shallowest prefix it will ever see, and installation is one-shot

`install_chunk_latency_model` calls `predicts_useful_shrink(base_chunk_size=max_num_batched_tokens, history_len=max_num_batched_tokens)` — requiring ≥5% shrink at `L` equal to one chunk.

Meanwhile `_poll_dynamic_chunking_calibration` sets `self._dynamic_chunking_enabled = False` **before** calling install, and the worker has already nulled its `_calibrator` inside `take_fit`. There is no second attempt.

So with 128K prompts at `max_num_batched_tokens=16384`, a model predicting 3% shrink at L=16K but 55% at L=112K is rejected and the feature is silently off for the rest of the process. The `first` / "refreshed" DEBUG branch at `:2058-2060` is written as if refits happen; they cannot.

# 8. [verified] PP stages can derive different chunk grids, hanging startup before the ready signal

```python
max_chunk = min(max_num_batched_tokens, max_model_len,
                config.num_kvcache_blocks * runner.block_size)
```

`num_kvcache_blocks` is set per-stage from that stage's own `get_num_blocks` (`engine_core.py:138-153`) — PP stages hold different layer counts and different free memory. `_sweep` is a blocking send/recv chain with no barrier and no broadcast of the grid.

On a memory-tight config where the KV term becomes binding on some stages but not others, the grids differ in values *and length*: stage A runs 24 forwards, stage B runs 23, or one stage takes the `len(chunk_grid) < MIN_PROFILE_SAMPLES` early return at `:745` and runs zero forwards while its neighbour blocks in `recv`.

The result is a hang inside `EngineCore.__init__`, before the ready signal — the "stuck between `load model runner success` and `ready`" signature that `wait_server_ready.sh` does not flag as an error, so it presents as a timeout with no diagnosis.

Fix: derive `max_chunk` only from the stage-invariant `min(max_num_batched_tokens, max_model_len)`, or broadcast the head's grid.

# 9. [verified] `_calibration_sweep_chunk` is a sizing query with a side effect

`self._calibration_sweep_alternate = not self._calibration_sweep_alternate` runs before the caller knows the request will be admitted. At `scheduler.py:1419` the call is made, then `:1425` (`chunk is None` → `waiting.appendleft`; break) and `:1428` (`block_manager.allocate` fails → `appendleft`; break) both requeue the seq with the arm already flipped.

One request bouncing on budget or allocation for N ticks flips the arm N times. The sweep is supposed to alternate *requests* between the full budget and budget/4; under requeue pressure it can degenerate to a single arm, which starves `MIN_CALIBRATION_CHUNK_SIZES=2` and lands in the same silent never-converges state as #6 — with no log saying why. Same hazard on the `offload_resume` path at `:1342`, which is a continuation of a request that already ran chunks at the other arm.

# 10. [verified] The first-chunk call site passes a quantity the comment 14 lines below says is wrong

Lines `:1251` and `:1346` pass `seq.num_cached_tokens`. Line `:1423` passes `num_cached_blocks * self.block_manager.hash_block_size`.

The existing comment at `:1437-1446` states verbatim: *"the hit the request kept, not the one `can_allocate` offered: allocate may disown the boundary, and `num_cached_blocks` is the pre-disown count. `seq.num_cached_tokens` also gets the unit right under DCP."*

Meanwhile the worker records its calibration sample as `batch.num_cached_tokens[0]`. So on a boundary the allocator disowns, the model is **queried** at an inflated prefix and **fitted** at the true one: the first chunk of every cache-hit request is sized too small and pushed toward `prefix_bounded_chunk`.

The author had limited choice here — `allocate()` runs at `:1428`, after the sizing — so at minimum this needs a comment naming the approximation.

---

The following four are **[reported]**: the mechanism is traced through the diff, but I did not open the far end of each call chain.

# 11. [reported] The FIFO assert was weakened into a tautology

The deleted assert compared `fwd_out.req_ids`, built independently in the worker at `model_runner.py:3076` from the sampler (`prepare_sampled_ids`). The new assert compares `completed_req_ids`, which is `tuple(batch.req_ids)` read straight out of the metadata the head itself pickled downstream (`:420`) — an echo over a FIFO transport, so it cannot disagree.

Under deferred output (`is_deferred_out`), `prepare_sampled_ids` returns the previous batch's ids, and on the first forward returns `[]`. That used to fire loudly as "PP token ordering violated: received []". Now the echo matches, `assert fwd_out is not None` passes (the object exists, it is just empty), and `Scheduler.postprocess` is handed a `ScheduledBatchOutput` with no rows for the seqs it was given — tokens silently dropped instead of a crash.

Cheap fix: keep the old assert too, inside the `needs_output` branch.

# 12. [reported] The startup sweep times downstream PP back-pressure into `b`, runtime calibration does not

`_sweep`'s timed window wraps `runner.forward(dummy)` **and** `runner.flush_pp_send()`, and `flush_pp_send` → `commit_pp_send_work` → `work.wait()` (`pp_comm.py:159-164`) blocks until the next stage posts its matching recv. So stage 0's `best_ms` ≈ `max(own compute, downstream compute)`, and "best of two" cannot remove it because both reps are coupled identically.

Runtime samples do **not** include it: `finish_sample` runs at `model_runner.py:2865`, before `commit_pp_send_work` at `:2876`.

`_fit_terms` then computes `overhead = latencies - self.linear_coeff*chunks` with an inflated `b` against latencies that exclude it → overhead biased negative → `a`/`gamma` negative → either the `terms.prefix < 0` refit drops the prefix column (gamma=0, the exact degeneration the docstring at `:620-624` warns about), or `attention_span <= roundoff` rejects every fit until `MAX_CALIBRATION_FIT_FAILURES`. Triggered by any unbalanced PP split where a downstream stage is slower than the head.

# 13. [reported] The calibration sweep arm is returned unaligned

`config.py` validates `min_chunk_size > 0` but never against block alignment. `predict()` is safe — it takes `max(alignment, min_chunk_size, ...)` then aligns down — but `_calibration_sweep_chunk` returns `max(budget//4, min_chunk_size)` with no alignment.

`_chunked_prefill_size` then computes `aligned = chunk - chunk % max(block_size, 64)`; with `--dynamic-chunking-min-chunk-size 32` that is 0, and the pre-existing `aligned or (0 if num_batched_tokens else chunk)` fallback hands back the raw 32 on an empty batch. A non-block-multiple mid-prefill chunk was previously only reachable when the whole remaining prompt was that short; it is now reachable with 100k tokens left.

Fix: validate `min_chunk_size % alignment == 0`, or align inside `_calibration_sweep_chunk`.

# 14. [reported] Test coverage

`tests/test_pd_pp.py:723` `stage.pp_transport.send_tokens.assert_not_called()` is now vacuously true: `pp_transport` is a `MagicMock`, so `send_completion` is itself a mock attribute and never reaches the real `send_completion` → `send_tokens` delegation. The assertion passes regardless of what the code does, so it can no longer detect a regression where the last stage acks a request-less batch — which matters precisely now that the head blocks on exactly one completion per metadata message. Should be `send_completion.assert_not_called()`.

Separately, `tests/test_pd_pp.py:648` sets `recv_completion.return_value = None`, so the restructured `while self._in_flight:` body in `pp_engine_core.py` never executes in any test in the repo — the changed behaviour is untested in both its old and new form. And no test anywhere constructs `DynamicChunkingWorker` or `_dummy_batch`, which is why #1 ships undetected.

# 15. [verified] Docs and comments

`docs/environment_variables.md` carries 74 `ATOM_` entries and **zero** hits for `DYNAMIC_CHUNK`; `docs/configuration_guide.md` likewise has no entry for the three new `Config` fields. CLAUDE.md names `atom/utils/envs.py` as the home of all `ATOM_*` definitions with `docs/environment_variables.md` as its guide, and the Docs-match-code rule requires the table be updated for new entries.

Two comment defects in the same PR, in the one part of a change that has no automated gate:

- `scheduler.py:566` replaces the accurate `_partial_prefill_count` comment ("Running seqs currently mid-prefill; counter lets `schedule()` skip the running-queue scan on pure-decode steps") with an unrelated "# Prefix-cache SWA correctness is enforced by BlockManager." — which duplicates a claim that already lives correctly at `:1380`, and drops the "Running / excludes waiting" semantics that the new `_partial_prefill_count + len(self.waiting)` arithmetic depends on for correctness.
- `scheduler.py:550` adds `self.pipeline_parallel_size = getattr(config, "pipeline_parallel_size", 1)`, which is never read anywhere in the file. The three `getattr()` defaults at `:550-556` are also unreachable — `Config` always defines those fields — so they would silently swallow a rename rather than failing loudly.

PR #2033, opened 2026-08-26
[feat](vllm-atom k3): support DSpark speculative decoding under DCP

PerryZhang01:
## Motivation

This PR supports dcp + dspark kimi k3 on vllm-atom backend

### Two runtime patches against vLLM

`dspark_dcp_patch.py`

- **Config guard lift**: vLLM's EngineArgs rejects speculative decoding whenever decode_context_parallel_size > 1. The patch allows the combination and keeps decode_context_parallel_size on the target parallel config, which create_draft_parallel_config otherwise drops.
- **Draft KV slot localization**:  Inside vLLM's DFlash speculator, _dcp_local_slots() maps the draft block's slots to this DCP rank's shard. Upstream computes pos // block_size, which is the wrong interpretation once the pool is round-robin sharded.

### Attention metadata

`metadata.py`

- Passes supports_dcp_with_varlen=True to the MLA builder. Without it vLLM clamps reorder_batch_threshold to 1 under DCP, sending every multi-token decode.
-  Derives DCP-local seq lens for the draft step.  The draft builds its own attention metadata in a path that never fills dcp_local_seq_lens, so they're computed from the global ones into a persistent buffer.
- Adds g_kv_indptr (global pre-shard cumulative KV lengths) and gates the round-robin path on dcp_world_size > 1 and max_qo_len > 1 and causal. A single-token decode sees all of its local KV and a bidirectional block masks nothing, so both stay on the plain kernel.
- Tells the persistent work-descriptor planner page_size=1 when round-robin is active, matching what mla_decode_fwd actually receives. The planner had been told 128, which silently scaled cprr's reconstructed global positions by the block size.

## Test Result

flexible-extrac：0.9538


valarLip (dismissed):

.github/workflows/atom-vllm-test.yaml
```diff
@@ -149,12 +149,12 @@ jobs:
...
+            extra_args: "--trust-remote-code --language-model-only --tensor-parallel-size 8 --decode-context-parallel-size 8 --kv-cache-dtype fp8 --max-model-len 16384 --max-num-seqs 64 --max-num-batched-tokens 16384 --gpu-memory-utilization 0.75 --block-size 128 --enable-prefix-caching --mamba-cache-mode align --compilation-config '{\"cudagraph_mode\":\"FULL_AND_PIECEWISE\"}' --additional-config '{\"online_quant_config\":{\"global_quant_config\":\"ptpc_fp8\",\"exclude_layer\":[\"lm_head\",\"model.embed_tokens\",\"*self_attn.[qkv]_conv1d*\",\"*block_sparse_moe.experts*\",\"*block_sparse_moe.routed_expert_*\",\"*vision_tower*\",\"*mm_projector*\"]}}' --speculative-config '{\"method\":\"dspark\",\"model\":\"/models/Inferact/Kimi-K3-DSpark\",\"num_speculative_tokens\":7}'"
```
valarLip: why  --gpu-memory-utilization 0.75

PerryZhang01: DCP 下 persistent_num_heads 会按 dcp_size 放大，每个 rank 承担全部 head（16 → 128，8×），DSpark 让 query 长度从 1 变成 8，而 packed_qo_len = 8 × 128 = 1024 跨过了 tiling 阈值，使 max_qo_tiles_per_batch 从 1 跳到 8，reduce_partial_map 又涨 8 倍。合起来相对无 DCP 的普通 decode 是几百倍。

native ATOM最大也只能开到0.85，vllm-atom如果减小graph capture size到64也能开到0.85。整体上vllm比atom的显存消耗还要高一点。

PR #2037, opened 2026-08-26
feat: support v4 mixed-schedule large conc opt

jiayyu:
## Test Result
省下 669 个 decode 步  →  理论收益  89.7 s
实测净收益                          3.3 s
                                 ────────
mixed 步吃掉的额外成本              86.4 s
                     每个 mixed 步  +115 ms

image

ci
image

## Submission Checklist


atom/model_ops/attention_mla.py
```diff
@@ -2550,7 +2584,138 @@ def forward_impl(
...
+                    kv_cache_dtype=self.kv_cache_dtype,
+                    scale=self._k_scale,
+                )
+
+            # Preallocate the combined pre-o_proj output [n_total, num_heads*v_head_dim]
+            # and write each segment straight into its slice (vLLM-style), so a single
+            # o_proj runs over the whole tensor and no torch.cat is needed.
+            attn_out = torch.empty(
+                (q.size(0), self.num_heads * self.v_head_dim),
+                dtype=torch.bfloat16,
```
ZhangLirong-amd: use `dtype=q.dtype` like non-mix?

atom/models/deepseek_v4.py
```diff
@@ -2872,34 +2881,186 @@ def forward_impl(
...
+                )
+            fc.attn_metadata = seg_md
+            ctx.is_prefill = is_prefill
+            if saved_input_ids is not None:
+                # Keep ctx.input_ids segment-consistent for anything downstream
+                # reads off it (e.g. the hash-MoE `_hash_topk`).
+                ctx.input_ids = saved_input_ids[lo:hi]
+            sl = slice(lo, hi)
+            tag = "mixed_seg[prefill]" if is_prefill else "mixed_seg[decode]"
+            with torch.profiler.record_function(f"{tag} n={hi - lo}"):
```
ZhangLirong-amd: no need `profiler` ?

atom/models/deepseek_v4.py
```diff
@@ -2872,34 +2881,186 @@ def forward_impl(
...
+        # the pre-refactor core allowed -- still far cheaper than the per-layer
+        # `torch.cat` it replaced (11.95 ms/step over 61 layers).
+        out = torch.empty(
+            (n_p + n_d, self.n_local_heads * self.head_dim),
+            dtype=hidden.dtype,
+            device=hidden.device,
         )
-        return self._attn_post(o, positions)
+
+        def _seg(lo: int, hi: int, seg_md, is_prefill: bool) -> None:
```
ZhangLirong-amd, line 3099: can we refactor this func?

valarLip:
## Review — mixed prefill+decode batching

Reviewed at `1e501a1c6` in an isolated worktree. Findings marked **[verified]** I read out or reproduced myself; **[reported]** ones match the code shape but I did not independently confirm. Measurement scripts are noted inline.

Three findings below are, in my read, merge blockers — and two of them fire on the default deployment shape, not an exotic config. After those, a structural pass on organization and cost.

---

# Blockers

## 1. With the flag on, the prefill-only early return becomes dead code — for pure-prefill batches too **[verified]**

`scheduler.py:1667`:

```python
if num_seqs_prefill > 0 and not self.enable_mixed_prefill_decode:
```

The block that condition guards does four things, and all four stop happening the moment the flag is on — **including on batches that contain no decode rows at all**:

- `is_final_chunk` is assigned only at `:1742`, inside the block. It is therefore always `None`, so `produces_output()` returns `True` at `:470` even for an all-middle-chunk batch, and `compute_logits` runs a step early.
- **`plan_midstep` has exactly one caller in the repo** — `:1718`, inside the block. (`grep -rn plan_midstep atom/` returns that call, the definition at `block_manager.py:2005`, and two comments.) Midstep state checkpoints silently stop being created entirely.
- `next_token_ids` and `_advance_prefill_on_schedule` likewise.

`ModelRunner._record_kv_cache_ready` also branches on `is_final_chunk is None`, so a PD consumer can be told a half-written prefix is ready. **[reported]**

The gate needs to be on whether *this batch* is mixed, not on whether the feature is enabled.

## 2. V4 `prepare_mixed` writes the host mirror of `cu_seqlens_q` and never uploads it **[verified]**

`deepseek_v4_attn.py:2611` writes `var["cu_seqlens_q"].np[: n_d_seqs + 1] = np.arange(...)`. There is no `copy_to_gpu` for `cu_seqlens_q` anywhere in `prepare_mixed` — I swept every `cu_seqlens_q` and `copy_to_gpu` occurrence in 2540-2720. And `prepare_decode` takes the device side as a bare view:

```python
# deepseek_v4_attn.py:2776-2777
# Uploaded once by `publish_cu_seqlens_q`; this is a view.
cu_seqlens_q_gpu = var["cu_seqlens_q"].gpu[: running_bs + 1]
```

So the device tensor still holds the **full-batch** cumsum (`[0, 4096, 4097, ...]`) while every host-side assert passes against the corrected `[0, 1, ..., 8]`. Kernels that segment q by it index thousands of rows into an n-row tensor.

The comment at `:2603-2610` states this write exists specifically to fix "GPU OOB in swa_write". The fix landed on the host only.

The dense twin does it correctly — `aiter_mla.py:1868` writes `.np`, then `:1873` `copy_to_gpu(d_bs + 1)`.

## 3. The mixed branch slices `q` but passes the per-token `q_scale` unsliced to both halves **[verified]**

`attention_mla.py`:

```python
2605:  q_p = q[:n_prefill]
2614:  prefill_q = proj_p(q_p, x_scale=q_scale)                        # q sliced, scale not
2675:  q_d = q[n_prefill:]
2680:  q_nope_d, q_rope_d = self._q_proj_and_k_up_proj(q_d, x_scale=q_scale)
```

`q_c_scale` is allocated as `torch.empty((M, scale_n), ...)` with `M = hidden_states.shape[0]` (`deepseek_v2.py:742` and `:772`) — **one row per token**. So the decode half dequantizes rows `[n_prefill:]` using the *prefill* rows' scales.

The two non-mixed branches (`:2723`, `:2794`) pass the full `q_scale` alongside the full `q` and are self-consistent; only the mixed branch separates them.

For the MXFP4 shuffled layout it is worse in both directions: `scale_m = ((M + 255) // 256) * 256` depends on `M` itself, so the prefill half misindexes too.

This is live on every FP8/FP4 DeepSeek checkpoint — the normal deployment. It does not crash and does not log; a prefill-only or decode-only accuracy run cannot see it. Which matters given the new accuracy cell (see §Tests).

---

# Correctness — also verified

## 4. The new assert in `_forward_piecewise_attention` is constant-folded by Dynamo **[verified]**

`deepseek_v4.py:2638` adds `assert not getattr(get_forward_context().attn_metadata, "is_mixed", False)` inside `_forward_piecewise_attention`, which sits under `@support_torch_compile class DeepseekV4Model` (`:4451`). It reads a plain Python attribute off a non-tensor object, so it is evaluated once at trace time against the warmup batch (never mixed) and baked out. Replays never evaluate it.

**The same file says so 20 lines above**, at `:2618`: *"Resolve to a plain bool: Dynamo folds it."*

Contrast `forward_impl`'s live `is_mixed` read at `:2888`, which is inside the opaque `torch.ops.aiter.v4_attention_with_output` custom op and does fire.

This also touches the repo rule *"NEVER modify `@support_torch_compile` decorated model files — instrument at call sites instead."* Related: `deepseek_v4.py` has **no COMPILE BOUNDARY block** naming which functions are traced; `deepseek_v4_dspark.py` does. Given this PR changes 161 lines in that file, adding one would be worth more than the assert.

## 5. Fix-then-sweep miss: one of three prefill budget sites was not converted **[verified]**

The PR converts Phase 1 (`:1325`) and Phase 2 (`:1484`) to `prefill_budget`, but the offload-prefill-resume branch still spends the full budget:

```python
# scheduler.py:1408
budget_remaining = self.max_num_batched_tokens - num_batched_tokens
```

On that path the decode reservation does not hold — which is the entire premise of the change.

## 6. `decode_token_reserve` uses a seat count and a width the decode loop does not **[verified]**

```python
# scheduler.py:1303
decode_token_reserve = min(n_decode_inflight * (self.mtp_k + 1), ...)
# scheduler.py:1776  — what the loop actually spends
spec_width = self.mtp_k if self.spec_decode_local else 0
```

`spec_decode_local = self.use_spec and pp_size == 1` (`:597`), so spec + `pipeline_parallel_size > 1` over-books by `(mtp_k+1)x` what decode can ever spend, silently, every step. `n_decode_inflight` is also capped at `max_num_seqs` while the loop only has `max_num_seqs - num_seqs_prefill` seats.

Separately: this formula budgets for speculative tokens, but **MTP + mixed is explicitly refused** by two of the gates in §7. The formula was written against a configuration this PR forbids.

## 7. Seven refusal points, five files, no config-time gate **[verified]**

```
attention_mha.py:209     assert not is_mixed            MHA models
attention_mla.py:2592    assert not is_sparse_mla       sparse MLA
attention_mla.py:2596    assert not use_prefill_mla     prefill-MLA
aiter_mla.py:1807        assert not self.is_sparse      sparse MLA   (again)
model_runner.py:502      assert not self.use_spec       MTP
model_runner.py:2466     raise NotImplementedError      MTP          (again)
backends.py:929          raise NotImplementedError      backend
```

Five `assert`s — **stripped under `python -O`**, leaving silently wrong numerics — and two `raise`s, with no principle separating which is which. All of them fire after the server is up and a batch has been built.

Two of them refuse *the same thing* (MTP) with different wording (`"(follow-up)"` vs `"(P2-M4 follow-up)"`); two more refuse sparse MLA in two different files. That is a capability matrix that has been scattered instead of written down. One table in `Config.__post_init__` that fails the launch would replace all seven and remove the `-O` exposure.

---

# Organization

`is_mixed` now appears across **12 files** — scheduler, block_manager, model_runner, forward_context, two TBO modules, four attention modules, embed_head, and the model. No module owns the concept: there is no one place that defines what a mixed batch is, which configurations may produce one, or how it is split.

The feature was added entirely inside the repo's largest files rather than in a new module:

```
deepseek_v4_attn.py   4567 -> 5024   (+457)    ~6x the 800-line soft ceiling
scheduler.py          3804 -> 4004   (+200)
attention_mla.py      3251 -> 3416   (+165)
deepseek_v4.py        4837 -> 4998   (+161)
aiter_mla.py          2979 -> 3120   (+141)
```

Those files were already over the ceiling before this PR — but several of the new pieces are self-contained (`prepare_mixed` x2, `_MixedDecodeView`, `_get_mixed_prefill_bank`, `split_mixed_token_midpoint`) and would sit naturally in their own module.

At function level (measured with an AST pass, PR vs `main`):

```
scheduler._schedule()               610 -> 721 lines
MLAAttention.forward_impl()         280 -> 411 lines    mixed is the 4th arm of the if/elif chain
deepseek_v4._attn_mixed()           new,  142 lines
deepseek_v4_attn.prepare_mixed()    new,  145 lines
```

`forward_impl` at 411 lines is 8x the guideline, and the mixed branch is ~120 lines inlined into it. Extracting `_forward_mixed()` would both bring the function back into range and give the mixed path a boundary that can be reviewed on its own.

**`_attn_mixed` reuses the downstream code by temporarily rewriting thread-local state** — `fc.attn_metadata`, `ctx.is_prefill`, `ctx.input_ids` are saved, overwritten per segment, and restored in a `finally` (`deepseek_v4.py:3001-3006`, `:3058-3062`). The restore is correct. But it means any future code that reads the forward context from inside a segment silently sees a fabricated batch, with no signature or type to warn about it.

---

# Cleanliness

## `_get_mixed_prefill_bank` reimplements `CpuGpuBuffer.clone()` — and drops `pin_memory` **[verified]**

```python
# backends.py:512-523 (new)            # atom/utils/__init__.py:739-748 (existing clone())
mirror = CpuGpuBuffer(                 new = CpuGpuBuffer(
    *val.cpu.shape,                        *self.cpu.shape,
    dtype=val.cpu.dtype,                   dtype=self.cpu.dtype,
    device=val.gpu.device,                 device=self.gpu.device,
                                           pin_memory=self.cpu.is_pinned(),   <-- missing
    with_numpy=hasattr(val, "np"),         with_numpy=hasattr(self, "np"),
)                                      )
mirror.cpu.copy_(val.cpu)              new.cpu.copy_(self.cpu)
mirror.gpu.copy_(val.gpu)              new.gpu.copy_(self.gpu)
```

Line for line the same, and `clone()`'s docstring states its purpose as *"per-in-flight-slot metadata rings so concurrent pipeline microbatches never share a staging buffer"* — the same purpose as this bank.

Dropping `pin_memory` is not cosmetic. The bank exists so the prefill half's H2D does not race the decode half's pinned writes; a **non-pinned** CPU buffer makes `copy_(non_blocking=True)` effectively synchronous, so the async behaviour the docstring claims quietly does not hold. `bank[name] = val.clone()` is both the dedup and the fix.

Separately, the bank is cached as one dict on the builder while `forward_vars` is a per-PP-microbatch ring (`model_runner.py:1383` builds one dict per in-flight microbatch precisely so one cannot overwrite staging another is reading). Both swap sites point at the same single bank. **[reported]**

## One new env var bypasses `envs.py` **[verified]**

This PR registers four new variables in `envs.py` (`ATOM_TBO_MIXED`, `ATOM_PROBE_TBO_MIXED`, `ATOM_PROBE_STEP_BUDGET`, `ATOM_NUMA_BIND`) and then reads a fifth, `ATOM_PROBE_MIXED_NORM`, via raw `os.environ.get` on a per-segment per-layer path (`deepseek_v4.py:3034`), with a function-local `import logging as _lg` inside the branch. The file does have pre-existing raw `os.environ` reads, so this is a local convention — but the PR does both things at once.

## A docstring contradicted 76 lines away **[verified]**

`deepseek_v4_attn.py:2679` — `prepare_decode` *"populates positions, **cu_seqlens_q**, ..."*
`deepseek_v4_attn.py:2603` — *"prepare_decode READS var["cu_seqlens_q"] (**it never writes it**)"*

These are about the same function, and the contradiction sits exactly where blocker §2 lives.

## Naming asymmetry

`split_mixed_token_midpoint` (public) sits next to `_split_prefill_token_midpoint` (private) in the same module, at 51% normalized line-level similarity — one exported, one not, with no stated reason.

---

# Performance

## 0.8 ms per mixed step of `record_function`, with the profiler off **[verified — measured]**

`deepseek_v4.py:3009` and `:3055` are ungated — no check that profiling is active, and both tags are f-strings formatted on every call. Three per layer (two inside `_seg`, one around the pair) x 61 layers = 183:

```
record_function enter/exit (profiler OFF, f-string tag):  4.361 us
x 183 per mixed step                                   =  0.798 ms
```

That is unconditional host time on every mixed step, forever, for a profiler that is off by default. The repo's `mark_trace` decorator exists for exactly this.

The `os.environ.get` above measures 0.353 µs — ~0.043 ms/step at 122 calls. Small, but free to fix by moving it to `envs.py`.

## The mixed prefill bank is allocated mid-serve **[reported]**

`_get_mixed_prefill_bank` duplicates every `CpuGpuBuffer` in `forward_vars` lazily on the **first mixed step** — after the KV cache has already been sized to fill VRAM. Independent of how large that turns out to be on a given config, warmup is where this allocation belongs, not a live step.

---

# Tests

`ATOM_TBO_MIXED=1 pytest tests/test_tbo_mixed_veto.py` **fails — reproduced** **[verified]**:

```
AttributeError: '_Batch' object has no attribute 'total_seqs_num'
2 failed, 3 passed
```

Both new catalog cells export that variable (`.github/benchmark/models.json:67`, `models_accuracy.json:517`), so any runner or shell inheriting it turns these into failures that read like a product bug. The file's docstring also still asserts the pre-PR premise: *"Mixed batches cannot be split today."*

Meanwhile the only new production branch in `local_tbo_precompute` — the `ATOM_TBO_MIXED=1` dispatch into `_precompute_mixed_token_split` — has no coverage at that level, so a wrong argument order or a renamed batch field on that line ships green. That is the line the test above dies on.

The new accuracy cell's `_baseline_note` says it is "the sibling plus `--enable-mixed-prefill-decode`" so "a delta isolates TBO splitting a mixed batch" — but it also drops `--index_cache_dtype fp4` (present on all nine other V4-Pro entries) and adds `--max-num-batched-tokens 16384`. Three differences, so a drift is not attributable, and the copied thresholds were calibrated on a config that was never run. Given blocker §3 is a silent-accuracy defect, this is the cell that would have to catch it.

---

# Also worth a look (not independently verified)

- `aiter_mla.py:1823` — the dense `prepare_mixed` swaps in `pf_bank` without the `pf_bank["cu_seqlens_q"]` re-sync the V4 builder added for exactly this (`deepseek_v4_attn.py:2584-2586`). The bank is built once and never re-synced, so its `cu_seqlens_q` freezes at the first mixed step's value. This is the only backend `prepare_mixed` claims to support.
- `attention_mla.py:2621` — the mixed branch hardcodes one KV-cache write kernel and one decode dispatch, dropping the seg-MLA layout, the shuffled-KV layout, and the whole DCP path that both non-mixed branches dispatch on. Under `--decode-context-parallel-size>1` every decode row in a mixed step would attend to 1/dcp of its context with no cross-rank LSE merge, and no assert covers it.
- `model_runner.py:2176` — the PCP override fires on `is_prefill and pcp_size > 1` and unconditionally replaces the mixed veto `(False, False, 0, 0)`, then routes to a prefill-only slice builder that drops every decode row from the forward.
- `scheduler.py:1813` — in a mixed step, `_preempt_one_running()` scans from the tail, where `_schedule_prefill_seq` just appended this step's admissions; the preempted seq stays in `scheduled_seqs` with its blocks freed, and `block_tables` silently loses a row, shifting every later row.
- `deepseek_v4_attn.py:4436` — `_MixedDecodeView.__getattr__` raises `AttributeError`, which `getattr(x, name, None)` swallows; `state_fork_srcs` then reads as `None` and forked rows get `state_slot_in == state_slot_out`. The guard needs to raise something `getattr` will not eat.
- `deepseek_v4_attn.py:409` — `assert len(_slots) == batch.total_seqs_num` asserts a property `state_slots_committed` does not have: it is built from a filtered subset (`scheduler.py:260-269`), not one entry per row. The view's positional slice depends on the same premise.
- `ubatching.py:258` — `ATOM_TBO_MIXED` lifts the mixed veto globally, but only the V4 builder grew the mixed redirect; the dense MLA, MHA and base builders did not.

One note on a claim I checked and can rule out: the new `prepare_mixed` functions are **not** copies of their prefill/decode siblings — normalized line-level similarity is 6-8%. They are genuinely new orchestration. Only `split_mixed_token_midpoint` vs `_split_prefill_token_midpoint` is meaningfully duplicated, at 51%.

---

The CPU unit suite is green on the branch (6185 passed; the nine failures are aiter/`gh`-dependent and fail identically without the PR).

**Suggested order:** §1-§3 block. Then §7 (one config-time capability table — it removes seven scattered refusals *and* the `python -O` exposure in one change), the `val.clone()` one-liner, and gating the `record_function`s. The `_forward_mixed()` extraction is worth doing while the branch is still new.

PR #1880, opened 2026-08-13
feat(lmcache): add DSV4 page and slot offload

yhl-amd:
## Summary
- route page-major DSV4 compressed KV through the standard LMCache chunked store/retrieve path
- persist full per-request compressor and SWA state as a fail-closed AOS1 sidecar
- add generation-safe TP/MultiConnector lifecycle handling, bounded staging/backpressure, and comprehensive regression coverage

## Test plan
- [x] `pytest -q tests/test_deepseek_v4_transfer_regions.py tests/test_kv_aggregator.py tests/test_lmcache_offload_config.py tests/test_lmcache_offload_connector.py tests/test_lmcache_offload_v4_page_slot.py tests/test_lmcache_offload_disk_integration.py tests/test_lmcache_offload_gpu_disk_e2e.py tests/test_multi_connector.py tests/test_page_region_codec.py tests/test_per_req_cache_decoupling.py tests/test_scheduler.py tests/test_slot_sidecar_codec.py tests/test_slot_sidecar_format.py tests/test_slot_sidecar_store.py tests/test_state_checkpoint.py tests/test_v4_pool_geometry.py` (781 passed, 7 prerequisite skips)
- [x] `ruff check` and `black --check` on all changed Python files
- [x] `python3 -m compileall` on changed packages and `git diff --check`
- [x] real DeepSeek-V4-Pro TP8 ROCm serve regression with standalone LMCache PAGE/SLOT CPU offload

## ROCm regression results

Validated at commit `4cf68dea` with DeepSeek-V4-Pro, TP8, FP8 KV/index cache, and standalone DSV4 PAGE/SLOT LMCache LocalCPU offload.

### GSM8K two-pass offload regression

Workload: GSM8K 20-shot, 1,319 questions, concurrency 32.

| Metric | Pass 1 | Pass 2 |
|---|---:|---:|
| Flexible exact match | 0.9507 | 0.9530 |
| Strict exact match | 0.9515 | 0.9538 |
| Wall time | 472 s | 218 s |

- Pass 2 was **2.17x faster** than Pass 1.
- Pass 2 completed 1,316 logical CPU loads, with `1,316 x 8 = 10,528` successful rank-local completions; the remaining 3 requests were served by the native HBM prefix cache.
- Load miss, load/save failure, OOM, traceback, and checkpoint drop/eviction counts were all zero.
- Accuracy did not regress.

### LMBenchmark CxS recompute vs offload

[...]

PR #2045, opened 2026-08-27
paged state cache 

ganyi1996ppo:
# feat(state-cache): per-slot allocation, a switchable demand rung, and a -1 interval

Rebased onto main's PAGE-backed checkpoint work (#1874, #1894, #1943, #1880), which
rewrote this subsystem underneath the branch. Three commits squashed to one, because
each re-conflicted against the new base *and* against the previous commit's
resolution; the reasoning from all three is preserved below.

## Summary

Four independent changes to state checkpointing, plus the reconciliation with main's
PAGE work:

1. **The slot, not the group, is the unit of state allocation.** A checkpoint no
   longer pays for speculation scratch it cannot use — at `--num-speculative-tokens 2`
   that was two thirds of every checkpoint's bytes.
2. **A prompt-end anchor** places a checkpoint where the next turn of a conversation
   actually resumes. It reads back **85.2%** of the time against the demand rung's
   **2.8%**, making it the highest-value placement in the system — and the only one
   left standing under `-1` + demand-off.
3. **`--state-checkpoint-interval-tokens -1`** turns off the interval ladder without
   turning off checkpointing.
4. **`--no-state-checkpoint-demand`** turns off the demand rung independently of the
   interval, so its write traffic can be measured against its reuse. Checkpoints are
   now snapshotted from a **full-length forward's interior states** rather than by
   cutting the prefill chunk onto each position.

Default behaviour is unchanged on every axis except one: DeepSeek-V4 loses a prefill
chunk per prompt that it was paying for and getting nothing back from (see
[DeepSeek-V4](#deepseek-v4)).

---

## 1. Allocate state slots per need, not by group

A "state cache group" was `1 + num_spec` slots wide and was the unit of everything:
allocation, admission, sizing, and the checkpoint index. But a checkpoint has no
speculation to roll back — it holds a committed state — so filing one cost a full
group and wasted `num_spec / (1 + num_spec)` of its bytes.

The slot is now the unit. `--state-checkpoint-slots 64` buys 64 checkpoints for 64
slots instead of 192, and the slots it no longer takes stay in the paged KV pool,
which is sized out of what is left after this.

This is only possible because a request's slots need not be adjacent, which the
kernels never required: the ssm kernel gathers each index out of the indices tensor
and the conv path is handed column 0 alone. Contiguity was *manufactured* by
`prepare_state_indices` writing `arange(base, base + width)`; it now writes the seq's
own slot list straight in.

**Renames** (mechanical, no behaviour change):

| before | after |
|---|---|
| `StateGroupPool` | `StateSlotPool` |
| `GroupRetirement` | `SlotRetirement` |
| `Sequence.per_req_cache_group: int` | `Sequence.state_slots: list[int]` (+ `state_slot` property = element 0) |
| `BlockManager.num_per_req_cache_groups` | `num_state_slots` (+ new `state_slots_per_req`) |
| `_attach_state_group` | `_attach_state_slots` |
| `lookup_group()` | `lookup()` |

`Sequence.state_slot`'s setter re-points `[0]` and preserves `[1:]`, because
speculation scratch persists across forwards — step N's accepted slot is step N+1's
initial state — so it belongs to the request, not to whichever slot it currently
commits into.

`--state-checkpoint-groups` still parses, as an alias for `--state-checkpoint-slots`.

## 2. A prompt-end anchor

The interval ladder guesses where reuse will resume. The prompt end does not have to
guess: in a multi-turn conversation the next turn *is* this turn's prompt plus the
reply, so the position a resume will land on is exactly where this prompt stopped.

`_record_checkpoint_end` reserves a checkpoint there, floored to the hash-block grid
(`Sequence.checkpoint_end_pos`). On the cc-traces it reads back **85.2%** of the
time — against **2.8%** for a demand rung and **0.0%** for the interval ladder.

Two consequences worth stating plainly, because the rest of this PR depends on them:

- **The anchor is what makes `-1` viable.** With the ladder off and the demand rung
  off, the anchor is the *only* thing placing checkpoints; without it that
  configuration would be checkpointing-off in all but name.
- **Every Kimi-K3 measurement on this branch is the anchor-only arm.** The K3 launch
  script sets `--state-checkpoint-interval-tokens -1` and
  `ATOM_STATE_CHECKPOINT_DEMAND=0`, so the anchor placed every checkpoint behind
  those numbers.

The anchor sits outside the grid and does not dominate the rung, nor the rung it —
one is the last position the grid could place, the other is where a resume actually
arrives. That is why `checkpoint_cut` has to loop over candidates and take the
*earliest*, rather than collapsing the grid to its rightmost rung as it could when
the rung was the only placement.

## 3. `-1` turns off the interval ladder without turning off checkpointing

The interval is a *guess* about where reuse will resume; a demand rung is a position
a request was actually refused at. Measured on the SemiAnalysis cc-traces, the 8192
ladder placed ~30× the writes of the demand rung alone and caught reuse the demand
already reaches — **0.0% of resumes landed on a ladder rung**. Every rung also costs
the prompt that keeps it an extra prefill chunk.

```
 >0   a rung every N tokens          (unchanged, still the default)
  0   state checkpointing off entirely (unchanged)
 -1   no interval rungs; the demand rung and prompt-end anchor still place them
```

`-1` rather than reusing `0`, even though "no interval" reads like zero: `0` is the
documented contract *and* it is reachable by accident — the grid snap rounds an
off-grid interval down and can land on 0, so a `--block-size` typo currently fails
safe. Giving 0 a second meaning would make that typo silently *enable* a caching
policy instead of disabling one.

Four sites, three of which are not the arithmetic you would guess:

- the clamp at `__init__` stops flattening the negative (`max(-1, ...)`, not `max(0, ...)`)
- the grid snap tests `> 0` rather than truthiness, or -1 snaps to -4 and warns about
  a flag the user set deliberately
- `checkpointers_at` cannot ask `pos % interval` under -1: **-1 divides every
  integer**, so that test admits *every* position as on-grid rather than none
- the decode path returns early under -1 for the same reason — `pos - last < -1` is
  true for every pos, which would checkpoint on every decode step

Swept over 76 prompt lengths × 3 requests: -1 costs 68 cuts against the ladder's 108
(−37%), and every cut is still kept in all three modes.

## 4. Make the demand rung switchable

A demand rung is **47% of all checkpoint writes** on the cc-traces and reads back
**2.8%** of the time, against **85.2%** for a prompt-end anchor. Every one of those
writes evicts something, so the rung may cost more than its reuse is worth — but the
CPU replay scores "delete the rung" and "keep it and demote it in LRU" *identically*
(86.9% → 88.2% both), which is the signature of a harness that models eviction order
and not the cost of the write itself. **Only hardware separates them**, which is what
the flag is for.

Gated independently of `--state-checkpoint-interval-tokens`, because the demand is
not part of the interval grid: it is the one placement a refused hit makes for itself.

The refusal is still *measured* when the placement is off — `num_wanted_hit_blocks` is
what `CacheStats` splits declined reuse by, and switching off a rung must not blind
the diagnostic that justifies it.

### Taking the state midstep instead of cutting the chunk

Placing a rung used to mean **shortening a prefill chunk onto it**: a checkpoint was
whatever state a forward held when it *ended*, so the only way to capture position P
was to make a forward stop at P. Each placement therefore cost a chunk boundary, and
that cost — not the bytes — is what made rungs expensive. It is where the ladder's
17.5% prefill regression came from, and why "one extra prefill chunk per prompt"
recurs throughout this PR as the unit of harm.

The chunked recurrence already computes the state at every chunk boundary internally;
it just threw the interiors away. `chunk_gated_delta_rule_fwd` now returns `h`
unconditionally, and `keep_intermediate_states=True` parks it on
`ChunkGatedDeltaRuleFunction.last_h` for `pop_last_intermediate_states()` to collect
(`atom/model_ops/fla_ops/chunk.py`). A backend that can read its own interiors takes a
snapshot **at every reserved position inside one full-length forward**, and the
forward runs to its natural end.

This is what `midstep_positions` is for, and it is the twin of `checkpoint_cut`: same
three candidates (grid rung, demand, anchor), but where the cut must pick *one* and
shorten onto it, `midstep_positions` returns *all* of them. `checkpoint_cut` returns 0
— no cut at all — once **every** class that applies to the sequence answers
`readable_midstep`. Every, not any: one unreadable class still needs the forward to
land on the position, and the readable ones lose nothing by being handed a position
they would have taken anyway.

Consequences:

- **Placements stop competing with prefill efficiency.** A rung, a demand and an
  anchor on the same prompt cost one forward between them, not three.
- **The demand rung's price drops to its write traffic**, which is what makes the
  A/B this flag exists for a clean question — the confound was the chunk cut.
- Interior states are **bit-exact** with the state a cut forward would have ended on,
  so this is a scheduling change, not an approximation. `readable_midstep` is a
  capability flag on `StateTransfer`, independent of `kind`: GDN answers True, the
  DeepSeek-V4 PAGE coordinator answers False (a page image is not readable midstep).

The two gates are a pair and change together. Suppressing the cut alone leaves
`checkpointers_at` refusing every off-grid position it is then handed — zero
checkpoints kept, no error, no warning, and a hit rate that quietly collapses.

---

## DeepSeek-V4

**Unaffected by the slot-vs-group change**, and that claim is checked rather than
asserted:

- DSV4 declares `entries_per_req=1` unconditionally (`deepseek_v4_attn.py:1318`). The
  MTP/DSpark lookahead widens *the slot* via `win_with_spec`; it never multiplies the
  count. So `state_slots_per_req == 1`, `pop_many(1)` pops the same index `pop()` did,
  and slot ≡ group exactly as before.
- `v4_pool_geometry.py` and `sub_pool_spec.py` are untouched, so the `_physical_slots`
  reversal and DSV4's pool size are byte-identical to main.
- DSV4 passes no `extra_entries`, so `--state-checkpoint-slots` is inert for it.

Worth stating plainly, because the naming invites the opposite reading: **DSV4 does
use a state slot** — one per request, for its compressor ring and sliding windows,
for the request's whole life. What it does not use a state slot for is a
*checkpoint*. The split is by live-state-vs-checkpoint, not by model:

| | live state | checkpoint |
|---|---|---|
| GDN / SSM (fork) | `StateSlotPool` | another **state slot** |
| DeepSeek-V4 (copy) | `StateSlotPool` | a **PAGE image** in the KV pool |

**The anchor, though, did reach DSV4 — and cost it.**
`PagedStateCheckpointCoordinator.applies()` is true for any V4 seq, so
`_record_checkpoint_end` reserved a prompt-end anchor and `checkpoint_cut` shortened
a prefill chunk onto it. But the coordinator files one pending checkpoint per seq
(`_pending[id(seq)]`, and it `del`s `boundary_blocks`), so the prompt-end checkpoint
landing a chunk later **overwrote the anchor before either was stored**.

Measured on the copy path: one extra prefill chunk per prompt for a hit rate that did
not move (identical re-send 0 blocks either way, continuation 11 either way). Silent,
too — the write really happened, it just got superseded, so no funnel counter
reported it.

The anchor is now gated on `keeps_interior_boundaries`: `StateSlotPool` answers True
(each boundary is its own slot in the index, so both survive), the PAGE coordinator
answers False. Asked as a *capability* rather than by naming the backend, so a future
multi-boundary copy class opts in by answering yes. DSV4 is back to main's one-cut
prefill; GDN keeps the anchor.

---

## Reconciliation with main

**Dropped**: the copy/pending-checkpoint path (`_commit_pending`, `record_copy`,
`take_copies`, `Sequence.pending_checkpoint`, `forget_pending` on the pool). DSV4
checkpointing moved to `PagedStateCheckpointCoordinator` on main, and `StateSlotPool`
now rejects `transfer.copies` outright. The fork path (GDN), where the measured wins
are, is kept in full.

`readable_midstep` is carried on main's `StateTransfer` in `state_runtime.py` (wire
format included) rather than on the branch's now-deleted copy.

**Two bugs this rebase exposed, both fixed here:**

1. `PagedStateCheckpointCoordinator` did not implement the midstep half of the
   `StateCache` protocol, so `checkpoint_cut` raised on every V4 batch. This was
   introduced *by this branch*, not latent in main — `readable_midstep` does not
   exist on main at all, and main's `checkpoint_cut` never consults it. A PAGE image
   is not readable midstep, so the three methods are the no-ops the protocol
   documents.
2. `_record_checkpoint_end` could place the anchor **past the last matchable block**.
   `can_allocate` stops one block short of the prompt (prefill must forward a block to
   produce logits), so a checkpoint filed under the final block's hash is one no scan
   ever looks up — and being stored, it evicted the ladder rung that would have served
   the resume, taking an identical re-request from **8 hit blocks to 0**. Capped at
   `(n_hash_blocks - 1) * hash_block_size`.

---

## New / changed user-facing surface

| flag | default | effect |
|---|---|---|
| `--state-checkpoint-slots N` | `0` | Extra STATE-pool slots beyond what in-flight requests need. One slot per checkpoint whatever `--num-speculative-tokens` says. `0` keeps the old coupling to `max_num_seqs`. `--state-checkpoint-groups` is the old spelling, still accepted. |
| `--state-checkpoint-interval-tokens -1` | `8192` | New third regime: no interval grid, demand + anchor still place checkpoints. `0` and `>0` unchanged. |
| `--no-state-checkpoint-demand` | on | Leaves the prompt-end anchor as the only placement. |
| `ATOM_STATE_CHECKPOINT_DEMAND=0/1` | unset | Overrides `--state-checkpoint-demand` for one run without editing a launch script. Unset costs nothing. |

Config fields: `state_checkpoint_slots: int = 0`, `state_checkpoint_demand: bool = True`.

## Risk

- **Default behaviour unchanged** for GDN/SSM and for DSV4, apart from DSV4 losing the
  anchor chunk it was paying for and not using.
- The rename touches `Sequence`, so out-of-tree code reading `per_req_cache_group`
  breaks loudly (attribute gone) rather than silently — except at the three
  `getattr(seq, ..., -1)` sites in the LMCache DSV4 connector, which were updated and
  which I checked have no other caller shape (no vLLM/SGLang bridge defines that
  attribute).
- Slots for one request are no longer contiguous. Verified the kernels never assumed
  otherwise; `prepare_state_indices` was the only place that manufactured the
  assumption.

## Test plan

**Automated (run on this branch):**

- `tests/test_state_checkpoint.py` — **171 passed**
- state/checkpoint + DSV4/LMCache suites together (`test_page_unit_checkpoint`,
  `test_state_transfer`, `test_lmcache_offload_connector`,
  `test_lmcache_offload_v4_page_slot`, `test_gdn_state_relocation`,
  `test_dsv4_checkpoint_store`, `test_v4_checkpoint_slot_copy`, `test_sub_pool_spec`,
  `test_block_pool`, `test_gdn_state_indices`, `test_per_req_cache_decoupling`) —
  **668 passed, 47 skipped**
- `tests/test_gdn_state_checkpoint_gpu.py` — **16 passed** on GPU (bit-exactness of a
  midstep checkpoint against a cut prefill, parametrized over seed / length /
  boundary). The original 56-combination MI355 sweep behind the `readable_midstep`
  accuracy argument predates this rebase and was not re-run here.
- Full suite: **45 failed / 2301 passed**, a strict subset of `origin/main`'s own 114
  pre-existing failures — **zero regressions**, verified by set difference against a
  clean `origin/main` worktree. (Per repo lore the suite is not CPU-clean; the
  comparison, not the absolute count, is the signal.)
- `black` clean; `ruff` reports no new findings on changed files (14 vs main's 21 on
  the same set).

**Notable new tests:**

- `TestPerNeedWidth` — runs at `entries_per_req=3`, where a request and a checkpoint
  differ in width and the old code could not tell them apart. Every pre-existing case
  ran at 1, where nothing distinguishes them.
- `test_a_last_boundary_only_class_is_not_anchored_for` — pins the DSV4 anchor fix.
  The failure it guards is silent and asymmetric: DSV4 would pay one extra forward per
  prompt for exactly zero hit rate, with nothing in the funnel counters saying so.
- `test_interval_zero_anchors_nothing` — `0` is off for all three placements, not just
  the grid.
- `tests/test_gdn_state_indices.py` — the slot list reaches the kernels unmodified.

**Accuracy**
gsm8k:
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9629|±  |0.0052|
|     |       |strict-match    |     5|exact_match|↑  |0.9636|±  |0.0052|


valarLip:
Reviewed at `4df58434b`. Substantial piece of work — the funnel counters, the `_validate_paged_state_op` checks and the single-entry-point discipline in `state_checkpoint.py` all read well. Findings below are ordered by consequence.

One framing point first, because it changes how the rest should be weighed:

> **The benchmark table measures Kimi-K3, which takes the PAGE path and therefore never executes the new Triton kernel or the midstep reservation machinery.** That machinery fires only for plain GDN backends (Qwen3-Next), which is where findings 3–6 live and where nothing was measured.

---

# Silent wrong output

## 1. `page_unit_checkpoint.py:490` — one boundary's hash is filed over another boundary's bytes

```python
:469   self._pending[(id(seq), h)] = (seq, h)   # re-keyed by this PR
:490   src_slot = seq.state_slot                # resolved at drain, one slot
:493   op = self.store.begin_store(h, src_slot)
```

The re-key to `(id(seq), hash)` is what lets several boundaries of one prompt survive to the drain — that is the point of `test_two_boundaries_of_one_prompt_both_stay_pending`, which sets `seq.state_slots = [0]` and asserts hashes 111 (block 4) and 222 (block 8) coexist in `_pending`. But `take_checkpoint_ops` then resolves a single `seq.state_slot` for all of them, and that live slot holds the state as of the **last** forward. The image indexed under hash 111 therefore contains the recurrent state at block 8.

A later request whose prefix ends at block 4 matches 111, restores it, and continues from a state four blocks ahead. Every check in `_validate_paged_state_op` (layout_id, image_bytes, unit count) passes, because only the content is wrong. Before this PR the seq-keyed dict made it last-write-wins, so the single stored image always matched its hash.

Reachable whenever two `checkpoint()` calls land in one drain — an idle step is enough (`state_maintenance_ops=None` when `scheduled_seqs` is empty, `scheduler.py:1828`). This is the DeepSeek-V4 / Kimi-K3 PAGE path, i.e. the one the PR ships and benchmarks.

## 2. `block_manager.py:1114` — `commit_midstep` publishes unconditionally; the runner declines to write on six paths

`_checkpoint_targets` drops a target on `not 0 <= dst < limit` (`gdn_attn.py:770`) and on `not (start + state_len <= p <= end)` (`:772`), returns None when `found` is empty (`:775`), and is skipped entirely when `batch.block_tables == []` (`:1162`); `attention_gdn.py:404` writes nothing when `pop_last_intermediate_states()` returns None; and `write_state_checkpoints` is only reachable inside `if gdn_metadata.num_prefills > 0`. None of the six report back, and `publish_midstep` does `_index(h, slot)` + `release(slot)` + `checkpoints_kept += 1` regardless.

Concrete: previous chunk ended at `num_cached_tokens=62`, a rung sits at p=64, `state_len=3` — `62 + 3 <= 64` is false, the target is dropped, and the freshly popped slot is filed under the 64-token prefix hash still holding whatever its last owner left. The next request that prefix-hits 64 tokens forks off it.

`_checkpoint_targets`' own docstring names this hazard for the `is_end` case, and the drop conditions immediately below reintroduce it. The comment at `:767-769` — "skipping degrades to no checkpoint, which is always safe" — is not true under this control flow.

## 3. `gdn_attn.py:751` — one index `i` over three differently-scoped sequences

- `state_save_all` / `state_slots_committed` are built over the **filtered** `state_seqs` (`scheduler.py:608-611` keeps only `seq.has_per_req_cache and seq.state_slot >= 0`)
- `num_cached_tokens` / `num_scheduled_tokens` are built over **every** scheduled seq (`plan_midstep` iterates `scheduled_seqs.values()`, `scheduler.py:1682`)
- `rows.append(i)` is consumed by the kernel as an index into `cu_seqlens` / `chunk_offsets`, which are built over the **non-spec (prefill)** rows

One scheduled prefill seq without a claimed state slot — the `-1` sentinel case the filter's own comment anticipates, reachable via offload/PD-parked and remote-KV-promoted requests — shifts every later row by one. `start`/`end` are then read from a different request, so the `start + state_len <= p <= end` window test passes or fails on the wrong lengths, and `rows` names the wrong sequence in `cu_seqlens`/`chunk_offsets`: one request's chunk-boundary state and conv window get copied into another request's checkpoint slot, published, and resumed from.

The existing `if i >= len(runtime_slots): continue` guard is dead (both filtered lists are always the same length) and guards the wrong list; `cached[i]` / `sched[i]` get no guard at all.

## 4. `state_checkpoint.py:81` — the interior SSM read floor-truncates to a 64 grid nothing enforces

`src = chunk_offsets[row] + off // CHUNK` with `CHUNK = 64`, but checkpoint positions are only guaranteed to be multiples of `hash_block_size`, which defaults to `kv_cache_block_size = 16` (`config.py:1542`).

Qwen3-Next at the default `--block-size 16`: an anchor at p=992 gives `off=992`, `992 // 64 = 15`, so the kernel copies `h[chunk_offsets[row] + 15]` — the state after 960 tokens — into the slot then published under the 992-token prefix hash. At p=16, `off // 64 == 0` copies the state *before* any token of this step (zeros on a cold prefill). And `off` is `p - num_cached_tokens`, where `num_cached_tokens` is an arbitrary chunked-prefill boundary, so even a 64-aligned `p` does not save it.

Nothing enforces the grid: `midstep_positions` asserts only `pos % hash_block_size == 0` (`block_manager.py:1085`), and this PR removed the mechanism that used to align chunk ends to rungs for this backend (`checkpoint_cut` now returns 0 when every applicable class is `readable_midstep`). `CHUNK_SIZE`'s own docstring states the invariant — "a checkpoint is takeable at multiples of this and nowhere else" — and no code enforces it. Either assert `off % CHUNK == 0` at the entry point, or restore the alignment.

## 5. `state_checkpoint.py:141` — two incompatible definitions of the conv window width

`write_state_checkpoints` takes `STATE_LEN = conv_state.shape[2]`, which `_state_shape` makes `conv_kernel_size - 1 + num_spec` (`gdn_attn.py:237-240`), while the producer's reachability guard uses `state_len = linear_conv_kernel_dim - 1` with no `num_spec` term (`gdn_attn.py:750`).

Any GDN model with `--num-speculative-tokens > 0` plus prefix caching: a target admitted at exactly `p == start + (W-1)` passes the host guard, but the kernel loads `x[end - (W-1+num_spec) + j]` — `num_spec` tokens further back than the guaranteed history. Inside the batch-concatenated `x` that is the previous sequence's tail; for the first row of the batch, `end - STATE_LEN + j` is negative — an unmasked out-of-bounds read before `x`'s base, since the conv half masks only the feature dim (`mask=dmask`), never the token axis.

The checkpoint is then written with a corrupt conv half and a correct SSM half — precisely the pairing the module docstring says the single entry point exists to prevent ("an SSM state at P paired with a conv window from elsewhere is silently wrong ... because it is findable").

---

# Crashes and leaks

## 6. `scheduler.py:239` — two new asserts turn a stats inconsistency into an engine crash

```python
assert num_cached_tokens <= num_wanted_tokens <= num_compressed_tokens
assert num_compressed_tokens <= num_reusable_tokens <= num_full_tokens
```

These run for every prefill seq from `_schedule_prefill_seq` (`:2151`). `num_wanted_hit_blocks` / `num_compressed_hit_blocks` are set only by `BlockManager.can_allocate` from the GPU block index, but `seq.num_cached_tokens` has four independent writers:

```
scheduler.py:663    self.num_cached_tokens = (...)
scheduler.py:1953   seq.num_cached_tokens = loaded                        # CPU-offload wake
scheduler.py:3332   seq.num_cached_tokens = num_tokens_computed
engine_core.py:953  seq.num_cached_tokens = assignment.num_cached_tokens  # PD assignment
```

`:1953` touches neither counter (verified). An LMCache/hybrid-offload resume that loads more prefix than the GPU index held gives `cached > wanted`, and the `AssertionError` propagates out of `schedule()`, killing the engine mid-run. These are logging counters — the old code reported a nonsensical rate, the new code aborts the server. It also vanishes under `python -O`, so the two builds differ in behaviour. A `logger.warning` + clamp fits the stakes better.

## 7. `gdn_attn.py:552` — the staging buffer is sized on an invariant this PR removed

`_checkpoint_descriptor_buffer` sizes the pinned staging at `2 * max_num_seqs * plan.num_spans`, on the stated rationale "A step can carry at most one store and one restore per sequence" — which was enforced by `_pending` being keyed on `id(seq)`, exactly what finding 1 describes this PR changing. `deepseek_v4_attn.py:1078` is verbatim the same.

With anchor placement on top of the grid rung and the demand rung, a PAGE seq that reaches two boundaries between two `take_state_maintenance_ops()` calls contributes two store ops. Once `len(store_ops) + len(restore_ops) > 2 * max_num_seqs`, `execute_paged_state_copies` trips `total > staging.np.shape[0]` and raises `RuntimeError("a step asked to copy N checkpoints, more than the M its descriptor was sized for")` from inside `build()` — a hard serving failure, not a degradation. Both docstrings still assert the old invariant, and no test covers a multi-boundary PAGE step.

## 8. `block_manager.py:728` — a reservation can outlive its forward and hold slots for the request's life

`commit_midstep` is reachable only from the tail of `hash_blocks`, which has three earlier returns: `not self.enable_prefix_caching`, `start >= end`, and `_chain_parent_hash(...)` returning None (a preempted/re-admitted seq whose earlier blocks lost their hash — the function logs an error and returns). `plan_midstep` is invoked only in the prefill branch of `Scheduler.schedule` (`scheduler.py:1682`).

So once a seq's **last** prefill chunk hits one of those returns, nothing ever drains `seq.midstep_reservations` again until `deallocate`. The slots popped by `reserve_midstep` stay off the free list through the entire decode. Admission gates on `has_free(state_slots_per_req)` and knows nothing about them, so the pool loses capacity it believes it has and starts evicting live checkpoints to serve admissions, with no counter attributing the loss. Meanwhile `state_save_all` re-emits the stale `(slot, position)` pairs on every subsequent decode step.

---

# Policy inversions and lying observability

## 9. `state_pool.py:520` — the speculative demotion is skipped exactly where it matters most

`if anchor and pos != anchor` demotes nothing when a seq has no anchor at all (`checkpoint_end_pos == 0`), so every guess is filed at the LRU tail as if it were knowledge — inverting the policy `mark_speculative` exists to implement. Same expression in `checkpoint` at `:558`.

`_record_checkpoint_end` returns without setting an anchor whenever `isinf(room)`, and floors `end` to 0 for any prompt shorter than the room step-back plus the `(_n_hash_blocks - 1)` cap — i.e. short prompts, the majority of agentic first turns. Those publish ladder rungs and demand rungs with `_speculative` never set, so they land at the LRU tail alongside genuine prompt-end anchors from long prompts and are spent after them. Under pool pressure the 85.2%-read-back anchors get evicted while the 2.8%-read-back rungs survive — the exact measured ordering the mechanism was added to invert. Silent: `checkpoints_evicted` climbs and the hit rate simply fails to improve.

Note also that `publish_midstep(reservations, seq=None)` documents its `anchor = 0` default as "the conservative half of the choice", but the effect of `anchor = 0` is to demote nothing.

## 10. `block_manager.py:1247` — `pool_pressure()` reads the wrong source under PAGE

It reads `self.state.checkpoint_fates()`, but when `checkpoint_spec is not None` (DeepSeek-V4, Kimi-K3 under PAGE — the exact configurations this PR targets) `__init__` sets `state_caches = (paged_state_checkpoints,)`, `self.state` is constructed with `StateTransfer.none()`, and nothing ever calls `self.state.checkpoint()`. `checkpoint_funnel()`, the very next method, correctly uses `self._state_checkpoint_cache`.

So `_log_pressure` prints `kept: 0, dropped: 0, evicted: 0, orphaned: 0` every log interval for the life of the server while `/debug/cache_stats` reports the real numbers from the other source. Two outputs with the same metric names disagreeing is how a tuning session concludes "checkpointing never fires" and turns the wrong knob. Fix is `self._state_checkpoint_cache.checkpoint_fates()`; the coordinator has no `occupancy()`, so only that half moves.

## 11. `offload/hybrid/dsv4/policy.py:142` — the `-1` sentinel is understood by one consumer of two

The new `-1` ("checkpointing on, no interval rungs") is understood only by `BlockManager`. The DSV4 offload policy — the other consumer of `state_checkpoint_interval_tokens` — clamps with `max(0, ...)`, collapsing `-1` into `0`, which for it means "no sidecar checkpoints at all".

A V4 run launched with `--state-checkpoint-interval-tokens -1` keeps placing anchor and demand checkpoints in the engine while the hybrid offload connector computes `checkpoint_interval = 0` and therefore `sidecar_interval = 0`, stopping sidecar boundary alignment and saving entirely. The two halves of the same feature disagree about whether checkpointing is on, and offload resume silently degrades to zero reuse with no warning.

This is one instance of a broader hazard: the sign now encodes three policies read with six different comparisons across the codebase (`== 0`, `!= 0`, `> 0`, `< 0`, `<= 0`, `max(-1, ...)`), with no single "is checkpointing on" predicate.

---

# Structure

## 12. `fla_ops/chunk.py:135` — `ChunkGatedDeltaRuleFunction.last_h` is class-level mutable state used as an out-of-band channel

Correctness rests entirely on "the consumer reads it immediately after its own forward". Anything interposing hands over the wrong tensor with no shape mismatch to catch it: TBO / dual-ubatch microbatching interleaving two GDN forwards (this repo runs TBO for V4 and gpt-oss), a second layer's forward issued before layer N's pop, an exception between the chunk call and the pop, or a draft/target model sharing the module. One microbatch's `h` written into another's checkpoint slots — correctly shaped, wrong tokens, undetectable downstream.

It is also invisible to Dynamo/CUDA-graph capture: under replay the Python attribute write never re-executes, so `pop` returns stale-or-None while the copy kernel is (or is not) replayed. The value belongs in the return tuple or in the metadata object the caller already owns.

## 13. `block_manager.py:516` — `seq.block_hashes` became the single point of failure, because both fallbacks were disabled in the same PR

**(a)** `checkpoint_cut` returns 0 when every applicable class is readable (`:1169`) and `checkpointers_at` excludes readable classes on the aimed path (`:1350`), so `midstep_positions` is the only remaining producer — and it skips any position with `nblocks > len(hashes)`. Any admission reaching `allocate()` without a full `can_allocate` pass (`_promote_ready_remote_kv_requests`, offload-connector resumption, `publish_loaded_prefix`) leaves `block_hashes` empty and produces zero checkpoints for the entire prompt, silently, with no counter moving. That is the coupled-gate collapse `checkpoint_cut`'s own docstring warns about, reached through a third door neither gate guards.

**(b)** The call sits above `if not self._has_page_units(...): return -1`, so a 128k prompt queued behind a full pool burns ~2000 xxhash rounds plus two temp allocations each, per waiting request, per scheduling pass, for a list that is discarded. Also, the comment at `:514` claims the function "reads `checkpoint_end_pos` to decide how far the chain has to reach" — it reads no such thing and unconditionally extends to `_n_hash_blocks(seq)`.

## 14. `fla_ops/chunk.py:115` — returning `h` unconditionally raises the default-config high-water mark

At the default `SUPPRESS_LEVEL < 3` this position previously returned `None`, so `h`'s only reference was the fwd frame and it was freed the instant fwd returned. Now the caller's `forward` frame holds the ~33 MB per-chunk state tensor (the PR's own figure for T=8192, H=8, K=V=128) alive for the rest of that layer's forward, even when `keep_intermediate_states=False` sets `last_h = None` one line later.

That is every GDN prefill with checkpointing off — the default — and every non-ATOM plugin caller (vLLM, SGLang, rtpllm). It raises the allocator high-water mark that KV cache sizing is computed against, for zero benefit. The comment "the kernel computes it either way, so returning it costs nothing" is true of compute but not of lifetime; keeping the conditional in the return tuple costs nothing.

---

# Tests: the new kernel and the whole midstep write path have zero executed coverage

`tests/test_gdn_state_checkpoint_gpu.py` (431 lines) — the only numerical verification of this PR's core — was added and then deleted inside the branch (commit `a04ce7f3e`, "remove gpu unit test"). It covered exactly the defects above: `test_interior_is_exact_not_approximate`, `test_end_checkpoint_does_not_read_the_next_sequences_chunk`, `test_varlen_interior_uses_its_own_sequences_base`, `test_untargeted_slots_are_left_alone`, `test_pop_consumes_so_a_later_step_cannot_read_a_stale_h`. It was CPU-safe (module-level `torch.cuda.is_available()` skip), so removing it bought CI nothing.

`test_gdn_state_indices.py` (154 lines) was the sole coverage of `prepare_state_indices`, the one place a slot number crosses from scheduler into kernel — `grep -rn prepare_state_indices tests/` now returns nothing. Its own docstring stated the stakes: every value written there is a valid index, so a wrong one reads another request's recurrent state with no shape check and no NaN.

The replacements, `tests/test_kda_layout_id.py` and `tests/test_kda_checkpoint_slot_copy.py` (630 lines), sit behind `pytest.importorskip("aiter")`, which the non-GPU CI runner never satisfies — the known failure mode where such modules rot silently. And `TestTheSpareSlotsGoBackToTheKvPool` monkeypatches `GDNStateMixin.state_spec` to a lambda returning a literal `extra_entries=32`, so it asserts on its own stub and would pass unchanged if production stopped reading `state_checkpoint_slots` altogether.

---

# Smaller items

- `[Cache Stats]` denominators changed under unchanged JSON key names (`hit`, `compressed_hit`), so pre/post-PR dashboard numbers are silently incomparable; `paged_hit` is bit-identical to `compressed_hit`; and `chunks_cut_for_end` never reaches `llm_engine`'s aggregation whitelist, so it is always zero in production.
- ~150 lines of the PAGE-copy driver are duplicated between `gdn_attn.py` and `deepseek_v4_attn.py` and have **already** diverged — `_validate_paged_state_op` uses a logical block bound in one and a physical one in the other.
- `_speculative` is not carried across `_rehome_checkpoint`.
- `_has_page_units` still budgets one checkpoint per seq while a seq can now file three.
- `_checkpoint_targets` does five pageable `torch.tensor(...)` H2D copies per prefill step, plus two function-local imports on the per-layer forward path (the latter against the repo's standing no-local-imports rule).
- `clear_index()` bypasses the new eviction counters.
- `_warn_if_unschedulable`'s rename turned `entries // entries_per_req == 0` into `entries == 0`, so a pool narrower than one request now starves silently.
- Docs: `per_req_cache_group` still appears in three guides, including a release snippet that would leak every state slot; the new CLI flags, `ATOM_STATE_CHECKPOINT_DEMAND`, and two `Config` fields are undocumented.

valarLip (approved):

PR #2046, opened 2026-08-27
[Gluon] MOE TP 

k50112113:
This adds Gluon MOE GFX1250 TP case

  ATOM_USE_TRITON_MOE=1
  ATOM_USE_TRITON_MOE_DECODE=1

with no duplication of weight scales for MOE, now flydsl and gluon uses the same weight scales

```
local-completions ({'model': '/data/deepseek-ai/DeepSeek-V4-Pro', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 65, 'max_retries': 3, 'max_gen_toks': 2048, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: 3, batch_size: 1
|gsm8k|      3|flexible-extract|     3|exact_match|↑  |0.9500|±  | 0.006|
|     |       |strict-match    |     3|exact_match|↑  |0.9507|±  | 0.006|
```

image
image

Co-authored-by: Boss2002n <nikhil.kodukula@gmail.com>


atom/model_ops/moe.py
```diff
@@ -1403,66 +1513,46 @@ def apply(
...
-        if self.use_triton:
+        if use_triton_now:
             from atom.model_ops.fused_moe_triton import (
                 triton_kernel_fused_experts,
                 triton_kernel_moe_forward,
             )
 
-            # Check if the model needs custom routing that triton routing()
-            # does not support (grouped topk, sigmoid scoring, bias correction).
+            use_triton_gfx1250_silu = self.is_gfx1250 and (
```
JiaoliangYu: The backend/layout decision should not be derived again here. Architecture, activation type, and weight layout are static properties established during initialization and weight preparation; only the prefill/decode phase is dynamic at runtime. Recomputing the layout condition independently in apply() duplicates policy and risks selecting a kernel that does not match the prepared weights if the two condition sets diverge later. Please derive and store one explicit layout/backend mode during weight preparation, then let apply() only decide whether the current phase uses the prepared Triton or FlyDSL path.

atom/model_ops/fused_moe_triton.py
```diff
@@ -108,6 +106,210 @@ def _resize_cache(x: torch.Tensor, v: tuple[int, ...]) -> torch.Tensor:
...
+        a13_scale,
+        None,
+        w1_bias,
+        routing_data,
+        gather_indx=gather_indx,
+        gammas=gammas if apply_router_weight_on_input else None,
+        swizzle_mx_scale=w13_swizzle_layout,
+        apply_swiglu=True,
+        alpha=1.0,
+        limit=swiglu_limit,
```
JiaoliangYu: `limit=0` means the opposite here than it did on the old path, and it makes the whole MoE layer output exactly zero.

The GGUU path this replaces used `fused_clamp_act_mul`, whose docstring says `swiglu_limit: clamp threshold, to skip set <= 0` -- 0 is the "off" sentinel. But `moe_gemm_a8w4`'s inlined `_swiglu` only tests `if limit is not None`, so `0.0` is taken as a real bound: `gate = min(g, 0)`, `up = max(-0, min(u, 0)) == 0`, and `s * linear` is identically 0. No error, no assert.

`apply()` passes `getattr(layer, "swiglu_limit", 0.0)`. DSv4 / GLM5 / MiniMax do set that attribute on the layer, which is why gsm8k comes out at 0.95 -- but `atom/models/deepseek_v2.py` (R1 / V3) and the Qwen models never mention `swiglu_limit` at all, so they get the 0.0 fallback. And this path is now the **default** for gfx1250 + SiLU; before this PR those models went through branch B's `fused_clamp_act_mul` and were fine.

Suggest normalising the sentinel on the ATOM side: `limit=swiglu_limit if swiglu_limit > 0 else None`. `_swiglu` tests `is not None`, so None is safe and also compiles both clips away.

Side note: `apply()` has two different fallbacks (`0.0` and `7.0`) and this function's signature defaults to `10.0` -- three values for the same knob. Worth unifying while you are here.

atom/model_ops/fused_moe_triton.py
```diff
@@ -108,6 +106,210 @@ def _resize_cache(x: torch.Tensor, v: tuple[int, ...]) -> torch.Tensor:
...
+
+    Defers the persistent decision to aiter's own selector instead of restating
+    its thresholds, so this stays correct if that heuristic moves. The selector
+    reports it as `persistent_iters` -- the count of N-tiles one program walks --
+    and the persistent kernel is the one launched when that exceeds 1, matching
+    moe_gemm_a8w4's own `config["persistent_iters"] > 1` test.
+    """
+    if routing_data is None or getattr(routing_data, "block_m", None) is None:
+        return False
+    return get_kernel_config_gluon(m, n, k, routing_data)["persistent_iters"] <= 1
```
JiaoliangYu, line 166: The implementation does not match the decision table in the docstring above.

The docstring says only `block_m == 16` is a YES and the prefill kernel is a no, but this line only excludes `persistent_iters > 1`. A `block_m` of 32 / 64 / 128 lands on `_moe_gemm_a8w4_prefill` and still returns True -- which is exactly the "silently WRONG, no assert on the aiter side" case the docstring warns about.

The docstring's premise is also stale: since aiter #4836 all three gluon kernels (`_decode_persistent`, `_decode`, `_prefill`) take `HAS_MX_OUT` and all three write `offs_y_mx`. So nothing computes wrong against current aiter, but then the guard is pure overhead (an extra `get_kernel_config_gluon` per GEMM1) and it needlessly disables the fusion for the persistent kernel.

Either drop the guard along with the docstring, or state the aiter version floor explicitly -- if you are pinned below #4836 this is a real bug. Separately, `get_kernel_config_gluon` does `cfg = tuned[key]` with no fallback, so a missing key raises KeyError here.

atom/model_ops/fused_moe_triton.py
```diff
@@ -108,6 +106,210 @@ def _resize_cache(x: torch.Tensor, v: tuple[int, ...]) -> torch.Tensor:
...
+
+    gammas = routing_data.gate_scal if routing_data else None
+
+    # Only gfx1250's gluon kernel consumes the WMMA-preshuffled weight; the
+    # CDNA triton kernel takes a plain (E, K, N) weight.
+    _preshuffled = (
+        (get_arch() == "gfx1250") if preshuffled is None else bool(preshuffled)
+    )
+
+    if envs.ATOM_USE_TRITON_MOE_A4W4 or act_quant == MoEActivationQuant.FP4:
```
JiaoliangYu, line 230: `ATOM_USE_TRITON_MOE_A4W4` is unreachable in any real configuration.

This function is only called with `use_triton_gfx1250_silu=True`, which implies gfx1250, which makes `_preshuffled` default to True, which makes the `assert not _preshuffled` just below fire unconditionally. On CDNA the function is never entered at all, so the flag is silently inert -- yet the assert at `moe.py:1121` presents it as "requires the Triton MoE path", which is misleading.

The only escape hatch, `preshuffled=False`, is never passed by any caller in the tree. Suggest either dropping the flag together with the a4w4 branch, or actually wiring up the CDNA side.

atom/model_ops/moe.py
```diff
@@ -1557,33 +1652,44 @@ def apply(
...
-                # Always-on shared expert(s) via a standalone dense GEMM,
-                # added to the routed output before the TP all-reduce.
+                # Always-on shared expert(s) via a standalone dense GEMM, added
+                # to the routed output before the TP all-reduce. aiter's routing()
+                # does not widen the top-k with the always-on slots, so without
+                # this the shared expert is silently dropped -- and the GUGU prep
+                # does not stash the dense slices _apply_shared_experts_dense
+                # needs, so refuse rather than lose it.
                 if layer.num_fused_shared_experts > 0:
+                    assert not use_triton_gfx1250_silu, (
```
JiaoliangYu: Suggest moving this assert up into `_process_weight_layout_after_loading`.

The condition is fully determined at weight-prep time (branch A does not stash the dense shared-expert slices). Leaving it in `apply()` means startup looks successful and the failure only surfaces on the first token -- and it runs on every forward, on the hot path.

This is also a regression relative to the current behaviour: `num_fused_shared_experts > 0` on gfx1250 used to work via branch B.

tests/test_mxfp4_triton_moe_decode.py
```diff
@@ -0,0 +1,379 @@
...
+    import atom.model_ops.fused_moe_triton as triton_mod
+
+    monkeypatch.setattr(
+        triton_mod, "triton_kernel_fused_experts", fake_triton_kernel_fused_experts
+    )
+    monkeypatch.setattr(
+        moe_mod, "routing_stub_unused", lambda *a, **k: None, raising=False
+    )
+
+    layer = SimpleNamespace(
```
JiaoliangYu, line 301: This `SimpleNamespace` is missing `use_fused_silu_gugu`, which `apply()` reads unconditionally inside the `use_triton_now` branch (`moe.py:1573`).

`test_decode_takes_triton_over_a_view_of_the_flydsl_weights` runs with `is_prefill=False` and so enters that branch -- it should fail with AttributeError. It would only pass on a machine where the aiter import fails and the whole module gets skipped.

atom/model_ops/fused_moe_triton.py
k50112113: fixed

atom/model_ops/fused_moe_triton.py
k50112113: docstring updated

atom/model_ops/fused_moe_triton.py
k50112113: this fix is already at PR 2047, which is closely related to this PR, so I can address this issue there

atom/model_ops/moe.py
k50112113: fixed

tests/test_mxfp4_triton_moe_decode.py
k50112113: fixed

k50112113:
@JiaoliangYu  Please help add 

```
Co-authored-by: Boss2002n <nikhil.kodukula@gmail.com>
```

in the merge comment section when you merge thanks

JiaoliangYu (dismissed):

JiaoliangYu (approved):

valarLip (approved):

JiaoliangYu:
Co-authored-by: Boss2002n <nikhil.kodukula@gmail.com>

PR #2047, opened 2026-08-27
[Gluon] MOE EP

k50112113:


valarLip:
AI review (advisory). Not a findings list — one narrower question: **which of these behavior changes to existing configurations are intentional?**

The PR reads as additive (new EP path, new env flag, new test file), but four existing configurations change behavior. Diffed against merge base `5b78f17b3f`. Three look deliberate and one looks like a regression; I can't tell them apart from the diff alone, and the PR body is empty.

Unchanged, for contrast: **gfx94x/gfx95x TP is byte-for-byte the same path.** `use_triton_gfx1250_silu` and `use_triton_ep` are both False there, so weight prep still takes branch B (`_swizzle_mxfp4`) and `apply()` still takes the same block. Same for any `ATOM_USE_TRITON_MOE=0` deployment.

---

### 1. EP no longer bypasses the modular kernel — deliberate?

```python
-    self.use_triton = gfx.startswith("gfx94") or (gfx95x and ATOM_USE_TRITON_GEMM)
+    use_triton_moe = not ep_moe and (gfx.startswith("gfx94") or ...)
+    self.use_triton    = use_triton_moe and not ep_moe
+    self.use_triton_ep = use_triton_moe and ep_moe
```

On base, `apply()`'s `if self.use_triton:` block (moe.py:1488) returns unconditionally and sits **before** `if self.fused_experts is None:` (1628). So gfx94x + EP, with `use_triton` defaulting True, returned from the TP Triton block carrying `expert_map` and **never reached `self.fused_experts`** — the mori dispatch/combine was skipped even when the transport had been initialised.

The new `not ep_moe` term makes EP fall through to the modular kernel instead. If that is the intent, it's arguably a fix for a pre-existing bug and worth saying so in the PR description, because the observable effect on every existing CDNA EP deployment is that the routed-expert backend changes with no log line. A one-line `logger.info` at selection time would make it diagnosable in the field.

Follow-on: re-enabling Triton there with `ATOM_USE_TRITON_MOE=1` behaves differently at the two EP entry points. The modular-kernel branch (`modular_kernel.py:511`) has no arch gate and runs on CDNA. The local, no-transport branch (`moe.py:1886`) is gated on `is_gfx1250` and so falls through to the assert at `moe.py:1953`. Intentional that only one of them is arch-gated?

### 2. `ATOM_USE_TRITON_MOE_DECODE` — the `is_guinterleave` condition is inverted

```python
-    if envs.is_set("ATOM_USE_TRITON_MOE_DECODE") and not self.is_guinterleave:
+    self.use_triton_decode = (self.use_triton or self.use_triton_ep) and envs.ATOM_USE_TRITON_MOE_DECODE
```

plus the new `assert self.is_gfx1250 and self.is_guinterleave and is_silu` in the prep. Three existing launch lines change:

| config | base | this PR |
|---|---|---|
| `ATOM_MOE_GU_ITLV=0` + `ATOM_USE_TRITON_MOE_DECODE=1` | enabled | armed at init, then **asserts at weight load** |
| `ATOM_MOE_GU_ITLV=1` + `ATOM_USE_TRITON_MOE_DECODE=1` | silently off | enabled |
| gfx95x without `ATOM_USE_TRITON_GEMM` + `..._DECODE=1` | enabled (base does not gate on `use_triton`) | off (new `use_triton or use_triton_ep` prerequisite) |

The first row goes from working to not starting. The new premise (one FlyDSL copy + a zero-copy Triton view, valid only where the two preshuffles agree) explains why GUGU is now *required*, so the inversion itself looks correct — but it is a breaking change to a documented flag, and the assert message doesn't tell an operator that the flag's meaning flipped.

Related: `envs.py` still says `ATOM_USE_TRITON_MOE_DECODE` is *"TP only; inert under EP"* while `moe.py:1114` says *"Now arms under EP as well as TP"*. One of the two is stale.

### 3. gfx1250 + SiLU TP moves from branch B to branch A

`use_triton_gfx1250_silu = self.use_triton and self.is_gfx1250 and is_silu`, and branch A ends in `return`, so this configuration now takes `moe_shuffle_weight(is_guinterleave=True)` + `shuffle_scale_moe` (GUGU, fused SiLU) where base took `_swizzle_mxfp4` (GGUU). Different layout, different kernel, different numerics — for a config that did not opt into anything new.

If that is the intended migration, it would be good to state it; the new test file compares branch A against the FlyDSL view, so it is self-consistent and would not catch a divergence from branch B.

### 4. Regression that falls out of 3: shared experts

Branch B stashes `layer.shared_w13_weight` / `shared_w13_weight_scale` / `shared_w2_*` for `_apply_shared_experts_dense`. **Branch A does not**, and returns early. So gfx1250 + SiLU + `num_fused_shared_experts > 0` now hits `assert not use_triton_gfx1250_silu` (`moe.py:1739`) on the first forward, with no fallback to branch B — a configuration that ran on main.

This one I don't think is intentional.

---

Everything above is from reading the diff plus the base revision; none of it was run on gfx1250 hardware, so please sanity-check before acting. Separately: `pytest tests/test_mxfp4_triton_moe_decode.py` on the PR head gives **2 failed, 5 passed, 4 skipped** — `test_decode_flag_only_narrows_the_triton_path[True-True-True-False]` asserts the EP case is False, which change 2 makes True; and `test_decode_takes_triton_over_a_view_of_the_flydsl_weights` raises `AttributeError: 'SimpleNamespace' object has no attribute 'use_fused_silu_gugu'` at `moe.py:1648`.

A longer list of findings (row-count bounds in `modular_kernel.py`, the mega+Triton path, `scale_kwidth`, an unreachable `combine_scatter_target` branch) is out of scope for this comment; happy to post it if useful.

k50112113:
Note: this branch predates the `shaoclee/gluon_moe_ep` "address comments" commit. 
EP merged TP at `c04eb64d` and `8b3515db` landed after, Two of the things you found are already fixed there and will arrive when `shaoclee/gluon_moe_ep` merges (flagged below). Please merge `shaoclee/gluon_moe_ep` (https://github.com/ROCm/ATOM/pull/2046) first 

1. Taking the logging suggestion. At selection time we now:

**decline Triton MoE under EP on anything other than gfx95x/gfx125x**, with a logger.warning saying so and falling back to the FlyDSL EP path 

logger.info naming the wrapper (a8w4/a4w4).

2. ATOM_USE_TRITON_MOE_DECODE 
Fixed — it now states the requirement inverted and what to change.

My intension is: ATOM_USE_TRITON_MOE_DECODE =1 is only meaningful when both kernels can read one weight copy up to `a .view()`. That's covered, in tests/test_mxfp4_triton_moe_decode.py.

Stale doc: fixed. 

4. gfx1250 + SiLU TP moving branch B → A — intentional
Intentional: branch B was not a gfx1250 path. Stating it in the description.

Your caveat stands and I'm not going to argue it away — the new test file compares branch A against the FlyDSL view, so it is self-consistent and structurally cannot catch a divergence from branch B. A branch-A-vs-branch-B numerics test is the missing piece; noted as follow-up.

5. Shared experts
The refusal is deliberate: branch A's prep genuinely does not stash the dense shared-expert slices _apply_shared_experts_dense consumes

So I fixed on shaoclee/gluon_moe_tp in `8b3515db`: the refusal moved to _process_weight_layout_after_loading

valarLip (approved):

PR #2050, opened 2026-08-27
Reduce mem pressure raised by multiple comm groups init

gbyu-amd:
## Motivation

When launching atom server with multiple parallel strategies, e.g., TP8 + DCP8, each different comm group asks for device memory buffer and could make the total non-torch memory allocation large, thus starving the available kv cache.

Running kimi-k3 with TP8+DCP8, non-torch memory takes 30GB out of the budget, only leaving 18.37GB for available kv cache:
Before:
```bash
[atom 21:34:21] Memory budget: total_gpu=287.98GB, free=63.91GB, utilization=0.86, budget=247.67GB, peak_torch=192.92GB, non_torch=30.21GB, cudagraph_est=0.40GB, safety=5.76GB, available_for_kv=18.37GB, block_bytes=1769472, num_kvcache_blocks=6832
```

By reusing the comm groups with this PR, non-torch memory now only takes 13GB out of the budget, almost doubles the kv cache:
After:
```bash
[atom 04:36:43] Memory budget: total_gpu=287.98GB, free=80.78GB, utilization=0.86, budget=247.67GB, peak_torch=192.91GB, non_torch=13.40GB, cudagraph_est=0.40GB, safety=5.76GB, available_for_kv=35.20GB, block_bytes=2138112, num_kvcache_blocks=13727
```

## Technical Details
An env var ATOM_REUSE_COMM_GROUPS is introduced to decide whether to reuse. 
Note that when enabled, aliased groups issue their collectives on one communicator, so those
collectives serialize against each other. That is already true for TP/DCP/EP, whose
collectives are issued in sequence from the forward pass.

## Submission Checklist


zejunchen-zejun (dismissed):

valarLip:
Reviewed at `3ef2c6f60`. The memory win is real and worth having, but the dedup as written removes expert parallelism on the common configurations, and it is on by default. Ordered by consequence.

---

## 1. EP can essentially never survive the alias — and aliasing it buys nothing

aiter decides whether a communicator is the EP one by substring-matching its name:

```python
# aiter/dist/device_communicators/base_device_communicator.py:123-124
self.is_ep_communicator = "ep" in unique_name
self.use_all2all = self.is_ep_communicator
```

and `_EP` is the **last** group aiter builds:

```
parallel_state.py:1853 _TP   1871 _DCP   1891 _PCP   1918 _DP   1939 _EP
```

Being last means it aliases onto whichever earlier group shares its rank set, and in the two common topologies that always happens:

- `--enable-dp-attention --enable-expert-parallel` (the recipe in `docs/distributed_guide.md:462`, which sets `tensor_parallel_size=1`, `data_parallel_size=tp*dp`): DP is built first with `[[0..7]]`, EP is `[[0..7]]` → `_EP` gets `unique_name="dp:N"` → `"ep" in "dp:N"` is **False**.
- `dp == pcp == 1` (e.g. `-tp 8 --enable-expert-parallel`): EP aliases TP → `"tp:0"` → **False**.

Either way `use_all2all=False`, `communicator_cuda.py:107` never builds the manager, and:

```python
# atom/model_ops/moe.py:598-599
all2all_manager = get_ep_group().device_communicator.all2all_manager
assert all2all_manager is not None
```

fires during weight load, once per MoE module. The gate that reaches it is `use_all2all_kernels` = `dp_size > 1 and use_ep and dp_logical_ratio == 1 and _has_module("mori")` (`moe.py:206-211`) — exactly the config above. `ATOM_REUSE_COMM_GROUPS` defaults to `"1"`, so this is out-of-the-box.

**The part that makes this cheap to fix:** the PR's own motivating win (TP8 + DCP8, dp=1, 17 GB) comes entirely from DCP no longer building its own pynccl + ca + qr. Every other alias in that configuration is a `world_size == 1` singleton, which never had communicators to share. **Excluding EP from the dedup keeps the full 17 GB and removes findings 1–4 outright.**

Two more shapes of the same root cause:

- **`--moe-backend mega`** is gated only on `enable_expert_parallel` (`config.py:1673`), no `dp>1` requirement. `flydsl_mega_experts.py:192-193` does `am = get_ep_group().device_communicator.all2all_manager; rank, world = int(am.rank), int(am.world_size)` → `AttributeError: 'NoneType' object has no attribute 'rank'` on the first forward. Note the comment right above it: that read is load-bearing *because constructing `MoriAll2AllManager` is what initializes the mori symmetric shmem heap*, so a None-tolerant patch would still leave the heap uninitialized.
- **In vLLM/SGLang plugin mode it is silent instead of loud.** `base_device_communicator.py:257-261` is `def prepare_communication_buffer_for_model(self, model): if not self.is_ep_communicator: return`. ATOM's native loader only errors by accident, because `loader.py:368` calls `init_prepare_finalize` directly and that path is skipped under `not is_sglang()`. When the host framework drives `prepare_communication_buffer_for_model` instead, the aliased EP group reports `is_ep_communicator=False` and the call is a no-op — no prepare/finalize installed, no diagnostic.

## 2. Simulated-TP's lies now leak into EP

`apply_simulated_tp` mutates the TP `GroupCoordinator` **in place** (`simulated_tp.py:185-198`): it rebinds `all_gather` / `_gather` / `reduce_scatter_tensor` / `reduce_scatter`, sets `group.world_size = logical`, and at `physical == 1` replaces `all_reduce` with the identity.

With `_EP is _TP`, `get_ep_group().world_size` reports 8 while `get_ep_group().ranks` has 4 entries, and EP all_reduce silently becomes a no-op. `_reject_unsupported` (`simulated_tp.py:89-115`) screens PP / PCP / DCP / DP / TBO / EPLB but has **no EP clause** — because before this PR EP was a distinct object. The module docstring's safety premise ("the one number we lie about is `get_tp_group().world_size`") no longer holds. Repro: `--fake-eplb -tp 8` on 4 GPUs.

## 3. A per-process env var gating a collective decision

`if not envs.ATOM_REUSE_COMM_GROUPS: yield; return` is evaluated independently in every worker. `GroupCoordinator.__init__` (`parallel_state.py:539-545`) calls `new_group` once per subgroup per rank, and that is collective over the default group. At tp=8/dp=1 the alias elides three singleton groups — 48 fewer `new_group` calls on that rank.

Multi-node DP where the second node's launcher does not inherit the env, or someone exporting `ATOM_REUSE_COMM_GROUPS=0` in one shell while bisecting, gives a rank-count mismatch on a collective → **silent startup hang**. Nothing broadcasts or asserts cross-rank agreement, though the world group already exists at that point.

## 4. One communicator, two streams — this repo already has that deadlock on file

The docstring's safety argument ("collectives are issued in sequence from the forward pass") is an assumption nothing enforces. `llm_engine.py:103-121` documents an existing TP↔PCP cross-communicator RCCL deadlock under TBO ("concurrent unordered collectives on comm/compute streams") and works around it by force-disabling TBO. Sharing one communicator between `_EP`/`_DCP` and `_TP` is the same hazard class one level worse: two concurrent `custom_all_reduce` calls also overwrite the same `ca_comm._pool["input"]` staging buffer — **silent wrong values rather than a hang**. Nothing here rejects TBO, dual-stream MoE, or disagg CU-masked streams when aliasing is on.

## 5. Which group owns the shared communicator is decided by aiter's build order, and nothing pins it

`built[key] = group` makes the first-built group win. Today TP is first, so the shared object is `"tp:0"`. If aiter ever reorders `initialize_model_parallel` to build EP before TP, the failure **inverts into something worse**: `_TP` becomes the EP coordinator with `unique_name="ep:0"`, `"ep" in "ep:0"` is True, and every TP-only run would construct a `MoriAll2AllManager` and initialize the mori shmem heap. A cross-repo reordering with no test on either side flips this.

---

## Smaller items

- **Positional forwarding pins aiter's exact 6-parameter signature.** `original(group_ranks, local_rank, backend, use_device_communicator, use_message_queue_broadcaster, group_name)`. aiter is a sibling repo this tracks closely — adding a parameter gives every user a startup `TypeError`, and reordering one binds arguments to the wrong slots *silently*. `*args, **kwargs` with `kwargs.get()` for the two decision fields removes the coupling.
- **The docstring's claim about isolated groups is not true of `_CUSTOM`.** `parallel_state.py:1991` does `_CUSTOM[gname] = init_model_parallel_group(..., group_name=f"custom_{gname}")`. Declaring two custom names over the same partition is exactly how a caller asks for two independent communicators for overlapping streams; with this patch `_CUSTOM["a"] is _CUSTOM["b"]`. Not reachable from ATOM today (`custom_group_config` is never passed), and cross-aliasing with std groups is prevented only incidentally by the `use_device_communicator` equality check, not by a deliberate guard.
- **Only one of three `init_dist_env` call sites is wrapped.** `atom/plugin/register.py:1129` and `atom/rollout/model_runner_ext.py:86` are not, so a globally-named env var means engine-mode-only. A memory number measured in one mode does not transfer, and someone flipping the flag to debug an EP hang may not be changing the path they are running.
- **The `mq_broadcaster` back-fill is unreachable**, as its own "somehow" comment concedes: only TP passes `use_message_queue_broadcaster=True`, and TP is always built first. If the order ever changed, the branch runs `MessageQueue.create_from_process_group` — a rendezvous over `cpu_group` — under a guard that is the object's *mutable state* rather than a rank-invariant input, so the one state that makes it reachable is the one where entering it can be asymmetric → startup deadlock. It also duplicates aiter's `1 << 22, 6` constants with no link. Separately, the inverse invariant is already broken: `_EP.mq_broadcaster` is now non-None (TP's) where it used to be None, so any `is not None` branch on a non-TP group changes behavior today.
- **No test, although the whole decision is pure Python over rank lists** — no aiter, no GPU. `grep -rn "group_reuse\|ATOM_REUSE_COMM_GROUPS" tests/` returns nothing, while `tests/test_simulated_tp.py` is the GPU-free precedent for the sibling module, and `from aiter.dist import parallel_state as ps` is already a function-local import so a `sys.modules` stub suffices. Worth pinning: which of TP/DCP/PCP/PP/DP/EP collide for a given `(tp, pp, pcp, dp, dcp)`; that the order-sensitive key does not alias permuted-but-equal rank sets; that a `use_device_communicator` mismatch refuses to alias; that the monkeypatch is restored when the body raises; that the flag-off path is a true no-op. Please also run it with the patch reverted — green on both sides would mean it is not testing the change.
- **Docs.** `ATOM_REUSE_COMM_GROUPS` is in neither `docs/environment_variables.md` nor the distributed env-var table at `docs/distributed_guide.md:337`; `atom/distributed/group_reuse.py` is missing from the source-file table at `:472`; `:37` still shows the unwrapped `init_dist_env`; and the Qwen3-235B TP8+EP example at `:424` is precisely the configuration where EP aliases TP.
- **The docstring states the wrong trigger.** It says aliasing hits "TP/DCP/EP at `tp == dcp == world_size`, plus the degenerate single-rank PCP/PP/DP". At the default `decode_context_parallel_size=1` DCP is itself a singleton and belongs in the degenerate bucket; and EP aliases TP whenever `dp == pcp == 1`, which has nothing to do with dcp. As written it implies EP only collides in a rare DCP config — the opposite of reality, and the reason finding 1 is easy to miss. Also `:81-86` logs `list(key[0])` (the first subgroup) rather than this rank's, so on a per-node TP layout rank 5 prints `ranks=[0,1,2,3]`; it should print `source.ranks`.

---

## Two ways out

1. **Minimal:** exclude EP from the dedup. Keeps the full 17 GB, removes findings 1–4.
2. **Right altitude:** fix it in aiter. `initialize_model_parallel` statically knows `ep_ranks == tp_ranks`, and already carries this kind of knob (`need_std_comm = custom_group_config is None`, `parallel_state.py:1844`). Sharing `cpu_group` / `device_group` / `pynccl_comm` / `ca_comm` / `qr_comm` while keeping **distinct `GroupCoordinator` objects with correct `unique_name`s** removes findings 1–5 at the root.

One coordination note: `atom/plugin/vllm/tp_group_reuse.py` already implements the safer shape for the same problem — a `GroupCoordinator` subclass reusing the underlying process groups, with `self.unique_name = "tp:0"` set explicitly at `:98` — and it assigns `aiter_ps._EP = getattr(vllm_ps, "_EP", None)` on a path that never enters this context manager. Two group-dedup schemes with opposite designs and no shared code will drift, and which one runs decides whether `get_ep_group()` has an EP-named communicator. Worth converging before either grows further.

gbyu-amd:
Seems not easy to implement it in ATOM, move it to aiter side: https://github.com/ROCm/aiter/pull/5111

gbyu-amd:
> Reviewed at `3ef2c6f60`. The memory win is real and worth having, but the dedup as written removes expert parallelism on the common configurations, and it is on by default. Ordered by consequence.
> 
> ## 1. EP can essentially never survive the alias — and aliasing it buys nothing
> aiter decides whether a communicator is the EP one by substring-matching its name:
> 
> ```python
> # aiter/dist/device_communicators/base_device_communicator.py:123-124
> self.is_ep_communicator = "ep" in unique_name
> self.use_all2all = self.is_ep_communicator
> ```
> 
> and `_EP` is the **last** group aiter builds:
> 
> ```
> parallel_state.py:1853 _TP   1871 _DCP   1891 _PCP   1918 _DP   1939 _EP
> ```
> 
> Being last means it aliases onto whichever earlier group shares its rank set, and in the two common topologies that always happens:
> 
> * `--enable-dp-attention --enable-expert-parallel` (the recipe in `docs/distributed_guide.md:462`, which sets `tensor_parallel_size=1`, `data_parallel_size=tp*dp`): DP is built first with `[[0..7]]`, EP is `[[0..7]]` → `_EP` gets `unique_name="dp:N"` → `"ep" in "dp:N"` is **False**.
> * `dp == pcp == 1` (e.g. `-tp 8 --enable-expert-parallel`): EP aliases TP → `"tp:0"` → **False**.
> 
> Either way `use_all2all=False`, `communicator_cuda.py:107` never builds the manager, and:
> 
> ```python
> # atom/model_ops/moe.py:598-599
> all2all_manager = get_ep_group().device_communicator.all2all_manager
> assert all2all_manager is not None
> ```
> 
> fires during weight load, once per MoE module. The gate that reaches it is `use_all2all_kernels` = `dp_size > 1 and use_ep and dp_logical_ratio == 1 and _has_module("mori")` (`moe.py:206-211`) — exactly the config above. `ATOM_REUSE_COMM_GROUPS` defaults to `"1"`, so this is out-of-the-box.
> 
> **The part that makes this cheap to fix:** the PR's own motivating win (TP8 + DCP8, dp=1, 17 GB) comes entirely from DCP no longer building its own pynccl + ca + qr. Every other alias in that configuration is a `world_size == 1` singleton, which never had communicators to share. **Excluding EP from the dedup keeps the full 17 GB and removes findings 1–4 outright.**
> 
> Two more shapes of the same root cause:
> 
> * **`--moe-backend mega`** is gated only on `enable_expert_parallel` (`config.py:1673`), no `dp>1` requirement. `flydsl_mega_experts.py:192-193` does `am = get_ep_group().device_communicator.all2all_manager; rank, world = int(am.rank), int(am.world_size)` → `AttributeError: 'NoneType' object has no attribute 'rank'` on the first forward. Note the comment right above it: that read is load-bearing _because constructing `MoriAll2AllManager` is what initializes the mori symmetric shmem heap_, so a None-tolerant patch would still leave the heap uninitialized.
> * **In vLLM/SGLang plugin mode it is silent instead of loud.** `base_device_communicator.py:257-261` is `def prepare_communication_buffer_for_model(self, model): if not self.is_ep_communicator: return`. ATOM's native loader only errors by accident, because `loader.py:368` calls `init_prepare_finalize` directly and that path is skipped under `not is_sglang()`. When the host framework drives `prepare_communication_buffer_for_model` instead, the aliased EP group reports `is_ep_communicator=False` and the call is a no-op — no prepare/finalize installed, no diagnostic.
> 
> ## 2. Simulated-TP's lies now leak into EP
> `apply_simulated_tp` mutates the TP `GroupCoordinator` **in place** (`simulated_tp.py:185-198`): it rebinds `all_gather` / `_gather` / `reduce_scatter_tensor` / `reduce_scatter`, sets `group.world_size = logical`, and at `physical == 1` replaces `all_reduce` with the identity.
> 
> With `_EP is _TP`, `get_ep_group().world_size` reports 8 while `get_ep_group().ranks` has 4 entries, and EP all_reduce silently becomes a no-op. `_reject_unsupported` (`simulated_tp.py:89-115`) screens PP / PCP / DCP / DP / TBO / EPLB but has **no EP clause** — because before this PR EP was a distinct object. The module docstring's safety premise ("the one number we lie about is `get_tp_group().world_size`") no longer holds. Repro: `--fake-eplb -tp 8` on 4 GPUs.
> 
> ## 3. A per-process env var gating a collective decision
> `if not envs.ATOM_REUSE_COMM_GROUPS: yield; return` is evaluated independently in every worker. `GroupCoordinator.__init__` (`parallel_state.py:539-545`) calls `new_group` once per subgroup per rank, and that is collective over the default group. At tp=8/dp=1 the alias elides three singleton groups — 48 fewer `new_group` calls on that rank.
> 
> Multi-node DP where the second node's launcher does not inherit the env, or someone exporting `ATOM_REUSE_COMM_GROUPS=0` in one shell while bisecting, gives a rank-count mismatch on a collective → **silent startup hang**. Nothing broadcasts or asserts cross-rank agreement, though the world group already exists at that point.
> 
> ## 4. One communicator, two streams — this repo already has that deadlock on file
> The docstring's safety argument ("collectives are issued in sequence from the forward pass") is an assumption nothing enforces. `llm_engine.py:103-121` documents an existing TP↔PCP cross-communicator RCCL deadlock under TBO ("concurrent unordered collectives on comm/compute streams") and works around it by force-disabling TBO. Sharing one communicator between `_EP`/`_DCP` and `_TP` is the same hazard class one level worse: two concurrent `custom_all_reduce` calls also overwrite the same `ca_comm._pool["input"]` staging buffer — **silent wrong values rather than a hang**. Nothing here rejects TBO, dual-stream MoE, or disagg CU-masked streams when aliasing is on.
> 
> ## 5. Which group owns the shared communicator is decided by aiter's build order, and nothing pins it
> `built[key] = group` makes the first-built group win. Today TP is first, so the shared object is `"tp:0"`. If aiter ever reorders `initialize_model_parallel` to build EP before TP, the failure **inverts into something worse**: `_TP` becomes the EP coordinator with `unique_name="ep:0"`, `"ep" in "ep:0"` is True, and every TP-only run would construct a `MoriAll2AllManager` and initialize the mori shmem heap. A cross-repo reordering with no test on either side flips this.
> 
> ## Smaller items
> * **Positional forwarding pins aiter's exact 6-parameter signature.** `original(group_ranks, local_rank, backend, use_device_communicator, use_message_queue_broadcaster, group_name)`. aiter is a sibling repo this tracks closely — adding a parameter gives every user a startup `TypeError`, and reordering one binds arguments to the wrong slots _silently_. `*args, **kwargs` with `kwargs.get()` for the two decision fields removes the coupling.
> * **The docstring's claim about isolated groups is not true of `_CUSTOM`.** `parallel_state.py:1991` does `_CUSTOM[gname] = init_model_parallel_group(..., group_name=f"custom_{gname}")`. Declaring two custom names over the same partition is exactly how a caller asks for two independent communicators for overlapping streams; with this patch `_CUSTOM["a"] is _CUSTOM["b"]`. Not reachable from ATOM today (`custom_group_config` is never passed), and cross-aliasing with std groups is prevented only incidentally by the `use_device_communicator` equality check, not by a deliberate guard.
> * **Only one of three `init_dist_env` call sites is wrapped.** `atom/plugin/register.py:1129` and `atom/rollout/model_runner_ext.py:86` are not, so a globally-named env var means engine-mode-only. A memory number measured in one mode does not transfer, and someone flipping the flag to debug an EP hang may not be changing the path they are running.
> * **The `mq_broadcaster` back-fill is unreachable**, as its own "somehow" comment concedes: only TP passes `use_message_queue_broadcaster=True`, and TP is always built first. If the order ever changed, the branch runs `MessageQueue.create_from_process_group` — a rendezvous over `cpu_group` — under a guard that is the object's _mutable state_ rather than a rank-invariant input, so the one state that makes it reachable is the one where entering it can be asymmetric → startup deadlock. It also duplicates aiter's `1 << 22, 6` constants with no link. Separately, the inverse invariant is already broken: `_EP.mq_broadcaster` is now non-None (TP's) where it used to be None, so any `is not None` branch on a non-TP group changes behavior today.
> * **No test, although the whole decision is pure Python over rank lists** — no aiter, no GPU. `grep -rn "group_reuse\|ATOM_REUSE_COMM_GROUPS" tests/` returns nothing, while `tests/test_simulated_tp.py` is the GPU-free precedent for the sibling module, and `from aiter.dist import parallel_state as ps` is already a function-local import so a `sys.modules` stub suffices. Worth pinning: which of TP/DCP/PCP/PP/DP/EP collide for a given `(tp, pp, pcp, dp, dcp)`; that the order-sensitive key does not alias permuted-but-equal rank sets; that a `use_device_communicator` mismatch refuses to alias; that the monkeypatch is restored when the body raises; that the flag-off path is a true no-op. Please also run it with the patch reverted — green on both sides would mean it is not testing the change.
> * **Docs.** `ATOM_REUSE_COMM_GROUPS` is in neither `docs/environment_variables.md` nor the distributed env-var table at `docs/distributed_guide.md:337`; `atom/distributed/group_reuse.py` is missing from the source-file table at `:472`; `:37` still shows the unwrapped `init_dist_env`; and the Qwen3-235B TP8+EP example at `:424` is precisely the configuration where EP aliases TP.
> * **The docstring states the wrong trigger.** It says aliasing hits "TP/DCP/EP at `tp == dcp == world_size`, plus the degenerate single-rank PCP/PP/DP". At the default `decode_context_parallel_size=1` DCP is itself a singleton and belongs in the degenerate bucket; and EP aliases TP whenever `dp == pcp == 1`, which has nothing to do with dcp. As written it implies EP only collides in a rare DCP config — the opposite of reality, and the reason finding 1 is easy to miss. Also `:81-86` logs `list(key[0])` (the first subgroup) rather than this rank's, so on a per-node TP layout rank 5 prints `ranks=[0,1,2,3]`; it should print `source.ranks`.
> 
> ## Two ways out
> 1. **Minimal:** exclude EP from the dedup. Keeps the full 17 GB, removes findings 1–4.
> 2. **Right altitude:** fix it in aiter. `initialize_model_parallel` statically knows `ep_ranks == tp_ranks`, and already carries this kind of knob (`need_std_comm = custom_group_config is None`, `parallel_state.py:1844`). Sharing `cpu_group` / `device_group` / `pynccl_comm` / `ca_comm` / `qr_comm` while keeping **distinct `GroupCoordinator` objects with correct `unique_name`s** removes findings 1–5 at the root.
> 
> One coordination note: `atom/plugin/vllm/tp_group_reuse.py` already implements the safer shape for the same problem — a `GroupCoordinator` subclass reusing the underlying process groups, with `self.unique_name = "tp:0"` set explicitly at `:98` — and it assigns `aiter_ps._EP = getattr(vllm_ps, "_EP", None)` on a path that never enters this context manager. Two group-dedup schemes with opposite designs and no shared code will drift, and which one runs decides whether `get_ep_group()` has an EP-named communicator. Worth converging before either grows further.

@valarLip — thanks for the very detailed review. As you suggested, I moved this to the "right altitude" (option 2) and reimplemented it in aiter: https://github.com/ROCm/aiter/pull/5111. No ATOM code changes are needed now. The core idea is exactly what you outlined: when several parallel groups span the same rank set, share    cpu_group / device_group / pynccl_comm / ca_comm / qr_comm while keeping distinct GroupCoordinator objects with correct unique_names.

Issue #25, opened 2025-12-05
Attribute https://github.com/GeeeekExplorer/nano-vllm


amathews-amd:
This project has significant overlap with https://github.com/GeeeekExplorer/nano-vllm , but makes no attribution. 

The nano-vllm project is under MIT license, and requires attribution - https://github.com/GeeeekExplorer/nano-vllm/blob/main/LICENSE 
`The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.`

PR #2051, opened 2026-08-27
[Model] GLM-5.3-Flash (glm5_next) — text path on MI355X

carlushuang:
Adds `atom/models/glm5_next.py` and the engine wiring behind it, so [GLM-5.3-Flash](https://huggingface.co/zai-org/GLM-5.3-Flash) (320B total / 18B active) loads and generates under ATOM on 4× MI355X.

Draft: the text path works and is validated, but context is capped at 2048 tokens and MTP/vision are not wired. See **Scope**.

```bash
python -m recipes.glm5_3_flash.atom_run --model /models/GLM-5.3-Flash -tp 4 \
    --kv-cache-dtype bf16 --enforce-eager --max-tokens 64
```

Needs aiter with `chunk_kimi_delta_attn` and `mla_decode_fwd(causal=...)` — `rocm/atom-dev:nightly_202608270231` or newer.

## What it is

45 text layers: 34 KDA linear-attention + 11 MLA, 4-stream mHC at every attention and FFN site, 288-expert MoE, block-FP8, NoPE throughout.

Mostly assembly of existing ATOM parts — KDA from `kimi_k3.KimiKDAAttention` + aiter `kimi_delta_attn`, mHC from `sparse_attn_v4.hc_split_sinkhorn` + aiter `mhc_*`, MLA via `MLAModules`, MoE via `FusedMoE`, clamped SwiGLU via `swiglu_oai_split`. Three decisions, each exact rather than approximate:

- **DSA layers run dense.** `index_topk=2048` / `index_kpool=4` selects 512 pools covering 2048 tokens, so at or below 2048 the indexer selects everything and dense causal MLA is numerically identical. `deepseek_v2` already gates sparsity the same way.
- **NoPE on a zero-width rope slice.** Zero-padding would also be exact but is impossible: `qk_nope_head_dim` is already 256 and CK caps head dims at 256.
- **KDA output gate folded at load.** GLM factorises it (`g_b_proj @ g_a_proj`) where Kimi-K3 has one `g_proj`; both linear, so materialising the product once lets `KimiKDAAttention` be reused unchanged.

The one new op is `model_ops/kpool_indexer.py` (k-pool compressed DSA indexer). Not on the forward path yet — dense MLA covers ≤2048 — but implemented and validated, ready for the sparse path.

## Validation

Per-layer hidden states vs the transformers reference, cosine of the mHC residual, 21-token prompt, TP4 — **≥0.9997 at all 45 layers, no degradation with depth**, which is the property that separates a sound port from a mis-wired one:

| layer | 0 | 3 | 11 | 27 | 43 | 44 |
| --- | --- | --- | --- | --- | --- | --- |
| cosine | .99972 | .99976 | .99999 | .99999 | .99998 | .99998 |

All 1190 parameters load; no checkpoint tensor dropped. The k-pool indexer selects byte-identical indices to `Glm5NextTextIndexer` on real weights across seq 7–3000 and with left padding, plus 7 CPU unit tests.

Perf (21-token prompt, TP4, `--enforce-eager`, batch 1, no MTP): TTFT 3.12 s, TPOT 45 ms, ~22 tok/s — against 4.25 tok/s for transformers + aiter FP8.

`recipes/GLM-5.3-Flash.md` §5–6 documents how to reproduce every number, plus two measurement traps that produce convincing-looking nonsense.

## Engine wiring

Each following existing precedent: `config.py` (multimodal text-config extraction + bare-`PretrainedConfig` load, as `kimi_linear` does), `model_runner.py` (arch registration; `is_kimi_linear()` — really "hybrid MLA + KDA" — now covers `glm5_next_text`), `llm_engine.py` (`per_req_cache_model_types`), `gdn_attn.py` (0-based KDA layer lists, fp32 temporal state).

## Scope

1. **Contexts beyond 2048** — needs the sparse k-pool path wired into the sparse MLA backend.
2. **`swiglu_limit` in the MoE.** Only the `flydsl`/`mori` paths plumb it; the default `standard`/CK path drops GLM's ±10 clamp. Measured: binds on under 0.001% of elements. Dense layers 0–2 unaffected.
3. **MTP draft layer** (checkpoint layer 45) and **vision tower**.
4. **CUDA graphs** — currently `--enforce-eager`.

## Upstream bugs found

Detailed in `recipes/GLM-5.3-Flash.md` §4; none block this PR. Two are transformers-side (it mis-quantizes all 68 KDA forget-gate linears for this checkpoint; `finegrained-fp8` won't compile on gfx950). The third is worth attention beyond this PR: **aiter kernels launch on the current CUDA device, not the tensor's**, so under a multi-GPU `device_map` they return NaN past the first device boundary — the model loads, runs, and emits garbage. ATOM proper is unaffected (one device per rank).

## Testing

`pytest tests/model_ops/test_kpool_indexer.py` — 7 passed, CPU only, no checkpoint. End-to-end and parity harness in `recipes/glm5_3_flash/` (needs the checkpoint and 4 GPUs). black + ruff clean on all touched files.

gsm8k accuracy:

| num_shot |flexible | strict|
| --- | --- | --- | 
| 3-shot  | 0.9682 | 0.9689 |
| 16-shot  | 0.9659 | 0.9666 |


valarLip:
Reviewed at `6085e6bf1`. Big new-model landing (19 files, +4852/−74), and two things check out that I want to record before the findings, because they are the parts that would have been expensive to get wrong:

- **The cross-repo aiter dependencies are consistent.** `mla_decode_fwd(causal=)`, `chunk_kimi_delta_attn`, `mhc_pre/post`, and the `deepgemm_fp8_paged_mqa_logits` `KVBlockSize % 16` constraint all match at block_size=64 / kpool=4.
- **The KDA gate-fold TP algebra is correct** — `g_a_proj` replicated, `g_b_proj` column-parallel, low-rank inner dim unsharded. The only gap there is the absence of a load-completeness check: the folded `g` shard has no checkpoint tensor of its own, so it escapes `_report_coverage`.

Findings ordered by consequence; I verified the load-bearing ones against the branch.

---

# 1. `relocate_state_slots` reads the same `pairs` two incompatible ways

`kimi_mla_gdn_attn.py:258`. Three pieces side by side:

```python
# base, gdn_attn.py:870 — pairs are RAW SLOT indices
for src, dst in pairs:
    destinations.append(cache[:, dst])

# the new override
super().relocate_state_slots(pairs)                    # moves by raw slot
span = 1 + self.num_spec
for src_group, dst_group in pairs:
    src, dst = src_group * span, dst_group * span      # moves again, by GROUP
```

and the buffer it is moving is allocated one row **per slot**:

```python
# kimi_mla_gdn_attn.py:245-252
out["kpool_tail_cache"] = torch.zeros((
    len(index_cache_layer_ids),
    entries.get(STATE_SLOT_CLASS, 0),   # <- slots, not groups
    2, self._kpool_size(), hf.index_head_dim,
), ...)
```

with the indexer addressing it by raw slot (`tail_cache[state_slot_idx[req_idx]]`, `glm5_next.py:1580`). So relocating slot 5→9 moves the KDA state correctly and then moves tail rows 20..23 → 36..39: **the relocated request's tail is never moved, and four unrelated requests' tails are overwritten.** IndexError near the top of the pool, silent cross-request corruption otherwise. `span = 1 + self.num_spec` is also wrong under replayssm, where `slots_per_req()` returns 1.

Latent until `StateSlotPool.record_relocation` (`state_pool.py:653`) is driven.

# 2. The kpool tail was added to the state entry but not to the checkpoint plane list

`state_spec()` widens the STATE entry by `_kpool_tail_bytes()` (`:221`) and `relocate_state_slots` was extended to move the tail, but the PAGE-copy checkpoint path was not: `GDNStateMixin._checkpoint_plane_shapes` (`gdn_attn.py:580`) still returns exactly two planes (mamba_k, mamba_v), `checkpoint_image_bytes` (`:623`) prices only those, and `_KimiMLAGDNCommon` overrides neither.

Prefix caching defaults ON, and both new CI entries pass `--no-enable_prefix_caching`, so this is untested. A request that publishes a checkpoint mid-pool (`seq_len % 4 != 0`), is evicted, and resumes into a different slot gets conv+ssm restored exactly while the tail rows hold the previous tenant's keys. `_assert_checkpoint_geometry_still_holds` cannot fire, because image and spec are both derived from the same incomplete plane list. The next pooled key is then compressed from two foreign tokens and quantized into the index cache permanently.

# 3. `closes` guards on the flat batch row, cancelling the cross-chunk reconstruction 15 lines above it

`glm5_next.py:1585`:

```python
closes = (abs_pos % kpool == kpool - 1) & (row >= kpool - 1)
```

The second term is a *flat batch row* test, and nothing enforces that chunk boundaries are pool-aligned. Chunked prefill, kpool=4: request A's chunk 2 starts at abs 103 and A is first in the batch, so that token sits at flat row 0. `abs_pos % 4 == 3` holds and `from_tail` has correctly reconstructed slots for abs 100..102 out of `tail_cache` — but `row >= 3` is False, so `closes` is False and **pool #25 is never written**. Nothing revisits it: that cache row keeps zeros (or a previous request's pooled key) for the rest of the sequence, and every subsequent top-k scores a garbage candidate. The `if n < kpool: return` at `:1561` is the same drop.

# 4. The kpool tail is addressed with the write slot, where KDA deliberately uses the in slot

`glm5_next.py:764` uses `gdn.non_spec_state_indices_tensor` — the **write** slot — for both the read of the in-progress pool and its write-back. `KimiKDAAttention._forward_impl` uses the separate `non_spec_state_indices_in_tensor` for exactly this case, and says why (`kimi_k3.py:1126-1140`):

> the prefix-cache hit that forks off a checkpoint ... **Reading the write slot there resumes from whatever the recycled group still held**

So on a prefix-cache hit that forks a KDA state checkpoint, the KDA recurrence resumes from the fork source while the indexer's partial pool is read from the freshly popped (recycled) slot — whatever request last owned it. The layer's compressed key for the current pool is built from another request's tokens, so the sparse selection is wrong for `index_kpool - 1` tokens after every fork. Silent; surfaces only as degraded long-context accuracy under prefix-cache load.

# 5. The custom op returns an input tensor while its fake returns a fresh fp32 one

`glm5_next.py:1970` — all four exits of `_sparse_attn_indexer_kpool` are `return weights`, and the fake at `:1993` is `return torch.empty_like(weights, dtype=torch.float32)`. That is both an aliasing violation and a dtype mismatch, and `deepseek_v2.py:2005-2009` avoids exactly this with a comment naming the hazard:

> Clone (not a bare return of qr) so the runtime output matches the fake's **fresh-tensor contract and never aliases an input**.

Under `--level 1/2/3`, functionalization sees a mutating op whose output aliases a live input: either torch.library's aliasing check raises at trace time, or inductor treats the returned buffer as fresh and reorders it. The caller discards the return value, so the op's only anchor is `mutates_args` — the write→read edge into MLA's `sparse_kv_indices_buffer` is what actually keeps it alive.

# 6. Incomplete sweep on the sparse-count axis

The PR extracted `_sparse_selected_counts()` (`aiter_mla.py:1305`) and routed three numpy sites through it, but two remaining producers of the same quantity still use the raw cap: `mtp_prepare_decode_mla_kernel` is passed `self.index_topk` (`:912`) and recomputes token-granular `min(kv_len, index_topk)` inside Triton, and the DCP path still does `torch.clamp(owned_counts, max=self.index_topk)` (`:1772`).

With `index_kpool > 1` the true per-row count is `min(pools, topk//kpool)*kpool + seq_len % kpool`, i.e. up to `index_topk + kpool - 1`. So on the MTP/spec-decode and DCP paths `sparse_kv_indptr` under-counts by up to `kpool-1` while `expand_pools_and_append_tail` wrote that many entries — MLA reads a truncated candidate set, and **the entries dropped are the always-selected trailing pool, i.e. the most recent tokens**. Neither new CI entry exercises MTP or DCP.

# 7. `torch.empty` output plus an early-return on `slot < 0`

`kpool.py:766` allocates with `torch.empty` while `_kpool_decode_stash_and_pool_kernel` (`:704`) does `if slot < 0: return`, leaving that row never written — and the caller gates the cache write on `closes = (pos % index_kpool) == (index_kpool - 1)` (`glm5_next.py:1916`), derived from `positions`, not from slot validity.

Negative slots are real: `gdn_attn.py:1205/1237` fill `non_spec_state_indices_tensor.gpu[num_decodes:]` with `PAD_SLOT_ID = -1` for CUDAGraph decode padding. A padded row whose `positions` entry happens to be ≡ kpool-1 (mod kpool) gets a valid non-negative slot from `pool_slot_mapping` (which only looks at pool_ids/positions), so **uninitialized device memory is FP8-quantized and written into the index cache as that pool's key**. The same sentinel is handled three different ways in three functions sharing one buffer; `tail_cache[state_slot_idx[req_idx]]` (`glm5_next.py:1580`) has no `>= 0` guard at all, so -1 wraps to the *last* slot rather than being skipped.

# 8. The `get_hf_config` change reaches every multimodal model, not just this one

`config.py:736` — the `except Exception` branch changed from `hf_config._multimodal_config = None` to always synthesizing a `PretrainedConfig` from the raw dict. That applies to every entry in `_MULTIMODAL_MODEL_TYPES`, and `_multimodal_config` is the sole source of `Config.multimodal_config` (`:1859`).

Four existing None-sentinel consumers: `qwen3_5.py:84` (fallback branch), `qwen3_5.py:762` (`raise ValueError('Qwen3.5 multimodal models require multimodal_config')`), `kimi_k3.py:1808/1830`, `multimodal.py:161`. Start Kimi-K3 or Qwen3.5-VL without `--trust-remote-code`, or on any transformers skew where AutoConfig throws: previously an actionable startup error, now a partially-populated object — so `qwen3_5.py:762` no longer raises and `kimi_k3.py:1830` silently takes the `media_placeholder_token_id` default 163605. **A clean startup error becomes a wrong-token-id run.** The synthesized rebuild can also itself raise on nested list-of-dicts, turning a previously-tolerated load into a crash.

# 9. `_ZeroRopePad.__getattr__` lets the DCP/QREP path reach straight through the wrapper

`glm5_next.py:1018` forwards every unknown attribute to the wrapped Linear, so `attention_mla.py:1004-1007`'s `_local_q_proj()` reads `self.q_proj.weight.data` and `self.q_proj.make_row_view(...)` from the **inner** Linear and calls the row view directly at `:1154` — the 64 zero pad lanes are never appended. The wrapper also hardcodes `num_heads = num_local_heads`, which is wrong for the `qrep_num_heads` call at `:1158`.

With DCP query replication on, `self._local_q_proj()(x, x_scale).view(-1, n_heads_out, self.qk_head_dim)` receives a `num_heads * 256` tensor and reshapes to `qk_head_dim == 320`: RuntimeError at startup, or silently re-sliced head boundaries if the element count happens to divide. The wrapper's docstring reasons only about the parameter path, not about callers that read `.weight` and rebuild the projection.

# 10. A falsy-empty-list test where the comment says `is not None`

`aiter_mla.py:1190`: `num_global_mla_layers = len(hybrid_mla_layers) if hybrid_mla_layers else num_hidden_layers`, with the code's own comment stating the intent is absence ("its absence means the dense one-row-per-layer layout"). Same at `:1210`.

This composes with a real asymmetry: `_normalize_glm5_next_config` (`glm5_next.py:167`) falls back to deriving the layer sets from `config.layer_types`, but `GDNStateMixin.__init__` (`gdn_attn.py:136`) reads only `lin.get('full_attn_layers', [])` with no fallback. So a config revision shipping `layer_types` but dropping or renaming `linear_attn_config.full_attn_layers` gives `full_attention_layers = []`: the model still builds 11 MLA + 34 KDA via the `layer_types` fallback, but the backend silently reverts to dense numbering — index-cache transfer regions numbered after 45 nonexistent KV regions instead of 11. Under P/D disaggregation every region id translated through `global_index_layer_ids` points at the wrong layer's bytes: raw-pointer RDMA into another layer's KV, surfacing as garbage decode rather than an error.

# 11. `Glm5NextModel` has no `@support_torch_compile`, and the recipe disagrees with CI

`glm5_next.py:1292` carries no decorator — `grep -n support_torch_compile atom/models/glm5_next.py` returns nothing — unlike its two direct templates (`kimi_k3.py:1534`, `deepseek_v2.py:3013`) and 24 other models. So the model never gets torch.compile or piecewise CUDAGraph at any `--level`.

Meanwhile the two new `models_accuracy.json` entries do **not** pass `--enforce-eager`, even though the recipe's launch command and §8 both say it is required. Either the decorator was forgotten — in which case GLM-5.3 forfeits the whole compilation pipeline and every perf number in the recipe is an eager number — or it was intentional, in which case the `--enforce-eager` in the recipe is cargo-culted. The two documents contradict each other and nothing in code resolves it.

# 12. The ue8m0 scale uses multiply-by-reciprocal where the reference divides

`kpool.py:403`: `_fwht_quant_kernel` computes `tl.exp2(tl.ceil(tl.log2(absmax * (1.0/448.0))))` while `quant_fp8_ue8m0_ref` (`:236`) uses `torch.exp2(torch.ceil(torch.log2(absmax / fp8_max)))`. `1.0/448.0` is not exactly representable, so the two disagree in the last ulp — and `ceil(log2(·))` is a step function on that value.

For any query row whose post-Hadamard absmax lands on or one ulp under a power-of-two multiple of 448, the kernel picks a scale one binade away from the convention aiter's own `indexer_k_quant_and_cache` uses for the **keys**. Query and key scales then differ by 2× for that row, its logits are 2× off, and top-k selects the wrong pools. Input-dependent, silent, and invisible to a ref comparison because the ref never round-trips through fp8. Note this repo already retired this exact `ceil` form once — `quant_v4.py:140-146` replaced it with `f32_to_e8m0` because it "differed ... by up to 1 binade". The fp32 result is also stored into a float8_e4m3fn output with no explicit `.to()`, unlike the sibling `_kpool_pool_rotate_kernel`.

# 13. `no_rope` is never True, so four guards on the shared MLA hot path are dead

`attention_mla.py:606`: `self.no_rope = mla_modules.qk_rope_head_dim == 0`. No model in the tree satisfies it — `glm5_next.py:1059` sets `qk_rope_head_dim = config.qk_rope_head_dim + _ROPE_PAD` = 0 + 64 and passes 64 into `MLAModules`. So `:2397`, `:2437`, `:2631` and the new `and not self.no_rope` term at `:2331` are dead branches.

The docstring at `:600-605` asserts the opposite ("there is no positional component at all ... the KV entry is just kv_lora_rank") and warns that rotating the slices is "at worst an out-of-bounds read", so the guard reads as load-bearing and will be preserved indefinitely. Rope **is** applied on GLM-5.3 today; it is harmless only because `_NoPositionalRotaryEmbedding.forward` (`kimi_k3.py:194`) returns q,k unchanged. And the MLAAttention arithmetic at `:937/1768/1958/2046/2338/2551/2610` works *because* the value is 64 — if anyone ever makes the flag true, every `kv_lora_rank + qk_rope_head_dim` site becomes 512 against a 576-wide cache.

# 14. Roughly a third of the new implementation is unreachable, and nine env vars bypass `envs.py`

Verified dead: `atom/model_ops/kpool_indexer.py` (210 lines) has **no production caller** — `grep -rn kpool_indexer atom/ tests/` finds only a docstring mention at `glm5_next.py:31` and the new 126-line test, **which is the only thing that test exercises**. Also `kpool.py:62-233` of `*_ref` oracles whose banner says they check the Triton kernels but which nothing imports; `_mla_ref_check` / `_mla_ref_check_decode` (`:838/905`, ~150 lines, never called — `_dbg_real_forward` additionally requires `ATOM_GLM5_DEBUG_STATS`, so `ATOM_GLM5_MLA_REF` alone does nothing); ~300 lines of `print()` bring-up probes. Write-only state: `kpool_always_select_tail` (`:598`), `_kpool_warned` (`:613`), `is_nope` (`:1062`), `o_norm_eps` (`:484`). The comment at `:566` — "The pooled scoring path is not implemented yet" — is falsified by this PR.

Concrete costs: a checkpoint setting `index_kpool_always_select_tail: false` is read into an unused field and the tail is appended anyway (silently wrong); the 550 shipped kpool lines have **zero** tests while CI reports the feature green via the dead module; and **`ATOM_GLM5_KPOOL` is not a debug knob — it sizes the index cache** — yet it is read via raw `os.environ`, invisible to `atom/utils/envs.py` (so `ATOM_GLM5_KPOL=0` is a silent no-op) and to `docs/environment_variables.md`. `GLM53_VISION_ATTN` and `GLM53_DISABLE_FUSED_MHC` do not even carry the `ATOM_` prefix. The PR touches no file under `docs/`, so `docs/model_support_guide.md` gets no row for `Glm5NextForConditionalGeneration` where `KimiK3ForConditionalGeneration` has one at line 44.

# 15. Duplication of existing components, and per-forward waste on a 45-layer hot path

Duplicated: `Glm5NextMoE` (`:367-450`) vs `glm4_moe.Glm4MoE` (`:85-197`) — ~90 lines identical down to the fp32 gate and the fusion-shared-expert block; `Glm5NextMLP` (`:332-362`) vs `minimax_m3.MiniMaxM3MLP` (`:197-231`); glm5_next_vl.py's rope / patch-embed / varlen-attn / RMSNorm vs qwen3_5_vl.py and kimi_k3_vl.py (~150 of 362 lines); the `SPARSE_INDEXER_LOGITS_BUDGET_MB` chunk loop (`:1838-1874`) verbatim from `deepseek_v2.py:1570-1613`; `hadamard128_ref` vs `quant_v4.rotate_activation`; the mHC wrapper vs `deepseek_v4.py:4038-4145`. That duplicated budget loop is the fix-then-sweep shape — the next tuning of issue #1376's OOM guard will land in `deepseek_v2.py` only.

Waste, all ×11 MLA layers × every step: `k_pe = torch.zeros(...)` allocated fresh per call (`:1186`) for a constant all-zero tensor; `_ZeroRopePad.__call__` memsets and copies a `[T,H,320]` buffer (`:1007`) that `_drop_rope_pad` immediately slices back to 256 on the prefill path; `torch.full((n_tokens, 2176), -1)` where deepseek_v2 uses `torch.empty`; `int(pool_cu[bs].item())` (`:1814`) is a D2H sync whose comment says "one per prefill" but is one per indexer layer, i.e. **11 pipeline stalls**; `_kpool_write_completed_pools` pools every token instead of every kpool-th (**4× FLOPs**, ~32 MB of transient gathers per layer at 8192 tokens); ~100 layer-invariant decode metadata kernels relaunched per step.

valarLip:
Second pass at `d43d3804c`, this time focused on structure — cleanliness, elegance, performance, and where the boundaries between concerns fall.

**First, the four from the last round are fixed** — I re-checked each against the branch rather than taking the commit messages for it:

| Previous finding | Now |
|---|---|
| `relocate_state_slots` read the same `pairs` as both group and raw-slot indices | Fixed — `tail[:, dst]` / `tail[:, src]`, and the `span = 1 + num_spec` arithmetic is gone |
| `closes` guarded on the flat batch row, cancelling the cross-chunk `from_tail` reconstruction | Fixed — `:1600` is now just `abs_pos % kpool == kpool - 1` |
| `torch.empty` output plus an early return on `slot < 0` | Fixed — `:1944` now gates on `state_slot_idx[:bs] >= 0` rather than on `positions` |
| The custom op returned an input tensor | Fixed — all five exits now `return result` |

Still open from that round: `Glm5NextModel` has no `@support_torch_compile` (`grep -c support_torch_compile atom/models/glm5_next.py` is still 0), and the two `models_accuracy.json` entries still do not pass `--enforce-eager` while the recipe says it is required.

---

# 1. Boundaries

**`atom/models/glm5_next.py` is 2033 lines** — 2.5× the 800-line ceiling in the repo's coding-style rules, from a file that did not exist before. It simultaneously holds:

- a ~500-line registered custom op (`_sparse_attn_indexer_kpool`, a 312-line function with **20 positional parameters**, plus its fake and the `direct_register_custom_op` call) — DSV4's equivalent lives in `model_ops/module_dispatch_ops.py`
- ~270 lines of bring-up debug harness
- config normalization
- the multimodal wrapper

The seams are already visible in the file: `1535-2033` is the custom op, `548-849` the indexer, `1003-1210` MLA, `226-450` mHC/MoE. Moving the op next to the kernels it drives (`atom/model_ops/glm5_next/`) is the one that buys the most, because it also moves the constants below.

**Three derived constants are duplicated across the model/backend boundary with "kept in sync" comments where a function belongs:**

| constant | the two copies | cost of divergence |
|---|---|---|
| `topk_out_width` | `glm5_next.py:629` vs `aiter_mla.py:289` | these are **a producer and the buffer it writes into** — divergence is an unguarded out-of-bounds device write |
| `aligned_index_dim` | `kimi_mla_gdn_attn.py:164` vs `aiter_mla.py:1011` | |
| index-cache allocation (~30 lines) | self-labelled "Mirrors `AiterMLAMetadataBuilder.allocate_kv_cache_tensors`" (`:336`) | `super()` would have given it for free |

Beyond that, `Glm5NextHyperConnection` re-copies the mHC fused/reference dispatch from `deepseek_v4.py:4030`, and `glm5_next_vl.py` re-copies `_sdpa_varlen` verbatim from `atom/diffusion/attention.py:57`, plus RMSNorm, vision rope and clamped SwiGLU from `qwen3_5_vl.py` / `kimi_k3_vl.py`.

**Nine env vars are read with raw `os.environ`** and registered in neither `atom/utils/envs.py` nor `docs/environment_variables.md`. Three of them (`GLM53_DISABLE_FUSED_MHC`, `GLM53_VISION_ATTN`) do not even carry the `ATOM_` prefix, which is how every env knob in this repo is discovered. Four bare `print()` calls remain at `:908`, `:950`, `:992`, `:1682`.

# 2. ~530 lines of dead or duplicate code ship as production — and CI's coverage is on the dead copy

**`atom/model_ops/kpool_indexer.py` (210 lines) has no production caller.** `grep -rn kpool_indexer atom/ tests/` returns exactly two hits: a docstring mention at `glm5_next.py:31`, and its own test file.

It is also a *second* implementation of the same pooling math added in the same PR, and the two already disagree: `kpool_indexer.build_kpools` anchors pools at the first real token (`:69-75`) while the production path addresses them absolutely as `abs_pos % kpool` (`glm5_next.py:1600`).

That matters more than the line count, because **the 7 tests in `test_kpool_indexer.py` are 7 of the only 9 new tests that run in CI at all** (see §4). So the bulk of this feature's enforced coverage validates an algorithm the engine never executes, and will stay green through any regression in the one it does.

**One correction to make here**, since I checked it directly: the `*_ref` family in `kpool.py:66-243` is not quite "zero callers" — `hadamard128_ref` and `quant_fp8_ue8m0_ref` are called by `compress_pools_ref` at `:128-129`. The accurate statement is that **they only call each other, and no test drives any of them**. That matters because the docstrings at `:20`, `:62`, `:289` and `:350` assert these are "the oracle the Triton kernel is tested against" — no such test exists, so a reviewer trusting those comments believes the fused kernels are parity-checked when they are not. Either wire them into a test or delete them and the claim together.

`_mla_ref_check` / `_mla_ref_check_decode` (`glm5_next.py:850-1002`, ~150 lines) are likewise never called.

# 3. Correctness

- **`kpool.py:59` hardcodes `torch.float8_e4m3fn` / `448.0`** while every buffer it interoperates with is allocated with aiter's **arch-dependent** `dtypes.fp8`. On gfx942 (MI300X) that maps to `float8_e4m3fnuz` — max 240, different exponent bias. `fwht128_quant_fp8` (`:428`) then produces an e4m3fn query that `deepgemm_fp8_paged_mqa_logits` reinterprets under the fnuz bias: every logit off by ~2×, and values in (240, 448] become inf/NaN. Nothing asserts. This passes here only because MI355X is gfx950, where the two happen to agree.
- **`aiter_mla.py:1220` — two numbering schemes collide.** `get_kv_transfer_regions` compacts the index-cache offset to `len(full_attention_layers)` but leaves `local_kv_layer_ids` as raw global model layer ids. For GLM-5.3 (`full_attention_layers = [3,7,...,43]`, `num_global_kv_layers = 11`) `block_region_consumer_indices` becomes `[3,7,...,43] + [11,...,21]`: 11/15/19 are emitted twice and 22..43 point past the 22 regions that exist. **The `len(local_kv_layer_ids) != num_layers` guard immediately below passes (11 == 11).** Under P/D the decode node pulls the wrong tensor for a layer — silent KV corruption, not a startup error. The identity mapping in the dense case is why it went unnoticed.
- **`glm5_next.py:711` — `Glm5NextIndexer.forward_impl` drops all PCP handling** that `deepseek_v2.Indexer.forward_impl:2192-2216` performs (`pcp_allgather_rerange` of `k` and of the key positions). The only PCP guard added is inside the kpool op, and with `index_kpool == 1` or `ATOM_GLM5_KPOOL=0` that branch is never taken — so the `NotImplementedError` at `:1732` never fires and the indexer scores this rank's 1/pcp queries against only this rank's 1/pcp keys, writing a partial k-cache with un-gathered positions.
- **`model_runner.py:2845` — `_pp_index_topk` uses the old stride.** This PR changed the per-row width of `_sparse_kv_indices_gpu` to `index_topk_out` (= `ceil128(index_topk + kpool - 1)`) whenever `kpool > 1`, but `_setup_pp_shared_indexer` still sets `_pp_index_topk = hf_config.index_topk` and the consumer at `:2987` slices with it. `is_deepseek_v32` is now True for GLM-5.3 (`:670`), so this path is **newly reachable**: with pp>1 and a shared-indexer layer at the boundary, every row after the first is read at the wrong offset.
- **`kimi_mla_gdn_attn.py:285` — `kpool_tail_cache` is in neither teardown list** (`model_runner.py:1069`, `rollout/memory_manager.py:147`). `release_kv_cache` / rollout sleep-wake frees every other pool and leaves this one resident, and `module.indexer.kpool_tail_cache` (bound at `:542`) keeps pointing at the *old* tensor after a wake reallocates the state pool — so a resumed request reads another tenant's partial pool.
- **`aiter_mla.py:284` — the documented off-switch does not reach the metadata builder.** `AiterMLAMetadataBuilder` derives `index_kpool` straight from `hf_config` and never calls `kpool.pooled_path_enabled()`, while the indexer's `use_kpool()` and `_index_rows_per_block()` both do. With `ATOM_GLM5_KPOOL=0` the builder emits pooled `sparse_kv_indptr` widths while the indexer runs token-granular — **the exact desync that predicate's own docstring says it exists to prevent** — and the only thing between that and wrong attention is `_assert_kpool_regime` raising, so the switch aborts rather than falling back, defeating the pooled-vs-dense A/B the recipe rests on. `_kpool_tail_bytes` (`kimi_mla_gdn_attn.py:224`) has the same gap, which additionally makes two builds with different `ATOM_GLM5_KPOOL` produce different state-slot widths and different checkpoint `layout_id`s (v1 vs v2) — mutually unreadable images.
- **`glm5_next.py:1385` — `packed_modules_mapping` merges the KDA mapping at the top-level multimodal class**, so `.q_proj/.k_proj/.v_proj -> .in_proj` is applied to every checkpoint name including `visual.*`. The identical collision on `.gate_proj/.up_proj` **was** noticed and worked around (`glm5_next_vl.py:_fused_gate_up`, whose docstring explains the hazard); the q/k/v axis got no equivalent. If the tower's on-disk attention is unfused, every such tensor is rewritten to a name no vision module owns, dropped with a warning, and `self.qkv` keeps its `nn.Linear` init — garbage vision embeddings while the text path looks healthy.
- **`glm5_next.py:782` — asymmetric guard.** The `tail_cache is None` branch 20 lines above (`:753-760`) asserts `context.is_dummy_run` before substituting a placeholder; the `state_slot_idx is None` branch silently builds `torch.zeros(n_tokens)`, mapping **every** request to state slot 0. The write mask `closes &= state_slot_idx[req_idx] >= 0` passes (0 >= 0), so nothing errors. The same assert used 20 lines above would catch it.
- **`glm5_next.py:1440` — the 362-line vision tower is unreachable but is built and weight-loaded on every deployment.** `_init_vision_tower` skips it only when `vision_config` is absent or the rank is not first-PP; the checkpoint is multimodal, so the 24-layer ViT is constructed and `model.visual.*` loaded on every TP rank — while nothing upstream produces `pixel_values` (the PR registers `glm5_next_text` as text-only at `llm_engine.py:616`). That VRAM comes out of the `gpu-memory-utilization` budget, i.e. it silently reduces max concurrency. Relatedly, `config.py:736`'s new fallback catches bare `except Exception`, so a `trust_remote_code` refusal or a malformed `config.json` also synthesizes a truthy `_multimodal_config`, destroying the `None` sentinel every other multimodal model uses to report "could not load".

**Carried over from the last round, still open:** `_ZeroRopePad.__getattr__` forwards `.weight` / `make_row_view` to the unpadded inner Linear so `MLAAttention._local_q_proj` reaches past the wrapper (and `Glm5NextMLAAttention` never applies `qrep_tp_override`, which `deepseek_v2.py:2374` does for exactly this path); the sparse-count sweep still leaves `aiter_mla.py:912` and `:1774` on the bare `index_topk`; and `no_rope` is still structurally always False, so the four guards this PR added to the shared MLA hot path are dead — with the identity rotary still imported as a leading-underscore private from `atom/models/kimi_k3.py:181` rather than living in `model_ops/rotary_embedding.py`.

# 4. Tests, and one stale docstring that matters

**Only 2 of the 6 new test files execute in CI; 49 of 58 new tests never run.** `test_glm5_next_kpool_state.py`, `test_kda_checkpoint_slot_copy.py` (34 well-written property tests) and `test_gdn_state_relocation.py` sit behind `pytest.importorskip("aiter")`; `test_mla_prefill_metadata.py` skips on an ImportError of `attention_mla`, which imports aiter and triton at module scope. So the kpool cross-step state and KDA checkpoint-copy coverage this PR adds is never enforced.

Two of the tests that do run are non-discriminating: `test_no_block_beyond_the_image_is_written` asserts `int(tail.sum()) == 0` on slices that are provably empty (image = 528 B, unit = 100 B, exactly 6 units == `N_LOGICAL`), and `test_prefill_tail_fork_copies_prior_rows_into_the_output_slot` allocates `tail` with `torch.zeros` and never seeds the destination slot, so it cannot distinguish "copied" from "happened to be zero".

**And `Glm5NextIndexer`'s class docstring at `:566` still says "The pooled scoring path is not implemented yet".** It is implemented, and `ATOM_GLM5_KPOOL` defaults to `'1'`, so it is the shipped default. That is the one line a reviewer would use to decide whether >2048 context is safe.

---

## Checked and cleared

Recording these so they are not re-litigated: moving the `cu_seqlens_q[:n_seqs+1]` cut into the `else` branch of `attention_mla.py:1799` is **not** a regression — the sparse branch overwrote it anyway (though both callers imply sparse, so that `else` is itself unreachable). And `is_deepseek_v32`'s new `or use_kimi_mla` term is inert for Kimi-K3, which has no `index_topk` in its config.

## What this PR gets right on reuse

Worth saying explicitly, because the duplication above is concentrated rather than pervasive: `Glm5NextKDAAttention` subclasses `KimiKDAAttention` with the low-rank gate folded at load, `Glm5NextIndexer` subclasses `deepseek_v2.Indexer`, and MoE / mHC / MLA all route through the existing `model_ops`. The re-copying is confined to the two off-path files (`kpool_indexer.py`, `glm5_next_vl.py`) and the constants at the model/backend seam — which is why §1 and §2 are worth doing now rather than after the next model lands on top of them.

valarLip:
Third pass, at `42ee4db3b`. The `relocate_state_slots` issue from round 1 is fixed — everything below is new. I verified the load-bearing items against the branch rather than the diff.

Unit suite on the PR head: 4902 passed / 0 failed (4866 on `main`); `black` / `ruff` clean.

---

# 1. The chunked-context K workspace is sized from the raw hf config, bypassing this PR's own `mla_kv_entry_dim`

`aiter_mla.py:548`:

```python
qk_head_dim = hf_config.qk_nope_head_dim + hf_config.qk_rope_head_dim   # 256 + 0
```

The helper this PR added at `aiter_mla.py:73` exists to prevent exactly this, and says so:

> A NoPE model (GLM-5.3-Flash, `qk_rope_head_dim == 0`) materializes the rope block at a padded width and holds it at zero so the standard 576-wide MLA kernels apply unchanged... **Sizing the cache from the raw config instead would allocate 512-wide rows under a 576-wide write.**

The fix was swept into the pool allocator and every flash-attn site, but not here. The same 256-wide buffer is also handed to `:1645`, `:1770` and `:2710`.

Consequence: `gather_kv_b_proj` gets a `k_out` with `KV_PeDim = qk_nope_pe_dim - qk_nope_head_dim == 0`, turning `tl.arange(0, KV_PeDim)` into `arange(0, 0)` — rejected by Triton at compile time — and the cache row stride becomes 512 under a 576-wide entry.

**Reachable by default.** Prefix caching is on (`config.py:1582`) and the chunked branch fires when `sum(batch.num_cached_tokens) > attn_prefill_chunk_size` (16384, `config.py:1546`) — 32 concurrent requests sharing a ~600-token system prompt is enough.

**CI cannot see it.** Both new entries in `models_accuracy.json` pass `--no-enable_prefix_caching` in `extraArgs`.

The PR's own header comment (`glm5_next.py:111-114`) and recipe §4d predict this crash, including the symptom (`NameError('kv_pe_data is not defined')`, "~15 s in").

# 2. Decode's index conversion writes 2176 columns into a region that is at most 2051 wide

`indexer.py:365` passes `NUM_TOPK_TOKENS=topk_out_width` (2176). `_convert_req_index_to_global_index_kernel` (`attention_mla.py:2711-2740`) gates the store on

```python
valid_col_mask = (indice_id < kv_len) & (indice_id < NUM_TOPK_TOKENS)
```

and stores at `out_offset = sparse_kv_indptr[b] + indice_id`. **Nothing bounds it by `sparse_kv_indptr[b+1]`.**

On `main` those two numbers agree exactly (`NUM_TOPK_TOKENS == index_topk == min(seq_len, 2048)`), which is why the missing upper bound has never mattered. Under kpool, `topk_out_width` is rounded up to 2176 while the decode `sparse_kv_indptr` is built from `_sparse_selected_counts` (`aiter_mla.py:2064`) and lands at 2048–2051.

Columns 2048..2175 of `topk_indices` are `-1`, so `out_val` becomes `0` and block 0 is stored over the head of request `b+1`'s sparse KV list — racing with `b+1`'s own program.

Reachable on any `bs > 1` decode with context > `index_topk`, which is exactly the regime the new `GLM-5.3-Flash-kpool-16shot` entry runs.

Worth connecting: that entry's own `_baseline_note` records that "two runs of identical code disagree on ~28 of 1319 questions in each direction". That is currently filed as run-to-run spread. A cross-request write racing between programs would produce the same signature, so it is worth ruling in or out before the number is treated as a noise floor.

# 3. `_is_kda()` was not widened alongside its two siblings in the same file

`gdn_attn.py`, all three on the same axis:

```
:129     if model_type in ("kimi_linear", "glm5_next_text"):     <- widened
:332-333     "kimi_linear", "glm5_next_text",                    <- widened
:373         == "kimi_linear"                                    <- not
```

`_is_kda()` (`:373`) is the sole feeder of `_replayssm_buffer_shapes()` (`:367`), which decides whether `g` is `(num_v_heads, cache_len)` or `(num_v_heads, cache_len, head_k_dim)` (`replayssm.py:111`).

`Glm5NextKDAAttention(KimiKDAAttention)` (`glm5_next.py:478`) reaches the record path with `is_kda=True` (`kimi_k3.py:1220`), so the kernel writes `head_k_dim=128` floats per position into a buffer sized for 1. There is no exception, because `stride_bufg_pos = buf_g.stride(2)` is just 1 on the 3-D tensor — it is a plain OOB device write. `_replayssm_bytes_per_slot` under-prices the state pool by the same factor.

**Does not require speculative decoding**: `ATOM_ENABLE_REPLAYSSM=1` overrides unconditionally (`gdn_attn.py:188-191`, `envs.py:237-241`), and `config.py` only rejects `speculative_config`.

# 4. `--load_dummy` raises on every KDA layer

`glm5_next.py:526`. Shard coverage is tracked only through the monkey-patched `record_input_shard` (`:500-506`) installed over `in_proj.weight.weight_loader`. But `loading_core.py:286-292` does `if load_dummy: continue` **before** dispatch, so that loader never runs and `_loaded_input_shards` stays empty.

`loader.py:361-367` then calls `process_weights_after_loading()` on every module unconditionally, so `:522-528` raises

```
RuntimeError: Incomplete GLM-5.3 KDA input projection: missing checkpoint shards [0, 1, 2]
```

for all 34 KDA layers, in all three `--load_dummy` modes (`empty`, `zero`, `xavier` — `arg_utils.py:262-272`).

The parent `KimiKDAAttention` has no record-based guard, so this makes GLM-5.3-Flash the only model in the tree that cannot start with `--load_dummy`, which is what capacity and profiling smoke runs use. Gating the check on `config.load_dummy` is enough.

# 5. The two debug knobs have guards that are wrong in opposite directions

`glm5_next.py:917`.

**(a) `ATOM_GLM5_FORCE_DENSE_MLA` has no length guard at all.** It sets `is_sparse=False, indexer=None, run_indexer=False` (`:922`, `:937-939`). Its own comment bounds correctness to "Below index_topk" — but the only length check, `_assert_kpool_regime` (`:632-644`), lives in `Glm5NextIndexer.forward_impl`, which force-dense never calls.

So a >2048-token request attends over all KV instead of the trained top-k selection, silently, while the knob is being read as an indexer-isolating A/B. That is the worst possible failure mode for a debugging tool: it changes the function under test without saying so.

**(b) Conversely, `_assert_kpool_regime` raises `NotImplementedError` from inside `forward_impl` (`:659`)** on a user-controllable prompt length. `engine_core.run_engine` has no per-batch handler, so the first oversized prompt kills the engine process and every concurrent request with it — while `envs.py:151-152` and `docs/environment_variables.md:78` both say "longer requests are refused", which reads as a per-request 4xx.

---

## Host-side and device-side row counts with nothing tying them together

`indexer.py:213`. The pooled prefill gather's destination row count comes from `attn_metadata.kpool_total_pools` (host), while the source bounds come from a `pool_cu` recomputed independently on device. There is no assertion that they agree, and a mismatch is an out-of-bounds read, not a quality loss.

`cp_gather_indexer_k_quant_cache` grids over **destination** rows and resolves each row's sequence by scanning `cu_seq_lens` into a `__shared__ int batch_idx[]` that is left **uninitialized** for rows past `cu_seq_lens[batch_size]`, then uses it as `block_table[batch_idx * num_blocks + ...]`. So `max_pools != pool_cu[-1]` in either direction indexes the block table out of bounds.

They agree today only because both branch identically on `has_cached` and read the same buffers — one edit to `full_seq_lens` (`aiter_mla.py:1387-1418`) breaks it silently.

Separately, `bs = cu_q.shape[0] - 1` (`:203`) is `running_bs` after `backends.py` pads the indptrs, whereas `attn_metadata.block_tables` was copied at `bs = batch.total_seqs_num_prefill`. Nothing ties the two.

Two cheap lines: assert `int(pool_cu[-1]) == max_pools`, and take `bs` from `pool_bt.shape[0]`.

## The folded KDA output gate has no load-coverage check, and the fold destroys the evidence

`glm5_next.py:534`. The q/k/v shards get a hard `RuntimeError` immediately above (finding 4's guard); `g_a_proj` / `g_b_proj` get nothing.

`ReplicatedLinear` / `ColumnParallelLinear` allocate with `torch.empty` (`linear.py:509-511`), and the loader's unloaded-parameter report is `logger.warning` only (`loading_core.py:427-439`) — its own docstring concedes "the destination keeps its init value ... and the forward pass is quietly wrong".

So if a checkpoint revision renames or nests `g_a_proj` / `g_b_proj`, then `g = g_b.float() @ g_a.float()` folds uninitialized memory into `in_proj[3*lp : 4*lp]`. The `assert g.shape == ...` passes — **it checks shape, and the shape is right; only the values are junk** — and `:544-545` then does `m.weight.data = m.weight.data.new_empty(0)`, so nothing downstream can re-check. Every KDA layer's output gate becomes garbage with no error at any point.

Track `g_a`/`g_b` coverage the way q/k/v are tracked, or fold-then-free only when both are confirmed loaded.

## `g_a_proj` / `g_b_proj` hardcode `quant_config=None` with no assertion

`glm5_next.py:508`. `kimi_k3.py:978-982` asserts `quant_type == QuantType.No` for `(in_proj, b_proj, f_a_proj)`; the two new modules are outside that tuple.

The checkpoint is block-FP8 e4m3 [128,128], and **the recipe itself (§4a, lines 125-127) records that this checkpoint's `modules_to_not_convert` uses a `model.layers.` prefix that does not match the real key prefix** — i.e. the exclude list is already known to be unreliable for these `self_attn` sub-projections.

If they were quantized, `default_weight_loader` takes the `numel() == numel()` branch (`loader.py:48-49`) and copies raw fp8 codes into a bf16 param unscaled, while the orphan `weight_scale_inv` is dropped with a warn-only `dropped_ckpt_keys` append. Silent wrong output gate on all 34 KDA layers.

Add the same `QuantType.No` assertion, or read `_effective_layer_quant` instead of hardcoding `None`.

## The config fallback can produce a model with zero full-attention layers, and the completeness check passes

`glm5_next.py:156`. If both `kda_layers` and `full_attn_layers` are absent and `layer_types` is empty:

- Block 1 (`:157-163`) sets `kda = all_layers - set(full)` with `full` empty → `kda` becomes all 45.
- Block 2 (`:164-169`) then reads that and sets `full = all_layers - set(kda)` → `[]`.

The check at `:172-181` tests duplicates, disjointness and completeness. "All vs nothing" satisfies all three. The `layer_types` cross-check at `:182` is skipped because the list is empty.

Result: `num_full_attn = 0`, every `Glm5NextDecoderLayer` takes `is_linear_attn = True` (`:995`), and the model builds with no MLA/DSA layers and no error.

The fallback exists precisely for config revisions that omit the redundant lists — which is exactly when it fires. A `num_full_attn > 0` check closes it.

## "How many KV entries the indexer selects" now has three implementations, one of them kpool-aware

`aiter_mla.py:937`. `_sparse_selected_counts` (`:1342-1369`) computes `min(pools, index_topk // kpool) * kpool + seq_len % kpool`. The other two sites implement the token-granular `min(kv_count, index_topk)`:

- `mtp_prepare_decode_mla_kernel` gets a bare `self.index_topk` (`:936`)
- `_apply_pcp_reindex` does `torch.clamp(owned_counts, max=self.index_topk)` (`:1824`)

Under kpool both under-count every row by the unscored tail, so `sparse_kv_indptr` rows start at the wrong offsets — wrong tokens attended, no crash.

Unreachable today only because `config.py:2038-2043` rejects speculative decoding and PCP for Glm5Next. Lifting either rejection is a config-file edit, not an attention change, and it makes this live. Worth declaring the axis in one helper rather than leaving two token-granular copies behind — this is the same shape as finding 3.

## `is_deepseek_v32` widening newly exposes Kimi-Linear to the sparse path

`model_runner.py:670` widened from `use_mla` to `(use_mla or use_kimi_mla)`, and `use_kimi_mla` now covers `kimi_linear` too. Since `use_mla` returns False for `kimi_linear` (`:883-899`), the bare `hasattr(hf_config, "index_topk")` was previously unreachable for it.

A Kimi-Linear config revision carrying `index_topk` — or `--hf-overrides '{"index_topk": N}'` — now makes `sub_pool_specs` price index-cache bytes into every page unit (`kimi_mla_gdn_attn.py:162`), `allocate_kv_cache_tensors` allocate a full fp8 index cache (`:341`), `_page_unit_regions` include those bytes in the checkpoint image geometry, and `AiterMLAMetadataBuilder` emit sparse metadata for a dense model — while `kimi_k3.py:661` hardcodes `indexer=None`, so nothing is ever bound or written.

Every consumer is a soft `if` / `getattr`; the only `raise` fires in the opposite direction. Gate on indexer presence, or on the GLM model-type set.

## `NoPositionalRotaryEmbedding` allocates ~128 MiB/rank it never reads

`glm5_next.py:1086` constructs it with `max_position_embeddings` (1M for this model) rather than `max_model_len`. `RotaryEmbedding.__init__` (`rotary_embedding.py:46-54`) builds and registers `[max_pos, 1, 1, rotary_dim // 2]` for both cos and sin: at `_ROPE_PAD = 64` and 1M positions that is 2 × 1048576 × 32 × 2 B = 128 MiB on device — while `NoPositionalRotaryEmbedding.forward` returns `(query, key)` immediately.

**The PR's own recipe measured this exact constant**: lines 331-333 record that sharing this cache across 11 MLA layers dropped `peak_torch` from 42.71 to 41.47 GiB, i.e. 127 MiB per copy. One copy is still there.

It cannot simply be deleted — the `ATOM_GLM5_KPOOL=0` path passes `cos_cache` / `sin_cache` at `:769-771`, though the adjacent comment says they are never read — but it can be sized to `max_model_len`, or to one row. `kimi_k3.py:640-643` already uses the cap elsewhere.

## An unanchored `weights_mapping` substring rewrite, dropped with a warning

`glm5_next.py:1194`: `"index_kpool_compress_gate": "index_kpool_compress_gate.weight"`.

`weight_names.py:181-183` applies each mapping entry as a plain `name.replace(part, replacement)`, before the `weight_scale_inv` rename. A checkpoint revision shipping the gate as a proper Linear (`...index_kpool_compress_gate.weight`) or adding a quant sidecar yields `...gate.weight.weight` / `...gate.weight.weight_scale`; `weight_dispatch.py:288-291` appends to `dropped_ckpt_keys`, and `loading_core.py:445-468` only warns — so the `ReplicatedLinear` param stays at `torch.empty` garbage and pooled selection degrades with no load-time failure.

Docs-match-code: `recipes/GLM-5.3-Flash.md:91-92` still claims "its whole `weights_mapping` is one rule: `model.language_model.` -> `model.`", but there are two entries.

---

## Tests

**Essentially none of the new test surface runs upstream.** `tests/test_glm5_next_kpool_state.py:12` (238 lines), `tests/model_ops/test_glm5_kpool_kernels.py:9`, `tests/test_gdn_state_relocation.py:20` and `tests/test_kda_checkpoint_slot_copy.py:31` all sit behind module-scope `importorskip("aiter")` / `importorskip("atom.models.glm5_next")`, and CI has no aiter/triton/GPU. So the k-pool kernels, the config normalizer, the tail-plane geometry and the checkpoint slot arithmetic have zero coverage on the runner that gates merges.

Three specific problems inside that:

- **`test_kda_checkpoint_slot_copy.py:199-206` states the opposite of what happens.** The comment says the second `importorskip` is placed there "so the mixin tests above still run" — a module-scope `importorskip` skips the whole module regardless of position, so all 34 tests vanish, not 14.
- **`tests/test_glm5_model_style.py` asserts on the source text of the code it tests** (`".item()" not in source`, `"torch.save" not in source` over `glm5_next.py` and `indexer.py`). That is the repo's own "a test may not grep the code it tests" rule; it is also trivially defeated and trivially false-positived by a comment.
- **`tests/test_mla_prefill_metadata.py` passes verbatim on `main`.** It covers `_forward_prefill_mla`, whose body is byte-identical between `main` and the PR head — so it certifies a fix that is not in this diff.

Findings 1, 2, 6 and the config-fallback finding are all reachable from pure-Python or single-kernel tests that need neither the model nor a GPU.

## Also confirmed

- **`kpool.py` ships four functions nothing calls** — `compress_pools_ref` (`:103`), `expand_pools_to_tokens` (`:125`), `expand_and_append_tail_ref` (`:156`), `append_tail_to_topk` (`:190`), about 118 lines. Their docstrings present them as oracles (`:271`: "`pool_compress_ref` is the oracle"), and the fused kernel's docstring at `:543` names two of them as the thing it fuses — but no test compares against any of the four. The fifth, `pool_compress_ref`, does have one caller (`tests/model_ops/test_glm5_kpool_kernels.py:29`), in a file CI never runs. Meanwhile `recipes/GLM-5.3-Flash.md:100` asserts "there is no second, dead indexer implementation".
- **The module docstring cites a symbol that does not exist.** `glm5_next.py:28` supports the PR's load-bearing claim — that dense is numerically identical at or below `index_topk` — with `` (``deepseek_v2._pcp_sparse_active``: ``max_seqlen_k > index_topk``) ``. A repo-wide grep for `_pcp_sparse_active` returns exactly one hit: that citation. The claim may well be true, but its stated evidence is not checkable.
- **`Glm5NextHyperConnection` forks DeepSeek-V4's mHC and drops both of its fused paths.** `deepseek_v4.py:4046-4050` binds `aiter.mhc_fused_post_pre` when available, and `:4098` uses `_rmsnorm_nw` (`triton_rmsnorm_nw.py`). Neither appears in `glm5_next.py` — two extra kernel launches per site, ~90 sites per forward.
- **`Glm5NextMoE` forks `Glm4MoE` and `Glm5NextMLP` forks `MiniMaxM3MLP`**, and `is_kimi_linear()` / `use_kimi_mla` no longer describe what they select (they now gate GLM-5.3 too — the same drift finding 10 exploits).
- **`config.py:2038-2043` rejects PCP, DCP, speculative decoding and TBO for Glm5Next, but not `pipeline_parallel_size > 1`**, where the hybrid backend keys a PP-local counter into global-layer-id maps.

## Refuted during verification

Recording these so they are not re-raised: ruff I001; the `clamp_min` slot-contamination theory; the seed-tail fork rows; CUDA-graph pad rows; the dropped `ENABLE_ALLREDUCE_RMSNORM_FUSION` guard; and the `layer_types` filter affecting other models.

valarLip:
Fourth pass, at `a883ea8b1`. The branch has moved only by an annotation commit since round 3, so this comment covers what is new rather than restating open items.

**Still open from round 3, not re-argued here:** `_is_kda()` not widened alongside its two siblings (`gdn_attn.py:373`); no load-coverage check on `g_a_proj` / `g_b_proj` before the fold (`glm5_next.py:526` guards q/k/v only); `_assert_kpool_regime` raising from inside the forward; the unanchored `weights_mapping` rewrite; `NoPositionalRotaryEmbedding` sized at `max_position_embeddings`; and the three divergent implementations of the selected-count. Happy to re-explain any of them.

---

# 1. Two new test files abort the entire CI unit-test run

`tests/test_mla_chunk_workspace_width.py:13` and `tests/models/test_glm_moe_router_dtype.py:13` import at module scope with no `importorskip`:

```python
from atom.model_ops.attentions.aiter_mla import mla_kv_entry_dim, mla_qk_head_dim
from atom.models.deepseek_v2 import _moe_router_dtype
```

Both modules pull native deps at import time on this branch:

```
aiter_mla.py:10-11    import triton
                      from aiter import (...)
deepseek_v2.py:30     from aiter import (...)
```

Reproduced by blocking `aiter` and `triton` on `sys.meta_path`:

```
FAIL  atom.model_ops.attentions.aiter_mla: ImportError: No module named 'triton'
FAIL  atom.models.deepseek_v2:             ImportError: No module named 'aiter'
```

`.github/scripts/run_unit_tests.sh` runs on a plain runner with, in its own words, "no GPU, no aiter/MoRIIO native libs", and `tests/conftest.py` states "Nothing here fakes a module". pytest's default — no `--continue-on-collection-errors` — is to abort before executing anything, so **all ~4140 unit tests report as not-run**, and the PR's own new GLM tests never execute either. Green locally, red-and-empty in CI.

The intent is right and both docstrings say so — "These are CPU-only: they pin the dtype decision and the numerical fact behind it, without needing a GPU or a 408 GB checkpoint" — the import path just did not follow. Either move `mla_kv_entry_dim` / `mla_qk_head_dim` / `_moe_router_dtype` into an aiter-free module, or defer the aiter import in their current homes.

# 2. `_moe_router_dtype` misses the MTP draft, and the correct predicate is in the same file with the reason

`deepseek_v2.py:349`:

```python
if getattr(config, "model_type", None) == "glm_moe_dsa":
    return torch.float32
```

Two other GLM predicates in this file carry a fallback, and the comment above one of them (`:2128-2131`) states exactly why:

> The MTP draft rewrites model_type to deepseek_mtp but retains the GLM-only `index_share_for_mtp_iteration` marker.

```python
is_glm52 = getattr(config, "model_type", None) == "glm_moe_dsa" or bool(
    getattr(config, "index_share_for_mtp_iteration", False)
)
```

`SpeculativeConfig._MTP_TYPE_MAP` (`config.py:1120`) maps `glm_moe_dsa -> deepseek_mtp`, so the target's router builds `e_score_correction_bias` at fp32 while its MTP draft keeps bf16 — the collapse this PR's own test `test_bf16_destroys_a_glm_shaped_correction_bias` says destroys the selection signal (238 distinct values onto 8).

Draft and target then pick different experts and MTP acceptance drops. The symptom is a throughput regression with GSM8K still inside threshold, which is the hardest shape to attribute later. (Per this repo's own guidance, acceptance rate is the probe that would catch it — GSM8K is not.)

# 3. The router-dtype change is unconditional on three shipped models, and the riskiest part is not gated pre-merge

`deepseek_v2.py:1213-1215`:

```python
router_logits = (
    self.gate(hidden_states)
    if self.router_dtype is None
    else self.gate(hidden_states, otype=self.router_dtype)
)
```

`self.router_dtype` resolves to `torch.float32` for every `glm_moe_dsa` model — GLM-5, 5.1 and 5.2, all already shipping. The model this PR adds does not use `DeepseekV2MoE` at all.

That is also the only explicit `otype=` call site in the tree that can carry fp32: `grep -n "otype="` over `atom/` returns `linear.py:978`/`:1016` (forwards), `deepseek_v4.py:1151`/`:1281` (comments), and this line. Everything else takes the `otype=dtypes.bf16` default at `linear.py:935`/`:965`. So `tgemm.mm(x, w, bias, otype=fp32)` on a bf16×bf16 GEMM is a first for this repo — if aiter's tuned GEMM has no fp32-output instantiation it raises on the first MoE layer; if it silently falls back, numerics change.

Combined with the bias-width change, expert selection shifts on three shipped models whose CI entries are `test_level: nightly` — i.e. the highest-risk part of this diff is neither gated pre-merge nor exercised by the new model's own runs. A `pr`-level entry on one GLM-5.x model, or a measured acceptance-rate comparison, would close that.

# 4. The fake-EPLB synthetic router table is still bf16 while the bias is now fp32

`moe.py:2984` builds the balance table at `torch.get_default_dtype()` (bf16) and `:4272-4273` substitutes it for `router_logits`.

Per the docstring added by this PR at `deepseek_v2.py:344-347`, aiter's `biased_grouped_topk` dispatches on `gating_output.dtype()` and reinterpret_casts `correction_bias` to that same `scalar_t`. So with fake-EPLB on any `glm_moe_dsa` model the kernel reads a 4-byte-per-element fp32 buffer as 2-byte bf16: garbage bias, wrong expert selection, and a half-length read of an `n_routed_experts`-long buffer.

This is the invariant the PR itself states, violated by a path the PR did not sweep.

# 5. `hybrid_mla_layers` treats an empty list as the hybrid layout

`aiter_mla.py:1237`:

```python
hybrid_mla_layers = getattr(runner, "full_attention_layers", None)
```

`[] is not None`, so an empty list takes the hybrid branch. `GDNStateMixin` assigns `full_attention_layers = [int(i) - offset for i in lin.get("full_attn_layers", [])]`, which is empty whenever the config omits the key — the same fallback shape as the config-normalizer gap from round 3.

Then `num_global_mla_layers == 0`, `num_global_kv_layers` counts only draft layers, and every index-cache region in `get_kv_transfer_regions` is numbered on top of the KV regions instead of after them. P/D disaggregated transfer writes index-cache bytes into KV rows.

The `len(local_kv_consumer_indices) != num_layers` check cannot catch it: both sides derive from the same empty list. The pre-PR expression (`num_global_kv_layers = num_hidden_layers + num_global_draft_layers`) had no way to be wrong.

# 6. `_drop_rope_pad` hands non-contiguous slices to five FA call sites

`attention_mla.py:909` returns `t[..., : self.qk_nope_head_dim]` straight into `flash_attn_varlen_func` at `:1285`, `:1403`, `:1457`, `:1587`, `:1722`, with no `.contiguous()`.

For GLM-5.3, `q` comes from `_ZeroRopePad` as `[T, 320*H]` and `k` from `torch.cat((k_nope, k_rope.expand(...)))`; after the slice both have `stride(-2) == 320` while `size(-1) == 256`.

If aiter/CK's varlen FA derives the row pitch from `size(-1)` rather than `stride(-2)`, every head after head 0 reads shifted data — a finite, plausible-looking, wrong attention output that no shape check catches. If it instead calls `.contiguous()` internally, this is a silent full copy of K per chunk per layer. Either answer is worth stating in a comment, since five shared sites now depend on it.

Two of the five also rebind `prefill_q` inside a loop, which is correct today only because the slice happens to be idempotent.

# 7. `_ZeroRopePad` is a non-Module callable in a slot typed `nn.Module | None`

`glm5_next.py:782` assigns it to `MLAModules.q_proj`, which shared MLA code dereferences as a Linear: `attention_mla.py:1003` reads `self.q_proj.weight.data` and `:1006` calls `self.q_proj.make_row_view(...)` on the QREP/DCP prefill path (also `:2427`). Both raise `AttributeError`.

Unreachable today only because `config.py` rejects DCP for Glm5Next — i.e. the guard that makes the wrapper safe lives in a different file with no comment linking them, and DCP is listed as planned follow-up. It is also invisible to every `nn.Module` walk: `.to()`, dtype/quant post-load traversal, `named_parameters`, and `atom/rollout/memory_manager.py:139`'s module sweep.

Separately, its `if isinstance(y, tuple): y = y[0]` discards a returned activation scale with no comment.

# 8. The new PR-level accuracy gate is measured to flake

`.github/benchmark/models_accuracy.json:475`. `GLM-5.3-Flash-kpool-16shot` is `test_level: pr` with `accuracy_threshold: 0.96`, and its own `_baseline_note` records repeat runs of identical code at 0.9613 / 0.9629 / 0.9644 / 0.9659 — the worst clears by 0.0013.

So every unrelated PR now runs an 8-GPU 16-shot GSM8K gate that the author has measured at roughly a 1-in-4 chance of failing on unchanged code. The standard response is re-run-until-green, at which point the gate stops carrying information about the pooled path it was added to protect. Either lower the threshold to sit below the measured spread, or move it to `nightly` where a flake costs a re-run rather than a blocked PR.

The note also refers to a short-context `GLM-5.3-Flash` entry that exists neither on `main` nor at this head.

Separately in that file, three unrelated entries (gpt-oss-120b ×2, Llama-3.3-70B) were reformatted `0.90 -> 0.9` — the fingerprint of a `json.load`/`json.dump` round-trip over a shared CI manifest, in a file where several entries embed JSON inside `--online_quant_config '{...}'`. Worth reverting those lines so the diff does not carry unrelated churn.

---

## Not a defect — recording it so a fifth round does not re-raise it

**`indexer.py:76`'s `safe_slots = read_slots[req_idx].clamp_min(0)` cannot read another request's tail.** This has now been proposed once and refuted once across review rounds, so here is the producer-side evidence.

`state_slot_idx_in` is `gdn.non_spec_state_indices_in_tensor` (`glm5_next.py:703-705`), filled at `gdn_attn.py:1021-1022`:

```python
src = fork_srcs[idx] if idx < len(fork_srcs) else -1
non_spec_state_indices_in[idx] = src if src >= 0 else committed
```

It is never negative: either a valid fork source, or `committed = slots[0]`. On the `with_spec` branch it stays at the `= 0` initialisation. So the `-1` sentinel never reaches the clamp, and the "publish gate checks the output slot, not the input slot" asymmetry has no input that exercises it.

Note this is a *different* clamp from the round-3 finding ten lines above it — `gather_idx = ((rows - (K-1)).clamp_min(0)[:, None] + offsets[None, :])` at `:65`, which clamps before adding the slot offsets. **That one is still open.**

## Smaller, all verified

- `stable_topk` stays False for `glm5_next_text` at tp8, because `Indexer` keys it on `glm_moe_dsa` — the same predicate gap as finding 2, third instance.
- `is_deepseek_v32` widened to `use_mla or use_kimi_mla` admits `kimi_linear` on a bare `hasattr` (latent: `/data/Kimi-K3/config.json` carries no `index_topk`).
- `record_input_shard` narrows `MergedColumnParallelLinear.weight_loader`'s declared `int | tuple | None` contract via `int(shard_id)`.
- Uninitialised `torch.empty` for `k_fp8` / `k_scale` / `pool_topk`.
- `triton_mla.py`'s new `**kwargs` silently swallows future keywords.
- `glm4_moe` / `kimi_k3` / `deepseek_v4` / `minimax_*` are all left on the lossy bf16 bias path this PR argues is wrong — if the argument holds, it holds for them.
- The DP all-gather round-trips the new fp32 router logits through bf16.
- `Glm5NextMoE.gate` is a bare `nn.Linear(requires_grad=True)` inside a `@support_torch_compile` model, bypassing `atom_parameter` / `weight_loader`.
- The mHC `% 512 or % 256` disjunct is dead.
- `config.py` matches `"Glm5Next" in str(a)` as a substring, and silently overrides a user block size using the raw rather than the effective `index_kpool`.

## Conventions

- `tests/test_glm5_model_style.py` asserts on the source *text* of the modules it tests.
- `tests/model_ops/test_glm5_kpool_kernels.py` gates on `importorskip("aiter")` + triton, so CI never runs it once — worth pairing with finding 1, since the fix for one is the diagnosis for the other.
- Four commits carry `Co-Authored-By: Claude`, several more `Co-authored-by: Cursor`.
- `glm5_next.py` is 1257 lines and `kpool.py` 790, with no stated exception to the 800-line soft ceiling.
- `docs/model_ops_guide.md` was not updated for the new `atom/model_ops/glm5_next/` package.

## Refuted during verification

Beyond the `clamp_min` item above: the `num_sched_blocks` truncation (it is exact); the "missing `ENABLE_ALLREDUCE_RMSNORM_FUSION` guard" (GLM-5.3 never uses that fusion); the KDA gate fold under quantization (the parent asserts `in_proj` is unquantized); and "the `glm5_full_attn_layers` branch is unreachable" (the model is built at `model_runner.py:751`, before the builder at `:786`).

valarLip (approved):

PR #2052, opened 2026-08-27
Clarify no-tools parser behavior and improve documentation

Jasen2201:
## Submission Checklist

### DIFF
--- a/atom/entrypoints/openai/serving_chat.py
+++ b/atom/entrypoints/openai/serving_chat.py
@@ -72,11 +72,10 @@ def normalize_chat_tools(tools: Any) -> Any:
 def _tool_parser_for_request(parser_cls, tools):
     """Return the parser this request actually needs.
 
-    SGLang only activates tool parsing when the request has effective tools.
-    Doing otherwise lets an ordinary ``<tool_call>`` quotation open an
-    unbounded region on a no-tools request, withholding the rest of the stream
-    until EOS. Keep parsers that also consume model-wide channel framing,
-    though: Kimi-K3 wraps every answer in such markers, even without tools.
+    A no-tools request cannot produce a dispatchable tool call. Disable parsers
+    whose markers only open tool-call regions so an unclosed marker remains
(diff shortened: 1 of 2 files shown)


valarLip (approved):

ThomasNing:
### Independent gfx950 validation

I validated the exact PR head `67d959853e03fa97302f744c177687268f87b56e` in a patch-free unified ATOM/AITER/SGLang image on an 8x MI350X (gfx950) host.

Tested ATOM path:

- Model: `amd/GLM-5.2-MXFP4`
- ATOM native server, TP=4, MTP with 3 speculative tokens
- FP8 KV cache and the production online quantization exclusions
- Candidate isolated to host GPUs 4-7 while the existing TP4 service remained on GPUs 0-3
- Image provenance: ATOM `67d95985`, AITER `636c6663`, SGLang `29481685`, no local patches

Results:

- `tests/entrypoints/test_serving_chat.py`: **98 passed**
- `TestNoToolsMeansNoToolReadAhead`: **3 passed**
- Exact-source/import/ATOM-entrypoint/SGLang-entrypoint verification: **passed**
- Live GLM-5.2 TP4+MTP server reached healthy state
- A no-tools streaming request returned the terminal `finish_reason` chunk, usage chunk, and `[DONE]`; no output-buffering stall was observed

Scope note: this validates the GLM no-tools path addressed by this PR. The open mixed framing/tool-parser concern (for example KimiK3Parser) is still valid and is not covered by the GLM result; a framing-only parser variant or equivalent marker filtering would close that generic gap.

CI note: the native ATOM `GLM-5.2-MXFP4 MTP` accuracy job passed. The currently blocking SGLang matrix jobs fail at **Docker Login**, and the vLLM GLM-5.2 job fails at **Download aiter wheel**, before model launch/inference. Those failures appear to be CI setup failures rather than evidence of a regression in this change.

This PR head also contains merged #2012, so advancing the runtime pin preserves the speculative-output truncation fix.

PR #2090, opened 2026-08-31
fix(mla): guard the new per-seq q-cum slice on kv_last_page_lens

ganyi1996ppo:
0b4f1ddba started cutting `paged_cu_seqlens_q` down to the per-seq width by reading `attn_metadata.kv_last_page_lens.shape[0]`. That field is None on the sparse path, so every rank dies on its first prefill batch:

  File "atom/model_ops/attention_mla.py", line 1743, in _forward_prefill_mla
    n_seqs = attn_metadata.kv_last_page_lens.shape[0]
  AttributeError: 'NoneType' object has no attribute 'shape'

Reproduced on GLM-5.2-MXFP4, TP=8, DCP=8: 8 requests arrive, the first prefill batch is scheduled, and ModelRunner1/8 aborts in layer 0 with zero completions.

None is correct there, not a metadata bug. Twelve lines below, `is_sparse_mla` overwrites all four of the values this block computes -- paged_cu_seqlens_q, paged_kv_indptr, paged_kv_indices and kv_last_page_lens -- with their sparse_* counterparts, and the DCP sub-branch sizes its indptr off `sparse_cu_seqlens_q.shape[0]`. The dense per-seq array is never read on that path, so a builder that only serves sparse attention has no reason to fill it. The new line dereferences it before the branch that would have discarded it.

Keep the slice for the dense path and skip it when the array is absent.

Verified: GLM-5.2-MXFP4 TP=8 DCP=8, 65536-token inputs -- 8/8 requests, 32 prefill batches, no traceback, 34.36 s (34.38-34.50 s on the runs before 0b4f1ddba landed).

## Submission Checklist

### DIFF
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -819,13 +819,14 @@ class PagedAttentionImpl(nn.Module):
             # Raw K/V is fed as a block_size=1 flash-layout cache, never shuffled.
             shuffled_kv_cache = False
 
-        # Number of requests this step scheduled, taken from the context rather
-        # than from a companion array's shape. prepare_prefill fills exactly
-        # cu_seqlens_q[:scheduled_bs + 1] and context_lens[:scheduled_bs] and
-        # pads the tail out to running_bs, so this is that width by definition.
-        # block_tables.shape[0] was an indirect proxy for it, and one that is
-        # optional besides -- the base builder uploads a table only when
-        # `has_cached`, so the old narrowing had to special-case its absence.
+        # Requests this step scheduled: the width prepare_prefill fills in
+        # cu_seqlens_q and context_lens before padding the tail to running_bs.
+        #
+        # context.scheduled_bs is batch.total_seqs_num, while the builder sized
+        # these arrays by total_seqs_num_prefill. The two differ only on a batch
+        # carrying decode rows, and such a batch leaves total_tokens_num_prefill
+        # at 0 -- hence is_prefill False, which is the branch this method is
+        # reached from. Every is_prefill batch sets the two counts equal.
         n_seqs = fwd_ctx.context.scheduled_bs
         unified_attention(
             q,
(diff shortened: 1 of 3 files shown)


atom/model_ops/attention_mha.py
```diff
@@ -819,16 +819,29 @@ def prefill_attention_triton(
...
+        block_tables = attn_metadata.block_tables
+        n_block_rows = None if block_tables is None else block_tables.shape[0]
         unified_attention(
             q,
             k_for_attn,
             v_for_attn,
             o,
             # Cut to the rows of the block table this indexes with them: the
             # prefill builder pads both past the requests it scheduled.
-            cu_seqlens_q=attn_metadata.cu_seqlens_q[: n_block_rows + 1],
```
valarLip, line 830: replace by fwd_ctx.context.scheduled_bs

ganyi1996ppo: done

atom/model_ops/attention_mla.py
```diff
@@ -1746,8 +1740,17 @@ def _forward_prefill_mla(
...
+        #
+        # Guarded on kv_last_page_lens: AiterMLAMetadataBuilder.prepare_prefill
+        # only fills it when a drafter exists or the batch has a prefix-cache
+        # hit, so a plain first-chunk prefill leaves it None. That was harmless
+        # while the value was merely passed through -- the sparse branch below
+        # overwrites all four of these, and the dense kernels take None -- but
+        # slicing it here raised AttributeError on the first prefill batch.
+        paged_cu_seqlens_q = attn_metadata.cu_seqlens_q
+        if attn_metadata.kv_last_page_lens is not None:
+            n_seqs = attn_metadata.kv_last_page_lens.shape[0]
```
valarLip: same fwd_ctx.context.scheduled_bs?

valarLip (changes requested):

atom/model_ops/attention_mla.py
ganyi1996ppo: done

valarLip (approved):

PR #2095, opened 2026-08-31
[gfx1250] route wo_a grouped LoRA through the FlyDSL A8W8 BMM

JiaoliangYu:
## Summary

Builds directly on [ROCm/ATOM#1833](https://github.com/ROCm/ATOM/pull/1833), preserving its gfx1250 `wo_a` integration and unified inverse-RoPE structure while replacing the A8W4 backend with A8W8.

- gfx1250 uses the FlyDSL A8W8 BMM by default; `ATOM_WO_A_USE_FLYDSL=0` explicitly restores the BF16 fallback.
- Eligible `wo_a` weights remain in their checkpoint FP8 format instead of being dequantized and requantized to MXFP4.
- Weight preparation only applies the required 16x16 preshuffle; the existing e8m0 128x128 block scale is retained.
- Forward reuses `inverse_rope_group_quant(..., quant_group_size=128, scale_layout="row")` and dispatches through `batched_gemm_a8w8_mxscale_bpreshuffle`.
- The existing gfx950 Opus mxscale path is unchanged.
- The new AITER symbol is guarded so non-gfx1250 environments remain importable during staged rollout.

Depends on [ROCm/aiter#5041](https://github.com/ROCm/aiter/pull/5041).

## Test plan

- [x] Black formatting and Python compile checks
- [x] Environment default/override coverage
- [ ] Install an AITER build containing ROCm/aiter#5041
- [ ] Run DeepSeek-V4 gfx1250 accuracy with the default A8W8 path
- [ ] Benchmark A8W8 versus `ATOM_WO_A_USE_FLYDSL=0` BF16 fallback


atom/models/deepseek_v4.py
```diff
@@ -2706,23 +2723,33 @@ def process_weights_after_loading(self) -> None:
...
+                # The A8W8 FlyDSL BMM expects a 16x16 preshuffled weight. Its
+                # block scale remains in [G, N/128, K/128] e8m0 layout.
+                shuffle_weights(w, layout=(16, 16))
             # Cached as module attrs so the forward skips the reshape and the
             # scale conversion on every call.
             self._wo_a_w_fp8 = w.data.view(G, N, K)
             self._wo_a_w_scale = w_scale
             self._wo_a_mxscale = True
             if self.layer_id == 0:
+                backend = "bpreshuffle FlyDSL" if self._is_gfx1250 else "opus"
```
junhaha666: There is no need to distinguish between the flydsl or opus backends. Just indicate whether it is a preshuffle.

atom/models/deepseek_v4.py
```diff
@@ -2986,13 +3013,23 @@ def _wo_a_grouped_lora(
...
             # (same layout as the old out= buffer); N is contiguous so the
             # flatten below is a free view.
-            y = batched_gemm_a8w8_mxscale(
-                x_fp8,
-                self._wo_a_w_fp8,
-                x_scale,
-                self._wo_a_w_scale,
-                dtype=o.dtype,
-            )
+            if self._is_gfx1250:
```
junhaha666: There can be dispatch by “preshuffle" flag

atom/models/deepseek_v4.py
JiaoliangYu: Updated the log to report only `preshuffle=True/False`, without exposing the FlyDSL/Opus backend distinction. Addressed in cbdfafca.

atom/models/deepseek_v4.py
JiaoliangYu: Updated the forward path to select the BMM callable using the local `preshuffle` flag and invoke it once with the shared arguments. Addressed in cbdfafca.

valarLip (dismissed):

PR #2097, opened 2026-08-31
build(mooncake): enable HIP dma-buf MR in release image

junyyang-amd:
## Purpose

On Crusoe ionic HCAs, registering HIP GPU KV memory through `ibv_reg_mr` falls back to the legacy GPU-VA path and fails with `EINVAL 22`. This PR builds Mooncake with HIP dma-buf support so TransferEngine can register HIP GPU memory through `ibv_reg_dmabuf_mr`, while preserving the runtime option to use `ibv_reg_mr` when required.

## Summary

- Pin upstream Mooncake at a commit that supports HIP dma-buf memory registration and build TransferEngine with `USE_HIP_DMABUF=ON`.
- Disable the unused Mooncake Store, Rust bindings, unit tests, and examples to keep the ATOM release image build focused and avoid HIP hipify test-header failures.
- Fail the image build when CMake does not enable HIP dma-buf support or the installed libraries do not reference `ibv_reg_dmabuf_mr`.

## Memory registration selection

- Enable `ibv_reg_dmabuf_mr` for HIP GPU memory (the image default):
  ```bash
  MOONCAKE_DISABLE_HIP_DMABUF=0
  ```
- Force the legacy `ibv_reg_mr` path:
  ```bash
  MOONCAKE_DISABLE_HIP_DMABUF=1
  ```
  Any non-zero value disables HIP dma-buf registration. On Crusoe ionic HCAs, forcing this path for GPU memory may reproduce `EINVAL 22`. CPU/host memory continues to use the normal `ibv_reg_mr` path.

`ibv_reg_dmabuf_mr` also requires Mooncake to be compiled with `USE_HIP_DMABUF=ON`; the release image enables this at build time.

## Test plan

- [ ] Build `docker/atom_release.dockerfile` with the default Mooncake configuration and verify the CMake dma-buf check passes.
- [ ] Run a Mooncake PD workload on a Crusoe ionic node with `MOONCAKE_DISABLE_HIP_DMABUF=0` and verify GPU MR registration succeeds without `EINVAL 22`.
- [ ] Set `MOONCAKE_DISABLE_HIP_DMABUF=1` and verify Mooncake selects the legacy `ibv_reg_mr` path.

### DIFF
--- a/.github/benchmark/models_atomesh.yaml
+++ b/.github/benchmark/models_atomesh.yaml
@@ -17,8 +17,6 @@ defaults:
     common:
       PYTHONUNBUFFERED: "1"
       ATOM_PD_RANK_MAPPING_POLICY: "${ATOMESH_PD_RANK_MAPPING_POLICY}"
-      HSA_NO_SCRATCH_RECLAIM: "1"
-      IONIC_LOCKFREE: all
 
 backends:
   atom:
--- a/.github/benchmark/models_atomesh.yaml
+++ b/.github/benchmark/models_atomesh.yaml
@@ -17,6 +17,8 @@ defaults:
     common:
       PYTHONUNBUFFERED: "1"
       ATOM_PD_RANK_MAPPING_POLICY: "${ATOMESH_PD_RANK_MAPPING_POLICY}"
+      HSA_NO_SCRATCH_RECLAIM: "1"
+      IONIC_LOCKFREE: all
 
 backends:
   atom:
--- a/.github/benchmark/models_atomesh.yaml
+++ b/.github/benchmark/models_atomesh.yaml
@@ -386,7 +386,7 @@ models:
         kv_cache_dtype: fp8
         block_size: 16
         enable_prefix_caching: false
-        gpu_memory_utilization: 0.9
+        gpu_memory_utilization: 0.8
         max_num_seqs: 256
         max_num_batched_tokens: 32768
         max_model_len: 32768
--- a/.github/benchmark/models_atomesh.yaml
+++ b/.github/benchmark/models_atomesh.yaml
@@ -386,7 +386,7 @@ models:
         kv_cache_dtype: fp8
         block_size: 16
         enable_prefix_caching: false
-        gpu_memory_utilization: 0.8
+        gpu_memory_utilization: 0.9
         max_num_seqs: 256
         max_num_batched_tokens: 32768
         max_model_len: 32768
--- a/.github/scripts/atomesh/pd_slurm_job.sh
+++ b/.github/scripts/atomesh/pd_slurm_job.sh
@@ -196,6 +196,7 @@ EOF
     -v "${RUN_DIR}":/run_logs/slurm_job-"${JOB_ID}"
     -v /mnt:/mnt
     -v /data:/data
+    -v /home/junyyang/models:/home/junyyang/models:ro
   )
 
   if [[ "${rank}" -eq 0 \
@@ -548,6 +549,7 @@ for execution_phase in "${EXECUTION_PHASES[@]}"; do
         -v /mnt:/mnt \
         -v /data:/data \
         -v /it-share:/it-share \
+        -v /home/junyyang/models:/home/junyyang/models:ro \
         "${nested_docker_args[@]}" \
         "'"${DOCKER_IMAGE}"'" \
         bash -lc "cd /workspace/ATOM && bash .github/scripts/atomesh/pd_server_atom.sh" \
--- a/.github/scripts/atomesh/pd_slurm_job.sh
+++ b/.github/scripts/atomesh/pd_slurm_job.sh
@@ -196,7 +196,6 @@ EOF
     -v "${RUN_DIR}":/run_logs/slurm_job-"${JOB_ID}"
     -v /mnt:/mnt
     -v /data:/data
-    -v /home/junyyang/models:/home/junyyang/models:ro
   )
 
   if [[ "${rank}" -eq 0 \
@@ -549,7 +548,6 @@ for execution_phase in "${EXECUTION_PHASES[@]}"; do
         -v /mnt:/mnt \
         -v /data:/data \
         -v /it-share:/it-share \
-        -v /home/junyyang/models:/home/junyyang/models:ro \
         "${nested_docker_args[@]}" \
         "'"${DOCKER_IMAGE}"'" \
         bash -lc "cd /workspace/ATOM && bash .github/scripts/atomesh/pd_server_atom.sh" \
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -1838,6 +1838,20 @@ class ModelRunner:
         return pool
 
     def allocate_kv_cache(self, num_kvcache_blocks):
+        # NCCL/RCCL communicators are created lazily on the first collective
+        # and allocate scratch via hipMalloc (not the PyTorch caching
+        # allocator). Do that collective *before* packing the GPU with KV,
+        # otherwise gpu_memory_utilization=0.9 leaves no room for those
+        # buffers and the post-alloc barrier OOMs. Skip world_size==1: a
+        # no-op barrier still forces communicator creation there.
+        if (
+            torch.distributed.is_initialized()
+            and torch.distributed.get_world_size() > 1
+        ):
+            warmup = torch.zeros(1, device="cuda", dtype=torch.float32)
+            torch.distributed.all_reduce(warmup)
+            del warmup
+
         pre_alloc = torch.cuda.memory_stats()["allocated_bytes.all.current"]
 
         config = self.config
@@ -1976,10 +1990,8 @@ class ModelRunner:
                     f"diff={diff_pct:.1%}"
                 )
 
-        # Skip on single-rank: a world_size==1 barrier is a no-op but still
-        # forces lazy NCCL communicator creation (CUDA-allocs its buffers),
-        # which can OOM/fail on single-card runs. The process group stays
-        # initialized so get_tp_group() and friends keep working.
+        # Align ranks after KV is resident. Communicator scratch was allocated
+        # by the pre-alloc allreduce, so this barrier should not hipMalloc.
         if (
             torch.distributed.is_initialized()
             and torch.distributed.get_world_size() > 1
--- a/atom/model_engine/model_runner.py
+++ b/atom/model_engine/model_runner.py
@@ -1838,18 +1838,6 @@ class ModelRunner:
         return pool
 
     def allocate_kv_cache(self, num_kvcache_blocks):
-        # NCCL/RCCL communicators are created lazily on the first collective
-        # and allocate scratch via hipMalloc (not the PyTorch caching
-        # allocator). Do that collective before packing the GPU with KV,
-        # otherwise the post-allocation barrier can OOM.
-        if (
-            torch.distributed.is_initialized()
-            and torch.distributed.get_world_size() > 1
-        ):
-            warmup = torch.zeros(1, device="cuda", dtype=torch.float32)
-            torch.distributed.all_reduce(warmup)
-            del warmup
-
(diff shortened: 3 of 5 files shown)


valarLip:
Reviewed at `97de76c18`. One file, but the pin swap has consequences outside it. I fetched the upstream tree at the new pin to check the load-bearing claims rather than reason about it.

---

# 1. Moving off the fork drops the QP `rd_atomic` clamp — the other half of what that fork branch was for

The fork branch is named `fix/ionic-mr-and-qp-resource-fixes`. This PR takes the MR half and leaves the QP half behind.

Fetched `kvcache-ai/Mooncake@bfca1ce`:

```
mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp   (1487 lines)
  grep -c 'rd_atomic|max_res_rd_atom'  ->  0
  grep -n 'uint32_t init'              ->  (nothing)

mooncake-transfer-engine/src/transport/rdma_transport/rdma_endpoint.cpp
  1185:    attr.max_dest_rd_atomic = 16;
  1221:    attr.max_rd_atomic = 16;
```

So upstream hardcodes 16 with no clamp against `device_attr.max_res_rd_atom`, and never fixed this — a GitHub code search for `max_res_rd_atom` in the repo returns nothing. The fork clamps in `rdma_context.cpp:667-695` and consumes it at `rdma_endpoint.cpp:611/632`.

AMD Pensando ionic advertises `max_res_rd_atom = 16` **device-wide**, so the first QP consumes the entire responder budget; at the default `num_qp_per_ep=2`, even a single endpoint's second `ibv_modify_qp(..., IBV_QPS_RTR)` returns EINVAL. Symptom on Crusoe ionic nodes: PD-disaggregation handshake failures — `Failed to modify QP to RTR ... Invalid argument`, `stage=kRtr`.

Net: this PR fixes MR-registration EINVAL 22 and reintroduces QP-setup EINVAL, on the same hardware. The clamp needs to be re-applied (a patch step in the Dockerfile, or upstreamed) or the fork kept as the base.

# 2. The RDMA smoke-test job hard-fails against the new pin

`.github/workflows/atomesh-benchmark.yaml:585-590` patches `rdma_context.cpp` with:

```python
old = "            uint32_t init = 16;"
new = "            uint32_t init = max_dest_rd_atomic_;"
...
raise RuntimeError(f"Expected exactly one Mooncake RDMA atomic setting, ...")
```

`old` does not exist in upstream `bfca1ce` (verified above), so the job dies with `... found 0 in /app/mooncake/.../rdma_context.cpp` the first time it runs against an image built from this PR. The workflow file is not in the diff, so the breakage is invisible when reviewing it.

Worth noting what this patch *is*: it replaces the hardcoded 16 with `max_dest_rd_atomic_`, i.e. the workflow has been hand-applying the same clamp the fork provides. That is independent evidence for finding 1 — someone already found that this hardware needs it.

# 3. The `ibv_reg_dmabuf_mr` verification can never fail

`docker/atom_release.dockerfile:198`:

```sh
grep -a -R -l PATTERN ... | head -n 1 || { echo ...; exit 1; }
```

In POSIX `sh` without `pipefail`, a pipeline's status is the *last* command's, and `head` always succeeds. Reproduced locally:

```
$ sh -c 'grep -l zzzz /etc/hostname | head -n 1 || echo GUARD_FIRED; echo "status=$?"'
status=0
```

`GUARD_FIRED` never runs. So if `USE_HIP_DMABUF` silently fails to compile into `rdma_transport` — hsa-runtime64 found at configure but the `PRIVATE` define missing on the object lib, say — the image ships with GPU MRs on `ibv_reg_mr` and the build passes green.

Secondary: `-l | head -1` accepts a match in **any** file under `/opt/venv`, including an unrelated wheel that bundles libibverbs, so even a working guard could false-pass. Suggest `grep -a -R -q ... || { echo ERROR; exit 1; }`, scoped to the installed `libtransfer_engine.so` / `mooncake/engine*.so`.

# 4. `docker/Dockerfile` carries the same block and was not swept

```
docker/Dockerfile:374  ARG MOONCAKE_REPO="https://github.com/Jasen2201/Mooncake.git"
docker/Dockerfile:375  ARG MOONCAKE_COMMIT="fix/ionic-mr-and-qp-resource-fixes"
docker/Dockerfile:427  cmake .. -DUSE_HIP=ON -DUSE_ETCD=ON     # no USE_HIP_DMABUF
```

`grep -n USE_HIP_DMABUF docker/Dockerfile` returns nothing. Per `/app/ATOM/CLAUDE.md`: *"Fix-then-sweep: after fixing a bug, immediately grep for the same pattern across the codebase and fix all occurrences in one pass."*

Two images now disagree about which Mooncake they contain and which MR path they take, and the divergence is undetectable at runtime. (`docker/Dockerfile` appears to be referenced by nothing in the repo — if it is dead, deleting it in this PR is better than leaving it to drift.)

---

## Smaller items

- **`:183` — the dma-buf configure gate only accepts the literal `ON`.** `if [ "${USE_HIP_DMABUF}" = "ON" ]` skips the check entirely for `on` / `1` / `TRUE` / `YES`, all of which CMake treats as true. So `docker build --build-arg USE_HIP_DMABUF=1` produces a dma-buf-enabled build whose verification is skipped — and combined with finding 3, such an image has no build-time verification at all. Either normalize the ARG, or invert to "skip only when explicitly OFF".
- **`:204` — the comment contradicts the line beneath it**, and an empty value silently disables the feature. Upstream parses it as `if (const char *env = getenv("MOONCAKE_DISABLE_HIP_DMABUF")) { if (std::string(env) != "0") return false; }` (`rdma_context.cpp:73-78`), so `""` disables dma-buf. A deployment doing `docker run -e MOONCAKE_DISABLE_HIP_DMABUF ...` with the host var empty, or a k8s entry with `value: ""`, silently reverts to `ibv_reg_mr` → EINVAL 22 on ionic, with only a `LOG(INFO)` as evidence. The PR body's "any non-zero value disables" does not read as covering the empty string.
- **`:205` — the new runtime env var is documented nowhere.** Repo-wide grep for `MOONCAKE_DISABLE_HIP_DMABUF` and `dmabuf` returns 0 hits outside this diff, and `docs/environment_variables.md` has no Mooncake entries at all. The only description lives in the PR body, which is not shipped, so an operator debugging ionic MR failures has no in-repo way to find the escape hatch. `recipes/pd_disaggregation_guide.md` is the other natural home.
- **`:170` — the `HSA_PREFIXS` loop is one `dirname` too shallow, and redundant either way.** Three `dirname`s on `/opt/rocm/lib/cmake/hsa-runtime64/hsa-runtime64Config.cmake` give `/opt/rocm/lib`, which is not a valid CMake prefix for that package (find_package would search `/opt/rocm/lib/lib/cmake/hsa-runtime64`). Four would give `/opt/rocm` — which is already the unconditional seed. And `mooncake-common/common.cmake:433-434` already does `if(USE_HIP) list(APPEND CMAKE_PREFIX_PATH "/opt/rocm/lib/cmake")`. So it is ~6 lines of dead, wrong shell that reads as load-bearing and will mislead the next person debugging a dma-buf configure failure. (`HSA_PREFIXS` is also a typo for `PREFIXES`.)
- **The build-time assertion is at the wrong altitude for what decides the behavior.** `isKernelDmabufSupported()` (`rdma_context.cpp:71-144`) requires `CONFIG_PCI_P2PDMA` and `CONFIG_DMABUF_MOVE_NOTIFY`, discovered by reading `/boot/config-$(uname -r)`, `/lib/modules/.../build/.config`, or `/proc/kallsyms`. Inside this container `/boot` and `/lib/modules` are typically empty, leaving `/proc/kallsyms` as the sole detector — and if it is unreadable or symbol-restricted the probe returns false, logs a WARNING, and falls back to `ibv_reg_mr`, reproducing EINVAL 22 despite a green build. Neither the image nor the docs give the operator a runtime check to run.
- **`:600` (workflow) — the smoke test re-configures with the exact combination this PR's own comment says is broken.** The new comment at `:165-167` states "HIP hipify copies tests/*.cpp into the build tree but not headers such as rdma_test_peers.h, so BUILD_UNIT_TESTS=ON fails", and the workflow does `cmake ... -DUSE_HIP=ON -DBUILD_UNIT_TESTS=ON`. Masked today by finding 2 aborting first; once that is fixed this is the next failure. It also targets `.../example/transfer_engine_validator`, which relies on `BUILD_EXAMPLES` defaulting ON — an assumption this PR is simultaneously flipping in the image build.
- **The dma-buf path has no test coverage.** The validator the smoke test runs is compiled in a separate tree (`/tmp/mooncake-rdma-smoke-build`) from a cmake invocation that omits `-DUSE_HIP_DMABUF`, so a green atomesh-benchmark run says nothing about whether `ibv_reg_dmabuf_mr` works on ionic. Combined with the three unchecked test-plan boxes, the feature has not been demonstrated end to end anywhere.
- **`:123` — the store/Rust build surface was turned off but its dependency install was left.** The comment justifies `WITH_STORE=OFF` / `WITH_STORE_RUST=OFF` as avoiding "build surface (cachelib, cargo) that nothing here loads", but `[MC 2/4]` still runs `bash dependencies.sh -y` and still downloads the ~250 MB Go 1.22.2 toolchain into `/usr/local/go`, whose only consumer was the store's Go bindings. The image keeps the cost and saves only compile time — either trim those in the same pass, or drop the "build surface" claim from the comment.
- **`:189` — compiler output is discarded on success and the logs ship in the image.** `make -j$(nproc) > /tmp/mooncake-make.log 2>&1` means a successful build emits nothing to the docker log, so all warnings (including any about the new code path) are lost; and neither `/tmp/mooncake-cmake.log` nor `/tmp/mooncake-make.log` is removed by the cleanup, which only does `rm -rf /app/mooncake/build /app/mooncake/.git`. `tee` plus deleting both in cleanup.
- **`:110` — the pin is a mid-stream commit, and no PR check builds this file.** `bfca1ce` is "[TransferEngine] Fix duplicate cleanup in transport tests (#3583)" — a test-only fix, not a release tag, so the tree carries whatever else was in flight. `docker/atom_release.dockerfile` is built by `docker-release.yaml:169` and two `workflow_dispatch`-gated accuracy workflows; none is a PR check, and none passes `--build-arg MOONCAKE_REPO/MOONCAKE_COMMIT`, so the ARG defaults are the effective pin. A typo'd cmake option would first surface in the nightly/release build. Worth pinning to a tagged release if one carries the dma-buf support, and dispatching `docker-release.yaml` manually before merge.

---

## Checked and cleared

Recording these so they do not get re-raised:

- **`ENV MOONCAKE_DISABLE_HIP_DMABUF=0` does not invert the feature.** Upstream's check is value-based (`std::string(env) != "0"`), not presence-based, so `=0` keeps dma-buf enabled. Only the empty-string case above is a real hazard.
- **`-DWITH_STORE=OFF` does not break `from mooncake.engine import TransferEngine`** (`mooncake_connector.py:59`, the repo's only real Mooncake import). The pybind `engine` module is guarded by `if(WITH_TE)` and `store` separately by `if(WITH_STORE)`, and the generated `__init__.py` has no eager store import. The one loose thread is `torchspec`'s `EagleMooncakeStore` (`atom/rollout/model_runner_ext.py:210-215`) — that package is not installed by any dockerfile here, so it is out of scope for this image, but worth confirming if it ever is.
- All five cmake option names (`WITH_TE`, `WITH_STORE`, `WITH_STORE_RUST`, `BUILD_UNIT_TESTS`, `BUILD_EXAMPLES`) exist upstream with those exact spellings; the gate string `HIP dmabuf MR registration enabled` matches the `message(STATUS)` verbatim and the two negative branches say "disabled", so it cannot false-positive; and pairing `WITH_STORE=OFF` with `WITH_STORE_RUST=OFF` is required (upstream `FATAL_ERROR` otherwise) — that part is right.

junyyang-amd:
Because kvcache-ai/Mooncake doesn't include QP fix from https://github.com/Jasen2201/Mooncake/tree/fix/ionic-mr-and-qp-resource-fixes, merge v0.3.14-rc1 as the new fork baseline. Now close the PR.

PR #2099, opened 2026-09-01
ci: prevent benchmark dashboard history loss

junyyang-amd:
## Summary
- Validate and normalize existing benchmark dashboard data before every benchmark action.
- Prevent parser failures from silently replacing dashboard history with a single result.
- Apply the protection to all ATOM, ATOMesh, vLLM, and SGLang dashboard writers.
- Fix OOM with killing leftover docker containers in vLLM/SGLang plugin preflight.

## Test plan
- [x] `python3 -m py_compile .github/scripts/prepare_dashboard_data.py`
- [x] Parse all workflow YAML files successfully.
- [x] Validate normalization against the intact 300-entry dashboard history.
- [ ] Verify the updated workflows in GitHub Actions.

### DIFF
--- a/.github/actions/prepare-dashboard-data/action.yml
+++ b/.github/actions/prepare-dashboard-data/action.yml
@@ -1,0 +1,48 @@
+name: Prepare dashboard data
+description: Normalize and publish an existing dashboard data file before benchmark action runs.
+
+inputs:
+  data-path:
+    description: Path to data.js on the gh-pages branch.
+    required: true
+
+outputs:
+  entry_count:
+    description: Number of benchmark run entries before the action writes.
+    value: ${{ steps.prepare.outputs.entry_count }}
+
+runs:
+  using: composite
+  steps:
+    - id: prepare
+      shell: bash
+      env:
+        DATA_PATH: ${{ inputs.data-path }}
+      run: |
+        set -euo pipefail
+        SCRIPT_PATH="$RUNNER_TEMP/prepare_dashboard_data.py"
+        cp "$GITHUB_WORKSPACE/.github/scripts/prepare_dashboard_data.py" "$SCRIPT_PATH"
+        CURRENT_SHA=$(git rev-parse HEAD)
+        git config user.name "github-actions[bot]"
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git fetch origin gh-pages
+        git checkout -B gh-pages origin/gh-pages
+
+        output=$(python3 "$SCRIPT_PATH" "$DATA_PATH" --write)
+        printf '%s\n' "$output"
+        entry_count=$(printf '%s\n' "$output" | awk -F= '$1 == "entry_count" { print $2 }')
+        if [ -z "$entry_count" ]; then
+          echo "::error::Dashboard helper did not report entry_count"
+          exit 1
+        fi
+
+        if [ -f "$DATA_PATH" ]; then
+          git add -- "$DATA_PATH"
+          if ! git diff --cached --quiet; then
+            git commit -m "ci(dashboard): normalize benchmark data format"
+            git push origin gh-pages
+          fi
+        fi
+
+        git checkout "$CURRENT_SHA"
+        echo "entry_count=$entry_count" >> "$GITHUB_OUTPUT"
--- a/.github/actions/validate-dashboard-data/action.yml
+++ b/.github/actions/validate-dashboard-data/action.yml
@@ -1,0 +1,25 @@
+name: Validate dashboard data
+description: Verify dashboard data was not reduced while publishing a benchmark result.
+
+inputs:
+  data-path:
+    description: Path to data.js on the gh-pages branch.
+    required: true
+  min-entries:
+    description: Minimum number of run entries expected after publishing.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      env:
+        DATA_PATH: ${{ inputs.data-path }}
+        MIN_ENTRIES: ${{ inputs.min-entries }}
+      run: |
+        set -euo pipefail
+        DATA_FILE="$RUNNER_TEMP/dashboard-data.js"
+        git show "gh-pages:${DATA_PATH}" > "$DATA_FILE"
+        python3 "$GITHUB_WORKSPACE/.github/scripts/prepare_dashboard_data.py" \
+          "$DATA_FILE" \
+          --min-entries "$MIN_ENTRIES"
--- a/.github/scripts/gpu_preflight_check.sh
+++ b/.github/scripts/gpu_preflight_check.sh
@@ -3,14 +3,16 @@
 #
 # This intentionally avoids importing ATOM or aiter. It records the visible GPU
 # process/memory state, then performs a minimal torch allocation on every visible
-# HIP device. If this fails, the node/container is already unhealthy before ATOM
-# participates in the run.
+# HIP device. Leftover docker cleanup is opt-in via GPU_PREFLIGHT_KILL_DOCKER=1
+# (vLLM/SGLang plugin CI). ATOM native CI leaves it off.
 
 set -euo pipefail
 
 CONTAINER="${1:-}"
 ENGINE="${2:-docker}"
 GPU_PREFLIGHT_ALLOCATION_MB="${GPU_PREFLIGHT_ALLOCATION_MB:-8}"
+GPU_PREFLIGHT_KILL_DOCKER="${GPU_PREFLIGHT_KILL_DOCKER:-0}"
+GPU_PREFLIGHT_KILL_WAIT_SECONDS="${GPU_PREFLIGHT_KILL_WAIT_SECONDS:-30}"
 
 case "$GPU_PREFLIGHT_ALLOCATION_MB" in
     ''|*[!0-9]*)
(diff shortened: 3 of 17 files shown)


wanzhenchn (dismissed):

valarLip:
Reviewed at the branch head. The script findings below are executed, not read — the probe is at the end.

Two things worth settling before the list, because they change what this PR is:

---

# 0. The change that actually ships is `skip-fetch-gh-pages: true`, not the semicolon

The PR is framed around a stray terminal semicolon corrupting `data.js`. I checked every `data.js` on `origin/gh-pages`:

| file | trailing bytes | validated by this PR? |
|---|---|---|
| `benchmark-dashboard/data.js` | `]\n  }\n}` | yes (7 of the 9 steps) |
| `atomesh-accuracy-dashboard/data.js` | `]\n  }\n}` | yes |
| `atomesh-mocker-dashboard/data.js` | `\n  }\n}` | yes |
| **`vllm-benchmark-dashboard/data.js`** | **`}\n};\n`** | **no step points at it** |

So the semicolon exists on exactly one file on gh-pages, and it is the only one none of the nine new steps validates. On all three validated paths `has_trailing_semicolon` is False, `normalize()` returns `changed=False`, nothing is written, `git diff --cached --quiet` succeeds, and no commit is made.

That is consistent with the writer: `storeDataJs` emits `SCRIPT_PREFIX + JSON.stringify(data, null, 2)` — no semicolon, no trailing newline — so the action can never produce one.

Two consequences:

1. **The normalize step's only observable effect today is creating a local `gh-pages` branch for `skip-fetch-gh-pages: true` to switch onto.** The protection is hypothetical; the regressions in findings 1 and 2 are not.
2. **`vllm-benchmark-dashboard/data.js` is the one piece of evidence about where a semicolon comes from, and it is uncovered.** `atom-vllm-benchmark.yaml:2247` sets `benchmark-data-dir-path: benchmark-dashboard`, so nothing in this repo writes that directory — it reads as a dead artifact. Worth confirming that, because if it is *not* dead there is an unidentified writer, and that is the actual root cause this PR is looking for.

Separately, the silent-wipe mechanism itself is untouched: `loadDataJs` swallows every error into `{...DEFAULT_DATA_JSON}` at `console.log` severity with the text "Could not find data.js". The durable fix is a post-write assertion that the entry count did not shrink, before the push — that catches every corruption shape, including the ones below.

# 1. `skip-fetch-gh-pages: true` turns a recoverable push race into a deterministic hard failure

In `dist/src/write.js@v1` the retry does `git reset --hard HEAD~1` and recursively re-enters `writeBenchmarkToGitHubPagesWithRetry`, whose only refresh is `await git.pull(...)` guarded by `if (!skipFetchGhPages && ...)`. With the flag true every arm is dead and it reaches the literal `console.warn('NOTHING EXECUTED', ...)` branch.

So: concurrent push → `[rejected]` → reset → rebuild the identical commit on the identical stale parent → rejected again, ten times, then `throw`.

**Where this bites hardest.** I checked all three conditions:

- 9 workflows each gain one `skip-fetch-gh-pages: true`.
- Only **7** of them carry `concurrency: group: gh-pages-deploy`. The two without it are `atom-sglang-accuracy-validation.yaml` (job `accuracy-dashboard`) and `atom-vllm-accuracy-validation.yaml`.
- Both of those run on `cron: '0 16 * * *'` — the same minute — and both write `benchmark-dashboard/`.

The only two dashboard jobs that are not serialized against each other now also have their conflict recovery disabled, and they are scheduled to collide daily.

# 2. `git fetch origin gh-pages` does not fast-forward an existing local `gh-pages`

`git fetch origin <branch>` updates `FETCH_HEAD` and opportunistically `refs/remotes/origin/gh-pages`, but leaves an existing local `gh-pages` ref alone; the following `git checkout gh-pages` lands on the stale ref.

Previously the action compensated — `src/git.ts` ran `git fetch <remote> gh-pages:gh-pages`, which force-updates the local branch. `skip-fetch-gh-pages: true` removes exactly that. So the run reads a stale `data.js`, appends to it, and the push is rejected — or an older snapshot is what gets carried forward.

The pre-existing publish step in the same file already defends against this: `atom-vllm-benchmark.yaml:2254` and `atom-sglang-benchmark.yaml:1670` use `git checkout -B gh-pages-data-only origin/gh-pages`. The new step drops that defense.

Fix is `git checkout -B gh-pages origin/gh-pages`, in all nine copies.

---

# 3. The validator waves through the exact input that causes the silent wipe

`prepare_dashboard_data.py:28` is a bare `json.loads(payload)`. Python accepts `NaN` / `Infinity` / `-Infinity`; V8's `JSON.parse` does not.

Executed:

```
nan.js   exit=0   /tmp/nan.js: valid
python json.loads('{"v": NaN}')  -> {'v': nan}
node   JSON.parse('{"v": NaN}')  -> REJECT  Unexpected token 'N'
```

So a `data.js` carrying a `NaN` benchmark value passes validation, `loadDataJs` then throws, the catch returns `{lastUpdate:0, repoUrl:'', entries:{}}`, and `storeDataJs` overwrites 22.9 MB of history with one entry — job green.

That is the failure this PR exists to prevent, and the new guard says "valid" on it. One line: `json.loads(payload, parse_constant=_reject)`.

# 4. The emptiness check is stricter than the consumer it protects, and there is no self-heal

`:35-38`:

```python
if not entries or not any(
    isinstance(value, list) and value for value in entries.values()
):
    raise ValueError("dashboard data contains no benchmark entries")
```

`loadDataJs` has no such notion — it does `JSON.parse` and would happily append to `{"entries":{}}` or `{"entries":{"Benchmark":[]}}`.

Executed: `window.BENCHMARK_DATA = {"lastUpdate":0,"repoUrl":"","entries":{}}` exits 1. Under `set -euo pipefail` that fails the Prepare step, the Store step is skipped, and new results are never published.

The part that makes this severe: **the only code that could repair the file is the step that refuses to run.** Every subsequent run of all 8 workflows fails the same way until a human hand-edits gh-pages. A benign dashboard state becomes a permanent red.

# 5. The two most likely corruption shapes raise `TypeError`, which the handler does not catch

`:71` catches `(OSError, ValueError, json.JSONDecodeError)`. `parse_data` raises `TypeError` at `:30` and `:34`.

Executed: `window.BENCHMARK_DATA = {"foo": 1};` prints a full traceback ending `TypeError: dashboard data must contain an 'entries' object`, never reaching the `print(f"{args.data_js}: invalid dashboard data: {exc}")` / `return 1` branch. Same for `= null;`, `= [1,2];`, `= "x";`, `= {"entries": []};`.

The exit code is still 1, so the job does fail — but the diagnostic that was written for this case is unreachable. Commit `f3b2fb7d` changed those two raises from `ValueError` to `TypeError` ("correct type exceptions") without updating the handler.

# 6. `payload[:-1]` strips exactly one semicolon, so the tool fails on the second application of its own bug

`:26`. Executed: `...};;\n` exits 1 with `Extra data: line 1 column 77`, and so does `...};\n// comment\n`.

Whatever external writer appends a `;` appends a second one on its next pass. At that point the tool that exists to repair the file instead blocks all 8 publish jobs (finding 4 — no self-heal). A normalizer should strip trailing `;` and whitespace repeatedly, or re-serialize.

# 7. The round-trip rewrites every line ending, contradicting the docstring

The module docstring says "preserving the existing formatting and history". `Path.read_text(encoding="utf-8")` opens with `newline=None`, so `\r\n` and lone `\r` become `\n` in memory; `write_text` at `:76` writes back `os.linesep`.

Verified byte-for-byte:

```
input   b'":1}]}]}};\r\n'
output  b'ue":1}]}]}}\n'
```

A CRLF `data.js` — hand-edited on Windows, which is precisely the external-writer scenario the semicolon guard assumes — is reflowed in full and committed to gh-pages as a whole-file 23 MB diff under the message "ci(dashboard): normalize benchmark data format".

---

## Smaller, all executed

- **`:55` — `description=__doc__.splitlines()[0]` raises before argument parsing under `-OO`.** Executed: `AttributeError: 'NoneType' object has no attribute 'splitlines'`. Latent today since the workflows call plain `python3`, but the failure is a traceback rather than a usage message. `description=(__doc__ or "").splitlines()[0] if __doc__ else None`.
- **`:78` — without `--write`, a file needing normalization exits 0.** Executed: prints `valid but requires normalization`, returns 0, file untouched. All 8 workflows pass `--write` so it is unreachable today, but the exit code contradicts the message by construction — any future use as a check (a pre-commit hook, a lint job over a committed `data.js`) gets green on the corruption case. A `--check` mode should return 1.

## Structure

**`atom-test.yaml:752` — the new Prepare step has no `if:` guard while the Store step it protects has `if: hashFiles('accuracy-benchmark-input.json') != ''`.** All seven other workflows use byte-identical conditions on the two steps; this is the one asymmetry. On a run where the transform produced no accuracy entries, Store is skipped but Prepare still fetches gh-pages, checks out a 23 MB tree, and — per finding 4 — can fail the `accuracy-dashboard` job over a dashboard state nothing was going to be published to anyway.

**The 15-line git block is pasted verbatim into nine steps, and the branch switch is not needed at all.** Nine copies (`atom-benchmark.yaml:261`, `atom-sglang-accuracy-validation.yaml:798`, `atom-sglang-benchmark.yaml:1636`, `atom-test.yaml:752`, `atom-vllm-accuracy-validation.yaml:1616`, `atom-vllm-benchmark.yaml:2220`, `atomesh-accuracy-validation.yaml:565`, `atomesh-benchmark.yaml:1385`, `atomesh-mocker-benchmark.yaml:259`) differing only in the `data.js` path and the `if:` guard — and each duplicates a fetch/checkout/restore block roughly 20 lines below it in the same job. Any fix to findings 1, 2 or the guard asymmetry has to land in nine places.

A read-only validation needs no branch switch, no commit, no `/tmp` copy and no detached HEAD:

```bash
git fetch origin gh-pages
git show origin/gh-pages:benchmark-dashboard/data.js | python3 .github/scripts/prepare_dashboard_data.py -
```

The `cp .github/scripts/prepare_dashboard_data.py /tmp/...` exists only because the branch switch deletes the script.

## Tests

`grep -rln "prepare_dashboard\|dashboard_data" tests/ .github/` returns nothing. The PR's test plan lists `python3 -m py_compile`, which is a syntax check.

This is pure, GPU-free string logic with an in-repo precedent — `tests/test_benchmark_catalog.py:14-25` already does `sys.path.insert(0, str(REPO/".github"/"scripts"))`. `normalize()`'s `text[: content_end - 1] + text[content_end:]` is exactly the off-by-one-prone surgery a unit test should pin, and **findings 3, 5, 6, 7 and both smaller items are each one line of a parametrized test.**

## Out of scope

**`pre-checks.yaml:64` — `-fail-on-error=true` → `-fail-level=any` is a behavioral no-op with a latent hard failure attached.** In reviewdog v0.21.0 `cmd/reviewdog/main.go:failLevel()`, `-fail-on-error=true` with a non-`github-check` reporter already resolves to `FailLevelAny`; the reporter here is `github-pr-review`, so the only effect is suppressing a deprecation warning.

Meanwhile `reviewdog/action-setup@e04ffabe` defaults to `reviewdog_version: latest` with no pin in the workflow, and `-fail-level` first exists in v0.20.2 — so pinning `reviewdog_version` to ≤ v0.20.1 makes Go's flag package print usage and exit 2 on every pre-checks run, unrelated to any lint finding.

It also does not do what its commit message ("resolve dashboard helper lint failures") claims: the new `.py` file's ruff findings fail identically under either flag.

## Trailers

Commits `5d804644` and `f3b2fb7d` both end with `Co-authored-by: Cursor <cursoragent@cursor.com>`, and the PR body's last line is `Made with [Cursor](https://cursor.com)`. Worth clearing before merge — these land permanently in main's history.

---

## Reproducing the script findings

```python
import json, subprocess, sys
from pathlib import Path

SCRIPT = ".github/scripts/prepare_dashboard_data.py"
TMP = Path("/tmp/dashcheck"); TMP.mkdir(exist_ok=True)

CASES = {
  "nan":     'window.BENCHMARK_DATA = {"entries":{"B":[{"commit":{"id":"a"},"benches":[{"name":"n","value":NaN}]}]}}',
  "typeerr": 'window.BENCHMARK_DATA = {"foo": 1};',
  "empty":   'window.BENCHMARK_DATA = {"lastUpdate":0,"repoUrl":"","entries":{}}',
  "dsemi":   'window.BENCHMARK_DATA = {"entries":{"B":[{"commit":{"id":"a"},"benches":[{"name":"n","value":1}]}]}};;\n',
  "crlf":    'window.BENCHMARK_DATA = {"entries":{"B":[{"commit":{"id":"a"},"benches":[{"name":"n","value":1}]}]}};\r\n',
}

for name, body in CASES.items():
    p = TMP / f"{name}.js"; p.write_bytes(body.encode()); before = p.read_bytes()
    r = subprocess.run([sys.executable, SCRIPT, str(p), "--write"], capture_output=True, text=True)
    out = (r.stdout + r.stderr).strip().splitlines()
    print(f"{name:8s} exit={r.returncode}  {out[-1] if out else ''}")
    if name == "crlf":
        print(f"         {before[-12:]!r} -> {p.read_bytes()[-12:]!r}")
```

Output:

```
nan      exit=0  /tmp/dashcheck/nan.js: valid
typeerr  exit=1  TypeError: dashboard data must contain an 'entries' object
empty    exit=1  ...: invalid dashboard data: dashboard data contains no benchmark entries
dsemi    exit=1  ...: invalid dashboard data: Extra data: line 1 column 77
crlf     exit=0  ...: removed terminal semicolon
         b'":1}]}]}};\r\n' -> b'ue":1}]}]}}\n'
```

junyyang-amd:
Thanks for the detailed review. The findings have been addressed in commit `143381d6`:

- Removed `skip-fetch-gh-pages: true`; the benchmark action keeps its normal fetch/retry behavior.
- Centralized dashboard preparation in a composite action using `git checkout -B gh-pages origin/gh-pages`, so an existing local branch cannot remain stale.
- Added the missing `gh-pages-deploy` concurrency lock to the vLLM and SGLang accuracy dashboard jobs.
- Changed all dashboard writers to `auto-push: false`; each now validates the generated data before the workflow pushes it. The validator rejects history shrinkage, malformed JSON, and JavaScript-incompatible `NaN`/`Infinity` values.
- Normalization now handles repeated terminal semicolons and preserves the original bytes/line endings. Type errors and check-mode failures are reported correctly.
- Consolidated the repeated workflow logic into reusable local actions and added unit tests for the parser/normalizer edge cases.
- Reverted the unrelated reviewdog option change from this PR.

Regarding `vllm-benchmark-dashboard/data.js`: the current repository workflows do not write that path; `atom-vllm-benchmark.yaml` writes `benchmark-dashboard/data.js`. It therefore remains an external or stale artifact and is not included in the reusable protection path.

zhuyuhua-v (approved):

PR #2106, opened 2026-09-01
[Fix] Allocate the Eagle3 MHA draft KV pool in flash layout for MiniMax-M3

yitingw1:
## Motivation

**Symptom** — MiniMax-M3-MXFP8 + EAGLE3 (8k1k, c=32) OOMs at startup in CI: one 12.37 GiB / 6.18 GiB allocation on top of an already-sized KV pool, at a 1.38% memory margin.

**Cause** — the draft's KV pool holds V as `[num_blocks, kv_heads, head_dim, block_size]`, the one V layout no prefill reader consumes, so `_gather_prefix_and_concat_kv` converts the **whole pool** with `permute().contiguous()` on every prefill carrying a cached prefix — 12.37 GiB moved to produce 297 MB. Only CI hits it: `ATOM_FORCE_ATTN_TRITON=1` is what sends `rope_cache` to the writer emitting that V.

**Fix** — allocate those drafts a flash (4D) pool; `unified_attention` reads it in both prefill and decode, so the gather is never entered. `Eagle3DraftBuilder._use_flash_layout` derives this from the module, mirroring `rope_cache`'s branch condition, so drafts already getting a working SHUFFLE V (Kimi-K2.5) keep it.

## Technical Details

`use_flash_layout` was dead — `False` at all three assignment sites since it was introduced — so setting it True makes every reader live at once. In particular it **swaps the draft's kernels, not just its allocation**: `paged_attention_triton` branches on `ATOM_USE_UNIFIED_ATTN or self.use_flash_layout`, and this config runs with `ATOM_USE_UNIFIED_ATTN=0`.

| | baseline | flash |
| --- | --- | --- |
| decode | context-partitioned split-KV (gluon PA) | `unified_attention`, no partition |
| prefill | `flash_attn_varlen_func` | `unified_attention` |

Both run `num_speculative_tokens` times per target step, over the draft's single attention layer (60 target + 1 draft).

**A pre-existing bug this reached** — `prefill_attention_triton` viewed raw K/V as a `block_size=1` cache, valid only with the fake table `TritonMHAMetadataBuilder` builds; MiniMax-M3 uploads a real page-128 table, so addressing became `block_id * 1 + offset` — in bounds, no fault, wrong rows, and only acceptance moves. Fixed by reading the paged cache `rope_cache` already wrote: −1.70 / −1.76 pp acceptance became −0.06 / −0.31 pp.

**A second gap** — `propose` reads `forward_vars["block_tables"].gpu` unconditionally, while producers upload only under their own conditions. An unrefreshed buffer is not `None`; it holds the previous batch's pages, in bounds, so the draft attends over the wrong KV and only acceptance moves. A null check cannot see that, so freshness is tracked: `CommonAttentionBuilder.build` clears a flag (sole path into `prepare_*`), `CpuGpuBuffer.copy_to_gpu` sets it, `propose` checks it. The flag sits on the **upload**, not on `prepare_block_tables` — those do not coincide (`gdn_attn.py:1020` packs and never uploads; `deepseek_v4_attn.py:4246` uploads without it), whereas all eleven upload sites go through `copy_to_gpu`. Checked **for flash drafts only**: a SHUFFLE draft's pure prefill takes the raw-varlen path and never reads `block_tables`. `_enter_decode_metadata` is left unguarded because `warmup_inputs` replays it with no batch; noted in a comment.

## Test Plan

- `tests/test_draft_kv_layout.py` (11) — **the only file here that runs on CI**, which has no aiter (`tests/conftest.py`). `eagle3_kv_builder` does `from aiter import dtypes`, so a test reaching the predicate through it is skipped on the very CI this PR fixes; the decision therefore lives in `atom/spec_decode/draft_kv_layout.py`, which needs no aiter. `test_the_predicate_imports_without_aiter` re-checks that in a subprocess with `aiter` shadowed, so the file cannot lapse back into being skipped. Verified: 11 passed with aiter shadowed, the two files below 2 skipped.
- `tests/test_eagle3_draft_layout.py` (15) — the binder: bound shapes, the recorded flag, both arms staying views of one allocation, both KV dtypes (`x = 16 // itemsize` reshapes K).
- `tests/test_block_tables_freshness.py` (12) — both ends of the freshness flag, the guard's raise and quiet cases, a SHUFFLE draft left alone, and that a later SHUFFLE layer cannot clear an earlier flash layer's flag.
- Defect injection on a scratch worktree, 7 injections. Layout: reverting `impl.use_flash_layout = flash` fails 3; swapping the arms' bodies fails 7. Freshness: deleting the guard call fails 1, `build` not clearing fails 4, `copy_to_gpu` not setting fails 2, dropping the flash-only narrowing fails 2, `= flash` instead of `or flash` fails 1. Invariant: deleting the prefill upload fails 1, bypassing `copy_to_gpu` fails 1. `test_propose_actually_calls_the_guard` exists because deleting the guard's call site left every other test green.
- gsm8k end to end, and a throughput A/B against the base commit with acceptance pinned on both arms (`--spec-decode-acceptance-rate`), so the delta is the kernels and not draft quality — the arms landed within 0.01 pp.

## Test Result

**Accuracy** — gsm8k flexible-extract, 1319 problems, TP4, fp8 KV, over `cd161f16`. No OOM in any run.

| n-shot | MNBT | cov | gsm8k | MTP accept |
| 5 | 32768 | 0.4% | 0.9007 ± 0.0082 | 70.01% |
| 20 | 32768 | 1.3% | 0.9401 ± 0.0065 | 72.56% |
| 5 | 1024 | 61% | 0.9030 ± 0.0082 | 69.86% |

**Performance** — output tok/s, same node, arms back to back, within-pair order alternating between repeats. "cov" is measured in the run, not inferred from the config.

| scenario | cov | shuffle | flash | delta |
| 8k1k, mbt 1024 | **87.45%** | 1128.3 (n=4, spread 0.9%) | 1543.9 (n=4, spread 2.0%) | **+36.8%**, ranges disjoint |
| 8k1k, mbt 32768 | 0.00% | 2562.6 (n=1) | 2556.2 (n=2) | −0.2% |
| 32k1k, mbt 32768 | 0.00% | 1057.6 (n=4, spread 1.1%) | 1047.1 (n=3, spread 1.4%) | −1.0%, ranges overlap |

Only the first row reaches the fixed path, and there every `flash` run beats every `shuffle` run (mean TTFT 5039 → 3286 ms, mean TPOT 23.42 → 17.49 ms). Where coverage is zero the copy never fires and only the kernel swap remains. One 8k1k `shuffle` run is excluded as an outlier, by name, in `summarize.py`.

## Known Limitations

- **A rope-less draft under `use_triton_attn` still gets the 4D V** — `rope_cache` falls through to `reshape_and_cache(asm_layout=False)`. None in-tree; pinned by `test_rope_less_draft_is_a_known_gap` so it is not read as deliberate.
- **`prefill_attention_triton`'s `else` branch is now dead** — marked `DEAD:` rather than deleted, to retire with the fake-table producer in `triton_mha.py`.
- **The PD-separation rebind cannot bind an Eagle3 MHA draft at all** — `_get_total_num_layers` excludes drafts owning their builder, so the index raises first.
- **The kernel swap costs about 1% where the copy never fires** — a real trade: configs that never carry a cached prefix pay for a fix they do not need. Avoiding it needs a coverage-aware predicate, unknowable at bind time.
- **One configuration is now killed by the guard rather than run** — flash draft + non-sparse `AiterBackend` target + `ATOM_USE_UNIFIED_ATTN=0` + prefill with no prefix satisfies no upload condition. Nothing in-tree hits it and CI will not go red, but this PR makes it reachable. Left loud and actionable; the real fix is the last item.
- **`assert_kv_layout_matches` checks K's rank, and this bug was in V** — it would not have caught this OOM, and three of its four call sites assign the flag on the line before, so they cannot fire. It guards a future binder, not the failure already found.
- **The predicate mirrors the *base* `rope_cache`** — `SparseMHAPagedAttentionImpl` overrides it wholesale and hardcodes SHUFFLE; safe today only because it always carries both norms. Noted in the docstring.
- **The SGLang plugin carries a line-by-line copy of the same binder** (`plugin/sglang/eagle3_llama_bridge.py:519`) and its M3 EAGLE3 CI config sets `ATOM_FORCE_ATTN_TRITON=1`. Dormant only because that bridge never sets `has_cached`.
- **The `block_tables` producers still each carry their own upload condition** — the flag makes a stale read loud but does not converge them. Converging means `block_tables` always non-`None`, the branch condition for ten readers across MLA, V4, TBO and the rtpllm plugin.

## Submission Checklist

- [x] Test added (3 files, 38 tests, all verified by defect injection; 11 run on CI, which has no aiter)
- [x] Accuracy validated (gsm8k, three coverages)
- [x] Throughput benchmark (3 scenarios, measured prefix-cache coverage, n=1–4 per arm)


valarLip:
Reviewed the branch head. I checked the load-bearing claims against the tree rather than reading the diff alone, and cite line numbers from the PR head.

The core idea is right — a draft whose writer emits a V nobody reads should get the layout its writer already produces, rather than paying an O(pool) `permute().contiguous()` to convert it back. What follows is about the blast radius of the flag, which is wider than the diff suggests.

---

# 1. Pure prefill now reads the paged cache, but no `block_tables` producer was widened to match

`attention_mha.py:802` adds `or self.use_flash_layout` to the condition that selects the paged cache. Every producer of `attn_metadata.block_tables` still gates on `ATOM_USE_UNIFIED_ATTN` alone:

| producer | condition |
|---|---|
| `backends.py:525` | `if has_cached:` |
| `aiter_attention.py:889-891` | `block_tables is None and envs.ATOM_USE_UNIFIED_ATTN and batch.block_tables` |
| `triton_mha.py:55` | `envs.ATOM_USE_UNIFIED_ATTN and batch.block_tables` |

So a flash-layout impl with `ATOM_USE_UNIFIED_ATTN` unset and no cached prefix reaches `unified_attention(..., block_table=None)` and dies at `block_table_stride = block_table.stride(0)`.

For the Eagle3 draft it is worse than a crash. `eagle_proposer.py:499` assigns unconditionally:

```python
attn_metadata.block_tables = var["block_tables"].gpu[:scheduled_bs]
```

That buffer is not refreshed this step, so the draft attends to a previous batch's pages — possibly already freed, always in bounds, no fault, only a degraded acceptance rate.

Two configurations escape today, both by accident:

- MiniMax-M3 — `aiter_attention.py:880` uploads a table under `_has_sparse_attention`.
- MLA targets — `aiter_mla.py:1350` refreshes on `hasattr(runner, "drafter")`.

Neither is written down as a contract, and the new predicate does not consult either. The cleanest fix is to make the upload condition the same predicate as the read condition — one term, one place.

# 2. `_dispatch_decode` can route a flash draft to a SHUFFLE-only kernel

`attention_mha.py:862`'s `if envs.ATOM_USE_UNIFIED_ATTN:` branch short-circuits **before** the `use_flash_layout` check at `:869`, so `paged_attention_persistent_asm` becomes reachable for a 4D pool.

Trigger: `ATOM_USE_UNIFIED_ATTN=1`, `ATOM_FORCE_ATTN_TRITON=0`, `kv_cache_block_size=256`, draft `head_dim != 128` (itself one of the new flash triggers), `sliding_window=-1`.

`:655` computes `page_size = int(k_cache.shape[3])`, which on a flash `[nb, block_size, nh, hd]` pool is `head_dim`; and `_view_v_cache_for_pa_decode_bf16_asm` (`:474`) unpacks `n, nh, head_dim, block_size = v_cache.shape` on a flash V while deriving `x` from `k_cache.shape[-1]`. Wrong `page_size` into `pa_persistent_fwd` / `pa_decode_bf16_asm` gives transposed or out-of-range KV reads.

Reordering the two checks at `:862`/`:869` is enough.

# 3. There is a second binder, and it does not know about the flag

`model_runner.py:4494`, `_bind_kv_cache_to_modules` — the PD-disaggregation IPC rebind — walks `self.drafter.model`, binds every non-MLA module to a 5D SHUFFLE view of the **target** pool (`:4495-4507`), and never touches `impl.use_flash_layout`.

So in a disaggregated decode process with an Eagle3 MHA draft that was made flash at allocate time, the impl keeps `use_flash_layout=True` while its `k_cache`/`v_cache` are rebound to 5D slices of `self.kv_cache`. `rope_cache` then calls `fused_qk_rope_reshape_and_cache(flash_layout=True)` against a 5D tensor, tripping aiter's `t_cache, block_size, kh_cache, dk_cache = key_cache.shape` unpack; and the decode reader is handed `shuffled_kv_cache=False` for a SHUFFLE tensor.

This PR introduces a flag whose only setter lives in the other binder. Nothing keeps the two in sync, and nothing asserts they agree.

---

# 4. The added term makes the condition tautological for its only caller — the `else` branch is now dead, and its dead partner cites it back

`prefill_attention_triton` has exactly one call site. Repo-wide grep:

```
attention_mha.py:768   def prefill_attention_triton(          <- the def
attention_mha.py:890       return self.prefill_attention_triton   <- the only caller
aiter_attention.py:886  # `prefill_attention_triton` reads ...   <- a comment
```

and `:889` is `if envs.ATOM_USE_UNIFIED_ATTN or self.use_flash_layout:`. The method body at `:802-806` is:

```python
if (
    envs.ATOM_USE_UNIFIED_ATTN
    or self.use_flash_layout
    or attn_metadata.has_cached
):
```

The first two terms are already guaranteed true by the dispatch. So `has_cached` is a dead term, and `:812-818` — plus the six-line contract comment at `:780-784` that explains it — can never execute. `SparseMHAPagedAttentionImpl` overrides `dispatch_backend` at `:1201` and never reaches here either.

Its dead partner is `triton_mha.py:69`'s fake `block_size=1` table, built only when `block_tables is None` **and** not UNIFIED — a combination `triton_mha.py:148` (`impl.use_flash_layout = False`) sends to `prefill_attention` instead.

The net effect is two dead constructs that now reference each other, so each reads as load-bearing to the next person. Either delete both, or (if the `else` is meant to stay reachable) the dispatch at `:889` is what needs to change, not the body.

# 5. `_use_flash_layout` hand-copies a decision that has already drifted twice elsewhere

`eagle3_kv_builder.py:118-131` replicates three separate decisions from `rope_cache` (`attention_mha.py:210-221`) into a different package, with no shared symbol and no assertion. The impl already stores the answer as `self.use_triton_attn` (`:215`).

That the condition moves is not hypothetical — there are already two drifted copies:

- `plugin/vllm/attention/layer_mha.py:264` — drops the env term, adds `block_size != 16`.
- `triton_mha.py:99-104` — the qk-norm pair, copied a third time.

The predicate is also applied to whatever `module.impl` happens to be. `SparseMHAPagedAttentionImpl` subclasses `PagedAttentionImpl` and overrides `rope_cache` with a writer that ignores the flag entirely; it escapes mis-classification only because it happens to carry both norms. Drift, or one subclass without qk-norm, hands a 4D pool to a SHUFFLE writer — same element count, no shape error, no fault, only an acceptance-rate drop.

The shape that removes the class of bug: a property on `PagedAttentionImpl` that `rope_cache` itself dispatches on and builders read, or — cheaper — a bind-time assert that `(k_cache.dim() == 4) == impl.use_flash_layout`.

# 6. The `rotary_emb is None` case still produces the exact shape this PR exists to eliminate — and a new test pins it as intended

`_use_flash_layout` returns False when `rotary_emb is None` (`:119`). But `rope_cache`'s `else` branch sets `asm_layout = False` when `use_triton_attn and v_cache.dim() != 5` (`attention_mha.py:337-338`), and the builder's non-flash V is 4D by construction.

So a rope-less draft under `ATOM_FORCE_ATTN_TRITON` (or `head_dim != 128`) writes NHD, sets `_cache_format = "NHD"`, and routes straight back into the whole-pool `permute().contiguous()` at `:435-440` on every prefix-carrying prefill — the exact cost this PR removes.

Two things make that hard to find later: the docstring at `:116` says *"The other two already write a SHUFFLE V"*, and `test_no_rope_draft_is_left_alone` (`tests/test_eagle3_draft_layout.py:109`) asserts `False` for this input. The next person to hit the same OOM gets a green test telling them the choice was deliberate.

---

## Tests

**None of the seven cases run in CI.** `tests/test_eagle3_draft_layout.py:33` is:

```python
eagle3_kv_builder = pytest.importorskip("atom.spec_decode.eagle3_kv_builder", ...)
```

and that module's line 4 is `from aiter import dtypes`. CI has no aiter, so all seven report skipped.

The file's own docstring acknowledges this ("the module under test imports aiter, so this skips on a CPU-only runner") — but the consequence is not acknowledged anywhere: the only gate on a decision whose failure mode is silent KV corruption contributes zero assertions upstream. `tests/aiter_stub.stubbed_aiter()` is a scoped context manager that deliberately removes the fake from `sys.modules`, so it does not rescue a module-level `importorskip`; `tests/conftest.py:89-94` documents this exact rot pattern ("CI has no aiter, so CI never saw them and nobody was told").

The predicate is a pure function over five plain attributes. Deferring the aiter import, or moving `_use_flash_layout` somewhere aiter-free, makes these real tests. Repo CLAUDE.md states the target: `python -m pytest tests/  # all tests (no GPU needed — mocks AITER and torch.cuda)`.

**The revert run is green.** Delete `:170` (`impl.use_flash_layout = flash`) and the entire PR reverts — `attention_mha.py:804` and `:889` stop firing — with all seven tests still passing. Every assertion in the file calls the static predicate `_use_flash_layout(...)`; none touches which view `build_kv_cache_tensor` actually binds. Swapping the bodies of the `if flash:` / `else:` arms — precisely the silent corruption the module docstring warns about — is also green.

A stub-runner test (a `torch.zeros` pool plus a `SimpleNamespace` config) asserting the returned k/v shapes **and** the flag for both branches is cheap and needs nothing beyond the module import.

## Fix-then-sweep: a line-for-line copy in the SGLang plugin

`atom/plugin/sglang/eagle3_llama_bridge.py:522`:

```python
v_cache = native_cache[1].view(num_blocks, num_kv_heads, head_dim, page_size)
```

Same page-major pool, same 5D-K / 4D-VHD-V pair. `grep -rn use_flash_layout atom/plugin/sglang/` returns zero hits.

This path drives the same `PagedAttentionImpl` (`models/minimax_m3.py:94`, `impl_cls = impl_cls or PagedAttentionImpl`) and the same `rope_cache`, and the SGLang MiniMax-M3-MXFP8 EAGLE3 TP4 entries set `ATOM_FORCE_ATTN_TRITON=1` (`sglang_models_accuracy.json:141`) — the first term of the new predicate.

It escapes the whole-pool permute today only because the bridge never sets `has_cached=True` (`:434-445` leaves it at its `False` default while setting `state=AttnState.PREFILL_PREFIX`). The moment chunked prefill or a cached prefix is reported there, it pays the full O(pool) conversion this PR just removed on the core path.

Repo CLAUDE.md, Critical Rules: *"after fixing a bug, immediately grep for the same pattern across the codebase and fix all occurrences in one pass"*.

## A second prefill dispatch opened silently

`attention_mha.py:112`:

```python
if not (self.use_flash_layout or envs.ATOM_USE_UNIFIED_ATTN):
    return False
```

This was effectively "UNIFIED only", because every prior assignment of the flag in-tree was `False` (`:98`, `triton_mha.py:148`, `layer_mha.py:154`). This PR is the first to set it `True`, which makes `_can_attempt_prefill_sink_asm` reachable for drafts.

Its other conditions require `head_dim == 64` — itself one of the things that makes `_use_flash_layout` return True — plus sinks on gfx1250. Such a draft now takes `prefill_attention` (`dispatch_backend:887`) ahead of the flash path, running `_gather_prefix_and_concat_kv` through its untyped 4D catch-all at `:441` with `per_token_quant` computed from the scalar `self.kv_scale` that `rope_cache` substituted at `:304`.

No test, no benchmark, and the builder's docstring does not describe what actually runs.

## The kernel changed in both directions, with no measurement

`attention_mha.py:497`. Pre-PR (flash `False`, UNIFIED off) a FORCE_TRITON draft decoded through `paged_attention_triton`'s else branch into `run_pa_decode_gluon` — a decode-specialised split-KV kernel with `get_recommended_splits` / context partitions. Post-PR `:497` selects `unified_attention(shuffled_kv_cache=False)`, which has no context-partition split. This runs `num_speculative_tokens` times per target step, per draft layer.

Prefill likewise moves from `aiter.flash_attn_varlen_func` (`dispatch_backend:891`) to `unified_attention`.

The stated win applies only to prefills carrying a cached prefix. Pure prefills and every decode step pay an unmeasured kernel change. The accuracy table also has no main-branch arm, so neither axis has a before/after. A single benchmark point on the OOM config, plus one on a non-flash draft to show it is untouched, would settle it.

## Budget vs allocation, and a scale tensor that is dead for exactly these modules

`sub_pool_specs` (`:59-69`) charges for the fp32 scale tensor only when `kv_cache_dtype == "fp8"`. `allocate_kv_cache_tensors` allocates it unconditionally (`torch.zeros(2, L, blocks, kv_heads, ...)`).

For bf16 draft KV the runner sizes `num_kvcache_blocks` from a per-block cost that omits `2*L*nh*block_size*4` bytes, then allocates them anyway — an unbudgeted allocation at the exact moment the pool is sized, roughly 1.6% of the draft pool, i.e. under the post-allocation mismatch warn threshold.

Separately: `rope_cache`'s flash writer sets `per_token_quant = False` and overwrites the bound scales with the scalar `self.kv_scale` (`attention_mha.py:303-304`). So for exactly the modules this PR sends down the flash path, `eagle3_kv_scale` is never written and never read — yet it is still budgeted (`:59-69`), bound (`:182-184`), and exported as 2*L zero-filled `KVTransferRegion`s (`:220-229`).

In a PR whose motivation is an OOM at 1.38% margin, that is the adjacent win left on the table.

## Smaller, confirmed

- **`:130` — inconsistent defensiveness inside one 20-line predicate.** `rotary_emb`, `q_norm`, `k_norm` are read via `getattr(..., None)`; `impl.sliding_window` and `impl.head_dim` are bare reads on the same object two lines later. The caller passes `getattr(module, "impl", None)`, and the first three guards advertise tolerance of heterogeneous impls — so an impl with `rotary_emb` but without the other two raises `AttributeError` inside `build_kv_cache_tensor`, i.e. a startup crash during KV binding, after the pool is allocated, with a traceback pointing at the KV builder rather than the model.
- **Three sources of truth for `head_dim`.** The pool is shaped with `draft_hf.head_dim` (`:38`, bare, no fallback), the predicate reads `impl.head_dim`, and the model derives it with a fallback (`eagle3_llama.py:83-86`). Nothing asserts the three agree.
- **`:83` — the re-paging expression is an identity.** `self.num_blocks = config.num_kvcache_blocks * runner.block_size // self.block_size` with `self.block_size = model_runner.block_size` (`:33`) is literally `config.num_kvcache_blocks`. The docstring claims the budget is "re-paged at the draft's own block size", which the expression cannot do — so a future change introducing a distinct draft block size will look already handled, while `slot_mapping` (`backends.py:508-524`) and `unified_attention`'s `block_size` (derived from `k.shape[1]`) still assume `runner.block_size`.
- **`:118` / `:169` — the `impl is None` arm and the `if impl is not None:` guard.** Control reaches here only past the `base_attention` / `use_mla` gate at `:139-142`, and the `False` assignment is a no-op against the `__init__` default at `attention_mha.py:98`. `test_non_attention_module`'s docstring also states a rationale the code does not have ("None for anything that is not an attention layer" — non-attention modules returned None thirty lines earlier). The file ships a documented, tested, permanently-green branch describing control flow that does not exist.

## Docs-match-code

Four comments in the files this PR edits are now false or misleading:

- **`:95-98`** still says the flag is "Set by the attention backend's `build_kv_cache_tensor` ... AiterBackend keeps this False". The new writer is in `atom/spec_decode/`, a different package, and it sets True under AiterBackend.
- **`:780-784`** still states unconditionally that pure prefill uses the raw-K/V `block_size=1` view with the fake table — now false for the case this PR added (and unreachable per finding 4).
- **`:809-810`** calls flash "the (default) 4D flash layout" when the default is `False` at `:98` and `triton_mha.py:148` re-asserts `False`.
- **The new log string and docstrings coin "VHD"** for a layout the codebase already names **NHD** — `self._cache_format = "NHD"` at `attention_mha.py:333`/`:369`, and the layout table at `:418-419`. An operator grepping the log token finds nothing.

---

## Checked and refuted

Recording these so they are not re-raised:

- **The `kv_scale` tensor is not on CPU.** `torch.set_default_device` in `model_runner.py:702`/`:758` puts it on CUDA.
- **bf16 q against an fp8 cache is not a mismatch.** `unified_attention` treats `QUERY_DTYPE` and `KV_CACHE_DTYPE` independently.

atom/spec_decode/eagle3_kv_builder.py
```diff
@@ -118,19 +124,44 @@ def build_kv_cache_tensor(self, layer_id: int, module):
...
-            self.block_size,
-            x,
-        )
-        v_cache = cache[1, idx].view(
-            self.num_blocks,
-            self.num_kv_heads,
-            self.head_dim,
-            self.block_size,
+        impl = getattr(module, "impl", None)
+        flash = self._use_flash_layout(impl)
```
valarLip, line 131: impl already have  impl.use_flash_layout = False/True ? like https://github.com/ROCm/ATOM/pull/2106/changes#diff-e357263dd54a3481a557c8818d9eda459522c154a8da8dbe4e8da7de557aa9eeL148

yitingw1: This is a reasonable question, and the code does indeed suggest this - `flash` is the content written to `impl.use_flash_layout`, but this writing statement is located 22 lines below. I have moved the writing statement next to the decision statement:
+ if impl is not None:
            impl.use_flash_layout = flash

yitingw1:
Closed as https://github.com/ROCm/ATOM/pull/2147 has solved OOM.

PR #2112, opened 2026-09-02
Adding support to define profiler window 

devalshahamd:
# Profiler window controls: `--profiler-delay-iters` / `--profiler-max-iters`

## Motivation

ATOM profiling is unbounded today: `/start_profile` records until `/stop_profile` arrives. On a long run that means an unreadable trace and a kineto export whose cost scales with recorded events x number_of_ranks — `capture-trace/SKILL.md` already notes this can OOM workers on an 8-rank capture. The only workaround was shortening the workload (`PROMPT_MULTIPLIER=1`), which is no help when the long run is the thing you need to profile. There is also no way to skip the ramp-up right after `/start_profile`.

vLLM and SGLang both have these functionalities today.

| | Configured | Enforced |
|---|---|---|
| vLLM | Launch flags: `ProfilerConfig.delay_iterations` / `max_iterations` | Worker, inside the forward (`gpu_worker.py:951`) |
| SGLang | Request body: `start_step` / `num_steps` | Scheduler, between batches (`scheduler.py:3822`) |
| This PR | Launch flags (vLLM) | Engine, between forwards (Similar to SGLang) |

## Design

A step is one trip through the EngineCore loop: schedule a batch, dispatch the forward, wait for the result. Every dispatch goes through one function, `AsyncIOProcManager.call_func(..., wait_out=True)`, which blocks until the workers answer — the moment a forward is provably done, and the only place the engine learns it. That return is where the window advances. `call_func` also carries weight updates and cache setup, so a three-name set picks out the forwards: `forward`, `prefill_forward` and `dummy_execution`, the last being a DP rank's no-op batch, counted so idle ranks keep pace with busy ones.

The counters live in `EngineUtilityHandler`, which already owns `start_profile` and `stop_profile`: steps left before recording, steps recorded, and whether recording is on. `EngineCore.__init__` points the manager's `on_forward_end` at the handler's `profiler_step`, so each completed forward advances the window by one. One handler per EngineCore means one window per engine.

`/start_profile` with no delay configured starts the profiler immediately, as today. With `--profiler-delay-iters D` it arms instead: store `D`, reply without touching the workers, and let each completed forward count it down, the one that reaches zero starting the profiler. From there each forward increments the recorded count, and at `--profiler-max-iters M` the handler stops the profiler itself, which triggers the export and writes the trace — no `/stop_profile` needed, and sending one anyway is harmless. Both counters reset on the next `/start_profile`.

Stopping runs a multi-second kineto export, synchronously. Advancing after `call_func` returns puts that work between two forwards, the same quiescent point `/stop_profile` has always used, rather than freezing one halfway through.

## Changes

| File | Change |
|---|---|
| `engine_utility.py` | The window logic: state fields, `_start_profiler_now` / `_stop_profiler_now` shared by the manual and automatic paths, and `profiler_step()`. |
| `async_proc.py` | `_FORWARD_FUNCS` plus an optional `on_forward_end` callback, fired from `call_func` on the `wait_out=True` branch. |
| `engine_core.py` | Passes the config through and wires the callback (3 lines). |
| `config.py`, `arg_utils.py` | The two fields, a `>= 0` check, and the CLI flags. |
| `api_server.py`, `llm_engine.py` | `/start_profile` now returns the engine's answer: the armed message when a delay is set, 409 when a window is already open. |
| `tests/test_profiler_window.py` | New, 11 tests. |
| `configuration_guide.md`, `capture-trace/SKILL.md`, `benchmark-guide.md` | Flag docs. |

`model_runner.py` is unchanged. Both flags default to `0`, which is today's behaviour. An iteration is one engine step, not one model forward, so under MTP/EAGLE a window covers more than `M` model executions — same as vLLM counting worker steps and SGLang counting batches. Each EngineCore has its own counter, so its TP ranks stop together while DP and disagg engines get separate budgets.

## Usage

```bash
python -m atom.entrypoints.openai_server --model <model> -tp 8 \
  --torch-profiler-dir /app/logs/traces/run1 \
  --profiler-delay-iters 100 --profiler-max-iters 200
```

```bash
curl -X POST http://127.0.0.1:8000/start_profile
# {"status":"success","message":"Profiling armed. Recording starts after 100 engine steps.","armed_after_iters":100}
```

Steps 101-300 are recorded and the trace is written automatically at step 300; no `/stop_profile` needed. Traces land where they always did, `<dir>/<worker_name>_ts_<timestamp>_<ms>.pt.trace.json.gz`. The delay counts steps, not seconds, so nothing is written if the workload ends first.

## Tests

`tests/test_profiler_window.py`, 11 tests, no GPU. They drive the real `EngineUtilityHandler` and real `call_func`; only the RPCs are faked. Covered: which step each RPC lands on for four `(delay, max)` combinations, rejection of a second `/start_profile`, counters resetting between requests, the hook firing only for forward names on waited calls, and `profile_active` tracking the window on a real `Scheduler`.

Two AST guards fail CI if a new engine RPC is neither classified as a forward nor excluded, or if a forward is dispatched without waiting for its output — `_FORWARD_FUNCS` is hand-maintained and `prefill_forward` arrived with PD disagg.

## Real-workload validation

MI355X (gfx950), TP=1, ROCm 7.2.4 / torch 2.10.0, base commit `171de45` with this change (`cdc306c`), `ATOM_PROFILER_MORE=0`. Each model was run three ways: `--profiler-max-iters 200` alone, the same behind `--profiler-delay-iters 100`, and with both flags unset to confirm nothing changed for existing users.

| Model | `max-iters 200` | with a 100-step delay | flags unset |
|---|---|---|---|
| gpt-oss-20b | 200 steps exact, 3.0 MB, 1.3 s export | pass | pass |
| Qwen3-8B | 200 steps exact, 3.1 MB, 1.4 s export | pass | pass |
| Qwen3-Next-80B-A3B-FP8, MTP (3 spec tokens) | 200 exact, both trace tracks | 200 exact | pass |
| PD disagg, 1 prefill + 1 decode GPU | 200 exact per role | — | — |

Every window closed at **exactly 200 engine steps**, on the CPU and GPU annotation tracks, with `/stop_profile` never called, and serving continued through every export. Profiled unbounded instead, the same workload on the same server produces a 70-101 MB trace with a 32-41 s export stall, against ~3 MB and ~1.3 s for the bounded window.

- **The delay is verified by contrast.** With it, the first recorded step is `decode[bs=32 tok=32 ...]`; without it, `prefill[bs=1 tok=224 ...]`.
- **MTP confirms a step is not a model execution.** `decode[bs=32 tok=128 d=32 spec=3]` is 32 sequences at 4 token positions each, so 200 steps span 800 target token positions plus draft passes (includes drafter and target verifier both in single step), and the counter still stops at 200.
- **PD disagg gets per-role budgets** The same max and delay interation is applied to prefill and decode independently. 
- **The redundant stopis no-oped.** Tested with InferenceX's `benchmark_serving.py` unmodified, which brackets profiled runs with start/stop: the profiler stopped after max iterations reached, the client's late stop_profile no-oped, one trace was written, 512/512 requests succeeded.


ChuanLi1101 (changes requested):
Requesting changes before this can be treated as AgentX/Hyperloom support:

- **Per-request window.** `/start_profile` ignores the body; `_handle_start_profile` drops `args`. CLI-only matches vLLM, not SGLang. Hyperloom needs optional `delay_iters` / `max_iters` on the request (CLI as defaults).
- **Atomic multi-engine start.** If one engine is already profiling, the others still start and the API returns 409. That splits DP/PD. 409 must mean nothing new started (rollback or two-phase).
- **PD/DP counters are independent.** Same delay on prefill vs decode is not “last N generation steps.” PD delay is untested in the PR table.
- **Failed auto-start.** If `_start_profiler_now()` raises when delay hits 0, pending is already 0 and the client never hears. Reset to idle and surface the error.

Nits: don’t re-enter `call_func` from `on_forward_end` (tick after the forward returns); auto-stop export still blocks the engine loop; rebase off fork `main`.

devalshahamd:
Thanks  @ChuanLi1101 for the feedback. I have made changes as per your feedback

**Per-request window.** `/start_profile` now takes an optional body with `delay_iters` / `max_iters`, with the CLI flags as defaults and a bodyless request behaving exactly as before. The values resolve into per-run fields rather than overwriting the config, so one request's window can't leak into the next. This will also help with cross-process PD runs. 

**Atomic multi-engine start.** Replaced the single broadcast with two rounds: a `reserve` that only checks and flags, no worker RPC, then a `commit` once every engine has cleared. Any conflict in the reserve round releases the reservations and returns 409 with nothing started, so the status code is now literally true. Went with reserve/commit rather than rollback because there's no per-engine synchronous stop to roll back with.

**PD/DP counters.** Agreed that a shared delay across a PD pair isn't "the last N generation steps". That's now said outright in `configuration_guide.md` and the trace recipe instead of left implied. In cross-process PD each role is its own server, so the per-request body already gives each one its own window.

**Failed auto-start.** Fixed. Added "try:except" for auto start of the profiler. A failure is then reported when stop_profile is triggered. 

**Re-entrancy (nit).** `profiler_step` only moves counters and records a deferred start/stop now; the RPC goes out from `process_queue`, which all four busy loops already call at the top of every iteration. I left the hook itself in `call_func`, since that's the one point all five forward dispatch sites funnel through.

**Export blocking the loop (nit).** vLLM and SGLang both block, the stall sits at a quiescent point between forwards either way. Making export async races kineto teardown against live workers, so it requires careful design, which we can follow up in later commits.

PR #2121, opened 2026-09-03
[ATOM][PD][DCP] Support MLA/Indexer cache transfer in P-CPP/D-DCP scenario

MengqingCao:
## Motivation
In the P-CPP/D-DCP disaggregated setup, the prefill producer stores globally contiguous MLA and DSA index caches, while each decode DCP rank owns an interleaved subset of tokens. Whole-block transfer therefore cannot preserve the consumer layout.

The DSA index cache also uses an MFMA preshuffled layout, so its per-token bytes are not contiguous and cannot use the MLA token-run RDMA path directly.

## Technical Details

```mermaid
flowchart TB
    Producer["CPP Producer<br/>Globally contiguous scheduler blocks"]
    Ownership["DCP4 ownership<br/>rank r owns tokens r, r+4, r+8, ..."]
    Plan["Shared DCPShardPlan<br/>source block/token → destination page/token"]

    Producer --> Plan
    Ownership --> Plan

    subgraph MLA["MLA Latent KV Transfer"]
        MLA_Source["Token-contiguous MLA cache<br/>576 bytes/token for GLM-5.2 FP8"]
        MLA_Runs["Generate source/destination token runs"]
        MLA_Coalesce["Coalesce adjacent contiguous runs"]
        MLA_RDMA["Mooncake RDMA<br/>token-run descriptors"]
        MLA_Dst["DCP-local MLA pages"]

        MLA_Source --> MLA_Runs
        MLA_Runs --> MLA_Coalesce
        MLA_Coalesce --> MLA_RDMA
        MLA_RDMA --> MLA_Dst
    end

    subgraph Indexer["DSA Index Cache Transfer"]
        Index_Source["Preshuffled producer page<br/>Key: [T/16, D/16, 16, 16]<br/>Scale: [T, D/128] × FP32"]
        Reconstruct["Group block_ratio physical pages<br/>into one scheduler-block view"]
        Gather["GPU gather using shared plan<br/>block ID + token tile + in-tile row"]
        Staging["Complete DCP-local staging pages<br/>repack keys + scales<br/>zero invalid tail and padding"]
        Fence["Synchronize GPU stream"]
        Index_RDMA["Mooncake RDMA<br/>full-page descriptors"]
        Index_Dst["DCP-local preshuffled index pages"]

        Index_Source --> Reconstruct
        Reconstruct --> Gather
        Gather --> Staging
        Staging --> Fence
        Fence --> Index_RDMA
        Index_RDMA --> Index_Dst
    end

    Plan --> MLA_Source
    Plan --> Index_Source
```

- Add explicit semantic roles for MLA KV and DSA index-cache transfer regions.
- Introduce a shared `DCPShardPlan` describing producer-to-consumer token ownership.
- Transfer MLA KV using coalesced token-run RDMA descriptors.
- Repack preshuffled DSA index data on GPU into complete DCP-local pages before Mooncake RDMA.
- Reuse a bounded per-worker staging pool, processing one index layer/chunk at a time.
- Support PP-local to consumer-global compact region mappings.

### Unit Tests
```
python -m pytest \
  tests/test_dcp_sharded_transfer.py \
  tests/test_mla_index_cache.py \
  tests/test_pd_pp.py -q
```
Result: 99 passed

### GSM8K Accuracy
Start the no-MTP CPP4→DCP4 topology:
```
Prefill: CPP4, GPUs 0–3
Decode: DCP4, GPUs 4–7
MTP disabled
Mooncake + LMCache enabled
Replicated index cache disabled
```

```bash
#!/usr/bin/env bash
set -euo pipefail

MODEL=/mnt/models/GLM-5.2-MXFP4
LOG_DIR=/tmp/atom_cpp4_dcp4
mkdir -p "$LOG_DIR"

MOONCAKE_LIB="$(python3 -c \
  "import sysconfig; print(sysconfig.get_path('purelib'))")/mooncake"

export LD_LIBRARY_PATH="${MOONCAKE_LIB}:/opt/rocm/lib:${LD_LIBRARY_PATH:-}"
export AITER_LOG_LEVEL=WARNING
export ATOM_MLA_PAGE_SIZE=1
export ATOM_ONLINE_QUANT_STREAMING=0
export ATOM_SPARSE_INDEXER_LOGITS_BUDGET_MB=2047
export ATOM_USE_TRITON_MLA=0

ONLINE_QUANT='{"global_quant_config":"ptpc_fp8","exclude_layer":["lm_head","model.embed_tokens","*.mlp.gate","*expert*"]}'

PREFILL_KV='{"kv_connector":"multi","connectors":[{"kv_connector":"mooncake","kv_role":"kv_producer","proxy_ip":"127.0.0.1","handshake_port":6301,"protocol":"rdma"},{"kv_connector":"lmcache_offload","kv_role":"offload"}]}'

DECODE_KV='{"kv_connector":"mooncake","kv_role":"kv_consumer","proxy_ip":"127.0.0.1","handshake_port":6301,"protocol":"rdma"}'

bash scripts/stop_atom_server.sh
rm -rf "${HOME}/.cache/atom/"*

# Prefill: CPP4, GPUs 0-3
HIP_VISIBLE_DEVICES=0,1,2,3 \
VLLM_PP_LAYER_PARTITION=20,20,20,18 \
LMCACHE_LOCAL_CPU=True \
LMCACHE_MAX_LOCAL_CPU_SIZE=256 \
LMCACHE_CHUNK_SIZE=256 \
nohup python -m atom.entrypoints.openai_server \
  --model "$MODEL" \
  --host 0.0.0.0 \
  --server-port 8010 \
  --trust-remote-code \
  --kv_cache_dtype fp8 \
  --block-size 16 \
  --max-num-seqs 512 \
  --max-num-batched-tokens 8192 \
  --online_quant_config "$ONLINE_QUANT" \
  --level 3 \
  --enforce-eager \
  -tp 1 \
  --pipeline-parallel-size 4 \
  --kv-transfer-config "$PREFILL_KV" \
  >"$LOG_DIR/prefill.log" 2>&1 &

# Decode: TP4 + DCP4, GPUs 4-7
HIP_VISIBLE_DEVICES=4,5,6,7 \
ATOM_DCP_REPLICATE_INDEX_CACHE=0 \
nohup python -m atom.entrypoints.openai_server \
  --model "$MODEL" \
  --host 0.0.0.0 \
  --server-port 8020 \
  --trust-remote-code \
  --kv_cache_dtype fp8 \
  --block-size 16 \
  --max-num-seqs 512 \
  --online_quant_config "$ONLINE_QUANT" \
  --level 3 \
  -tp 4 \
  --decode-context-parallel-size 4 \
  --cudagraph-mode FULL \
  --kv-transfer-config "$DECODE_KV" \
  >"$LOG_DIR/decode.log" 2>&1 &

until curl -sf http://127.0.0.1:8010/v1/models >/dev/null; do sleep 5; done
until curl -sf http://127.0.0.1:8020/v1/models >/dev/null; do sleep 5; done

# PD router
nohup atomesh launch \
  --host 0.0.0.0 \
  --port 8000 \
  --pd-disaggregation \
  --prefill http://127.0.0.1:8010 6301 \
  --decode http://127.0.0.1:8020 \
  --policy random \
  --backend atom \
  --disable-circuit-breaker \
  >"$LOG_DIR/mesh.log" 2>&1 &

until curl -sf http://127.0.0.1:8000/v1/models >/dev/null; do sleep 5; done

echo "Server ready: http://127.0.0.1:8000"
echo "Logs: $LOG_DIR"
```

Run GSM8K 20-shot:
```
lm_eval \
  --model local-chat-completions \
  --model_args model=/mnt/models/GLM-5.2-MXFP4,base_url=http://127.0.0.1:8000/v1/chat/completions,num_concurrent=64,max_retries=3,tokenized_requests=False \
  --tasks gsm8k \
  --num_fewshot 20 \
  --batch_size auto \
  --apply_chat_template \
  --fewshot_as_multiturn \
  --gen_kwargs max_tokens=16384,temperature=0,top_p=1 \
  --log_samples \
  --confirm_run_unsafe_code
```

### Results:

```
local-chat-completions ({'model': '/mnt/models/GLM-5.2-MXFP4', 'base_url': 'http://localhost:8000/v1/chat/completions', 'num_concurrent': 64, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({'max_tokens': 16384, 'temperature': 0, 'top_p': 1}), limit: None, num_fewshot: 20, batch_size: 1
|gsm8k|      3|flexible-extract|    20|exact_match|↑  |0.9719|±  |0.0045|
|     |       |strict-match    |    20|exact_match|↑  |0.9727|±  |0.0045|
```

## Limitations

- DCP is currently supported only on the decode side. A Mooncake prefill producer must not enable DCP because the transfer plan assumes that producer blocks contain tokens in globally contiguous order.
- A future optimization can replace this path with a fused Triton kernel that gathers key and scale planes, reorganizes physical pages, and clears invalid tail tokens in one launch. Performance gains have not yet been measured.

## Submission Checklist


yitingw1:
`uild_dcp_index_gather_plan`/`gather_dcp_preshuffled_index_pages` (the new MFMA-tile repack, and the highest-risk code in this PR) has zero test coverage, and the reused plan_sharded/_coalesce lost the unit tests PR #2008 had for them.

yitingw1:
The MLA-latent plan_sharded() path has no guard against the producer itself running with dcp_size > 1 — PR #2008 had an explicit raise RuntimeError here since plan_sharded assumes the producer's own blocks are globally contiguous; a sharded producer would silently corrupt the relayout instead of failing loudly. Worth adding back, at least as an assert.

MengqingCao:
@yitingw1 Thanks for your suggestions, your concerns are all addressed in https://github.com/ROCm/ATOM/pull/2121/changes/eb2b6a1f40577ad876370eb09a80d16b7d3d25b4, PTAL.

atom/kv_transfer/disaggregation/types.py
```diff
@@ -97,6 +97,12 @@ def key(self) -> ConnectorCompletionKey:
         return self.channel, self.operation_id
 
 
+# Region roles the DCP relayout dispatches on: an MLA latent region is always
+# interleave-sharded, a DSA index region may instead be replicated whole.
```
yitingw1: Please update here.

MengqingCao: fixed in https://github.com/ROCm/ATOM/pull/2121/changes/a106c33f1f22e47b0f63556b0e5764477509d838, thanks!

yitingw1 (dismissed):

valarLip:
Reviewed at `19e047b3d`. The core numeric machinery checks out — the preshuffle gather's advanced-index/reshape/permute chain, `coalesce_contiguous`, and the `off * dcp_size` slicing all round-trip correctly. Everything below is about the contract around it.

## 1. Regression: PD incremental transfer is now disabled on every symmetric-DCP request

```diff
-            if remote_hash_block_size != self.hash_block_size:
+            if remote_hash_block_size != self.block_size:
```

The producer side was not changed to match: `mooncake_connector.py:318` still computes `self.hash_block_size = self.block_size * self.dcp_size`, and `:456` still advertises that product on the wire. The `elif` on the very next line still divides by `self.hash_block_size`.

So with producer dcp=2 / consumer dcp=2 / block_size=16, the producer advertises 32 and the consumer compares it against 16. Mismatch on every request, "PD incremental transfer disabled ... falling back to full transfer" on every request, prefix reuse across the P/D boundary silently lost. Throughput only, no error — which is why it will not show up in a correctness test. The old comparison was correct for symmetric DCP.

The other direction is a new false positive: a producer with block_size=8, dcp=2 advertises 16 and is accepted by a consumer with block_size=16, dcp=4. `num_computed_blocks = num_cached_tokens // 64` then slices `remote_block_ids[off*4:]`, but those ids are the producer's rank-local interleaved pages, not global-contiguous blocks. The producer-side ceil check passes too, so the wrong prefix transfers with no diagnostic.

Suggest advertising the producer's `dcp_size` (or its raw `block_size`) as its own wire field rather than inferring it back out of a product.

## 2. Relaxing the src/dst count guard opens a path that was never taught about DCP

`_execute_transfer:1333` changed from `len(src) != len(dst)` to `expected_dst_blocks = -(-len(src) // consumer_dcp_size)`. But `_execute_transfer:1357` dispatches to `_execute_block_slot_transfer` when `_has_slot_regions`, and that function (`:1693`) contains no reference to DCP at all — it still does whole-block `zip(src_block_ids, dst_block_ids)` at `:1731`.

With slot regions (DSV4 compressor state) plus a dcp=4 consumer: src=8, dst=2 now satisfies the ceil-div check, so the abort no longer fires. Producer blocks 0 and 1 are copied *whole* into consumer pages 0 and 1 — unsharded, wrong tokens — and blocks 2..7 are dropped. That path also never sees the new `consumer_region_roles` / `consumer_block_bpb` / interleave validation, and the producer still sends write-done afterwards, so the consumer decodes on corrupt KV rather than timing out.

The equality guard was the only thing protecting this path. It needs either DCP support or an explicit rejection.

## 3. The token relayout's polarity is inverted: role-less regions are opted in, not refused

In the region loop the branches are:

```
role == INDEX_CACHE_ROLE and staging  ->  staged
sharded_runs is None                  ->  whole-block
otherwise                             ->  token relayout
```

So *every* region whose `semantic_role` is not `INDEX_CACHE_ROLE` gets the MLA per-token interleave under DCP, including regions that never declared one.

`model_runner.py:2062` does `transfer_tensors.block_regions.extend(draft_regions)`, and `eagle3_kv_builder.py:161-178` builds those with no `semantic_role`. With a DCP consumer the fp8 `eagle3_kv_scale` regions have a `unit_bytes` of a few bytes per block, so `bpb % self.block_size` at `:1578` is non-zero and every request dies inside a send worker with `RuntimeError: Region N stores B bytes per block, which block_size 16 does not divide`. The draft KV regions themselves divide evenly, so they silently get an interleave the consumer does not shard by. `docs/context_parallel_guide.md:645` lists spec decode under DCP as supported.

A region that has not declared a role should be *refused* under DCP, not given the MLA treatment by default.

## 4. `bpb % block_size` proves divisibility, not token contiguity

The comment at `:1506-1507` — "Safe in token units because DCP only runs on MLA, which stores a token contiguously" — is the actual precondition, and it is enforced nowhere. The divisibility check is not a proxy for it.

`stages_sharded_index` keys on the literal `INDEX_CACHE_ROLE`, which only `aiter_mla.py` emits. A DeepSeek-V4 producer tags its CSA indexer regions `f"{role_prefix}.layer_{id}"` (`deepseek_v4_attn.py:1916`), so staging never engages and those regions fall through to the token-unit path — while `deepseek_v4_attn.py:1302` documents that layout as block-major `[rows*index_head_dim FP8][rows*4 fp32 scale]`, not token-contiguous. `bpb % block_size == 0` holds, so nothing fires: silently misaligned KV, garbage indexer top-k, no exception and no log line. `aiter_attention.py:730` sets no role at all. V4 also has no equivalent of the `aiter_mla` `dcp_world_size > 1` producer guard.

## 5. `_pd_producer_connectors` defaults a missing `kv_role` to producer, and the guard it feeds is a hard startup failure

```python
connector.get("kv_role", "kv_producer") == "kv_producer"   # aiter_mla.py:368
```

The offload connectors default `kv_role` to `"offload"` in their own validator (`_offload_common.py:32`). So `--kv-transfer-config '{"kv_connector":"lmcache_offload"}'` on a DCP decode node is classified as a P/D producer and `aiter_mla.py:1379` refuses to start the server with "P/D transfer requires an unsharded producer".

Two more reachable cases:

- `{'kv_connector':'multi','connectors':[]}` — `.get("connectors") or [transfer_config]` returns the wrapper dict, which has no `kv_role`, so it is a "producer" too.
- The guard keys only on `dcp_world_size > 1`, not on whether an `index_cache` exists. It therefore fires for plain dense MLA, where none of the new relayout machinery is reachable, and it kills symmetric-DCP P/D (producer dcp=4 + consumer dcp=4), which was block-for-block correct before this PR.

The parametrized test only covers `kv_role` spelled out explicitly, so none of the three is caught.

## 6. `scales_per_token` contradicts `aligned_index_cache_dim` in the same file

`aligned_index_cache_dim` (`aiter_mla.py:110`) reserves `index_head_dim + 4` bytes — exactly **one** fp32 scale per token, regardless of head dim. `deepseek_v4_attn.py:494` states the same rule as `_index_row_bytes = index_head_dim + 4`.

The gather (`:338`) instead computes `scales_per_token = index_head_dim // quant_block_size`, default 128. The two agree only at `index_head_dim == 128`.

At `index_head_dim = 256` the gather reads two scales per token, so `source_bytes[:, key_bytes : key_bytes + block*8]` spans the real `block*4` scale plane plus `block*4` bytes of row padding, and `reshape(B, block_size, 2)` lands token *t*'s scale at destination position *2t*. `aligned_index_cache_dim(256) = 272` leaves enough room that no bounds error fires — the indexer just gets scrambled scales.

The `index_head_dim % quant_block_size` check at `:291` reads like it covers this, but it passes for every multiple of 128. `quant_block_size` is a defaulted parameter no caller passes. Latent today; armed for the next model.

## 7. Static config is validated per-request on a send worker, and the failure wedges the pair

`self.block_size % 16` and `_prepare_sharded_index is None` are known at startup — right next to the existing `dcp_world_size > 1` check at `aiter_mla.py:1380`. `interleave` is known at handshake. All three are raised per-request at `:1527` instead.

Launch a DCP consumer with `--dcp-config '{"interleave_size": 2}'` (documented as supported, `docs/context_parallel_guide.md:337-340`) against a DSA producer: `stages_sharded_index` becomes True (2 < 16) and `:1534` raises. `_execute_transfer`'s blanket `except Exception` logs it at `:1420` and returns without write-done. On the consumer, `_pending_recv`, `_pending_recv_blocks`, `_pending_recv_expected`, `_pending_recv_stages` and `_pending_recv_nonce` are cleared only by an arriving write-done (`:2017-2058`) — nothing expires them. Every request wedges permanently holding its KV blocks, the producer leaks `_completed_prefills` / `_transfer_refcount`, and the pair has to be restarted.

A config error should be a launch failure, not an unexplained decode-side hang.

## 8. `KimiAiterMLAGDNMetadataBuilder` inherits the staging path it cannot satisfy

`kimi_mla_gdn_attn.py:502` is `class KimiAiterMLAGDNMetadataBuilder(_KimiMLAGDNCommon, AiterMLAMetadataBuilder)` and neither base overrides `get_kv_transfer_tensors`. Kimi-K3 / GLM-5.3-Flash allocates `index_cache` as `(layers, num_scheduler_blocks, rows_per_block, aligned)` (`:365`) and its `_index_cache_block_bytes` override returns `_index_rows_per_block() * aligned * 1` (`:216`) — no `block_ratio`, and fewer rows than tokens under `index_kpool > 1`.

The staging pool is sized from that override, then handed to `gather_dcp_preshuffled_index_pages`, which asserts `source.shape[1] * block_ratio == scheduler_block_size` and recomputes `page_bytes = scheduler_block_size * aligned * elsize`. Both are false. The first DCP consumer request dies in a send worker with `ValueError: Source physical page R × ratio B does not match scheduler block S`. `docs/context_parallel_guide.md:649` lists hybrid KDA + MLA (Kimi-K3) as DCP-supported, and there is no startup-time rejection.

## 9. MoRIIO is counted as a producer and has zero DCP awareness

The PR's own test asserts `{'kv_connector':'moriio','kv_role':'kv_producer'} -> True`, yet `grep -c dcp moriio_connector.py` returns **0**: no `consumer_dcp_size`, no `semantic_role`, no `index_staging`, and its `register_kv_caches` (`:187`) never reads `transfer_tensors` at all.

So a MoRIIO producer allocates `pool_size × 256 × page_bytes` of VRAM that nothing can reach — inside `get_kv_transfer_tensors`, i.e. after the KV budget was planned and inside model_runner's pre_alloc/post_alloc cross-check window — and still does whole-block writes into a DCP-sharded consumer, which is exactly the silent corruption this PR fixes for Mooncake. No guard, no `NotImplementedError`. The staging pool is also sized with the default 16 while moriio spawns 4 threads (`:126`), which contradicts the new constant's comment that the two "cannot drift independently".

## 10. Missing column guard on the staging slice

`output = staging[:dst_pages, :page_bytes]` (`aiter_mla.py:314`) has a `dst_pages > staging.shape[0]` guard but no `page_bytes <= staging.shape[1]` one. Python slicing clamps silently.

The slot width comes from `_index_cache_block_bytes` (`:1266`: `stride(0) * elsize * block_ratio`) while the gather recomputes `page_bytes = scheduler_block_size * source.shape[2] * source.element_size()` (`:305`). Provably equal for the base builder, but the method's own docstring at `:1260-1263` says overriding is expected and Kimi already does. When they disagree, `output` loses trailing columns, `output.zero_()` clears only the truncated prefix, and `_execute_staged_index_layer_chunk` still RDMAs the full `bytes_per_page` from `slot.data_ptr()` — shipping uninitialized `torch.empty` bytes into the consumer's index cache. Worth adding next to the other three validations.

## 11. `build_dcp_shard_plan` re-derives a rule that is already declared

`sharded_transfer.py:134`:

```python
global_token = ((local_token // interleave_size) * dcp_size + dcp_rank) * interleave_size
```

That is `dcp_ops.py:815 dcp_global_pos` minus its `+ (local_index % S)` term, which is zero only because the plan samples run *starts* — a precondition stated nowhere. `aiter_mla.py:1597/2179` (this same file) already imports `dcp_local_index` / `dcp_owner_rank` for slot mapping, and `block_manager.py:576-582` warns explicitly that "an inverse derived by hand would drift from the allocator as soon as that arithmetic moved".

If the canonical rule changes, attention follows it and the RDMA relayout does not: tokens land where the old rule says the consumer owns them. No crash, no count mismatch, just wrong attention output. `tests/test_dcp_sharded_transfer.py:38-46` computes its expectation with the identical formula and never references `dcp_global_pos`, so the test stays green through the drift.

## 12. Performance: per-token descriptors and a default-stream sync in the inner loop

**Descriptor blowup.** `DCPShardPlan.token_runs` explicitly cannot merge (consecutive runs are `interleave*dcp` apart at the source but `interleave` apart at the destination). At block_size=64, dcp=8, a 32K prompt, 61 MLA layers: 61 × 4096 ≈ 250k descriptors of 576 B each per consumer rank, versus 61 × 64 ≈ 3.9k of 36 KB before. `_rdma_write_with_retry` batches at 4096, so that is ~61 synchronous batches of message-rate-bound writes instead of one. Host side, the three `.tolist()` calls at `:1587-1589` build ~750k Python ints (~27 MB) per in-flight request, and `sum(sizes)` at `:1598` walks all of them as an eager argument to a `logger.debug` that evaluates at any log level — with 16 send workers that is hundreds of MB of transient host allocation under the GIL on the prefill node.

**Default-stream sync.** Send workers get `initializer=torch.cuda.set_device` but no dedicated stream, so `gather_sharded_index` enqueues its zero/gather/permute/copy on the device default stream — the same one the prefill forward uses — and `torch.cuda.current_stream().synchronize()` at `:1656` waits for the entire in-flight forward. It sits in the innermost loop (`:1617` -> `:1618` -> `:1656`), so a 61-index-layer DSA model with 8 consumer DCP ranks does ~488 full default-stream syncs per prefill request with zero repack/NIC overlap (each worker holds one slot, so the 16-slot pool cannot double-buffer). Separately, model work is issued on `async_execute_stream` (`model_runner.py:4340`) and the TBO `_prefill_streams` / `_decode_streams`; no `wait_event` links those to the worker's stream, so read-after-write rests on an incidental end-of-step host sync — the same class of race the comment at `:1801-1803` documents for `_gather_slot` under TBO.

## 13. Staging pool: one buffer, N free lists, sized by `max()` when concurrency is `sum()`

`get_kv_transfer_tensors()` is called once (`model_runner.py:2053`) and `MultiConnector.register_kv_caches` (`multi_connector.py:267-268`) hands the same `KVTransferTensors` to every sub-connector, each of which then does `self._index_staging_free = list(range(tt.index_staging_pool_size))` (`:783`) over the same buffer. `_build_subconnectors` does not forbid two P/D-producer entries. With two, worker A of connector 1 and worker B of connector 2 both pop slot 0, both call `gather_sharded_index(pool_idx=0)`, and each RDMAs a page image half-overwritten by the other — silent corruption, no error.

Even without aliasing, `_pd_staging_pool_size` returns `max(worker_counts)` while total senders is the `sum`. The surplus falls into `_acquire_index_staging_slot`'s unbounded `while True: time.sleep(0.001)` (`:1139`) with an un-rate-limited warning per (chunk × index layer).

## 14. None of the new tests run in CI

`tests/test_mla_index_cache.py:24` is `pytest.skip(..., allow_module_level=True)` on `import aiter`, and `.github/scripts/run_unit_tests.sh` runs on a plain CPU runner with no aiter. So `test_preshuffled_index_gather_reorganizes_pages_and_zeros_tail`, `test_index_staging_requires_a_pd_producer`, `test_pd_staging_pool_matches_producer_worker_count`, `test_pd_staging_pool_rejects_invalid_worker_count` and `test_dcp_pd_producer_is_rejected_before_transfer_planning` — roughly 200 lines of new coverage — never execute.

None of them need aiter: the gather test runs on `torch.device('cpu')` and the two pool tests are pure dict parsing. The fix pattern is already in this PR — `tests/test_dcp_sharded_transfer.py` is a separate aiter-free file that does run.

Two gaps in the gather test itself, on top of that: it uses `scheduler_block_size=16`, so `token_tiles == 1` and `src_token_tile` is identically zero — the tile axis is untested — and it passes a staging tensor of exactly `(dst_pages, page_bytes)` rather than a real `(256, page_bytes)` pool slot, so neither the slot-width tail nor the multi-chunk `slice_pages` path is covered.

MengqingCao:
@valarLip  Thanks for the detailed review comments! I've addressed them in https://github.com/ROCm/ATOM/pull/2121/commits/7f34c427d1cf8ade88732170d402aca72054bd00  now. The changes is big,  thus I list all the changes here to describe it clearly:

### 1. Incremental page geometry validation
#### Fix:

Producer now explicitly sends block_size and dcp_size.
Incremental transfer is supported only for unsharded producers or symmetric DCP.
Mismatched configurations disable incremental transfer and fall back to full transfer.
Legacy hash_block_size compatibility is retained.

#### Code locations:

atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:331-357
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:427-448
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:473-496
tests/test_pd_pp.py:144-277
### 2. Slot regions do not support DCP

#### Fix:

Explicitly reject slot region + DCP consumer.
Prevent silently producing corrupted KV data.
#### Code location:

atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1405-1438
### 3. Undeclared regions incorrectly enter MLA relayout

#### Fix:

Only mla.kv regions may use token relayout.
Only dsa.index_cache regions may use index staging.
Other roles fail explicitly.
#### Code locations:

atom/kv_transfer/disaggregation/types.py:124-150
atom/model_ops/attentions/aiter_mla.py:1229-1252
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1673-1717
### 4. V4 indexer cannot be validated by divisibility alone

#### Fix:

V4 indexer regions no longer enter MLA token relayout.
Only regions explicitly marked as dsa.index_cache enter staging.
#### Code locations:

atom/model_ops/attentions/deepseek_v4_attn.py:1900-1917
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1647-1665
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1690-1723
### 5. Incorrect producer configuration classification

#### Fix:

Only Mooncake and MoRIIO are classified as P/D producers.
multi configurations are expanded correctly.
The overly strict “DCP producer cannot start” restriction was removed.
Symmetric DCP keeps whole-block transfer; only producer dcp=1 → consumer dcp>1 uses relayout.
#### Code locations:

atom/kv_transfer/disaggregation/pd_producer.py:13-73
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1631-1646
tests/test_pd_producer_config.py:13-55
### 6. Incorrect scales_per_token calculation

#### Fix:

Each token now always uses exactly one fp32 scale.
Tests cover head dimensions 128 and 256.
Code locations:

atom/kv_transfer/disaggregation/index_staging.py:131-145
tests/test_dcp_index_staging.py:14-104
### 7. Request-time configuration errors permanently hang the consumer

#### Fix:

Validate static configuration at startup or during handshake.
Send success=False for all request-time failures.
The consumer falls back to local prefill after receiving the failure notification.
#### Code locations:

atom/model_ops/attentions/aiter_mla.py:1349-1366
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:850-861
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1340-1528
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:2098-2145
atom/model_engine/scheduler.py:1768-1803
### 8. Kimi GDN does not support index staging

#### Fix:

Disable DCP index staging for Kimi GDN.
Do not allocate a staging pool or perform gather.
Non-DCP Kimi P/D remains supported.
#### Code locations:

atom/model_ops/attentions/kimi_mla_gdn_attn.py:502-504
atom/model_ops/attentions/aiter_mla.py:1108-1115
atom/model_ops/attentions/aiter_mla.py:1349-1354
### 9. MoRIIO does not implement DCP relayout

#### Fix:

Allocate the staging pool only for Mooncake producers.
MoRIIO no longer consumes staging memory.
Explicitly mark MoRIIO producer → DCP consumer as unsupported.
#### Code locations:

atom/kv_transfer/disaggregation/pd_producer.py:13-18
atom/kv_transfer/disaggregation/pd_producer.py:58-65
atom/model_ops/attentions/aiter_mla.py:1349-1367
### 10. Staging-slot width validation

#### Fix:

Validate page_bytes <= staging.shape[1].
Raise ValueError when the condition is not satisfied.
#### Code locations:

atom/kv_transfer/disaggregation/index_staging.py:83-109
tests/test_dcp_index_staging.py:109-120
### 11. Unified DCP layout formula

#### Fix:

Move the canonical DCP formulas into dcp_layout.py.
Make the shard plan call dcp_global_pos.
Re-export the functions from dcp_ops.py for compatibility.
#### Code locations:

atom/distributed/dcp_layout.py:12-47
atom/kv_transfer/disaggregation/sharded_transfer.py:99-140
atom/model_ops/dcp_ops.py:23-28
### 12. Descriptor count and CUDA stream synchronization

#### Fix:

Generate and send descriptors incrementally, with a maximum of 4096 entries per batch.
Compute sum(sizes) and other statistics only when logger.isEnabledFor(DEBUG) is true.
Use a private CUDA stream.
Establish compute/transfer dependencies with ready events and wait_event().
Remove the full default-stream synchronization from the inner loop.
#### Code locations:

atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:822-829
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1599-1761
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1762-1771
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:1808-1845
atom/model_engine/model_runner.py:3290-3309
atom/model_engine/model_runner.py:3380-3402
Not fixed: DCP token-run descriptors cannot be directly merged because adjacent runs are not contiguous on both the source and destination sides. Reducing the descriptor count further would require staging the MLA KV and transferring complete pages, which is outside the scope of this correctness fix.

### 13. Multiple connectors sharing the staging pool

#### Fix:

Allow only one Mooncake producer connector per process.
Size the staging pool according to that connector’s worker count.
#### Code locations:

atom/kv_transfer/disaggregation/pd_producer.py:76-101
atom/model_ops/attentions/aiter_mla.py:1366-1379
atom/kv_transfer/disaggregation/mooncake/mooncake_connector.py:822-829
### 14. New tests skipped because of the aiter import

#### Fix:

Move producer configuration tests to test_pd_producer_config.py.
Move index-staging tests to test_dcp_index_staging.py.
Add independent CPU tests for shard plans, layout, and incremental geometry.
#### Code locations:

tests/test_pd_producer_config.py:1-122
tests/test_dcp_index_staging.py:1-120
tests/test_dcp_sharded_transfer.py:1-120
tests/test_pd_pp.py:144-277

valarLip (approved):

PR #2122, opened 2026-09-03
fix(minimax-m3): unblock fp8 KV cache and EAGLE3 spec decode on the vLLM plugin

PerryZhang01:
Two fixes that together let MiniMax-M3 serve under `--kv-cache-dtype fp8` with an EAGLE3 draft on the vLLM plugin backend. Neither one alone is enough: the first unblocks prefill, the second unblocks decode.

## 1. The indexer cache must carry a real fp8 dtype

Serving MiniMax-M3 through the vLLM plugin backend with `--kv-cache-dtype fp8` kills EngineCore on the first request:

    triton.compiler.errors.CompilationError: at 73:17
        qk = tl.dot(q, k, out_dtype=tl.float32) * sm_scale_log2e
    Both operands must be same dtype. Got bf16 and uint8

`MiniMaxM3SparseIndexerCache` takes its KV-cache spec dtype from vLLM's `kv_cache_dtype_str_to_dtype()`, and vLLM maps every fp8 kv-cache-dtype to `torch.uint8` -- a byte buffer its own kernels reinterpret. ATOM's `_index_block_score_kernel` dispatches on `k.dtype.is_fp8()` instead, so a uint8-labelled index cache sends it down the bf16 branch, where it dots bf16 against uint8 and fails to compile.

The bytes in that buffer are already fp8: aiter's `fused_qknorm_idxrqknorm` writes them under `kv_cache_dtype="fp8"`, and the native ATOM server allocates the same cache as `dtypes.d_dtypes["fp8"]`. Only the torch dtype label was wrong, so label it `aiter.dtypes.fp8` (`torch.float8_e4m3fn`). The element stays one byte, so vLLM's page-size accounting is unchanged, and vLLM already allows a true fp8 dtype in this position -- its own `"fp8_inc"` entry maps to `torch.float8_e4m3fn`.

Scoped to the indexer cache deliberately: the main sparse and dense KV caches keep their uint8 label because the aiter paged kernels read them as raw bytes.

## 2. A decode segment must be uniform before a decode kernel may have it

With the above fixed, M3 + EAGLE3 still dies part way into a long run:

    atom/model_ops/minimax_m3/index_topk.py minimax_m3_index_topk_decode
    AssertionError: total_q 121 not divisible by max_query_len 4

Three ATOM decode paths recover a request from a flat query row by dividing:

- the M3 index-topk kernels use `row // max_query_len`, and the row's causal cutoff `seq_len - max_query_len + tok + 1`
- aiter's gluon paged decode reshapes q to `[q.shape[0] // max_query_len, max_query_len, ...]`

All three only hold when every decode request contributes exactly `max_query_len` rows, and none of them checked. Speculative decode breaks the assumption on both models. On the target, a request that joins the batch without draft tokens contributes one row while its neighbours contribute `num_spec + 1`, and vLLM keeps all of them in the decode segment because each query length is still within the reorder threshold (121 = 30*4 + 1). On the draft, a request contributes however many tokens the previous step accepted, so the segment is ragged by construction -- which surfaces as

    aiter/ops/triton/gluon/pa_decode_gluon.py
    RuntimeError: shape '[27, 4, 1, 16, 128]' is invalid for input of size 225280

(110 rows where 27 requests * 4 = 108 were assumed).

`_uniform_decode_query_len()` measures the segment, and all three builders consult it:

- `MinimaxM3SparseAttentionMetadataBuilder.build()` takes the uniform decode fast path only when the segment really is uniform; otherwise those requests go to the prefill kernel, which derives causality from `cu_seqlens_q`/`context_lens` and accepts variable query lengths.
- `AiterMhaMetadataBuilderForVllm.build()` routes a ragged decode segment to the extend path, which is varlen. Decode requests sort before extends, so widening the extend segment covers them without reordering.
- `AiterMhaMetadataBuilderForVllm.build_for_drafting()` -- the one drafting actually goes through. Its docstring claimed "during EAGLE/MTP drafting all requests are uniform decodes" and it only tested for prefills, so it never reached `build()`. It now falls back to `build()` when the batch is ragged.

Both mixed-batch branches also stopped reporting a `max_query_len` they had not measured: the M3 one passed `reorder_batch_threshold` (always `num_spec + 1`) and the MHA one a max over the segment. A plain-decode segment under a spec-decode threshold takes 4 there, and with a row count that happens to divide by 4 the kernels mis-map every row silently instead of asserting. `tests/plugin/test_minimax_m3_decode_uniformity.py` pins that case (`[4, 4, 6, 2]` sums to 16, divides by 4, and is still ragged).

The kernel-side assert stays as a backstop, now spelling out the invariant.

## 3. CI

The M3 accuracy cell ran with `--kv-cache-dtype auto` and no speculative config, so it exercised neither fix -- the index cache never reached the fp8 label path and the decode segment was uniform by construction. It now runs fp8 KV plus an EAGLE3 draft (`Inferact/MiniMax-M3-EAGLE3-GQA`, `num_speculative_tokens=3`); nothing else about the cell changed. The threshold stays 0.93, and the native accuracy catalog lists 0.9469 for MXFP4 + EAGLE3 at that same threshold.

## Test Result

Fix 1, on 4x MI355X (gfx950) TP4, `amd/MiniMax-M3-MXFP4`, vLLM 0.27.2.dev0+g6e448d0ea, `--kv-cache-dtype fp8`: before, the first chat request kills EngineCore; after, the index cache is allocated as `(num_blocks, 128, 128) torch.float8_e4m3fn`, chat completions answer normally, and a 90K-token 90%-prefix-hit aiperf run completes clean (32/32 requests, TTFT p50 2.28 s, TPOT p50 14.25 ms at concurrency 8).

Fix 2, on 4x MI355X TP4 with `MiniMax-M3-MXFP8` + `Inferact/MiniMax-M3-EAGLE3-GQA` (`num_speculative_tokens=3`), gsm8k 5-shot chat, full 1319 questions:

| build | flexible-extract | strict-match | outcome |
| --- | --- | --- | --- |
| fp8 KV + EAGLE3, before this PR | -- | -- | `total_q 121 not divisible by max_query_len 4` |
| fp8 KV + EAGLE3, after | **0.9530** ±0.0058 | 0.9538 | no errors, mean acceptance length 3.27 |
| fp8 KV, no spec decode (regression) | 0.9484 ±0.0061 | 0.9492 | unchanged from 0.9477 before this PR |
| auto KV, no spec decode | 0.9447 ±0.0063 | 0.9454 | -- |

The no-spec numbers confirm the uniform fast path still behaves; the spec-decode run keeps its acceptance length, so drafting is doing real work rather than degenerating to plain decode.

`tests/plugin/test_minimax_m3_decode_uniformity.py`: 13 cases, CPU only.

## Submission Checklist

### DIFF
--- a/atom/plugin/vllm/attention/minimax_m3_attnetion.py
+++ b/atom/plugin/vllm/attention/minimax_m3_attnetion.py
@@ -65,6 +65,40 @@ def minimax_m3_sparse_attention(
     return layer._forward_with_output(qkv, positions, output)
 
 
+def _index_cache_torch_dtype(kv_cache_dtype: str, model_config) -> torch.dtype:
+    """Torch dtype for the MiniMax-M3 index cache.
+
+    Everything except fp8 is vLLM's own mapping. fp8 is the exception, and it
+    belongs here rather than in ``kv_cache_dtype_str_to_dtype``: that table is
+    vLLM's, it is shared by every layer of every model, and its ``fp8 ->
+    torch.uint8`` entry is deliberate -- vLLM's kernels take an fp8 KV cache as
+    a byte buffer and reinterpret it, and so do the aiter paged kernels behind
+    ATOM's main sparse and dense caches (see
+    ``_page16_shuffle_cache_for_sparse_kernel``, which does the ``.view()``
+    itself). Relabelling fp8 globally would change those caches too.
+
+    The index cache is read by a kernel that dispatches on the tensor instead:
+    ``_index_block_score_kernel`` branches on ``k.dtype.is_fp8()``, so a
+    uint8-labelled cache goes down the bf16 branch and dots bf16 against uint8
+    (a Triton compile error on the first request). The bytes there are already
+    fp8 -- aiter's ``fused_qknorm_idxrqknorm`` writes them under
+    ``kv_cache_dtype="fp8"`` -- so only the label was wrong. ``dtypes.d_dtypes``
+    is the same handle the native server resolves this cache through
+    (``atom.model_ops.attentions.aiter_attention._resolve_index_cache_dtype``),
+    which keeps the two paths on one arch-correct fp8 dtype instead of a
+    hard-coded ``torch.float8_e4m3fn``.
+
+    The element stays one byte either way, so vLLM's page-size accounting is
+    unchanged, and vLLM already allows a real fp8 dtype in this position -- its
+    own ``fp8_inc`` entry maps to ``torch.float8_e4m3fn``.
+    """
+    from vllm.utils.torch_utils import kv_cache_dtype_str_to_dtype
+
+    if str(kv_cache_dtype).startswith("fp8"):
+        return dtypes.d_dtypes["fp8"]
+    return kv_cache_dtype_str_to_dtype(kv_cache_dtype, model_config)
+
+
 class MiniMaxM3SparseIndexerCache(nn.Module, AttentionLayerBase):
     """Key-only index cache owned by MiniMax-M3 sparse attention."""
 
@@ -76,7 +110,6 @@ class MiniMaxM3SparseIndexerCache(nn.Module, AttentionLayerBase):
         kv_cache_dtype: str,
     ) -> None:
         from vllm.v1.attention.backend import AttentionType
-        from vllm.utils.torch_utils import kv_cache_dtype_str_to_dtype
 
         super().__init__()
         atom_config = get_current_atom_config()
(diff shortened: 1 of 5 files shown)


atom/plugin/vllm/attention/minimax_m3_attnetion.py
```diff
@@ -86,9 +86,20 @@ def __init__(
         self.attn_type = AttentionType.DECODER
         self.attn_backend = SparseMHAIndexerBackend
         self.kv_cache_dtype = kv_cache_dtype
-        self.kv_cache_torch_dtype = kv_cache_dtype_str_to_dtype(
-            kv_cache_dtype, vllm_config.model_config
-        )
+        if str(kv_cache_dtype).startswith("fp8"):
```
ganyi1996ppo: Why not handle all of this in `kv_cache_dtyoe_str_to_dtype`

PerryZhang01: vllm kv_cache_dtype_str_to_dtype fucn will transfer str 'fp8' to torch.uint8

PR #2128, opened 2026-09-03
feat(dp): DP-vocab-sharded greedy argmax for speculative drafting

ZhangLirong-amd:
before in draft model DP:
image

after sharded gemm:
image

## Submission Checklist

### DIFF
--- a/atom/model_ops/embed_head.py
+++ b/atom/model_ops/embed_head.py
@@ -272,11 +272,18 @@ class ParallelLMHead(VocabParallelEmbedding):
 
         For greedy speculative drafting only the argmax is needed, so each rank
         reduces its own vocab shard to ``(max_val, global_idx)`` and we all-gather
-        just those ``[N, 2]`` (tp small) instead of the O(vocab) logits. Token
-        selection is identical to a full-logits ``argmax``: the values compared
-        are the same bf16 logits (fp32-packed exactly), and tie-breaking matches
-        the lowest global index — ``torch.max`` picks the lowest local index, and
-        ``argmax`` over ranks picks the lowest rank (== lowest vocab range).
+        just those ``[N, 2]`` (tp small) instead of the O(vocab) logits.
+
+        On the TP path the GEMM is unchanged, so this is bitwise-identical to a
+        full-logits ``argmax`` -- the values compared are the same bf16 logits
+        (fp32-packed exactly), and tie-breaking matches the lowest global index
+        (``torch.max`` picks the lowest local index, ``argmax`` over ranks the
+        lowest rank == lowest vocab range). The DP path (``ATOM_DP_DRAFT_ARGMAX``,
+        via ``_dp_sharded_logits(mode="argmax")``) reshapes the GEMM ([M, V] ->
+        [dp*M, V/dp]), which ``tgemm`` may tile differently, so its logits are not
+        bitwise-identical to the replicated ones and a near-tie can flip. The pick
+        itself is still exact over whatever logits it is given; only the GEMM's
+        rounding differs.
         """
         if out is not None:
             assert out.shape == x.shape[:-1], (
@@ -287,10 +294,13 @@ class ParallelLMHead(VocabParallelEmbedding):
                 out.dtype == torch.long and out.device == x.device
             ), "argmax out must be an int64 tensor on the input device"
         # Pure-DP draft: shard the vocab across the DP group instead of a
-        # replicated full-vocab GEMM. Env first so a disabled build skips the
-        # context lookup.
-        if envs.ATOM_DP_DRAFT_ARGMAX and self._can_use_dp_sharded_argmax(
-            get_forward_context().context
+        # replicated full-vocab GEMM. Skip plugin mode -- its caller decides the
+        # collective count per chunk (a mismatch would deadlock DP) -- and skip
+        # the context lookup entirely when the env is off.
+        if (
+            envs.ATOM_DP_DRAFT_ARGMAX
+            and not is_plugin_mode()
+            and self._can_use_dp_sharded_argmax(get_forward_context().context)
         ):
             return self._dp_sharded_logits(x, "argmax", out)
         logits = tgemm.mm(x, self.weight, self.bias)  # [N, vocab/tp]
@@ -444,19 +454,30 @@ class ParallelLMHead(VocabParallelEmbedding):
     def _can_use_dp_sharded_argmax(self, context) -> bool:
         """Whether a draft step may run the DP-sharded argmax.
 
-        Every DP rank must reach the same verdict, so every gated value is
-        DP-agreed. `running_tokens_are_unified` marks a rectangular draft (a
-        ragged one falls back); `running_tokens` is bounded because the hidden
-        gather outgrows the weight-read saving past ATOM_DP_DRAFT_ARGMAX_MAX_ROWS.
+        Total predicate -- returns False, never raises, outside a DP-reduced
+        rectangular pure-DP draft. Every DP rank must reach the same verdict, so
+        every gated value is DP-agreed: `running_tokens_are_unified` marks a
+        rectangular draft, `running_tokens` is bounded past
+        ATOM_DP_DRAFT_ARGMAX_MAX_ROWS (the gather outgrows the weight-read win),
+        and `dp_metadata is not None` is the proof `running_tokens` was actually
+        reduced across DP -- absent under SGLang dp-attention (where aiter's DP
+        group is >1 but ATOM's data_parallel_size is 1) and single-GPU. That
+        check also gates the raising `get_dp_group()`: dp_metadata is built via
+        the DP group, so its presence means the group exists.
         """
-        if not envs.ATOM_DP_DRAFT_ARGMAX or self.tp_size != 1:
+        if not envs.ATOM_DP_DRAFT_ARGMAX or is_plugin_mode() or self.tp_size != 1:
+            return False
+        fc = get_forward_context()
+        if (
+            context is None
+            or fc.dp_metadata is None
+            or not getattr(context, "is_draft", False)
+            or getattr(context, "is_prefill", False)
+            or not getattr(context, "running_tokens_are_unified", False)
+            or int(context.running_tokens) > envs.ATOM_DP_DRAFT_ARGMAX_MAX_ROWS
+        ):
             return False
         dp_group = get_dp_group()
-        if dp_group.world_size <= 1 or self.num_embeddings % dp_group.world_size != 0:
-            return False
         return (
-            context is not None
-            and getattr(context, "is_draft", False)
-            and getattr(context, "running_tokens_are_unified", False)
-            and int(context.running_tokens) <= envs.ATOM_DP_DRAFT_ARGMAX_MAX_ROWS
+            dp_group.world_size > 1 and self.num_embeddings % dp_group.world_size == 0
         )
--- a/atom/model_ops/embed_head.py
+++ b/atom/model_ops/embed_head.py
@@ -250,7 +250,7 @@ class ParallelLMHead(VocabParallelEmbedding):
                 last_indices = attn_metadata.cu_seqlens_q[1:] - 1
                 x = x[last_indices].contiguous()
             if self._can_use_dp_sharded_head(context):
-                return self._dp_sharded_logits(x)
+                return self._dp_sharded_logits(x, envs.ATOM_DP_LM_HEAD_MODE)
         logits = tgemm.mm(x, self.weight, self.bias)
         if self.tp_size > 1:
             use_custom = envs.ATOM_USE_CUSTOM_ALL_GATHER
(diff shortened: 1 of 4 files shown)


valarLip:
Reviewed at `41a3fa133`. Small diff, so I read all of it rather than sampling.

The idea is sound — a draft's replicated `[N, V]` GEMM is weight-read bound at tiny M, so sharding the vocab and exchanging only the packed `[N, 2]` is the right trade. What follows is about the gate, not the trade.

---

# 0. The new gate is the sibling gate with each of its three safety properties removed

`_can_use_dp_sharded_argmax` is a near-copy of `_can_use_dp_sharded_head`. Side by side:

| | `_can_use_dp_sharded_head` | `_can_use_dp_sharded_argmax` |
|---|---|---|
| call site wrapped in `is_plugin_mode()` | **yes** — `forward():243` guards the whole block | **no** — `compute_argmax_token:291` calls it bare |
| `is_prefill` exclusion | **yes** (`:350`) | **no** |
| `dp_metadata is not None` as the final check | **yes** (`:355`) | **no** |

Plus: the feature defaults to **on** (`ATOM_DP_DRAFT_ARGMAX` → `"1"`), and the only test change in the PR turns it **off**.

Those three are not independent findings. They are three faces of one thing: **the new path is reachable from a context the sibling deliberately excludes, where none of its DP-agreement premises hold.**

## Why it matters, concretely

**Plugin mode.** `atom/plugin/sglang/dflash_lm_head_bridge.py:153-159` calls `compute_argmax_token` once per 256-row chunk of `hidden_states.shape[0]` — a **per-rank** row count. With the flag on and `tp_size == 1` under `--enable-dp-attention`, rank A issues 2 pairs of `dp_group.all_gather` and rank B issues 3. That is an unrecoverable DP deadlock, and no amount of care inside `_dp_sharded_logits` can prevent it, because the collective count is decided by the caller.

Even short of that: `local_rows` there is a ≤256 chunk while `max_rows` is ATOM's `context.running_tokens` left by an unrelated forward, so either the tripwire at `:382` fires mid-serve or the pad/slice at `:386`/`:413` returns the wrong rows.

`forward()` already solves this by refusing to reason about a plugin's forward context at all. The argmax entry point needs the same refusal.

**The dropped `dp_metadata` check.** The sibling ends on that line, and it is the only proof that `running_tokens` was actually DP-reduced. `Drafter._publish_draft_shape` (`drafter.py:590-593`) returns early when `parallel_config.data_parallel_size <= 1` and never sets `running_tokens_are_unified` or `dp_metadata`.

Under SGLang's `--enable-dp-attention`, TP is folded into the DP size, so **aiter's `dp_group.world_size > 1` while ATOM's `data_parallel_size` is 1**. `dp_metadata` is None, `running_tokens` is a per-rank number that was never reduced, each rank computes a different `max_rows` at `:379`, and the fixed-size `all_gather` at `:391`/`:403` is posted with mismatched sizes. The dropped line was the one gate that caught this.

**The dropped `is_prefill` exclusion.** The sibling's own comment gives the reason, and it is worth deciding explicitly rather than by omission: `forward()` slices `x = x[last_indices]` on prefill (`:249-250`), `compute_argmax_token` does not. So the sibling's exact rationale does not transfer — but the conclusion might still hold for a different reason, and right now nothing records which. In plugin mode specifically, `atom/plugin/sglang/runtime/forward_context.py:167` returns `running_tokens_are_unified = True` unconditionally when TBO and dp-attention are off, including extend batches, and `_running_tokens` then returns `scheduled_tokens` — so `local_rows == max_rows`, the assert passes, and every rank all-gathers a differently-sized hidden tensor with no error.

# 1. `get_dp_group()` raises where the predicate should return False

`aiter/dist/parallel_state.py:1679`:

```python
def get_dp_group() -> GroupCoordinator:
    assert _DP is not None, "data parallel group is not initialized"
```

`_can_use_dp_sharded_argmax` calls it (`:453`) after only two cheap checks. So any `tp_size == 1` process that reaches `compute_argmax_token` without aiter's `_DP` built crashes with `AssertionError` on its first draft step, instead of taking the replicated path a predicate named `_can_use_...` implies it would.

The PR supplies its own evidence for this. `tests/test_lm_head_argmax_out.py:17-19`:

```python
# Exercise the replicated argmax path; the DP-sharded draft path needs an
# initialized DP group, which this unit test does not stand up.
monkeypatch.setattr(embed_head.envs, "ATOM_DP_DRAFT_ARGMAX", False)
```

The env defaults on, so that is a workaround for the raise, not a test-environment detail. A predicate should be total: guard with a non-raising accessor, or reuse `is_plugin_mode()`/a `_DP is not None` probe so it can only ever return False.

# 2. On by default, and the only test disables it

`ATOM_DP_DRAFT_ARGMAX` defaults to `"1"`. The single test change sets it to `False`. Net coverage for the new path: zero.

Two parts of it need neither a GPU nor a real DP group:

- **The verdict table.** `_can_use_dp_sharded_argmax` is pure predicate logic over `(tp_size, world_size, num_embeddings % dp_size, is_draft, running_tokens_are_unified, running_tokens vs MAX_ROWS)` — a fake context plus a stubbed `get_dp_group` covers every row, including the "no DP group" case that currently raises.
- **The offset arithmetic.** `start = dp_rank * max_rows` and the rank-major `.view(dp_size, dp_size*max_rows, 2)` unpack can be checked against a hand-built tensor with a fake `all_gather` that concatenates on dim 0. That is exactly the wrong-offset class this kind of change gets wrong, and nothing would currently catch it.

This file already stubs `tgemm.mm` the same way, so the scaffolding exists.

# 3. Token-identity is claimed in five places and is no longer true on this path

`compute_argmax_token`'s docstring says token selection is identical to a full-logits argmax because "the values compared are the same bf16 logits (fp32-packed exactly)". That holds for the TP path, where the GEMM is unchanged and only the reduction is sharded.

The DP path changes **both** GEMM dimensions: `M` from `local_rows` to `dp_size * max_rows`, `N` from `V` to `V/dp`. `tgemm` selects tiling and split-K by shape, so the logits are not bitwise identical to the replicated path, and a near-tied top-2 rounds differently.

That matters more here than it would elsewhere. This repo has a recorded precedent where a ~1ulp difference in an lm_head reduce produced *stable* token flips once amplified by MTP spec-acceptance — and where synthetic unit tests showed `cos=1.0` while end-to-end acceptance moved by points. Which is to say: a numerical-equivalence unit test would not settle this, and the PR ships no acceptance-rate or GSM8K numbers.

Claims that need either retraction or evidence: `embed_head.py`'s own docstring, `deepseek_mtp.py:278`, `eagle3_llama.py:488`, `eagle_proposer.py:681-683`, `plugin/vllm/model_wrapper.py:1120`.

Given the flag is default-on, one acceptance-rate A/B on a DP+MTP config is the thing that would make this mergeable as-is.

---

## Docs

`grep -rn ATOM_DP_DRAFT_ARGMAX docs/` returns nothing, and neither does `ATOM_DP_LM_HEAD_MODE` — so the one number an operator has to retune (`~256 is the V4-Pro crossover`) is discoverable only by reading `envs.py`. Per Fix-then-sweep, the pre-existing gap for the neighbouring var belongs in the same pass.

**`envs.py:357-360` is now false.** It reads:

> every other case — TP>1, DP off, prefill, **ragged/draft rows** — auto-falls back to the replicated path via `_can_use_dp_sharded_head`

Draft rows are now precisely the case that does *not* fall back. Someone debugging a DP draft hang will grep this comment, conclude the draft path is inert, and look elsewhere.

**`ATOM_DP_LM_HEAD_MODE="default"` no longer opts out.** It is documented as the "explicit opt-out" from the DP-sharded head, but the draft argmax runs the same shard-and-gather machinery under a separate, independently default-on flag. An operator who hit a problem and turned the knob off still gets DP-sharded lm_head collectives on every draft step. Either fold the draft path under the same mode string, or make `"default"` disable both.

Also stale in `embed_head.py`: `:313` "(config ② all-gather / ③ all-to-all)" no longer enumerates `mode`'s domain; `:316` "Precondition (checked by `_can_use_dp_sharded_head`)" — there are two gates now; `:329` "The strategy ... is read from `ATOM_DP_LM_HEAD_MODE` there" — it is now passed in as `mode` from `:253`.

## Smaller, all read in the diff

- **The all-gather is `dp_size`× oversized** (`:403`). Every rank receives every shard's values for *every* rank's rows, and discards all but `[start, start+local_rows)` at `:411`. At dp=8, max_rows=256 that is 128 KiB on the wire where 16 KiB suffices, and the `argmax(dim=0)` + `gather` + `.to(long)` run over 2048 rows to keep 256. `all_to_all_single` — used by the sibling branch fifteen lines below — is the primitive for exactly this. In a hot draft loop whose whole justification is shaving a weight read, part of the saving is spent back on the collective it introduces, and it makes `MAX_ROWS`'s crossover `dp_size`-dependent for no reason. The docstring's "exchange only `[Σrows, 2]`" understates the traffic by `dp_size`.
- **The argmax branch is a verbatim second copy of the TP branch** (`:305-311` vs `:400-414`): `lm_head_argmax_pack` → `all_gather(dim=0)` → `.view(G, -1, 2)` → `[:,:,0].argmax(0)` → `[:,:,1].gather(...)` → `.to(long)` → `out.copy_`. Only the group, the pack offset and a trailing slice differ. Two copies of a tie-breaking-sensitive reduction whose correctness argument is written down once — and a future fix to one (NaN handling, or int32 packing if a vocab ever passes 2^24) leaves the other wrong, silently, on the path whose only symptom is an acceptance drop. `_argmax_over_shards(packed, group, use_custom)` called from both.
- **`out` is silently dropped by the two non-argmax modes.** The all2all branch returns a fresh reshape (`:430-434`) and allgather a fresh slice (`:440`); neither honours it. Benign today since `forward()` passes `out=None`, but the signature advertises `out` for all modes and only the docstring restricts it — and a caller using `out` as a CUDAGraph-persistent buffer (the `deepseek_mtp.py:282` pattern) would replay stale contents with no error. `assert out is None or mode == "argmax"`, or move `out` into an argmax-only helper.
- **Name-matches-function: `_dp_sharded_logits` no longer returns logits.** With `mode="argmax"` it returns `[local_rows]` int64 ids; otherwise `[local_rows, vocab]` floats. One `str` parameter, two disjoint return types. A reader seeing `return self._dp_sharded_logits(x, "argmax", out)` reasonably appends `.argmax(-1)`. A shared `_dp_gather_and_project(x)` prefix plus typed tails removes the mode string entirely.
- **`mode` is an unvalidated magic string** with a silent fall-through: a typo lands in the allgather branch at `:436` and returns `[N, V]` floats where the caller expected `[N]` ids.
- **The argmax branch returns a non-contiguous view** of the full `[dp_size*max_rows]` buffer, while the sibling allgather branch calls `.contiguous()` on its slice for the same reason (`:440`). `out=None` is the common path — `deepseek_v4_mtp.py:208`, `eagle3_llama.py:493`, `eagle3_deepseek_mla.py:392`, the `qwen3_*_mtp` and `glm4_moe_mtp` / `mimo_v2_mtp` heads all call without `out=` — so the returned `[local_rows]` pins the whole buffer alive for the lifetime of the draft ids.
- **`ATOM_DP_DRAFT_ARGMAX` is read twice per call** (`:291` and `:450`), and `envs.py` re-evaluates the lambda on every attribute access — no caching. Per draft step, K times per target step, on the host path this repo has been actively trimming. The second read is redundant with the call-site check.

## One thing I checked and it is fine

The call-site comment "Env first so a disabled build skips the context lookup" is **correct** — Python's `and` short-circuits, so a disabled build never calls `get_forward_context()`. The real (much smaller) cost is that an *enabled* build at `tp_size > 1` pays the context lookup before the predicate can reject on `tp_size`; moving `tp_size` into the call-site condition would fix that, but it is a micro-optimisation, not a defect.

ZhangLirong-amd:
Thanks for the careful read — right on every point. Addressed in `91ace1e`:

**#0 / #1 — the gate was reachable where its DP-agreement premises don't hold.** `_can_use_dp_sharded_argmax` is now a total predicate mirroring the sibling's safety:

- **plugin mode** — both the `compute_argmax_token` dispatch and the gate bail on `is_plugin_mode()`, so a per-chunk caller can't drive a mismatched collective count.
- **`dp_metadata is not None`** — added as the DP-reduction proof; it is None under SGLang dp-attention (aiter DP group > 1 while ATOM `data_parallel_size` is 1) and single-GPU, i.e. exactly the mismatched-`all_gather` deadlock you described.
- **the raising `get_dp_group()`** — now runs last, behind the `dp_metadata` check (dp_metadata is built via the group, so its presence proves the group exists). The predicate returns False, never raises; the unit-test monkeypatch that worked around the raise is dropped.
- **`is_prefill`** — excluded, matching the sibling.

**#2 — coverage.** `tests/test_dp_sharded_argmax.py` (CPU, no DP group): the full gate verdict table (including the no-DP-group row that used to raise) and the rank-major offset arithmetic of the argmax exchange, against a fake dim-0 all-gather.

**#3 — identity + acceptance.** You're right that it isn't bitwise-identical on the DP path (the GEMM reshapes `[M, V] -> [dp*M, V/dp]`). Softened the docstring: the TP path stays bitwise-identical; the DP path is an exact pick over logits that differ only in GEMM rounding. On the acceptance question you flagged — DeepSeek-V4-Pro, `-tp 8 --enable-dp-attention` (pure DP), `--method mtp --num-speculative-tokens 3`:

| `ATOM_DP_DRAFT_ARGMAX` | accept_len | accept_rate |
|---|---|---|
| 0 (off) | 2.73 | 57.60% |
| 1 (on) | 2.72 | 57.43% |

That gap is within run-to-run variance (off itself spans ~2.73–2.76 across runs), so I've kept it default-on. Happy to flip to opt-in if you'd rather the conservative default.

The four external "token-identical" comments (`deepseek_mtp`, `eagle3_llama`, `eagle_proposer`, `model_wrapper`) describe the general greedy-argmax contract, which stays behaviorally intact here — I left those and clarified the one in `embed_head`.

valarLip (approved):

PR #2129, opened 2026-09-03
[ATOM Plugin CI] Fix the K3 DSpark DCP8 MLA head-width, capture-layout and cudagraph-break defects

Phi-C:
Three defects on the `Accuracy (Kimi-K3 DSpark DCP8 TP8)` cell, which has never been green — it was added in #2033 and that PR's own CI shows it failing. Net change: three plugin files plus one threshold line. The cell keeps `cudagraph_mode: FULL_AND_PIECEWISE`.

Everything below is measured on 8xMI355X with the cell's own arguments, the same aiter main nightly CI installs, and the K3 / K3-DSpark checkpoints.

## 1. Plan MLA work descriptors for the layers the builder serves

`AiterMlaMetadataBuilderForVllm` took its query-head count from `config.model_config.hf_config`, which always describes the **target**. A DSpark draft owns its own KV cache group and therefore its own builder, so the draft's persistent work descriptors were planned for Kimi-K3's **96** query heads while the draft's MLA decode dispatches at **64**. Under DCP those gathered widths differ, and a plan describing a different kernel than the one that runs leaves that kernel indexing partials nobody planned.

Fixed by reading the width off the layers the builder was handed — what ATOM native's `DSparkProposer` deliberately does ("read it rather than recomputing, so the work descriptors planned here describe the kernel that will actually run") — and passing `min_query_heads` along with it. Verified: draft plans and dispatches at 64, target unchanged at 96.

## 2. Capture: install the self-consistent KV layout ATOM native installs

vLLM hands `build_for_cudagraph_capture` batches whose sequences are 0 or 1 token long. The plugin's implementation was `return self.build(0, common_attn_metadata)` — a passthrough. ATOM native's is not: it installs a minimal self-consistent layout and names both failures it prevents.

> *"g_kv_indptr is the only thing telling the cprr kernel how long each sequence is GLOBALLY … capturing with it left at its allocation value **walks the kernel off the KV list (illegal access)**"*
> *"page_count == 0 but kv_last_page_lens == 1 … underflows (0 - 1 -> 0xFFFFFFFF) … the kernel never exits and **cudagraph capture hangs**"*

Both were reproducible here. The plugin now applies native's invariant to the degenerate capture batches — one local token per rank, i.e. a global length of `dcp_world_size` — and leaves the uniform-decode capture shapes, which already carry a full-length context, alone.

**Effect**: capture used to die at `Capturing dspark CUDA graphs (FULL): 0%`. It now completes PIECEWISE 100%, FULL 100%, dspark 100%, `Graph capturing finished`.

## 3. Break MLA out of the breakable cudagraph

With capture fixed, the server still died in vLLM's post-capture kernel warmup with an illegal access. Root cause: K3 runs on vLLM's **breakable cudagraph** path (`vllm/config/vllm.py` turns on `VLLM_USE_BREAKABLE_CUDAGRAPH` for `KimiK3ForConditionalGeneration`, compilation mode goes to `NONE`, so there is **no FX splitting**). One stream capture swallows the whole forward, and **only operators carrying `@eager_break_during_capture` break out of it and re-execute eagerly on every replay**. The break points are declared per operator; miss one and that layer is sealed into the graph with its `block_table`, `seq_lens` and DCP work descriptors frozen at capture time. Any later batch whose shapes differ then indexes through stale metadata.

`atom_vllm_mla_attention` was missing it, and also allocated `output` inside the op body — the decorator requires writing into the **caller's** buffer, otherwise the address changes on every replay and the captured segment downstream still reads the capture-time one. The KDA mixer in `models/kimi_k3.py` already had the decorator, with a docstring spelling out why; this is the other half of the same omission.

The op now takes `output` from the caller (`mutates_args=["output"]`, returns `None`) and delegates to a helper carrying the decorator, which re-fetches its metadata through `_get_layer_context(layer_name)` inside the break — passing metadata in as an argument would freeze it at capture time and defeat the fix. The `FULL` path is unaffected: the decorator is a passthrough when `cudagraph_runtime_mode == FULL`.

**Falsifiable check, predicted before it was run**: `eager_breaks` must equal the number of decorated layers. K3 has 93 (69 KDA + 24 MLA).

| | `BreakableCUDAGraphCapture` | server | capture pool |
|---|---|---|---|
| before | `graphs=70, eager_breaks=69` (= KDA only) | `died` at 280s, illegal access | 21.27 GiB |
| after | `graphs=94, eager_breaks=93` (= one per layer) | `ready` at 301s | 11.00 GiB |

## 4. Threshold 0.95 -> 0.94, on measurements

Full 5-shot gsm8k, 1319 questions, concurrency 64, same machine and harness throughout:

| configuration | run | flexible-extract | strict-match |
|---|---|---|---|
| **this cell, this PR** — DCP8, `FULL_AND_PIECEWISE` | 1 | **0.9530** | 0.9530 |
| **this cell, this PR** — same config again | 2 | **0.9484** | 0.9484 |
| this cell, `cudagraph_mode: NONE` (no graphs taken) | 1 | 0.9492 | 0.9492 |
| this cell, `NONE` again | 2 | 0.9500 | 0.9507 |
| **ATOM native**, TP8 + DCP8 + 7 spec tokens | 1 | 0.9500 | 0.9477 |
| non-DCP `Kimi-K3 DSpark TP8` — CI grades this one as **passing** at 0.95 | 1 | 0.9469 | 0.9462 |

Two things follow.

**The plugin's DCP8 path is not losing accuracy, with or without graphs.** It matches ATOM native question for question at this resolution; every row above sits within ~5 questions out of 1319 of every other, inside the +/-0.006 standard error.

**0.95 has no margin for anyone.** Native itself measures exactly 0.9500, this cell straddles the line across repeats, and the non-DCP cell that CI already passes measures 0.9469 here. A threshold the reference implementation only just reaches is a coin flip per run, not a regression detector; it was picked for a cell that had never produced a number. 0.94 sits about 1.4 standard errors below the lowest of the six runs above — enough to catch a real regression, not so tight that a green run depends on which side of the noise it lands.

## History

The first two commits' premise — an OOM answerable by cudagraph capture sizes and `--gpu-memory-utilization` — is disproven; their workflow edit is reverted, so the cell's arguments are unchanged from `main` apart from the threshold. The post-capture fault that #2033 left open is fixed here in section 3, not deferred.


Phi-C:
Sharper localisation of the fault that survives capture, for whoever picks it up.

With the capture fix in this PR, all three capture phases complete. What dies next is vLLM's **post-capture kernel warmup**, and serialised kernels (`AMD_SERIALIZE_KERNEL=3`) place it exactly:

```
warmup.py:369 warmup_kernels -> warmup.py:338 _run_decode_step
  -> gpu_worker.py:1079 execute_model -> model_runner.py:1418 execute_model
  -> cudagraph_utils.py:419 run_pw_graph
  -> breakable_cudagraph.py:214 replay        <- faults here
```

`warmup_kernels` runs five decode shapes. Instrumenting it to announce each one shows the failure is **step 0** — the plainest of them: all 15 requests with draft tokens, 15 x 8 = 120 query rows, which vLLM pads to the 128 capture bucket. Not the mixed spec / non-spec shapes that follow, and not a graph-free path: it is the piecewise graph being replayed on that padded batch. `cudagraph_mode: NONE` runs the same warmup step eagerly and passes.

Ruled out along the way, each by measurement rather than reading: KV indices out of range (checked `idx_max` against the pool at every decode), the planner overflowing the buffers `get_mla_metadata_info_v1` sized (canary tail on each), descriptor buffers left as `torch.empty` garbage (poisoned them deliberately), tensor contiguity, `page_size`, `max_split_per_batch` (16 vs native's AUTO), clamping empty and padded rows on the local side, the global side, and both together, routing degenerate batches off `cprr` (that trades the illegal access for the non-terminating kernel `mla_min_query_heads` warns about), padding rows to a split granularity, aiter's custom all-gather, and the DSpark DCP slot rewrite (its slots stay inside the pool: max 576001 against 585216).

A single-GPU replay of the failing decode call — same kernel, same shapes, every degenerate local/global and per-rank combination, with dirty buffers — passes every time. The trigger needs the real 8-rank setup plus a graph having been captured.

Repro is in place if useful: a seconds-per-iteration single-GPU harness for the decode call, and a full server repro at ~5 min per iteration once the weights are in page cache.

PerryZhang01:
Closing and reopening this PR: it was created with the wrong GitHub account that happened to be the active `gh` login on my machine. The branch, the commits and their author (perzhang@amd.com) are unchanged; reopening it under my own account so the PR author matches. Link to the replacement follows in the next comment.

PerryZhang01:
Replacement: #2160 (same branch `perzhang/k3-dspark-cudagraph-mem`, same head commit `af27f4a9`). Please review there.

PR #2138, opened 2026-09-03
build(rust): commit Cargo.lock for atom/mesh — unblocks all image builds broken by tinyvec 1.13.0

sunway513:
## Summary

Commits `Cargo.lock` for the two Rust crates under `atom/mesh` and stops `.gitignore` from excluding it. Without a lockfile every build re-resolves Rust dependencies from crates.io, and a release published mid-afternoon today broke all ATOM image builds.

## What broke

`tinyvec 1.13.0` fails to compile in this tree:

```
error: cannot find macro `vec` in this scope
   --> /root/.cargo/registry/src/.../tinyvec-1.13.0/src/tinyvec.rs:710:21
710 |       TinyVec::Heap(vec![A::Item::default(); len])
    |                     ^^^
note: `vec` is imported here, but it is a module, not a macro
   --> .../tinyvec-1.13.0/src/tinyvec.rs:3:18
  3 | use alloc::vec::{self, Vec};
error: could not compile `tinyvec` (lib) due to 1 previous error
ERROR: Failed building editable for atom
```

It reaches us through `atom/mesh`'s dependency graph, not directly.

## This is not an ATOM regression

The timeline rules that out:

- `tinyvec 1.13.0` published to crates.io at **2026-09-03T21:13:39Z**
- ATOM image build reached the Rust compile and failed at **2026-09-03T21:30:59Z**, 17 minutes later
- The two Nightly Docker Release runs earlier the same day, at **13:49** and **12:20**, both succeeded

Nothing in ATOM changed between those runs. The dependency resolution did, because there is nothing pinning it.

## Why the lockfile was missing

`atom/mesh/.gitignore` line 4 is `Cargo.lock`. That is the right default for a pure library, and `atom/mesh` does expose a `[lib]`. It also declares `[[bin]] atomesh`, and `setup.py` runs `cargo build --release` to produce that binary during the ATOM install. For a package that ships a binary, the Rust guidance is to commit the lockfile. The ignore rule looks like it came from a library template and was never revisited.

There is already a workflow that assumes the file exists. `atomesh-mocker-benchmark.yaml` builds its cache key with:

```
hashFiles('atom/mesh/Cargo.lock', 'atom/mesh/Cargo.toml', 'atom/mesh/mocker/Cargo.toml')
```

With the file absent, `hashFiles` contributes nothing for it and the cache key silently degrades. Committing the lockfile fixes that too.

## Changes

- `atom/mesh/.gitignore`: drop the `Cargo.lock` line, keep the `target/` rules.
- `atom/mesh/Cargo.lock`: new, generated with cargo 1.94.0, with `tinyvec` held at `1.12.0`.
- `atom/mesh/mocker/Cargo.lock`: new. `atomesh-mocker-benchmark.yaml` builds this crate with `--manifest-path atom/mesh/mocker/Cargo.toml`, so it carries the same exposure.

## Verification

`cargo build --release` in `atom/mesh` completes against the committed lockfile, on `rocm/atom-dev:latest` with cargo 1.94.0 / rustc 1.94.0.

Note for anyone reproducing: the build needs a raised file descriptor limit. At the default `nofile=1024` it fails partway through with `Too many open files (os error 24)` on `aws-lc-rs`, `base64` and `zeroize`, which is unrelated to this change.

## Trade-off worth stating

A committed lockfile means dependency updates become deliberate rather than automatic. Someone has to run `cargo update` periodically, and a `cargo update` PR is the right place to notice a breaking upstream release, rather than finding it in a failed release build. That is the intended effect.

`tinyvec 1.12.0` is a pin of convenience, not a judgement that 1.13.0 is permanently wrong. When upstream resolves it, `cargo update -p tinyvec` lifts it.


valarLip:
Reviewed at `448ee0a8f`. The commit message is unusually good — it carries its own evidence (the 13:49 success / 21:13 publish / 21:30 failure sequence), and it states its own limits ("a pin of convenience, not a claim that 1.13.0 is permanently wrong").

**To head off a misreading up front:** upstream has since resolved the tinyvec breakage, and that removes the *urgency* but not the *reason* for this PR. Without a lockfile every build re-resolves 531 crates from crates.io, and the timeline in the commit message is what that costs — an image build broken by a third-party publish with no ATOM change. The next crate to ship a bad version reproduces it exactly, and there is no guarantee the next one is fixed within five hours. **Land the lockfile; just refresh the pin first.**

---

# 1. Upstream resolved the tinyvec breakage — repin rather than ship a backward pin

Both lockfiles hold `tinyvec 1.12.0`. Checked externally:

```
crates.io:  1.13.2   2026-09-04T02:29:11Z   (not yanked)
            1.13.1   2026-09-04T01:26:57Z
            1.13.0   2026-09-03T21:13:39Z   <- the one in the commit message
            1.12.0   2026-07-10T20:03:55Z   <- what this PR pins
```

And the fix is in it — `Lokathor/tinyvec` commit `77dae049`, 2026-09-04T02:26:20Z, three minutes before 1.13.2 was published:

> On some rustc versions this only resolves the module (type namespace), so unqualified `vec![...]` calls in this file fail with 'cannot find macro `vec` in this scope'.

It imports `Drain` directly instead of the `vec` module and fully qualifies both `vec!` invocations. That is exactly the error in the commit message.

The commit message already prescribes the action — "`cargo update -p tinyvec` lifts it once upstream resolves" — and upstream resolved a few hours before this review. Doing it now matters more than it looks: a lockfile cannot carry a comment, `tinyvec` is not a direct dependency (it arrives via `url` → `idna_adapter`/`unicode-normalization`), and nothing in the tree would ever tell a future reader why a transitive crate sits one minor behind. Run it against both lockfiles.

# 2. The new cache key omits the new lockfile — the same defect this PR diagnoses, one file over

`.github/workflows/atomesh-mocker-benchmark.yaml`, lines 71 and 152, identical:

```yaml
key: atomesh-mocker-cargo-${{ runner.os }}-${{ hashFiles('atom/mesh/Cargo.lock', 'atom/mesh/Cargo.toml', 'atom/mesh/mocker/Cargo.toml') }}
```

Both cargo invocations in that workflow resolve through `atom/mesh/mocker/Cargo.lock` — `:79` (`cargo test --manifest-path atom/mesh/mocker/Cargo.toml`) and `:162` (`cargo build --manifest-path .../mocker/Cargo.toml`) — and that is the one input the key does not hash.

The commit message states the problem for the other file in its own words:

> Its cache key is `hashFiles('atom/mesh/Cargo.lock', ...)`, which with the file absent contributes nothing and silently degrades the key.

A later `cargo update` inside `atom/mesh/mocker/` changes the graph, leaves the key byte-identical, and restores a stale `target`. Two lines, and it belongs in this PR since this PR is what creates the file.

# 3. Nothing passes `--locked`, so the committed lockfile is advisory

This is the mechanism that lets the whole class recur, and it is the one change that would have turned the original incident into a one-line CI failure instead of a broken image.

Every build and test path omits it:

- `setup.py:33` — `subprocess.run(["cargo", "build", "--release"], cwd=mesh_dir, ...)`, the path `docker/Dockerfile:474`, `atom_release.dockerfile` and `atomesh-accuracy-validation.yaml:318` all reach
- `atomesh-mocker-benchmark.yaml:78`, `:161`, `:165`
- `atom/mesh/tests/run_atomesh_harness_test.sh:57-61`

So a dev who bumps `axum = "0.8.6"` to `"0.9"` in `Cargo.toml` without regenerating the lock gets a silent re-resolve of the affected subgraph and an in-place rewrite of the now-tracked lockfile. No job fails, CI is green, and the file in git no longer describes what shipped.

With `--locked` that same edit fails loudly at the first build with "the lock file needs to be updated". Four sites.

---

## Structural: two lockfiles for one dependency graph

`atom/mesh/mocker/Cargo.toml:15` declares `mesh = { package = "atomesh", path = ".." }`, but `atom/mesh/Cargo.toml` has no `[workspace]` table — its sections are `package / features / lints.rust / lib / bin / dependencies / build-dependencies / dev-dependencies / profile.release / profile.ci`. So cargo treats `mocker` as its own workspace root, and this PR commits 10,792 lines resolving the same graph twice with nothing keeping them in sync.

They agree today by accident of having been generated in one sitting: 467 shared crates, zero version disagreements, and every difference (`axum 0.7.9`, `tower 0.4.13`, `prost 0.13.5`, `indexmap 1.9.3`, `hashbrown 0.12.3`, `async-stream`, `serial_test`) traces to mesh's `tonic-v12` dev-dependency. Nothing enforces it — including the commit message's own prescription, which updates one lock and leaves the other.

Compounding it: `mocker/Cargo.toml` has no `[profile]` section, so mesh's `opt-level = "z"` / `lto = "fat"` (`Cargo.toml:94-96`) is ignored when `atomesh` is built through mocker. **The library the mocker benchmark measures is optimized differently from the one that ships.** That is worth knowing independently of the lockfile question.

`[workspace] members = ["mocker"]` collapses this to one lockfile and would let a follow-up delete one of the two files added here.

## Nothing will ever refresh the pins

Verified absent repo-wide: `.github/dependabot.yml`, `.github/dependabot.yaml`, `renovate.json`, `deny.toml`, and any `cargo audit` / `cargo deny` step.

Before this PR every build re-resolved and picked up patch-level security fixes for free. After it the set is frozen indefinitely, and the plan is "someone runs `cargo update` periodically". The pins are clean today — `ring 0.17.14`, `rustls 0.23.43`, `rustls-webpki 0.103.15`, `aws-lc-rs 1.18.1`, `openssl 0.10.81`, `tokio 1.53.1`, `h2 0.4.19`, `hyper 1.11.1` — no known advisory and nothing a full major behind. But the next ring or rustls CVE is not picked up until a human remembers.

A `dependabot.yml` with two `cargo` entries (`atom/mesh`, `atom/mesh/mocker`) is about twelve lines and turns the freeze from a liability into a reviewable stream. It does not have to be this PR, but it should be the next one.

## Smaller, all verified

- **`build.rs:15` declares only `rerun-if-changed=Cargo.toml`.** Declaring any `rerun-if-changed` opts out of cargo's default "rerun on any change in the package directory", so after the `cargo update -p tinyvec` this PR prescribes, dependencies recompile but `build.rs` does not re-run — `ATOM_MESH_VERSION` / `BUILD_TIME` / `GIT_COMMIT` / `GIT_STATUS` / `CARGO_VERSION` (consumed at `src/version.rs:11-18`, surfaced by `--version-verbose` and by `src/python.rs:254`) stay stamped from the previous resolution. Now that `Cargo.lock` is tracked and deliberately edited, add `cargo:rerun-if-changed=Cargo.lock`.
- **Neither manifest declares `rust-version`,** and there is no `rust-toolchain.toml` at the repo root or in `atom/mesh/`. Both lockfiles declare `version = 4`, which stabilized in cargo 1.78. A contributor on an older toolchain gets "lock file version 4 requires -Znext-lockfile-bump" rather than an actionable "package requires rustc 1.78". Related: `atomesh-accuracy-validation.yaml:297` only installs the pinned 1.94.0 when cargo is **absent** (`if ! command -v cargo`), so a `atom_base_image` that already ships an older cargo previously worked and now hard-fails at `:318`. Freshly built images are safe (`docker/Dockerfile:435-443` installs unconditionally); only the reuse-the-image path is exposed.
- **`MANIFEST.in:2` already said `include atom/mesh/Cargo.lock`.** While the file was gitignored that line was a dangling no-op emitting "warning: no files found matching" on every sdist from a clean checkout — which is the strongest independent evidence the ignore rule was wrong, and worth putting in the PR body. The reverse is now true for the new file: `setup.py:54` uses `use_scm_version=True` and `pyproject.toml:37` sets `include-package-data = true`, so `atom/mesh/mocker/Cargo.lock` ships (+131 KB) without being declared. Declare it or prune it.
- **The mesh lockfile freezes `native-tls` and a system OpenSSL alongside rustls,** contradicting the manifest's deliberate `default-features = false` opt-out: `native-tls 0.2.18`, `openssl 0.10.81`, `openssl-sys 0.9.117`, `openssl-probe`, `tokio-native-tls` — traced to `hf-hub 0.4.3` and `ureq 2.12.1` (via `llm-tokenizer`), which pull native-tls with default features. The lockfile does not cause this, but it now freezes it: an `openssl-sys` CVE needs a manual `cargo update` on a C dependency the team may not know it links. Worth a follow-up issue rather than a change here.
- **`tonic-v12` (dev-dependency, `Cargo.toml:91`) is the sole reason** mesh.lock carries 531 packages to mocker's 520 — `axum 0.7.9`, `tower 0.4.13`, `indexmap 1.9.3`, `hashbrown 0.12.3`, `prost 0.13.5`, `async-stream`. None of it ships in the release binary. Two consequences worth noting before adding `cargo audit`: it would block CI on advisories in the old axum 0.7 / tower 0.4 line that cannot reach production, and the `hashFiles` cache key now churns on dev-only movement. Gating the 0.12 compatibility test behind a feature would cut both.
- **`build.rs:89` stamps `GIT_STATUS` from `git status --porcelain`,** which un-ignoring makes newly reachable: a cargo-induced lockfile rewrite now marks the tree dirty and stamps `ATOM_MESH_GIT_STATUS=dirty` into the binary. Low impact today (`docker/Dockerfile:484` runs plain `atomesh --version`, which does not print it), and note the flip side — paired with `--locked` this becomes a genuine drift detector rather than noise.
- **Cosmetic:** `build(rust):` is outside the repo's declared type list (`feat, fix, refactor, docs, test, chore, perf, ci`) and has no precedent — `^build` over the last 2000 subjects returns 0, against `fix 45 / feat 26 / perf 20 / ci 16` in the last 400. `ci(rust):` or `chore(rust):` matches both.

## Checked and refuted

Recording these so they are not raised later: `pkg-config` and `libssl-dev` **are** installed in both Dockerfiles (`:438` / `:169`); and `atom/mesh/Cargo.lock` **is** exercised in CI, by `atomesh-mocker-benchmark.yaml:165` and by the accuracy workflow's `ATOM_MESH_BUILD=1 pip install -e .`.

PR #2146, opened 2026-09-06
[MiniMax-M3][LMCache] Drive ATOM's byte-level KV offload from the vLLM plugin

XiaobingSuper:
## What this does

Lets vLLM drive ATOM's existing byte-level LMCache offload, so the **vllm-atom plugin path gets the KV offload tier the native engine already has**. MiniMax-M3 is the model that forced it; nothing here is M3-specific except the one attention hook.

Validated end to end on 4×MI355X (TP=4, `amd/MiniMax-M3-MXFP4`, fp8 KV), with **cudagraph and HBM prefix caching both enabled**:

| pass | supplied by | gsm8k 20-shot, 200q |
|---|---|---|
| cold | all recomputed | 0.92 |
| replay | **621,312 / 656,081 tokens (94.7%) via `external_kv_transfer`** | **0.935** |

**No LMCache upgrade needed** — validated on the image's stock **LMCache 0.4.5**, where the replay pass scores 0.945 against 0.945 cold with 621,312 tokens served externally, and the round trip is bit-exact. (0.4.5 is what rejected `LMCacheConnectorV1` on M3 with `unsupported kv_caches format with list depth 1 and tensor dimension 4`; that is the path this branch does not use.) It also runs on 0.5.5rc4.

Plus a targeted correctness check (markers planted at 10/50/90% depth, HBM flushed with filler prompts in between): 11,008 tokens restored entirely from LMCache, all three markers reproduced.

## Does the tier actually earn its place

Paired A/B, same seeds, same server flags, `--kv-transfer-config` the only difference. Synthetic prefix-reuse load (80,896-token shared prefix + 9,270 unique tail, 32 requests, concurrency 2, KV capped to 262,144 tokens):

| | LMCache on | LMCache off |
|---|---:|---:|
| Total cache read | **76.65%** | 55.17% |
| Supplied by the offload tier | 629,504 tok | 0 |
| **TTFT** | **1,247 ms** | 1,766 ms |
| Output throughput | 133.19 tok/s | 125.18 tok/s |

**TTFT 29% lower, throughput 6% higher.** Run-to-run noise measured on this configuration is 0.11% (TTFT) and 0.86% (throughput), so the effect is far outside it.

On agentic replay (`semianalysis_cc_traces_weka_062126`, 8 trajectory sessions, 900 s) total cache read is **97.12%**, which is the figure comparable to published agentic results (~96%). The synthetic number above is not a worse result — it is a different ceiling, because a per-request unique tail can never hit.

**Pool sizing dominates any measurement of this tier.** The same code reads 76.65% on a 262,144-token KV pool and 56.81% on a 524,288-token one; on the default 7.6M-token pool the tier is queried 2,578,390 times and hits zero, because HBM alone already absorbs 91% and the remainder is content appearing for the first time. A larger pool making the tier look useless is the fingerprint of this mistake, and it is easy to read as "the tier does not work". The recipe carries the budget formula.

## Why not point vLLM at LMCache's own connector

LMCache's GPU connectors accept only the clean NHD/HND family and pick **one** format per model. M3 registers three physical layouts at once (dense K/V interleaved, sparse K/V in separate regions, plus a DSA index cache), and on ROCm the paths that could express that are unavailable anyway: the per-layer-format connector (V3) is off by default upstream and hangs on M3, and the multi-process path imports `cupy`, which LMCache's `platform/rocm` does not provide.

ATOM already solved this for its native engine by never asking LMCache to understand the layout: `DenseKVByteCodec` gathers whole paged blocks into a chunk-major `uint8` blob and LMCache stores opaque bytes. That codec is reused verbatim here — **no change to LMCache, no change to vLLM**. This branch is the adapter that lets vLLM drive it.

## Shape of the change

- `kv_cache_layout.py` — vLLM hands a connector a flat `{layer_name: tensor}` and says nothing about what is inside. M3's sparse layers are not contiguous as a whole (`stride(1)` jumps the entire K region), so the tensor cannot be one segment, but `t[:, 0]` and `t[:, 1]` each are — exactly the `k_cache`/`v_cache` pair the codec expects. Dense layers stay whole; the DSA index caches fold into their owning layer.
- `connector.py` — `KVConnectorBase_V1` wrapper. Layer-granular hooks are inert on purpose: ATOM moves a whole request's blocks per transfer, and blocking a forward on a save would put the offload tier on the critical path it exists to keep clear.
- `seq_view.py` — presents a vLLM `Request` as the seq ATOM's scheduler expects, as **one stable instance per request** (ATOM compares object identity to detect a recycled request id).
- `offload_config.py` — derives ATOM's config from `VllmConfig` rather than duplicating block size and role.
- `dense/connector.py` — two small additions on the ATOM side, see below.
- `recipes/atom_vllm/MiniMax-M3-LMCache-Byte-Offload.md` — launch command, a four-check verification list with two counter identities, the pool-sizing budget and reachable-ceiling formulas, and seven operational gotchas.
- 36 unit tests, built on M3's real strides.

## Five defects found while validating (each is a commit with its own reasoning)

1. **fp8 KV scales did not travel.** Every warm hit came back as fluent garbage while every metric reported success. M3's sparse cache stores fp8 mantissas whose scale is per token **and** per head, and that table lives on the attention layer — `register_kv_caches` hands over the caches only. A restored block was dequantised against whatever its previous occupant left behind; both sides are the right size and dtype, so nothing can fail loudly. The layer now reports its own scales (`get_kv_transfer_scales`, named after the native path's `get_kv_transfer_tensors`), with two hard errors so this cannot go quiet again: a layer holding multi-element scales it cannot report, and a reported scale whose leading axis is not the block axis.
2. **A promised load ATOM would not issue deadlocked the engine.** A lookup hit is not a decision to load — ATOM drops hits that HBM already covers, that are not chunk aligned, or that fall below its transfer floor, and its native scheduler asks `should_park_for_load_after_alloc` before parking anything. vLLM has no second chance: `async=True` parks the request until the **worker** reports `finished_recving`, and a load that is never issued is never reported. Both routine drops fire constantly in the configurations we want (the floor defaults to 8192 tokens; with prefix caching on, vLLM's block-aligned frontier is regularly not chunk aligned). Symptom: EngineCore spinning in `schedule()`, every GPU idle, no log line at all.
3. **Worker completions never reached the scheduler half.** `update_connector_output` was unimplemented, so nothing on the scheduler side ever finished — and every request whose save was still in flight kept its `SeqView`, which pins that request's prompt token ids. At M3's context lengths that is most of a megabyte retained per request.
4. **A watchdog** for (2): a promise that goes undispatched now names itself in one ERROR line instead of hanging silently.
5. **The save frontier could outrun the block table.** Every save on a long prompt failed with `LMCache token range exceeds ATOM block table: needed_blocks=N+1, available_blocks=N` — 88 times in one ISL-90k run, zero in any gsm8k-sized one. ATOM sizes a save from `num_cached_tokens` and hands `block_table` to LMCache alongside it; on the native path those live on one `Sequence` and advance together, but here they arrive through two independent vLLM callbacks (`scheduler_output` vs `update_state_after_alloc`) and separate by a block under chunked prefill. A chat-sized prompt is allocated in one go, which is why every earlier test passed. Capping the frontier is the semantics rather than a guard: KV past the block table is not in any block this connector knows about. Hit rate is unchanged by the fix (inside the 2.64-point run-to-run spread), so this is correctness, not throughput.

`save_finished_by_request` / `load_finished_by_request` are added to `DenseOffloadScheduler` for (3): vLLM's `KVConnectorOutput` carries request ids as plain strings, while `save_finished`/`load_finished` deliberately refuse a raw id once the lifecycle has an exact operation identity. Rather than weaken that guard for everyone, the plugin path gets a way to resolve the identity ATOM already parked.

## Depends on

`fix/m3-vllm-plugin-index-cache-fp8-dtype` — M3 with `--kv-cache-dtype fp8` needs the index cache to carry a real fp8 dtype, or the index-topk Triton kernel dots bf16 against uint8 on the first request. Independent of this branch; this one only depends on the result.

## `FULL_AND_PIECEWISE` was broken on M3, and is fixed here

An earlier revision of this description told readers to use `cudagraph_mode=FULL`. **That constraint is gone** — `bbfd3846` in this branch is the fix, and every measurement above runs on `FULL_AND_PIECEWISE`.

The defect was that M3's sparse attention lacked `@eager_break_during_capture`. vLLM auto-enables `VLLM_USE_BREAKABLE_CUDAGRAPH` for this architecture, so there is no FX splitting: one stream capture drives the whole forward and only decorated ops end a segment. Without it, 57 of M3's 60 attention layers were captured wholesale, freezing that batch's token counts, `block_table`, `seq_lens` and topk indices into every replay. A cold prompt prefills past the largest captured size and runs eagerly, so it answers correctly; reuse a prefix and the short remainder lands inside a captured size and answers fluently from the wrong KV — so **any M3 workload with prefix caching on was exposed**, and it stays invisible while the hit rate is near zero. Measured without LMCache, same gsm8k twice: replay **0.19 → 0.925** after the fix. Captured segments go 3 → 60.

The op now writes into a caller-owned `output` buffer, because the decorator replays the Python kernel on every breakable-cudagraph replay and a tensor allocated inside would land at a new address each time while captured segments still read the address recorded at capture.

Decode is untouched: full decode graphs dispatch with `cudagraph_runtime_mode == FULL`, which the decorator passes through. On a 1K1K load the cost of the extra segments is below the measurement floor — `FULL_AND_PIECEWISE` 1,490/1,497 tok/s against `FULL` 1,482/1,495, a 0.34% mean difference against 0.46%/0.87% run-to-run spread, i.e. **too small to measure, which is not the same as zero**.


atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -0,0 +1,407 @@
...
+        """
+        out = self._worker.get_finished()
+
+        finished_recving = {_req_id_of(c) for c in out.finished_loading}
+        failed = {_req_id_of(c) for c in out.failed_loading}
+        if failed:
+            logger.warning(
+                "ATOM LMCache offload: load failed for %s; recomputing", sorted(failed)
+            )
+            finished_recving |= failed
```
yhl-amd, line 206: [P1] Preserve failed-load invalidation

Adding failed loads to `finished_recving` does not make vLLM recompute them. In the pinned vLLM revision, `num_computed_tokens` is truncated only when `KVConnectorOutput.invalid_block_ids` identifies the failed blocks; a bare `finished_recving` notification is treated as a successful receive and the whole external prefix is cached. The scheduler callback below also calls `load_finished_by_request()` for this same ID. A partial or failed LMCache retrieve can therefore resume attention over stale/uninitialized KV and silently corrupt output. Please retain the failed request's allocated block IDs, expose them through `get_block_ids_with_load_errors()`, and carry success/failure separately so ATOM settles the operation through `load_failed()` rather than `load_finished()`.

atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -0,0 +1,407 @@
...
+
+        self._saved_awaiting_finish |= {_req_id_of(c) for c in out.finished_saving}
```
yhl-amd, line 208: [P1] Forward internal save completions before request termination

Holding every `finished_saving` notification here until the request finishes prevents the scheduler-side `DenseOffloadScheduler` from clearing `_save_inflight`. While that entry remains, `build_connector_meta()` skips every later save for the request, so a chunked long prompt stops offloading after its first emitted frontier. It also breaks final cleanup: `request_finished()` sees the stale inflight entry and retains `_save_tracker`; after this adapter finally emits `finished_sending`, vLLM frees the blocks, but only the `SeqView` is dropped, leaving the tracker able to schedule another save from freed/reassigned block IDs. Please transport internal `SaveOperationId` completions immediately via `KVConnectorWorkerMetadata` (as vLLM's offloading connector does), while reserving `finished_sending` solely for releasing a request whose final save is still pending.

atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -0,0 +1,407 @@
...
+
+        A load is published to the scheduler through ``get_finished`` only once
+        every block of the request has landed, so there is no partially-loaded
+        layer for a forward to wait on.
+        """
+
+    def save_kv_layer(self, layer_name: str, kv_layer, attn_metadata, **kwargs) -> None:
+        """Inert: saves are issued per request from ``build_connector_meta``."""
+
+    def wait_for_save(self) -> None:
```
yhl-amd, line 171: [P1] Fence asynchronous saves when vLLM preempts a request

Making `wait_for_save()` a no-op is only safe if preemption is handled separately. vLLM can preempt a running request, immediately free/reassign its KV blocks, and does not call `request_finished()` on that path. This adapter neither overrides `handle_preemptions()` nor carries `scheduler_output.preempted_req_ids` in its metadata, so the ATOM background save may continue reading those reused blocks and persist another request's bytes under the old key; its scheduler-side save tracker also retains the obsolete block table. Please propagate preemptions and wait/cancel/fence affected saves before block reuse, or block in `wait_for_save()`.

atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -0,0 +1,407 @@
...
+        finished_recving = {_req_id_of(c) for c in out.finished_loading}
+        failed = {_req_id_of(c) for c in out.failed_loading}
+        if failed:
+            logger.warning(
+                "ATOM LMCache offload: load failed for %s; recomputing", sorted(failed)
+            )
+            finished_recving |= failed
+
+        self._saved_awaiting_finish |= {_req_id_of(c) for c in out.finished_saving}
+        self._finished_awaiting_save |= set(finished_req_ids or ())
```
yhl-amd, line 209: [P2] Do not retain finished IDs for requests that never issued a save

This adds every completed request to `_finished_awaiting_save`, but an ID is removed only if it later appears in `finished_saving`. Short requests with no aligned chunk, and every request in `kv_consumer` mode, therefore remain in this set for the lifetime of the worker. Besides the unbounded growth, reusing such a request ID lets a save from the new request intersect the stale finish marker and emit `finished_sending` while that request is still running, which makes vLLM try to free a non-finished request. Please track only requests for which the scheduler actually deferred a save, or explicitly retire no-save completions.

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Agreed, and confirmed against the pinned vLLM revision before changing anything: `_update_requests_with_invalid_blocks` is the only thing that moves `num_computed_tokens`, and `_update_waiting_for_remote_kv` caches the whole external prefix for any id not in `failed_recving_kv_req_ids`. So the old code did exactly what you describe — woke the request and then served never-written blocks as if they held KV.

Both halves of your suggestion are implemented:

**Naming the blocks.** The worker now records, per failed load, the GPU blocks the load was supposed to fill and did not: `OffloadWorkerMixin._record_load_error_blocks()` in `atom/kv_transfer/offload/_offload_common.py`, drained by `take_load_error_blocks()` and returned from the new `AtomLMCacheOffloadConnector.get_block_ids_with_load_errors()`. The dense worker calls it at both real failure sites in `_do_load_req` — the non-chunk-aligned HBM boundary and the partial `ret_mask` — plus the generic `_guard` catch, so an exception mid-copy is covered too.

The recorded range is `block_ids[hbm_cached_tokens // bs : ceil(lmcache_cached_tokens / bs)]`. Blocks below the HBM frontier are deliberately excluded: they hold valid KV and may be shared with another request, so invalidating them would be a correctness bug in the other direction.

One subtlety worth flagging for review: `bs` here is the **virtual** block size, not `kv_cache_block_size`. `BlockGPUConnector` sets `self.block_size = self.virtual_block_size` (`kv_cache_block_size * decode_context_parallel_size`) as its token→block-id mapping grid, so under DCP one scheduler block id covers one virtual global block while the codec moves a rank-local physical page. Using the physical size would index different entries of the very same table. `test_dense_worker_records_the_blocks_a_failed_load_left_unfilled` pins this with `block_size 4 × dcp 2 = 8`.

**Settling through `load_failed`, not `load_finished`.** You are right that routing a failure through `load_finished_by_request()` was wrong for a second, independent reason: `load_finished` pops `_load_save_floors`, which is the record that the `[HBM, LMCache)` range is *not* persisted. Popping it means the recomputed chunks are never saved — a silent, permanent hole in the cache rather than a visible failure. There is now a `load_failed_by_request()` alongside the other two, and `update_connector_output` routes by whether the worker reported a failure for that id this step.

Failed loads still join `finished_recving`, since that is what releases the request from `WAITING_FOR_REMOTE_KVS` and the alternative is a hang — but they no longer travel alone.

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Agreed on both consequences, and the first one is worse than "stops offloading after the first frontier": because `_save_inflight[sid]` was cleared only via `finished_sending`, and `finished_sending` was gated on request finish, a chunked long prompt offloaded its **first** chunk and then nothing else for the life of the request — which is the case M3 exists for.

Implemented as you suggested, via `KVConnectorWorkerMetadata`:

- `AtomOffloadWorkerMetadata` carries two `dict[str, int]` — `saved` and `load_failed` — and implements `aggregate()` by summing per request id. `build_connector_worker_meta()` drains and returns it right after `get_finished()`, so it carries exactly the completions that call observed, and returns `None` on a quiet step rather than pickling an empty object per rank.
- `update_connector_output` absorbs it on the scheduler half and calls `save_finished_by_request()` per completed save, immediately — no longer gated on anything.
- `finished_sending` is now reserved for exactly what you describe: the scheduler half's own release list for requests whose free was deferred, echoed back once.

Counts rather than sets, and fired only at world-size quorum (`divmod(pending + count, world_size)`): `KVOutputAggregator` folds one metadata object per rank, and firing on the first rank to report would drop `should_defer_free()` while a slower rank is still reading those blocks — the same class of bug this comment is about, one layer down.

On the `request_finished` half: `_collect_releases()` re-runs `should_defer_free()` every step and calls `self._scheduler.request_finished(seq)` a second time at the moment the release is actually emitted. That call is what pops `_save_tracker`, so the window you describe — blocks freed, tracker still live, save scheduled against reassigned block ids — is closed. `has_pending_push_work()` keeps the engine stepping while a deferred free is still owed, otherwise an idle server never calls `build_connector_meta` again and silently loses KV capacity.

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Agreed — and confirmed the premise in the pinned tree: `_preempt_request` frees the blocks inside `schedule()` and there is no `request_finished` on that path. Worth naming precisely what the failure is: an in-flight save keeps reading blocks that now belong to somebody else and persists the **new occupant's bytes under the preempted request's LMCache key**. That is a poisoned cache entry, not a lost one — every later request that hits that key gets wrong KV, and nothing about it looks like a failure.

Fenced rather than left to `wait_for_save()`, in three parts:

1. **Propagation.** `build_connector_meta` reads `scheduler_output.preempted_req_ids` and carries them in `AtomOffloadMetadata.preempted_req_ids`.
2. **Fence on the worker half.** `handle_preemptions()` is overridden and calls `OffloadWorkerMixin.wait_for_requests()`, which joins exactly that request's in-flight copy futures (tracked via `_track_job` at both `submit` sites). vLLM calls `handle_preemptions` at the top of `execute_model`, before `_update_states` and before the forward — verified it is a valid fence point.
3. **Scheduler-half state.** `_handle_preempted` calls `cancel_pending_load()` for a load that was queued but not yet dispatched, and `SeqView.reset_for_preemption()` to forget the block table and the frontier. The save loop sizes its next store from exactly those two fields, so a stale pair is the same poisoning bug via a different route.

Deliberately **not** reset: anything recording what is already persisted. LMCache keys chunks by token content, not by placement, so a chunk stored before the preemption is still a valid hit afterwards and must not be stored twice. That distinction is the subject of `test_preemption_forgets_placement_but_not_what_was_stored`.

Fencing rather than cancelling is the deliberate choice: the already-persisted chunks stay valid, so the recomputed prefix is not re-saved.

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Agreed, and the id-reuse consequence you point out is the sharper end of it: `_update_from_kv_xfer_finished` does `assert req_id in self.requests` and then `_free_blocks`, so emitting `finished_sending` for a recycled id either trips the assert or frees a live request's blocks.

`_finished_awaiting_save` and `_saved_awaiting_finish` are both gone. The replacement tracks only what the scheduler actually deferred:

- `request_finished()` adds an id to `_deferred_frees` only when it returns `True` — i.e. only when vLLM is genuinely holding blocks for a pending ATOM save. A request with no aligned chunk, and every request in `kv_consumer` mode, returns `False` and is never entered anywhere.
- `_collect_releases()` moves ids from `_deferred_frees` to `_releases_in_flight` when `should_defer_free()` goes false, and `update_connector_output` discards them from `_releases_in_flight` on the echo. Both sets are bounded by the number of live deferred requests, and an id leaves for good once released — `test_a_release_is_emitted_exactly_once` pins that.
- Save reports are matched by request id **and** only fire `save_finished_by_request` at world-size quorum, so a late report from a slow rank cannot complete a newer lifecycle either.

atom/plugin/vllm/kv_transfer/offload_config.py
```diff
@@ -0,0 +1,147 @@
...
+    def __getattr__(self, name: str) -> Any:
+        if self._inner is not None:
+            try:
+                return getattr(self._inner, name)
+            except AttributeError:
+                pass
+        return getattr(self._outer, name)
+
+    def __repr__(self) -> str:
+        return f"_HFConfigView(inner={type(self._inner).__name__}, outer={type(self._outer).__name__})"
```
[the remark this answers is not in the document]
PerryZhang01: Fixed — the `__repr__` return is wrapped in parentheses. `black --check atom/ tests/` is clean across the tree (741 files).

PerryZhang01:
Rebased onto `main` (`bf9b129fc`). Two conflicts, both from #2147's extraction of `ChunkedOffloadSchedulerBase` into `atom/kv_transfer/offload/chunked_scheduler.py`: the scheduler half of `DenseOffloadScheduler` moved wholesale, so the three `*_by_request` helpers this PR adds now live in the new base class instead — which also makes them available to the other chunked schedulers rather than only the dense one.

The rebase also surfaced a real incompatibility worth calling out, fixed in its own commit:

**#2166's early block release writes attributes on the seq object.** `ChunkedOffloadSchedulerBase.request_finished` freezes a finishing request's placement onto the seq as `_offload_finished_block_ids` / `_offload_finished_cached_tokens`, so a final save can still be dispatched after vLLM has handed the blocks back. ATOM's own `Sequence` has a `__dict__` and absorbs that silently. `SeqView` is slotted, so it raised `AttributeError` — on *every* completed request, out of `request_finished`.

Both names are now in `__slots__`, deliberately left unset in `__init__` because that path tests for them with `hasattr` and an unset slot is absent the same way a missing attribute is. `reset_for_preemption` drops them as well: frozen placement is still placement, and a preempted request's is as stale as its live block table.

The regression test calls the real unbound `ChunkedOffloadSchedulerBase.request_finished` against a `SeqView` rather than a fake scheduler, so it will also catch the next attribute the base class decides to stash on a seq.

---

Post-rebase verification: `black --check atom/ tests/` clean (741 files); ruff reports nothing new on the touched files. The full suite was run on this branch and on `origin/main` in the same container, and the failure lists were compared rather than assumed — this branch adds **zero** failures (the 12 `FAILED`/`ERROR` lines it shows are all present on `main` too; `main` additionally showed 10 that did not reproduce here).

yhl-amd (approved):

PR #2149, opened 2026-09-07
feat(v4): DeepSeek-V4-Flash-Vision support

HaonanWang98:
Adds the vision path for DeepSeek-V4 checkpoints that ship a ViT + aligner. The tower is a pure-torch port of the checkpoint's reference `inference/vision.py`; vision is built conditionally on `vision_n_layers > 0`, mirroring the reference `Transformer.__init__`.

Two things make this model differ from a plain embedding substitution:

- MoE routing: image tokens select experts with a second per-layer bias (`gate.bias_vl`). `select_experts` only takes a `[E]` vector, so a Triton kernel does the per-token choice.
- Attention: tokens inside `[IMAGE_START, IMAGE_END]` attend both ways. The prefill kernel is index-driven with no causal mask, so this is a per-token `(extend_start, extend_count)` widening of the index range, not a kernel change. Text forwards pass neither and stay bit-identical.

Multimodal code also moves to its own `atom/multimodal/` package, with the per-model builders under `models/` and a Protocol for the builder contract.

Not supported: MTP/DSpark drafting, PCP, TBO, chunked multimodal prefill, TP-sharded ViT.

## Submission Checklist

### DIFF
--- a/atom/config.py
+++ b/atom/config.py
@@ -2077,6 +2077,45 @@ class Config:
         # 16. Same reasoning and same mechanism as the V4 override above: the
         # BlockManager and slot_mapping assume one global block size, so it is
         # set here and the attention builder sizes the index cache from it.
+        # ----- DeepSeek-V4 vision: reject unverified feature combinations -----
+        # Collected here rather than as scattered asserts so a request never
+        # reaches a half-supported path. Each of these silently corrupts rather
+        # than failing on its own.
+        if int(getattr(self.hf_config, "vision_n_layers", 0) or 0) > 0:
+            if self.enable_prefix_caching:
+                # Turned off rather than rejected: it is on by default, so
+                # raising would make every vision checkpoint fail to start.
+                # The block hash covers token ids only, and an image block's
+                # sentinel run is byte-identical for any two images of the same
+                # grid -- one request could be served another's image KV. A hit
+                # also restarts the prompt mid-block, which the in-image
+                # visibility math cannot express.
+                logger.warning(
+                    "Disabling prefix caching: DeepSeek-V4 vision block hashes "
+                    "cover token ids only, and image placeholder runs collide "
+                    "across different images of the same grid."
+                )
+                self.enable_prefix_caching = False
+            unsupported = []
+            if self.enable_tbo or self.enable_tbo_decode:
+                unsupported.append("TBO")
+            if self.prefill_context_parallel_size > 1:
+                unsupported.append("PCP")
+            from atom.utils import envs as _envs
+
+            if _envs.is_set("ATOM_USE_TRITON_MOE") and _envs.ATOM_USE_TRITON_MOE:
+                # `FusedMoE`'s Triton path checks `custom_routing_function is
+                # not None` and then calls aiter's `routing()` -- it never
+                # invokes the callable. Image tokens would route with the text
+                # bias and the hash layers would lose `tid2eid` entirely, with
+                # fluent wrong output and no warning.
+                unsupported.append("ATOM_USE_TRITON_MOE=1")
+            if unsupported:
+                raise ValueError(
+                    "DeepSeek-V4 vision checkpoints do not support "
+                    f"{', '.join(unsupported)}."
+                )
+
         is_glm5_next = any("Glm5Next" in str(a) for a in arches)
         if is_glm5_next:
             unsupported_features = []
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -39,9 +39,9 @@ if TYPE_CHECKING:
 from atom import SamplingParams
 from atom.model_engine.arg_utils import EngineArgs
 from atom.model_engine.llm_engine import _load_tokenizer
-from atom.model_engine.multimodal import build_multimodal_inputs
 from atom.model_engine.request import RequestOutput
 from atom.model_engine.sequence import new_token_ids
+from atom.multimodal import build_multimodal_inputs
 from atom.utils.arg_parser import FlexibleArgumentParser
 from atom.utils.gc_utils import (
     freeze_gc_heap,
--- a/atom/entrypoints/openai/api_server.py
+++ b/atom/entrypoints/openai/api_server.py
@@ -662,7 +662,25 @@ def _get_multimodal_processor():
     global processor, model_name
     if processor is None:
         logger.info(f"Loading multimodal processor from {model_name}...")
-        processor = AutoProcessor.from_pretrained(model_name, trust_remote_code=True)
+        try:
+            processor = AutoProcessor.from_pretrained(
+                model_name, trust_remote_code=True
+            )
+        except (OSError, ValueError, KeyError) as exc:
+            # Some checkpoints ship no `preprocessor_config.json` and do their
+            # own preprocessing inside the registered input builder
+            # (DeepSeek-V4-Flash-Vision is one). Without this the request 500s
+            # inside AutoProcessor before any model-specific handling runs, so
+            # the server could not serve an image the offline example can.
+            from transformers import AutoTokenizer
+
+            logger.info(
+                f"No HF processor for {model_name} ({exc}); "
+                "falling back to the tokenizer."
+            )
+            processor = AutoTokenizer.from_pretrained(
+                model_name, trust_remote_code=True
+            )
     return processor
 
 
--- a/atom/examples/multimodal_inference.py
+++ b/atom/examples/multimodal_inference.py
@@ -9,7 +9,7 @@ from transformers import AutoProcessor
 
 from atom import SamplingParams
 from atom.model_engine.arg_utils import EngineArgs
-from atom.model_engine.multimodal import build_multimodal_inputs
+from atom.multimodal import build_multimodal_inputs
 from atom.utils.arg_parser import FlexibleArgumentParser
 
 parser = FlexibleArgumentParser(
@@ -18,7 +18,7 @@ parser = FlexibleArgumentParser(
         "Generic image+text multimodal offline inference using the native ATOM engine.\n"
         "Validated with Qwen3.5 and Kimi-K3. The script relies on the model's\n"
         "Hugging Face processor and chat template, plus the architecture-specific\n"
-        "input builders in atom.model_engine.multimodal for processors that do not\n"
+        "input builders in atom.multimodal for processors that do not\n"
         "follow the Qwen convention."
     ),
 )
@@ -59,8 +59,15 @@ def main():
     # Force eager mode and single-batch cudagraph sizes for simplicity
     args.cudagraph_capture_sizes = "[1]"
 
-    # Load processor (handles media preprocessing and chat template)
-    processor = AutoProcessor.from_pretrained(args.model, trust_remote_code=True)
+    # Load processor (handles media preprocessing and chat template). Some
+    # checkpoints ship none and preprocess inside their input builder instead.
+    try:
+        processor = AutoProcessor.from_pretrained(args.model, trust_remote_code=True)
+    except (OSError, ValueError, KeyError) as exc:
+        from transformers import AutoTokenizer
+
+        print(f"No HF processor for {args.model} ({exc}); using the tokenizer.")
+        processor = AutoTokenizer.from_pretrained(args.model, trust_remote_code=True)
 
     images = [Image.open(path).convert("RGB") for path in args.image]
 
--- a/atom/model_engine/llm_engine.py
+++ b/atom/model_engine/llm_engine.py
@@ -13,8 +13,8 @@ from transformers import AutoTokenizer, PreTrainedTokenizerFast
 
 from atom.config import Config
 from atom.model_engine.engine_core_mgr import CoreManager, DisaggCoreManager
-from atom.model_engine.multimodal import get_mrope_input_positions
 from atom.model_engine.sequence import Sequence
+from atom.multimodal import get_mrope_input_positions
 from atom.sampling_params import SamplingParams
 from atom.utils import envs
 
--- a/atom/model_ops/attentions/deepseek_v4_attn.py
+++ b/atom/model_ops/attentions/deepseek_v4_attn.py
@@ -426,6 +426,15 @@ class DeepseekV4AttentionMetadataBuilder(CommonAttentionBuilder):
         self.index_head_dim = getattr(hf, "index_head_dim", 128)
         self.window_size = getattr(hf, "sliding_window", 128)
         self.index_topk = getattr(hf, "index_topk", 1024)
+        # Vision checkpoints attend BIDIRECTIONALLY inside an image span, which
+        # widens the per-token sliding window. Zero on text-only V4, where every
+        # visibility term below collapses and the causal formulas are untouched.
+        self.vision_max_n_token = (
+            int(getattr(hf, "vision_max_n_token", 384))
+            if int(getattr(hf, "vision_n_layers", 0) or 0) > 0
+            else 0
+        )
+        self.vocab_size = int(getattr(hf, "vocab_size", 0))
         self.rope_head_dim = getattr(hf, "qk_rope_head_dim", 64)
         # MTP-portion of compress_ratios. `prepare_mtp_decode`'s direct-kernel
         # fast path only handles SWA (ratio=0) draft layers; non-zero ratios
@@ -2877,6 +2886,9 @@ class DeepseekV4AttentionMetadataBuilder(CommonAttentionBuilder):
             attn_metadata.state_slot_out_cpu,
             scheduled_bs,
             scheduled_tokens,
+            image_visibility=self._image_visibility(
+                batch, extend_lens_np, scheduled_tokens
+            ),
         )
 
         # ----- PCP: reindex per-query metadata to this rank's 1/W shard -----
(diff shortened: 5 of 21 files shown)


valarLip:
Reviewed at `a268ede2d`. Verified against the checked-out branch.

Cleared and not repeated below, so a re-review need not redo them: the `.gate.bias_vl` before `.gate.bias` ordering in `weights_mapping` is safe (after the first `str.replace` the name no longer contains `.gate.bias`), and no `weights_mapper` prefix / `.scale` rule / `_is_past_last_layer` touches a `vision.*`, `aligner.*` or `image_*` name; the new `extend_start` / `extend_count` kwargs cannot shift any positional caller (all four call sites pass keywords); `_cos_sin_table`'s `@lru_cache` is not an aliasing hazard because `get_vision_cos_sin` always copies via `torch.cat`; `grid_tokens` and `build_image_block` do compute the same pair-transpose order; and `tokenizer.encode`'s default `add_special_tokens=True` matches the pre-existing text path at `llm_engine.py:750`.

---

# 1. Blocking: the vLLM plugin's V4 path is broken outright, text-only included

`deepseek_v4.py:4691` now passes three positional arguments:

```python
h = self.model(input_ids, positions, inputs_embeds)
```

The plugin's model class takes two:

```python
# atom/plugin/vllm/models/deepseek_v4.py:378
def forward(self, input_ids: torch.Tensor, positions: torch.Tensor) -> torch.Tensor:
```

and `:404` rebinds `deepseek_v4_base.DeepseekV4Model = DeepseekV4ModelVllm`, so under the plugin `self.model` *is* that class. The plugin wrapper overrides neither `forward` nor `_forward_impl`.

Every DeepSeek-V4 forward under the vLLM plugin therefore raises `TypeError: forward() takes 3 positional arguments but 4 were given` — **including on text-only checkpoints**, since `inputs_embeds=None` is still passed positionally. `support_torch_compile.__call__` does `sig.bind(self, *args)` against the subclass signature, so eager and compiled both fail.

This is a signature change without a subclass sweep. Adding `inputs_embeds: torch.Tensor | None = None` to `DeepseekV4ModelVllm.forward` (and forwarding it to `super().forward`) is the minimal fix.

# 2. Prefix caching is never disabled for multimodal sequences — only chunking is

`scheduler.py:1300` calls `can_allocate(seq)` unconditionally, and the new guard governs chunking only:

```python
atomic_prefill = getattr(seq, "multimodal_data", None) is not None
if not atomic_prefill and self.enable_chunked_prefill and ...:
```

There is no multimodal guard anywhere in `block_manager.py`, and `enable_prefix_caching` defaults to True. On any prefix-cache hit, `ScheduledBatch` stages `seq.token_ids[num_cached_tokens : ...]`, so `input_ids` starts mid-prompt. Three fatal paths, all inside the EngineCore worker (which kills every co-scheduled request):

- any hit at all → `chunk_start > 0` → the new `AssertionError` at `deepseek_v4_attn.py:3810`
- a cut inside the image block → `ValueError("unbalanced image span")` from `image_visible_spans`
- a cut after it → `ValueError(... for 0 IMAGE token slots)` from `merge_multimodal_embeddings`

A shared system prompt of one hash block is enough to trigger this.

Worse is the case where it does *not* crash. `block_manager.py:375`:

```python
h.update(np.asarray(token_ids, dtype=np.int64).tobytes())
```

The hash covers token ids and nothing else. The sentinel run `vocab_size + type` is byte-identical for any two images of the same grid, so request B can be served request A's image KV — silently, with no length or count mismatch anywhere to catch it.

# 3. The `chunk_start` guard is batch-wide, so a co-scheduled text prefill trips it

```python
# deepseek_v4_attn.py:3810
if chunk_start_pt.any():
    raise AssertionError("a multimodal prefill reached the attention builder with chunk_start > 0 ...")
```

`chunk_start_pt = np.repeat(chunk_start_per_seq_np, token_num_per_seq)` spans every sequence in the batch. The scheduler makes only the *multimodal* sequence atomic; it still admits, in the same batch, a text sequence continuing a chunk. So `--enable-chunked-prefill` plus one image request raises on the first mixed batch, blaming chunking that did not happen to the image request.

The condition needs to be evaluated over tokens of sequences carrying image sentinels only.

# 4. The Triton MoE path uses `custom_routing_function` as a boolean and never calls it

```python
# moe.py:1496
needs_custom_routing = (
    use_grouped_topk
    or scoring_func != "softmax"
    or e_score_correction_bias is not None
    or custom_routing_function is not None
)
if needs_custom_routing:
    from aiter.ops.triton.moe.moe_routing.routing import routing
    ... routing(router_logits, ..., bias=e_score_correction_bias, ...)   # returns; callable never invoked
```

Same shape at `:1444` for `use_triton_decode`. `self.use_triton` defaults to **True on gfx94x** (`moe.py:1097`) with no env var, and is forced on by `ATOM_USE_TRITON_MOE=1` — a workaround this repo actively uses for the V4 fused-MoE NaN.

This PR routes all 43 layers of a vision checkpoint through that hook. On those configs every image token is routed with `gate.bias` instead of `gate.bias_vl`, and the three hash layers lose `tid2eid` entirely. Fluent, wrong output, no warning. Note also that `MoE.__init__` assigns `self.experts.custom_routing_function` *after* construction, which bypasses the one guard at `moe.py:2817` that would have rejected the unsupported combination.

# 5. Vision checkpoints allocate `inputs_embeds` on every forward, including CUDAGraph-replayed decode

```python
# deepseek_v4.py:5069
if self.has_vision and inputs_embeds is None:
    inputs_embeds = self.embed_input_ids(input_ids)
```

This is unconditional, so at the default `--level 3` every decode step goes `_piecewise_cg_active()` → `self.model(input_ids, positions)` → a fresh allocation. `cuda_graph.py:204-207` records `entry.input_addresses = [x.data_ptr() for x in args]` at capture and replays against those addresses; the equality assertion at `:286` is gated on `if self.is_debugging_mode`.

So this works only while the caching allocator happens to hand back the same block, and reads a stale embedding buffer the moment anything perturbs allocation order — gibberish with no error. Text-only V4 does not have this problem because `embed` stays inside the compiled region, where piece 0's inputs are the persistent `forward_vars[...].gpu` slices.

Secondary cost: the vocab-parallel embed and its TP all-reduce now run eagerly outside the graph on every decode step.

# 6. The `AutoProcessor` fallback was added to the example but not to the server

`api_server.py:661` `_get_multimodal_processor()` does a bare `AutoProcessor.from_pretrained(model_name, trust_remote_code=True)`, and it is invoked at `:1675` and `:744` — both *before* `build_multimodal_inputs`.

The PR's own docs and `deepseek_v4_vl._resolve_tokenizer` state that this checkpoint ships no `preprocessor_config.json`, and `atom/examples/multimodal_inference.py:64-71` gained `except (OSError, ValueError, KeyError)` → `AutoTokenizer` for exactly that reason. On any transformers build without the internal AutoTokenizer fallback, every `/v1/chat/completions` image request 500s inside AutoProcessor and none of the PR's model-specific error messages fire — i.e. the server recipe in `recipes/DeepSeek-V4-Flash-Vision.md` cannot serve an image, while the offline example works.

# 7. Thinking mode diverges between the image and text paths, two independent ways

`deepseek_v4_vl.py:732`:

- **Default is inverted.** `mode = "thinking" if enable else "chat"` when neither `thinking_mode` nor `enable_thinking` is present. The text path for the same model builds its adapter with `defaults={"thinking_mode": "thinking"}` (`chat_encoders.py:89-91`), and `merged_kwargs["thinking_mode"]` is only written when the request explicitly sets `thinking` / `reasoning_effort`. So a client sending no `chat_template_kwargs` gets thinking **on** for text and **off** for images on the same server. The PR's own recipe prices this: ZeroBench subquestions Pass@1 21.0 thinking vs 14.7 not.
- **It reads a key the server never writes.** The server writes `merged_kwargs["thinking_effort"]` (`api_server.py:1646`); `_resolve_thinking` reads `chat_template_kwargs.get("reasoning_effort")`. Always None on the server path, so `reasoning_effort` is silently dropped on every image request — including the `max` effort the recipe's ZeroBench numbers were collected at.

# 8. `_image_visibility` is a predicate that mutates state, and every early return fails open

`deepseek_v4_attn.py:3683` writes `self._forward_has_images` as a side effect. Its first guard is:

```python
tokens = getattr(batch, "scheduled_tokens", None)
if tokens is None or tokens.shape[0] < total_tokens:
    return None
```

which degrades quietly while every other invariant in this PR raises. `_forward_has_images` stays False, no widened window is built, and `build_ubatch_prefill_metadata`'s guard at `:3042` (read via `getattr(self, ..., False)`) disarms — producing exactly the outcome its own comment describes: "the model answers fluently off half a picture." A typo in the attribute name would fail open the same way.

The guard's correctness also rests on an undocumented temporal coupling: every prefill must reach `_build_paged_prefill_meta` before any ubatch split, in the same forward, on the same builder instance. `batch.multimodal_data` is a stateless channel for the same fact and is already visible to the splitting caller — and `Config.__post_init__` (`config.py:2082-2095`) already hosts this exact startup-rejection pattern for GLM-5-Next, which is where TBO/PCP + vision belongs.

# 9. `sentinels[offsets.clamp(min=0)]` clamps one bound; the sibling twelve lines up clamps the other

`deepseek_v4.py:5039`. `sentinels` has exactly 5 rows, and `embed_input_ids` clamps the *upper* bound for the identical reason. Any `input_ids` value `>= vocab_size + 5` is an out-of-bounds gather — `offsets = 70720` into a `[5, dim]` tensor for a raw id of 200000 on a 129280-vocab checkpoint. `input_ids` is not validated against the vocabulary on the way in (`/v1/completions` accepts raw ids). On GPU that is an async device-side assert or an OOB read surfacing as a `Memory access fault` in an unrelated later kernel.

Separately, the `sentinels[...]` gather and the `torch.where` each allocate a full `[N, dim]` tensor (~64 MB at 8192×4096 bf16, ~1 GiB at `max_num_batched_tokens` 131072) on the peak-memory step, and `num_slots = int(image_mask.sum())` forces a D2H sync that blocks the host until the 32-layer ViT and aligner complete. An `index_put_` over the sentinel positions avoids all three.

# 10. `extend_start` / `extend_count` are the only kernel arguments with no validation

`paged_prefill_indices.py:344`. Every other tensor argument gets `.dim()` and `.shape[0] >= T` / `>= T+1` assertions. The kernel does unmasked `tl.load(extend_start_ptr + t)` / `tl.load(extend_count_ptr + t)` for `t` in `[0, T)`.

A caller passing an int64 array — the natural numpy default; `image_aware_extend_window` only casts to int32 on its final line — makes every load read half of two adjacent elements. A short array reads past the allocation. Either way the resulting `ext_start_row` values are non-negative garbage, so they pass `CHECK_NEG_ONE_SENTINEL` in `paged_prefill.py`, index arbitrary rows of the per-forward `kv` tensor, and produce fluent wrong output rather than a fault.

# 11. `load_checkpoint_encoding` caches one module process-wide and ignores `model_path`

`deepseek_v4_vl.py:430`:

```python
with _encoding_lock:
    if _encoding_module is not None:
        return _encoding_module
```

The guard never compares its argument, and the `sys.modules` key is the fixed string `"atom_dsv4_encoding"`. Any process touching two V4 vision checkpoints — an eval harness comparing revisions, a test module, a heterogeneous disagg/PP setup — applies the first checkpoint's roles, thinking modes and DSML tool rendering to the second's weights. That is the stale-template accuracy regression the function's own docstring says it exists to prevent, moved from disk into memory. Key both the cache and the `sys.modules` entry on `model_path`.

# 12. The four sentinel embeddings bypass `atom_parameter` and are uninitialized

`deepseek_v4.py:4971` uses bare `nn.Parameter(torch.empty(dim))` — no `atom_parameter`, no explicit dtype. Every other parameter in this file goes through `atom_parameter(torch.empty(..., dtype=...))` (see `MoE.__init__` at `:3628-3640`), which is the single control that forces `requires_grad` off under `ATOM_REQUIRES_GRAD` and pins the dtype.

So `torch.stack([...])` in `merge_multimodal_embeddings` builds an autograd graph on any path not wrapped in `no_grad`, and the dtype is whatever `torch.get_default_dtype()` happened to be. More seriously `torch.empty` is uninitialized: if `image_start` / `image_pad` / `image_newline` / `image_end` are absent or spelled differently in a checkpoint, every one of those positions gets raw memory, and the only signal is a single `logger.warning` from `loading_core.py:425`.

Related: `DeepseekV4ForCausalLM.load_weights` (`:4797`) indexes only `self.model.named_parameters()`, so on that path (documented as "used by tests + when ModelRunner is bypassed") the vision, aligner and sentinel weights are never loaded at all, silently.

# 13. `_as_pair` is defined twice in `kimi_k3_vl.py`

Line 35 validates (`isinstance(value, Sequence) and len(value) == 2`, raising a named `ValueError`); line 506 does `return (int(value[0]), int(value[1]))` unconditionally. Python binds the later one, so every existing caller — `KimiK3PatchEmbed` (`:137`) and the tower builders (`:403`, `:469`) — now gets the unvalidated variant, and a malformed 3-tuple in `preprocessor_config.json` truncates silently instead of raising.

Neither ruff nor the unit tests flag a same-module redefinition of a private function. Delete the copy at 506.

# 14. Module-scope PIL import in a module the model constructor and the attention builder both import

`deepseek_v4_vl.py:41` is `from PIL import Image, ImageOps`. `pyproject.toml` lists no pillow in the base dependencies — only in the optional `diffusion` extra (`:20`). This is now the only module-scope PIL import under `atom/` outside `examples/`.

`api_server.py:634-642` defers its own PIL import with a comment stating exactly why: "Pillow is not a declared dependency, so a module-scope `from PIL import Image` made the whole server module unimportable wherever it is absent — which is the non-GPU CI runner." `deepseek_v4_vl.py` is imported by `DeepseekV4ForCausalLM._build_vision` (`deepseek_v4.py:4958`) at model-construction time and by `deepseek_v4_attn.py:3695/3804` per prefill, so a base install fails to build a V4 vision model with an ImportError rather than a named error.

The PR is aware of the weight — `deepseek_v4.py:5079` duplicates `_IMAGE_SENTINEL = 2` as a literal specifically to avoid importing this module — and then imports it from the hot path anyway.

# 15. Docs and tests: three dangling references, a fictional test, and zero CI coverage

Verified individually:

- `atom/models/deepseek_v4_mm.py` and `atom/model_engine/deepseek_v4_mm.py` **do not exist** (the real module is `atom/models/deepseek_v4_vl.py`). Referenced at `recipes/DeepSeek-V4-Flash-Vision.md:37` and twice in the comment at `deepseek_v4.py:5079-5082`.
- `tests/test_deepseek_v4_vl.py` **does not exist**, yet `recipes/...:167` tells the reader to run it and `deepseek_v4.py:5082` cites `::test_image_sentinel_code_agrees` from it as what keeps the duplicated `_IMAGE_SENTINEL = 2` in sync with `deepseek_v4_vl.IMAGE`. Nothing does.
- `recipes/...:50` names `atom/model_ops/triton_vl_topk.py`; the real file is `triton_mm_topk.py`.
- `recipes/...:63` and `:96` describe `_finalize_prefill_chunk(atomic=True)`, but `scheduler.py:1977` is `_finalize_prefill_chunk(self, seq, start, chunk)` with no such parameter.
- `recipes/...:102-105` says "every image request so far has run at tp=1" while `:124` reports chartqa 2500 samples at `tp=8`.
- The rename left `docs/model_support_guide.md:157`, `recipes/Kimi-K3.md:79` and `atom/diffusion/registry.py:8` pointing at the deleted `atom/model_engine/multimodal.py`. No row was added to CLAUDE.md's Detailed Guides table, `docs/model_support_guide.md`'s models table, or `docs/model_ops_guide.md`'s ops table.
- `_build_paged_prefill_meta`'s docstring still asserts `extend_count[t] = min(token_pos_in_chunk[t] + 1, win)`, which the vision path falsifies. The sibling docstring in `paged_prefill_indices.py` *was* updated.

On coverage: the PR's only tests are 116 lines in `tests/test_prefill_indices_paged.py`, which `pytest.skip(allow_module_level=True)`s without CUDA (`:15-19`), while `.github/scripts/run_unit_tests.sh` runs on a CPU-only runner. CI therefore executes **zero** tests for the 955-line vision module, the 173-line Triton kernel, the new `atom/multimodal/` package, or the scheduler guard — even though `image_visible_spans` and `image_aware_extend_window` are pure NumPy and testable on CPU.

Within that one GPU test: `_visibility_case` hardcodes `width = WIN + max_img` instead of calling `image_window_width`, and passes `seq_lens=None`, so the per-sequence span resolution the docstring credits with preventing cross-request visibility has zero coverage. And `test_visibility_lets_image_tokens_read_forward`'s `if start_idx <= t < end_idx` / `elif t < start_idx or t > end_idx` leaves `t == end_idx` — the one token whose validity comes from the `| is_end` disjunct — with no assertion.

---

## Cleanup, not blocking

- `_hash_topk` was not refactored onto the new `_routing_dest` it was extracted from; the two copies already differ in target device.
- `triton_mm_topk.py`'s `IS_HASH` branch is a near-verbatim copy of `triton_hash_topk.py` down to the `g > 20.0` softplus cutoff and the `1e-20` renorm floor, but with different `num_warps`.
- `_segment_attention` / `RMSNorm` / `Block` / `ViT` largely duplicate `kimi_k3_vl.py`'s tower.
- `image_block_perm` and `build_image_block` each compute the same pair-transpose order, and `build_image_block` returns a `perm` its only caller discards.
- `topK.mm_topk` is a 10-argument passthrough; `MultiModalInputBuilder` is dead code whose signature matches neither real builder.
- `safe_resize`'s `while` loop has no floor and can drive `solve_resize_ratio` into `math.sqrt` of a negative number from attacker-controlled image dimensions.
- `deepseek_v4_vl.py` is 955 lines against the 800-line soft ceiling. There is a clean seam — the dependency-free NumPy visibility math — that would also resolve the PIL import in #14.

PR #2150, opened 2026-09-07
docker: ROCm 10 base (pip SDK) + wheel release flow

zufayu:
## Summary

First cut of the ROCm 10 delivery (ticket: ATOM v0.1.7 for ROCm 10.1, GA 2026-10-05):

- **rocm10-base stage** in `atom_release.dockerfile`: ubuntu:24.04 + AMD stable wheel channel (TheRock pip distribution) instead of the rocm/pytorch apt image. Stable 10.0.0 main line (torch 2.11.0+rocm10.0.0, triton 3.8.0+git4cff872c), all versions ARG-ified for the 10.1 switch. Single multi-arch image keeps a device payload per GPU_ARCH entry.
- **Component adaptations**: AITER_USE_SYSTEM_TRITON=1, guarded torch 2.11 torch.Stream backport (ROCm/aiter#4817), MORI rocm_sysdeps registration, final stack tripwire (torch must stay +rocm10, no nvidia-*-cu* leakage, pip check clean).
- **atom-release.yaml** (new): ATOM wheel build (ATOM_MESH_BUILD=1, built inside the release image) + GitHub Release publishing. First wheel flow for ATOM.
- **docker-release.yaml**: rocm10 entry (stable main line, -rocm10 tag suffix, latest untouched) + rocm101-nightly tracking entry (nightly channel, build+golden-test only, nothing pushed).

## Version policy

- Main line: stable channel 10.0.0 (the combination sglang validated on their rocm1000 flavors)
- 10.1 tracking: nightly channel 10.1.0a<date>, exists to surface 10.1 breaks before GA; retire once 10.1 RC/GA lands on stable

## Test plan

- [x] `--build-arg BASE_IMAGE=rocm10-base` build on an MI355X (gfx950) node
- [x] golden output test (simple_inference) — 4/4 models pass on the ROCm 10.0 image
- [x] rocm101-nightly tracking run — 4/4 models pass on the 10.1 nightly image
- [ ] wheel build + GitHub Release draft

Functional results, both images, MI355X (gfx950), TP4:

| model | ROCm 10.0 | ROCm 10.1 nightly |
|---|---|---|
| MiniMax-M3-MXFP4 | PASS | PASS |
| GLM-5.2-MXFP4 | PASS | PASS |
| DeepSeek-V4-Pro | PASS | PASS |
| Kimi-K2.6-MXFP4 | PASS | PASS |

Pass means all four of: exit code 0; one completion per prompt, count > 0; no
"No available shared memory broadcast block"; "All EngineCores shut down" present.
These are checked explicitly rather than through the release golden test, which
passes unconditionally today (`diff ... || true` swallows the exit code — fixed in #2168).

Note for anyone reproducing on 10.1: that image has no `rocm-smi` (10.1 dropped
rocm_smi), so arch checks must go through
`torch.cuda.get_device_properties(0).gcnArchName` instead.

cc @zufayu

---

### Known: performance of the ROCm 10 line

Measured on an MI355X (gfx950) node, paired A/B on one node in one
sitting, same benchmark parameters, only the image differing. Control is a
clean `rocm/atom-dev` build (ROCm 7.2.4); subject is the ROCm 10.0 image this
PR builds. GLM-5.2-MXFP4, 1024/1024, tp4:

| concurrency | ROCm 7.2.4 | ROCm 10.0 | Δ |
|---|---|---|---|
| 4 | 384.6 tok/s | 301.2 tok/s | **-21.7%** |
| 8 | 584.0 | 534.9 | **-8.4%** |
| 16 | 1014.4 | 876.9 | -13.6% |
| 32 | 1556.2 | 1243.3 | -20.1% |

Run-to-run spread on these figures is 0.05-0.6% (three repeats per point,
80-160 requests each), so the gap is 14-70x the noise. DeepSeek-V4-Pro and
MiniMax-M3 show the same shape; on GLM and V4-Pro the gap closes at
saturation (c=64 lands within noise), while MiniMax stays ~7% down. TTFT is
unaffected or better, TPOT tracks the throughput loss — the shape points at
per-step overhead rather than slower kernels.

This does **not** isolate ROCm: the two images differ in ROCm, torch, triton,
ATOM (47 commits) and aiter all at once. Isolating it needs the same
ATOM+aiter built against both ROCm versions.

It is also not a reason to hold this PR. The ROCm 10 flavor is opt-in — bare
`latest` keeps pointing at the ROCm 7 build, so no existing user moves. The
numbers matter for the later, one-line PR that moves that pointer, and are
being tracked separately.

For completeness: the ROCm 10.1 nightly line built by the `rocm101-nightly`
flavor performs the same as 10.0 (all five concurrencies within 2.2%, four
within 0.6%), and passes the same functional checks on gfx950.


zufayu:
Ready for review — rebased onto current `main`, conflict resolved.

### What changed since the last review request

1. **Merged `main`** (the branch was `CONFLICTING`; it is `MERGEABLE` now).
   The only conflict was `docker/atom_release.dockerfile`: `main` added a Triton
   pin (#2173) at the same two spots where this branch adds its ROCm 10 blocks.
   Both sides are kept, and `main`'s pin is scoped to the ROCm 7.x line:

   ```dockerfile
   RUN if [ "${ATOM_BASE_IMAGE}" != "rocm10-base" ] && [ -n "${TRITON_PIN_VERSION}" ]; then \
   ```

   The ROCm 10 flavor installs Triton from the ROCm 10 SDK channel
   (`ROCM_TRITON_VERSION`) and already asserts `rocm10` is in the Triton
   version. Force-reinstalling the `rocm7.2.0` wheel there would break the
   stack and trip that assert. **Non-rocm10 builds are byte-for-byte unchanged.**

2. **Corrected the 10.1 tracking-line comment** — the example build-args did not
   match the combination actually validated on MI355X (torch `2.13.0`, not
   `2.12.0`; Triton `git4cff872c`, not `gitc01b6774`). Comment-only.

### About the one expected red check

`Accuracy (Kimi-K2.7-Code-MXFP4)` is expected to stay red. **It is red on
`main` too** — please don't hold the merge on it.

Sampling every `atom-test.yaml` run in the last 100 via the Actions API, that
job reached a terminal state 28 times today: **14 failures, 0 successes**,
across 8 distinct runners (healthy `mia1-p01-g33/g40/g41/g42` included) and
across nine branches plus `main` (`main` at 09-14 05:16). Reruns do not clear
it.

We also rebuilt the CI environment off-fleet and ran the CI workload end to
end — the exact CI image digest
(`rocm/atom-dev@sha256:cb6246f3325f3526dbf7a30a3144e63ba60ae3e373f6590231c768d5ca7488dc`),
this PR's head, the aiter wheel from the failing run, GSM8K 3-shot over
`/v1/completions`, 1319 prompts at concurrency 65:

**1319/1319 completed, 0 errors, 0 `Memory access fault`.**

That closes the last open variable (the base image). aiter, the ATOM commit
range, the model weights and the workload shape had already been ruled out by
experiment. Root cause is still unknown and we are not attributing it to any
change — but nothing in this PR touches the MXFP4 MoE path, and the check is
failing independently of it.

Happy to split anything out if that makes review faster.

zufayu:
CI update on `547497f90`: **26 green, 1 red.**

The two other accuracy jobs that were red on the previous head —
`GLM-5.2-MXFP4 MTP` and `Kimi-K3 DSpark` — both **pass** on this run. They were
runner-side flakes, as suspected.

That leaves 12 of 13 accuracy jobs green, and a single outstanding check:
`Accuracy (Kimi-K2.7-Code-MXFP4)`, the one that is also red on `main` and that
the comment above covers.

No code changes since the last comment — this is the same head, just the
completed run.

valarLip (approved):

PR #2151, opened 2026-09-07
feat(dspark): capture the Kimi-K3 draft block into a CUDA graph

whx-sjtu:
### `feat(dspark)`: capture the Kimi-K3 draft block into a CUDA graph

K3 was the one drafter declaring no capturable pass. Now declared as V4 is: one
two-part pass, backbone recorded, head plus sampler as epilogue. But K3's
block metadata is built, not inherited: the Triton index generator and
`get_mla_metadata_v1` take their work from the batch; neither may sit inside
a recording, so `_k3_build_block_metadata` builds it outside the capture into
buffers preallocated at `max_num_seqs`. Its pad tail carries one page at block 0
and slot `-1`, because a zero-page row underflows the split-KV stage1 kernel's
`full_pages` arithmetic.

### `fix(dspark)`: a real token id from the Markov argmax on a NaN row

`dspark_markov_argmax` tie-breaks by keeping the lanes equal to the maximum,
with `vocab_size` as the sentinel for the rest. A NaN row equals nothing, so the
sentinel survives both reduces and the op returns `vocab_size` — an id past the
`[V, r]` table `markov_w1` then gathers, aborting the queue with
`HSA_STATUS_ERROR_EXCEPTION`. Reachable on every start: `warmup_model()` runs
before `allocate_kv_cache()`, so warmup drafts against uninitialized memory read
as fp8 e4m3, where `0x7F` is NaN (4 of 14 rows all-NaN). Now answers 0,
as `torch.argmax` does. Shared with V4; finite input unchanged.

## Validation

Before:
image

image

After:
image

image

Kimi-K3 tp8, recipe defaults, full GSM8K at 64 concurrency:

| drafting | strict | acceptance |
|---|---|---|
| captured | 0.9613 | 53.9% |
| eager | 0.9682 | 54.1% |

Indistinguishable at ±0.005 stderr; per-position acceptance falls monotonely
from 90% to 20%, none near zero. A trace shows 216 of 216 draft passes
replaying. Startup: `origin/main` 0/3, this PR 4/4. 166 unit tests pass.


atom/spec_decode/dspark_proposer.py
```diff
@@ -625,200 +637,369 @@ def propose(
...
+        if n_pad:
+            borrowed = block_tables[scheduled_bs:, 0]
+            saved_page0 = self._blk_saved_page0[:n_pad]
+            saved_page0.copy_(borrowed)
+            borrowed.zero_()
+        kv_indices_generate_triton(
+            block_tables,
+            self._blk_kv_indices,
+            kv_indptr,
+            block_size,
```
ZhangLirong-amd: [P1] Kimi draft block metadata should use the MLA page size, not runner.block_size.

_build_paged_block_metadata() builds token-level KV metadata: kv_indptr is a cumsum of ctx_lens and kv_last_page_lens is all ones, which matches page_size=1. However the new graph path passes self.runner.block_size to get_mla_metadata_v1(page_size=...). In the documented Kimi DSpark launch we use --block-size 128 while ATOM_MLA_PAGE_SIZE defaults to 1, and the runtime MLA decode path later passes ATOM_MLA_PAGE_SIZE to the kernel. That means the persistent work metadata is planned for 128-token pages while the KV indices/indptr are laid out as page-size-1 metadata.

Please derive this from the attention metadata builder, e.g. use builder.block_ratio for kv_indices_generate_triton() and builder.block_size for get_mla_metadata_v1(page_size=...), or explicitly assert ATOM_MLA_PAGE_SIZE == 1 and pass page_size=1 here. Otherwise long-context Kimi DSpark graph replay can read the wrong KV ranges or consume stale indices.

whx-sjtu: The three arguments this points at are unchanged from `main` — the diff only relocates them. On `main` the same call sits at `dspark_proposer.py:770-771` with `page_size=block_size` / `kv_granularity=max(block_size, 16)`, and `kv_indices_generate_triton` already receives `runner.block_size`. `git diff origin/main...HEAD` shows no edit to any of the three, so this isn't behavior the graph path introduces.

On the failure mode: I don't think the over-read can happen. In `csrc/kernels/mla/metadata/v1_2_device.cuh` the work bounds are in **index-entry** units and are clamped by the indptr itself — `curr_kv_end = p_seqlens_kv_indptr[b+1]`, then `work_info.kv_end = min(kv_start + remain_kv_blocks * kv_granularity, curr_kv_end)` (L542-544) — independently of `page_size`. `page_size` feeds exactly one field, `work_info.kv_offset` (L545-549). And the decode kernel does its own page→token conversion from its **compile-time** `T::kPageSize` (`mi35x_v32_fwd_decode_m16x4_fp8_fp8.cuh:281-293`), which comes from the decode call's `page_size=envs.ATOM_MLA_PAGE_SIZE` (`attention_mla.py:2314`), i.e. 1 — not from what the planner was told. So the indices consumed stay inside the written region.

What is genuinely wrong is `kv_offset`: on any work item that does not end at the batch tail it comes out `block_size`× too large. Its only consumer is `causal_offset = max(qpos_off_from_last - kv_offset, 0)` (same file, L299), so an over-large value saturates it to 0 — no clamp — which is what this block wants, per the note already in `attention_mla.py:2290-2291` ("a non-causal block (DSpark drafts bidirectionally: every query legitimately sees every KV row, so there is no mask to place)"). Separately, `kv_granularity=128` vs the builder's `16` changes how finely a sequence is split across CUs, which is a load-balance question rather than a correctness one.

So it works today, but it is resting on that saturation rather than on the arithmetic being right, and you're correct that the derivation contradicts the builder's own convention (`self.block_ratio` for the generator at `aiter_mla.py:1034`, `page_size=self.block_size` / `kv_granularity=max(self.block_size, 16)` at `aiter_mla.py:2476-2480`, with `self.block_size = ATOM_MLA_PAGE_SIZE if > 1 else 1`). I'd rather fix it in its own PR against `main`: it is pre-existing code, and moving `kv_granularity` from 128 to 16 changes the draft decode's split plan, so it needs its own accuracy and performance run — folding it in here would invalidate the GSM8K numbers already posted on this PR.

ZhangLirong-amd (dismissed):

valarLip:
AI review (advisory). Posting only the must-fix set — findings that corrupt another request's KV or silently produce wrong drafts, with no error surfaced. Everything else from the pass is omitted here.

The first four are one class: **the paged block writes the KV it attends**, so a bad slot is not a bad read, it is a write into a page the row does not own.

---

### 1. `dspark_proposer.py:767` — a dummy run replays the previous step's slot mapping

`_build_paged_block_metadata` early-returns on `is_dummy_run` without refreshing any paged buffer, but `_propose_with_draft` now calls `self.block.run(running_bs, **staged)` unconditionally, and `DraftGraph.is_captured` deliberately ignores `is_dummy_run`.

`ModelRunner.dummy_execution()` (`model_runner.py:1192`) builds `is_dummy_run=True`, `total_seqs_num_decode=1`, `seq.block_table=[0]`, and it `produces_output()` — so `postprocess` → `propose_draft_token_ids` → `_propose_with_draft` runs. The early return leaves `_blk_slots` / `_blk_ctx_lens` / `_blk_kv_indptr` / `_blk_kv_indices` and the ps work descriptors holding the **last real step's** values (only `_blk_positions` is overwritten, above the return). `running_bs` is always a warmed capture rung, so the graph replays and `fused_qk_rope_concat_and_cache_mla` writes the dummy's draft KV through real requests' slots.

Eager sub-case (`ATOM_DRAFT_CUDAGRAPH=0`, or an unwarmed rung) is worse: `attn_metadata` still describes the target's 1-seq dummy (`slot_mapping` of `mtp_k+1` rows, `cu_seqlens_q` of 2) while the block runs `running_bs * T` query rows → out-of-range reads/writes.

This is new: on main, K3 had no graph, no staging and no padding, so nothing read those stale buffers.

Suggest gating on *"the paged pool is bound"* rather than on `is_dummy_run`.

### 2. `dspark_proposer.py:940` — the new clamp guards the gather, not the slot write

`page_idx.clamp_(max=block_tables.shape[1] - 1)` feeds the **slot write**, not just the read.

Anchor near `max_model_len - 1`, `T=7`, `block_size=64`: `block_positions = anchor+1 .. anchor+7` divide to a column `>= block_tables.shape[1]`; the clamp pins them to the last column, whose entry for that request was never written (0 by allocation, or a stale id). Both branches then do

```python
slots.copy_(gather(block_tables, 1, page_idx))
slots.mul_(block_size)
slots.add_(remainder(block_positions, block_size))
```

— the offset uses the **original** position, so it wraps to a small in-page offset. `attn_metadata.slot_mapping = slots.view(-1)` and the block scatters `T` rows of draft latent KV into page 0 (or another request's page) at that offset.

The docstring's *"costs nothing real … their draft tokens cannot survive verification"* covers only the read path. These rows want slot `-1` — the same drop sentinel the pad tail three lines away already uses.

### 3. `dspark_proposer.py:816` — `ctx_lens` is not bounded by the same limit

`ctx_lens.copy_(global_lens)` with `global_lens = anchor_positions + (1 + T)` is never clamped to the block table's width, so `kv_indptr` promises more token indices than `kv_indices_generate_triton` actually wrote.

For the same near-`max_model_len` row that motivated finding 2, `kv_indices_generate_kernel` masks its store with `store_mask = load_mask[:, None] & ...` where `load_mask = block_cols < n_cols` (`block_convert.py:172-179`) — the trailing entries are simply **not written**. `kv_indptr[1:] = cumsum(ctx_lens)` still counts them, so the MLA decode dereferences whatever `_blk_kv_indices` held from a previous step: a different request's token slots, or zeros on the first step. No OOB signal.

This is the other half of the fix in finding 2: clamp `ctx_lens` to `n_cols * block_size` (local units under DCP).

### 4. `dspark_proposer.py:852` — the block-table borrow has no `try/finally`, and leaves two disagreeing descriptions of a pad row

The pad tail zeroes column 0 of the runner-owned `forward_vars["block_tables"].gpu`, runs the Triton index generator, then restores.

**(a)** Any raise between `borrowed.zero_()` and `block_tables[scheduled_bs:, 0].copy_(saved_page0)` — a first-seen Triton grid failing to JIT, an OOM, an assert — leaves rows `[scheduled_bs, running_bs)` of the **target's** block table permanently pointing at page 0. The comment itself notes *"a row that is padding now may be real next step"*: that request then reads page 0 as its own first page. Silent cross-request KV, no error.

**(b)** Even on the success path the two descriptions of a pad row disagree — `block_tables` says the restored (stale) id, `kv_indices` says page 0. Any decode that addresses through `block_tables` rather than `kv_indices` (`triton_shuffle_mla_decode_fwd` takes `attn_metadata.block_tables`, `attention_mla.py:2145`) reads a possibly freed or reassigned page.

A reserved dummy page id, or a proposer-owned block-table buffer, avoids mutating another component's live state every step.

---

### 5. `dspark_proposer.py:671` — the "persistent kernel" premise the capture rests on is never asserted

The soundness argument for capturing at a `1 + T` context is *"the MLA decode runs the PERSISTENT (ps=1) kernel"*, but nothing checks it. `should_use_persistent_mode` (`attention_mla.py:536`) is False for `ATOM_MLA_PAGE_SIZE > 1`, for DP without `dpa_persistent_supported`, and for DCP on anything but gfx950.

MI300X (gfx942) with `--dcp > 1`: `dcp_persistent_supported()` returns `get_gfx() == "gfx950"` → False, so `use_persistent_mode` is False and `_forward_decode` (`attention_mla.py:2245-2277`) drops `work_meta_data` / `work_indptr` / `reduce_*` to `None` and runs the split-KV path with a fixed host `num_kv_splits = max(1, 16 // dcp_world_size)`. The descriptor buffers that were supposed to carry the per-batch context length no longer reach the kernel, so a recording made over a `1 + T` token context is replayed against 100k-token ones — silently wrong drafts (acceptance collapse), no crash.

Either assert persistent mode before capturing, or warm at a representative `max_seqlen_k`.

---

Everything above is from reading the diff plus the surrounding call sites; none of it was reproduced on hardware, so please sanity-check the reasoning before acting on it. The remaining findings (warmup coverage, the NaN sentinel's sweep gap and altitude, stale docstrings, and the missing tests for the new paged-block helpers) are lower severity and left out of this comment.

whx-sjtu:
Four of the five are fixed in 7af12211, with six regression tests (each checked to fail on 4c720ef7 and pass after). Two I'm not changing, with reasoning below.

### 4(b) — not changing

The consumer named isn't reachable for this model. `attention_mla.py:2146` is on the Triton MLA decode, which needs `ATOM_USE_TRITON_MLA=1` **and** `ATOM_USE_TRITON_MLA_SHUFFLE_KV=1` (both default `0`), and enabling that pair asserts `--block-size 64` (`aiter_mla.py:301-305`) while Kimi-K3 runs `--block-size 128`. That path also hardcodes `causal=True`, which a bidirectional block cannot use. The aiter persistent path this block does run addresses through `kv_indices`, not `block_tables`. Worth stating as an invariant, but there's no live reader to break.

### 5 — not changing, and I'd want evidence before adding the assert

The gate reads as you describe, but `dpa_persistent_supported` currently returns `True` unconditionally, so DP can't break it. That leaves `ATOM_MLA_PAGE_SIZE > 1` and `DCP>1 on non-gfx950` — and `recipes/Kimi-K3.md:5` says "This guide targets **AMD MI355 (gfx950) only**, `-tp 8`", with no DCP flag anywhere in it.

The stated consequence also looks stronger than the code supports. On the non-persistent path the decode still reads `kv_indptr` / `kv_indices` / `context_lens`, and only their *addresses* are baked into the recording — `_build_paged_block_metadata` refreshes the contents every step. `num_kv_splits` is a constant host int (`max(1, 16 // dcp_world_size)`). So the per-batch context length does still reach the kernel; I can't see what the recording would have frozen.

I'm happy to add the assert if you'd rather have the premise enforced than documented, but as written it would refuse a configuration I can't demonstrate is broken. If you know of a shape the non-persistent split-KV path takes from the capture rather than from the descriptors each step, say which and I'll assert on it instead.

whx-sjtu:
Full GSM8K re-run on 7af12211, same server config as the numbers posted earlier (TP8, DCP1, fp8 KV, block 128, T=7, `--cudagraph-mode FULL`, `ATOM_DRAFT_CUDAGRAPH=1`, max-num-seqs 32), 5-shot, 64-concurrency, all 1319 problems, no `--limit`:

| filter | 4c720ef7 (before) | 7af12211 (after) |
| --- | --- | --- |
| flexible-extract | 0.9606 ± 0.0054 | 0.9583 ± 0.0055 |
| strict-match | 0.9613 ± 0.0053 | 0.9583 ± 0.0055 |

Down 0.23 and 0.30 points, against a standard error on the difference of about 0.8 points — so roughly 0.3σ, which is what I'd expect given none of the four fixes changes behaviour on this workload: the DP dummy path needs DP>1, the two block-table bounds need a request within T tokens of `max_model_len`, and the `finally` only runs on a raise. Zero failed or retried requests, and no errors or asserts in the server log.

Draft acceptance is healthy and no position is starved — cumulative 57.5% at 5.03 tokens/forward, distribution `{0: 8.9%, 1: 10.7%, 2: 12.3%, 3: 11.7%, 4: 11.0%, 5: 10.0%, 6: 9.5%, 7: 25.9%}`.

valarLip (approved):

PR #2154, opened 2026-09-07
feat(kimi-k3): support P/D disaggregation

wanzhenchn:
## Motivation

Kimi-K3's KV cache is in two halves and the MLA base class describes one: the 24 full-attention layers live in the paged pool the base class walks, the 69 KDA layers in the slot-indexed recurrent-state pool that `build_kv_cache_tensor` tags `per_request_state=True`. A request's KDA state is not derivable from its MLA blocks.

Under P/D disaggregation the transfer carried only the paged half, so decode ran 69 of 93 layers from a zeroed recurrent state. **That still decodes fluent text, so nothing downstream could tell** — not the connector, which transfers whatever regions it is handed and reports success either way, and not GSM8K.

Six distinct failures sat behind that one symptom:

1. `get_kv_transfer_tensors()` returned only the base class's block regions, so the KDA state pool was never registered for transfer at all.
2. The producer skipped Phase 2 whenever either slot index was negative and returned success — but Phase 1 had already moved the block regions by then.
3. Prefix checkpointing can move the producer's state slot after admission, so the admission-time slot names the wrong bytes.
4. The producer left deferred output on, so the consumer processed the handoff token twice and the KDA recurrent state advanced twice.
5. The producer still speculated: it padded every sequence with `mtp_k` placeholders and sized its decode window to `mtp_k + 1`, holding `num_tokens` `mtp_k` short of `max_tokens` so the handoff never fired and the producer kept decoding a request whose drafts do not exist.
6. Decode shapes did not match the scheduled batch, so MLA query width and the staged decode token range could disagree.

### Engine

**Hybrid cache registration** — `atom/model_ops/attentions/kimi_mla_gdn_attn.py`, `gdn_attn.py`

- `_KimiMLAGDNCommon.get_kv_transfer_tensors()` joins both halves: the MLA base's block regions plus the KDA slot regions.
- Refuses a backend that cannot carry the slot half.
- Joined in the hybrid class, not `GDNStateMixin` — only a hybrid has a base answer to extend; pure-GDN backends own no block regions.
- `gdn_attn.py` gains the region geometry: a slot resolves to its own bytes, stops at its own layer's end, plus a region for the write cursor.

**Slot identity across the transfer** — `atom/kv_transfer/disaggregation/types.py`, `mooncake/mooncake_connector.py`

- `ReqMeta` gains `remote_slot_index` (producer source), kept distinct from `local_slot_index` — the consumer's allocation overwrites the latter with its destination.
- The producer publishes its final `state_slot` at finish time, not the admission-time slot, so a prefix-checkpoint move between admission and handoff no longer reads the wrong bytes.
- The TP path uses that final slot instead of whatever the request payload happened to carry.

**Transfer integrity** — `mooncake/mooncake_connector.py`

- A stateful transfer arriving without a final source slot raises instead of proceeding.
- Phase 2 fails closed when slot regions were registered but the transfer carries no slot.
- Scoped to the contradiction only — registered slot regions mean there is state to move — so backends with nothing slot-indexed registered keep the existing skip. Same shape as the pre-existing SWA guard.

**Backend and role resolution** — `atom/kv_transfer/disaggregation/factory.py`, `types.py`, `__init__.py`

- Two predicates replace `bool(kv_transfer_config)` checks that were wrong for K3's topology.
- `resolve_pd_backend()` names the backend that will actually transfer, unwrapping `multi` so `multi[mooncake + lmcache_offload]` resolves to mooncake, and returning `None` for an aggregated node whose only KV config is the local LMCache offload tier.
- `kv_config_has_producer()` recurses into `multi`; shared by the scheduler and the model runner, which branch on the same fact.

**Scheduling under P/D** — `atom/model_engine/scheduler.py`

- A P/D producer is excluded from speculation for the same reason `PrefillScheduler` is: it prefills, samples T0 and hands off, so it never proposes.
- The drafter stays loaded — the producer still writes the draft's context KV into the shared pool at prefill.
- The consumer's first decode is padded to the full verify window regardless of what the producer sent, so T0 lands at the window head instead of the slice reaching back into the prompt and verifying prompt tokens as drafts.
- A request whose producer sent no drafts keeps its empty `spec_token_ids`; the consumer proposes for itself.

**Forward shapes and output timing** — `atom/model_engine/model_runner.py`, `atom/utils/forward_context.py`, `atom/model_ops/attentions/aiter_mla.py`

- Deferred output disabled on prefill producers, so the consumer processes the handoff token exactly once.
- MLA query width derived from per-forward batch metadata rather than assumed.
- Falls back to eager when no exact CUDA-graph shape exists — declining the graph, not the batch.
- The deferred decode path returns the flat width it actually staged; slicing by the scheduler's original total could drop draft rows and leave attention metadata wider than `input_ids`.

## GSM8K accuracy

5-shot, `local-chat-completions`, full 1319/1319 samples, threshold **0.94**.

| Config | Eval conc. | strict-match | flexible-extract | stderr |
|---|---:|---:|---:|---:|
| 1P1D TP8 DCP1, DSpark-7 | 4 | **0.9492** | 0.9500 | ±0.0060 |


atom/model_engine/scheduler.py
```diff
@@ -3799,6 +3852,7 @@ def _schedule(self):
                 scheduled_spec_decode_tokens=scheduled_spec_decode_tokens,
                 cu_stream_fraction=self.cu_fraction,
                 state_maintenance_ops=self.block_manager.take_state_maintenance_ops(),
+                is_first_decode_without_local_prefill=first_decode_flags,
```
whx-sjtu, line 3855: Why we need this new param to override original `seq.is_first_decode for seq in seqs.values()`?

atom/utils/forward_context.py
```diff
@@ -308,7 +310,23 @@ def decide(
...
+        shape_captured = (
+            graph_shapes is None or (running_bs, max_seqlen_q) in graph_shapes
+        )
+        use_cudagraph = (
+            not is_prefill
+            and unified
+            and not enforce_eager
+            and on_ladder
+            and shape_captured
+        )
```
whx-sjtu, line 329: Why we need this? For Prefill nodes, do we still run with FULL Graph or just enforce eager?

atom/model_engine/scheduler.py
wanzhenchn: You're right — it was redundant, removed.

atom/utils/forward_context.py
wanzhenchn: FULL graph stays on — this declines the graph for one unrecorded shape, not for the batch and not for the node.
- Prefill batches were already excluded by not is_prefill, unchanged. This is about decode batches on a prefill node.
- A recording is keyed (running_bs, max_seqlen_q). A producer still has speculative_config set, so the drafter is built and capture records only (bs, mtp_k+1).

This PR excludes the producer from speculation (num_spec_step=0), so its decode forwards run max_seqlen_q=1. On the ladder by batch size alone, self.graphs[(bs, 1)] is a KeyError. graph_shapes=None (eager / PIECEWISE / pre-capture) keeps the old behavior exactly.

Yuechguo (approved):

PR #2157, opened 2026-09-07
feat: add GLM-5.3-Flash MTP support

cubezhang:
# feat: add GLM-5.3-Flash MTP support

GLM-5.3-Flash checkpoint layer 45 was previously discarded, so the model could
only serve without speculative decoding. This change registers a GLM-specific
MTP model, reuses the existing NextN predictor prologue, and constructs layer 45
with GLM's k-pool MLA and clamped SwiGLU MoE behavior.

The pooled indexer now preserves position-addressed key/gate history across
speculative verification and rejection, including ragged concurrent requests.
Draft graph staging handles GLM-5.3's two-dimensional NextN residual even though
the backbone config contains `hc_mult=4`. The existing GLM-5.2/DeepSeek MTP path
keeps its original decoder behavior through the default predictor-layer hook.

## Validation

- Black formatting and Ruff checks pass for the changed MTP implementation and tests.
- 61 focused GLM/MTP tests pass, including the shared prologue, input masking,
  top-k sharing, decode contracts, scheduling, GLM routing, and k-pool behavior.
- The final physical KV-slot mapping kernel test passes on ROCm.
- A standalone image built directly from `upstream/main` plus this commit
  completed real-model TP8 MTP3 text serving and the 8/8 quality smoke suite on
  8× MI308X.
- MTP1/MTP3 natural acceptance lengths 0–3, forced rejection, and concurrent
  short/long requests were validated during bring-up.

## Measured performance

On a fixed natural-language workload with TP8, 512 output tokens, and BF16
KV/index cache:

| Configuration | Concurrency | Output tok/s | Change vs no MTP |
| --- | ---: | ---: | ---: |
| No MTP | 1 | 104.67 | — |
| MTP3 | 1 | 181.54 | +73.44% |
| No MTP | 8 | 409.96 | — |
| MTP1 | 8 | 559.98 | +36.59% |

This branch is based directly on `upstream/main` (`5a9c206`).


yitingw1:
Please fix conflicts

cubezhang:
Okay,let me check.

yitingw1:
Nice work — the k-pool history design is clearly reasoned and the new code stays contained in two new modules, so DeepSeek and GLM-5.2 keep their existing paths. Reviewed at `e67793a8`. A few things worth addressing before merge.

## 1. `indexer.py:146` — the first dispatch clause never fires, and would be wrong if it did

`num_spec_decodes` is only defined on `GDNAttentionMetadata` (`forward_context.py:729`, `gdn_attn.py:72`), but line 134 reads `fwd.attn_metadata`, which is the MLA-side `AttentionMetaData` — so `getattr(..., 0)` always falls back to 0 and the first clause never fires. The ragged-query condition is doing all the work, which makes the "Prefer explicit scheduler metadata" comment misleading about which path is actually taken.

More importantly, that clause carries no `not context.is_prefill` guard. If some metadata does start exposing `num_spec_decodes`, a prefill batch would be routed into `run_speculative_kpool_indexer` and bypass the prefill path below. Hoisting the guard makes it correct either way:

```python
is_speculative_verify = not context.is_prefill and (
    getattr(attn_metadata, "num_spec_decodes", 0) > 0
    or attn_metadata.max_seqlen_q > 1
)
```

## 2. `kimi_mla_gdn_attn.py:200` — the ring-size bound is the correctness floor but has no derivation or test

If the ring is too small, a rejected suffix overwrites history residues that are still going to be read, and the symptom is a silent scoring error rather than a crash. Right now the bound rests on one comment ("Two speculative windows plus the incomplete pool survive rejection"), and it has no direct test even though it is a pure function of `(pool, num_speculative_tokens)`.

Could the reasoning go in the docstring — specifically which positions are still live at which point — and is `next_pow2` required by the ring addressing or just a performance choice? Worth noting the capacity cost in the PR description too: at `pool=4, k=3` the per-request k-pool tail goes from 4 to 16 rows, which the perf table doesn't mention.

## 3. `speculative.py:300` — chunk scratch is sized from `max_model_len`, not the batch

`max_pools` comes from `max_model_len` rather than `metadata.max_seqlen_k`, so the per-chunk `logits` buffer is sized for the worst case regardless of actual request lengths — at `max_model_len=128k, pool_size=4` that is 16 MiB per 128-row chunk, reallocated for every chunk, every indexer layer, every step, plus the `.contiguous()` copy of the gathered `block_tables`.

The comment says to optimize only after parity, so this isn't blocking. Is the static bound deliberate (fixed shapes for cudagraph capture), or just the simplest thing that worked? If the latter, sizing from `max_seqlen_k` and hoisting `logits` / `selected_pools` into preallocated buffers would cut both the footprint and the per-step allocation churn.

## 4. `eagle_proposer.py:562` — unrelated debug-path change

This early `return` (and the matching warmup skip in `drafter.py:229`) changes the semantics of the `ATOM_DEBUG_FORCE_SKIP_DRAFT_MODEL` diagnostic: previously it filled `-1` and continued through the rest of `propose`, so any metadata side effects further down still ran. The `drafter.py` side has a clear rationale in its comment, but neither change relates to GLM-5.3 MTP support. Could it be called out in the description, or split out? It's easy to miss a debug-path behavior change inside a PR titled "add GLM-5.3-Flash MTP support".

## 5. Test coverage — none of the new logic is reachable from non-GPU CI

Both new test files skip at module level on `AITER imports require a visible ROCm device`, and 3 of the 5 cases in `test_glm5_kpool_speculative.py` additionally require `torch.cuda.is_available()`. Some of it doesn't need a GPU: `get_query_request_indices` is pure `torch.bucketize` semantics, and `_kpool_history_size` has no test at all today. Splitting those into CPU-runnable cases would let Pre-Checkin catch regressions in exactly the parts that fail silently.

By the way, please fix CI UT failures.

cubezhang:
Thanks for the detailed review. I’ll push the updates. Thanks again for your careful review.

cubezhang:
Sorry, I’m still getting familiar with the repository and missed some validation steps earlier.

yitingw1:
Please fix conflicts

yitingw1:
No need to merge main frequently. It seems that CI has some err now.

cubezhang:
Got it.

yitingw1 (dismissed):

atom/spec_decode/eagle_proposer.py
```diff
@@ -139,9 +139,13 @@ def _declare_draft_graphs(self):
...
         draft_hf = self.speculative_config.draft_model_hf_config
         # DeepSeek-V4 carries the mHC residual, so its hidden is [N, hc, dim]
-        # rather than [N, dim]. `hc_mult` is absent on every architecture that
-        # does not, which is exactly the two-dimensional case.
-        hc = getattr(draft_hf, "hc_mult", None)
+        # rather than [N, dim]. GLM inherits hc_mult from its backbone, but
+        # its NextN block carries an ordinary two-dimensional residual.
+        hc = (
+            None
+            if draft_hf.architectures[0] == "Glm5NextMTPModel"
```
valarLip: any other way to keep this place clean enough?

valarLip (changes requested):

atom/spec_decode/eagle_proposer.py
cubezhang: Thanks for the suggestion. I moved the draft-graph shape and buffer-reuse contracts onto the MTP model, removing the GLM architecture-name check from EagleProposer. I also added regression coverage for GLM, DeepSeek, and DeepSeek-V4.

valarLip (approved):

PR #2160, opened 2026-09-08
[ATOM Plugin CI] Fix the K3 DSpark MLA head-width, its breakable-cudagraph eager break, and the fused Markov argmax NaN escape

PerryZhang01:
Three real defects in the plugin's Kimi-K3 path — two in MLA, one in the fused DSpark Markov sampler — plus a threshold line. **With all three, the `Accuracy (Kimi-K3 DSpark DCP8 TP8)` cell passes: four consecutive runs of its own CI configuration score 0.9484 / 0.9515 / 0.9560 / 0.9598, every one of them 1319/1319 with no fault.** It has been red since it was added in #2033.

## Scope

Fixes 1 and 2 alone do **not** make the cell pass — the illegal access (HSA 0x1016) on the trailing decode batch survives them, which is what an earlier revision of this PR reported as an unexplained fault outside its scope. That fault is now identified and fixed: it is fix 4 below, an out-of-range token id returned by the fused Markov argmax on NaN rows. Fixes 1 and 2 remain independently correct and are kept; one of them restores an upstream contract the plugin was violating.

Still open, and **not** addressed here: whatever produces the NaN draft logits in the first place. Fix 4 makes the op crash-safe and contract-correct; it does not fix the NaN. It is now localised, though — to the draft's layer-0 `gate_up_proj` fp8 GEMM, which emits ~1.6e38 from unchanged, bounded inputs on a trailing batch. See fix 4.

## 1. MLA metadata builder used the target's head width — `metadata.py`

`AiterMlaMetadataBuilderForVllm` read its query-head count from `model_config.hf_config`, which always describes the **target**. A DSpark draft owns its own KV cache group and its own builder, so the draft planned persistent work descriptors for K3's **96** heads while its MLA decode dispatched at **64** — the kernel then indexes DCP partials nobody planned.

Now read off the layers the builder was handed (as ATOM native's `DSparkProposer` does), with `min_query_heads` passed through to `mla_dcp_kernel_num_heads`. Verified: draft plans and dispatches at 64, target stays 96.

## 2. MLA was missing the breakable-cudagraph eager break — `ops.py`, `layer_mla.py`

K3 runs on vLLM's breakable cudagraph path (`VLLM_USE_BREAKABLE_CUDAGRAPH`, compilation mode `NONE`, no FX splitting), where **only ops carrying `@eager_break_during_capture` re-execute eagerly on replay**. Upstream applies it to every attention entry point — `unified_mla_attention_with_output`, `unified_attention_with_output`, `kimi_k3/nvidia/mla.py::_attention`, the KDA mixers — and K3's AMD path has no `mla.py` upstream, because the plugin's `atom_vllm_mla_attention` is what stands in for it. It was the one such op without the decorator.

The op now takes `output` from the caller (`mutates_args=["output"]`) instead of allocating it — the decorator's documented requirement, since a tensor allocated inside would land at a new address each replay — and delegates to a decorated helper that re-fetches metadata via `_get_layer_context` inside the break.

**Effect is limited to PIECEWISE batches.** The decorator short-circuits when `get_forward_context().cudagraph_runtime_mode == CUDAGraphMode.FULL`, so under `FULL_AND_PIECEWISE` the uniform-decode batches are unaffected by this change.

| | breakable-cudagraph capture | server |
|---|---|---|
| before | `graphs=70, eager_breaks=69` (KDA only) | died at 280s, illegal access |
| after | `graphs=94, eager_breaks=93` (one per layer) | reaches ready |

93 = 69 KDA + 24 MLA, predicted before the run. Both rows use the same launch arguments, so the comparison holds; but see the measurement caveat below.

## 4. The fused Markov argmax could return an id one past the vocab — `dspark_markov_sample.py`

This is the cause of the illegal access described under Scope above.

`dspark_markov_argmax` picks the winning id in both reduce stages with
`tl.where(vals == max, id, vocab_size)`, where `vocab_size` is the tie-break
sentinel that loses the following `tl.min`. No NaN lane satisfies `==`, so on a
row whose biased logits carry a NaN nothing displaces the sentinel and **the op
returns 163840 — one past the table**.

The caller does not treat that value as data. `wrapped_sample_sequential` feeds
each sampled id straight back into the next block position's `markov_w1`
gather, an `nn.Embedding(163840, 256)`, so the return is an out-of-bounds read
on a device queue: no exception, just an asynchronous
`HSA_STATUS_ERROR_EXCEPTION 0x1016` attributed to whatever launch happens to be
in flight. That matches every symptom the cell has shown — the fault lands on a
tail batch, at a different launch each time, and **it does not fire every
run** — whether a NaN row lands in the trailing batch at all depends on which
requests share it, which is why the cell has looked flaky rather than broken.
It also explains why both unfused
spellings complete the same run (`torch.argmax` cannot return an out-of-range
index) while clamping the ids merely hides it.

Both stages now carry NaN candidates alongside the `==` reduce and prefer them,
reproducing `torch.argmax`: NaN orders above every number, lowest index wins
ties, an all-NaN row returns 0. The store additionally clamps to
`vocab_size - 1` — nothing above can reach the sentinel any more, so that only
bounds a future reduce that regresses the property, somewhere the cost of being
wrong is an illegal access rather than a bad token.

| | this cell, CI arguments |
|---|---|
| before, fused (7 runs) | illegal access on the tail batch in **6 of 7**; the seventh completed, 0.9469 |
| before, `ATOM_DSPARK_FUSED_MARKOV_SAMPLE=0` | 0.9545, 1319/1319 |
| before, `ATOM_DSPARK_MARKOV_ARGMAX_UNFUSED=1` | 0.9568, 1319/1319 |
| **after, fused (4 runs)** | **0.9484 / 0.9515 / 0.9560 / 0.9598, 1319/1319 each** |

`tests/model_ops/test_dspark_markov_sample.py` pins both halves of the contract
— ids equal to `torch.argmax`, and ids inside `[0, V)` — for NaN arriving
through each of the three inputs. Its four NaN cases fail against the pre-fix
kernel and pass after; the other nine pass either way.

**Not fixed here: where the NaN comes from.** A NaN reaching this op means the
draft logits are already garbage. That is a separate defect, still open, and
this PR does not fix it — but it is no longer unidentified. Device-side counters
along the draft's path (per site: NaN rows, inf rows, and a running max finite
magnitude; three reductions, no host sync, so acceptance is left alone — 4.27 /
4.34 / 4.83 against a 4.29-4.78 unprobed baseline) localise it to a single GEMM.

- **Whole rows, not stray elements.** `base_nan_elems` came to `1109 x 163840`
  exactly — 1109 complete logit rows, no remainder.
- **Born inside the draft, not inherited from the target.** At the burst the two
  tensors crossing over from the target — the concatenated auxiliary hidden
  states and the projected context — carry 0 NaN rows, while the draft output
  carries 1109.
- **It appears in one step, inside layer 0's dense MLP, with that MLP's inputs
  unchanged.** Splitting the MLP into its four stages, one rank:

  | step | `r0` residual | `m0_xs` act scale | `m0_gu` gate_up out | `m0_act` SiLU out | `m0_dn` MLP out |
  |---|---|---|---|---|---|
  | 2380 | ~308 | ~0.0121 | ~11 | ~40.2 | ~1.68e+03 |
  | 2400 | ~308 | ~0.0121 | **~1.61e+38** | ~3.12e+38, 441 inf rows | 533 NaN + 961 inf rows |

  Both MLP inputs are flat across all 2460 steps; `gate_up_proj`'s output moves
  37 orders of magnitude at step 2399. A GEMM cannot amplify a bounded input
  that far, so the fp8 `gate_up_proj` GEMM itself is producing the value. On two
  of the eight ranks the NaN/inf shows up directly in that GEMM's output
  (156 NaN + 15 inf rows, first at step 2408) rather than one stage later.
- **The rest follows mechanically.** `SiluAndMul`'s multiply overflows the
  huge-but-finite output to inf; `down_proj`'s per-token quantisation of a row
  containing inf gives that row scale inf, which makes the GEMM emit the
  *whole* row as NaN. `h0` equals `m0_dn` exactly, and `r0` — which already
  carries the attention output — stays at 0 NaN rows, so attention and the
  target side are both clean. ATOM's own three fp8 quant Triton kernels all
  guard the reciprocal (`tl.where(scale > 0, 1/scale, 0)`), so they are not the
  source either.
- **It bursts on trailing batches**, first bad step 2399 of ~2460 — the same
  population as the illegal access this PR fixes.

What is left open is why that GEMM produces 1.6e38 from bounded inputs. Which
kernel it is can be read off the dispatch without running anything: the layer is
`(N=3584, K=7168)` per rank, all three Aiter fp8 kernels need
`VLLM_ROCM_USE_AITER` (default off, and this cell sets no `env_vars`) and in any
case have no tuned entry for that shape, and the ROCm/per-tensor kernels cannot
take per-token x per-channel scales — which leaves
`RowWiseTorchFP8ScaledMMLinearKernel`, i.e. `torch._scaled_mm` rowwise on
hipBLASLt. That derivation is static; confirming it at runtime is the next step.
Across all 8 ranks that GEMM's output splits into three groups, which says
something about the mechanism:

| ranks | max finite magnitude | NaN+inf rows in the GEMM output |
|---|---|---|
| 4 of 8 | 1.61e38, byte-identical on all four | none — the overflow happens downstream in SiLU |
| 3 of 8 | 3.39e38, the largest finite bf16 | yes, from step 2408 |
| 1 of 8 | 10.5 — never goes bad | none |

An earlier revision read the 3.39e38 as reinterpreted bytes. That is
**withdrawn**: random bit patterns do not land on the same value on three ranks,
nor reproduce 1.61e38 byte-for-byte on four. The 3.39e38 is better read as a
saturating fp32->bf16 cast, so on those three ranks the true magnitude is
unknown; the 1.61e38 is an ordinary representable bf16 and is the true value.

What the value cannot be is correct arithmetic. The draft ships bf16 weights and
is quantised online (`ptpc_fp8`), so the per-channel scales come from the
checkpoint: layer 0's `gate_proj` and `up_proj` have **no zero rows** and
per-channel row maxima in [0.147, 1.445], with **no outlier among the 8 TP
shards**. With |W| <= 1.45, K = 7168 and the measured activation scale 0.0121
(so |x| <= 0.0121 * 448 ~ 5.4), the output is bounded by ~5.6e4 — and the
observed normal magnitude is ~11. 1.61e38 exceeds that bound by 33 orders of
magnitude, so whatever enters the multiply is not the intended operands. The
uniformity across shards also rules out a per-shard weight or scale explanation
for the 4/3/1 split.

The single clean rank is also informative: its `down_proj` output and layer-0
residual counters match the other seven exactly, because `down_proj` is
row-parallel and all-reduces the poison in. So the origin is on 7 of 8 ranks,
not all of them, and the 1.61e38 group bursts first (step 2399) with the 3.39e38
group following at 2408.

None of this affects the numbers above — drafts are verified by the target, so a
bad draft costs acceptance rate, not accuracy; the probed run scored 0.9553.

## 3. Threshold 0.95 -> 0.94

Full 5-shot gsm8k, 1319 questions, concurrency 64, 8xMI355X, **the cell's own
launch arguments** (default capture buckets, `FULL_AND_PIECEWISE`, TP8 + DCP8,
7 speculative tokens). The gate reads `exact_match,flexible-extract`
(`.github/scripts/atom_oot_test.sh:466`), so that is the column below.

**Four consecutive runs of this PR's code, same arguments, nothing varied:**

| run | flexible-extract | strict-match | completed | MTP acceptance |
|---|---|---|---|---|
| `nanfix_Q` | **0.9484** | 0.9469 | 1319/1319, no fault | 54.0%, len 4.78 |
| `nanfix_O` | 0.9515 | 0.9515 | 1319/1319, no fault | 50.8%, len 4.55 |
| `nanfix_P` | 0.9560 | 0.9560 | 1319/1319, no fault | 54.6%, len 4.83 |
| `nanfix_N` | 0.9598 | 0.9598 | 1319/1319, no fault | 52.5%, len 4.67 |

min 0.9484, max 0.9598, **spread 0.0114**, mean 0.9539.

**`nanfix_Q` is the load-bearing row: a fully correct run — all 1319 questions
answered, zero faults — that scores below 0.95.** A 0.95 gate does not merely
sit close to the measured spread of a correct build; it has been observed to
fail one. That is a direct observation, not an inference from a standard error.

**0.94 clears the lowest observation by 0.0084** — about three quarters of the
observed spread — and still catches any regression worth catching. If a later
run comes in under 0.94 the honest response is more data, not another nudge;
the neighbouring cell above already runs at 0.93.

Why the runs differ at all: lm_eval decodes greedily and speculative drafts are
verified by the target, so every row is measuring the same model — draft-side
differences move throughput, not answers. The acceptance column bears that out,
and rules out "the draft got better/worse" as the explanation: it spans
50.8-54.6% and does **not** track the score (P accepts the most at 54.6% and
scores 0.9560; Q accepts 54.0% and scores lowest; N accepts the second least
and scores highest). What is left is the target's own run-to-run numerics.

Per-run lm_eval standard error is +-0.0054 to +-0.0061. That is the error on one
measurement and cannot be turned into a probability of tripping the gate; the
four repeats above are the direct evidence, and they are what the threshold is
set from.

For reference, ATOM native (TP8 + DCP8 + 7 spec tokens) measures 0.9500 — 0.95
would sit exactly on the value the reference implementation produces.

**Provenance caveat.** These run directories did not snapshot a source hash at
launch. What supports them is after the fact: the op file's mtime (10:49:44)
precedes every server start, its md5 matches the committed blob, and each
server log self-attests that neither bypass was active (`argmax REPLACED` 0
lines, `fused Markov` present, no `ATOM_DSPARK_FUSED_MARKOV_SAMPLE=0`). Strong,
but recorded post-hoc — worth fixing in the harness, not worth re-running four
hours of gsm8k over.

**Withdrawn.** An earlier revision argued for 0.94 from two runs that differed
by 0.0046 and were launched with `cudagraph_capture_sizes=[8,16,32,64,128]`
plus `--async-scheduling` and `--load-format fastsafetensors`, none of which the
CI cell sets — restricting the buckets leaves tail batches (40/48/56) without a
FULL graph, so they never exercise the failing path. That argument was
off-configuration and the table above replaces it. Also withdrawn: the
intermediate revision that argued from "the lowest correct run clears 0.95 by
only 0.0015" — `nanfix_Q` makes that margin argument unnecessary.

For context, three earlier runs under the same CI arguments but with a
draft-side variation instead of the fix: 0.9477 (clamp workaround), 0.9545
(`ATOM_DSPARK_FUSED_MARKOV_SAMPLE=0`) and 0.9568
(`ATOM_DSPARK_MARKOV_ARGMAX_UNFUSED=1`). Together with the four above they span
0.9477 to 0.9598. They come from a parallel investigation
(`pr2160_evidence_summary.txt`); the clamp row in particular changed the
acceptance rate and recorded no per-run source hash, so treat them as context
rather than as part of the four-run measurement above.

A fourth run in that group, at 0.9469, was previously listed here as a second
clamp workaround. It was not: its helper's clamp function is defined and never
called, so that run measured the unpatched build. It is counted instead as the
one pre-fix run that completed, in section 2.


PerryZhang01:
### Accuracy run artefacts — `Accuracy (Kimi-K3 DSpark DCP8 TP8)`

Run on 8xMI355X with the cell's own launch arguments, this branch at `f989d7381` (vision fix applied as the identical local patch at measurement time), aiter = CI latest-main wheel `amd_aiter-0.1.1.dev1+g7d2f6a51a`.

**`exact_match,flexible-extract` = 0.9598** (gate: 0.94) · 1319/1319 · no fault in the server log.

<details>
<summary>lm_eval summary table</summary>

```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9598|±  |0.0054|
|     |       |strict-match    |     5|exact_match|↑  |0.9598|±  |0.0054|
```

</details>

<details>
<summary>lm_eval results JSON (full, as written by the harness)</summary>

```json
{
  "results": {
    "gsm8k": {
      "name": "gsm8k",
      "alias": "gsm8k",
      "sample_len": 1319,
      "exact_match,strict-match": 0.9598180439727066,
      "exact_match_stderr,strict-match": 0.005409439736970532,
      "exact_match,flexible-extract": 0.9598180439727066,
      "exact_match_stderr,flexible-extract": 0.005409439736970532
    }
  },
  "group_subtasks": {},
  "configs": {
    "gsm8k": {
      "task": "gsm8k",
      "dataset_path": "openai/gsm8k",
      "dataset_name": "main",
      "training_split": "train",
      "test_split": "test",
      "fewshot_split": "train",
      "doc_to_text": "Question: {{question}}\nAnswer:",
      "doc_to_target": "{{answer}}",
      "unsafe_code": false,
      "description": "",
      "target_delimiter": " ",
      "fewshot_delimiter": "\n\n",
      "fewshot_config": {
        "sampler": "default",
        "split": "train",
        "process_docs": null,
        "fewshot_indices": null,
        "samples": null,
        "doc_to_text": "Question: {{question}}\nAnswer:",
        "doc_to_choice": null,
        "doc_to_target": "{{answer}}",
        "gen_prefix": null,
        "fewshot_delimiter": "\n\n",
        "target_delimiter": " "
      },
      "num_fewshot": 5,
      "metric_list": [
        {
          "metric": "exact_match",
          "aggregation": "mean",
          "higher_is_better": true,
          "ignore_case": true,
          "ignore_punctuation": false,
          "regexes_to_ignore": [
            ",",
            "\\$",
            "(?s).*#### ",
            "\\.$"
          ]
        }
      ],
      "output_type": "generate_until",
      "generation_kwargs": {
        "until": [
          "Question:",
          "</s>",
          "<|im_end|>"
        ],
        "do_sample": false,
        "temperature": 0.0
      },
      "repeats": 1,
      "filter_list": [
        {
          "name": "strict-match",
          "filter": [
            {
              "function": "regex",
              "regex_pattern": "#### (\\-?[0-9\\.\\,]+)"
            },
            {
              "function": "take_first"
            }
          ]
        },
        {
          "name": "flexible-extract",
          "filter": [
            {
              "function": "regex",
              "group_select": -1,
              "regex_pattern": "(-?[$0-9.,]{2,})|(-?[0-9]+)"
            },
            {
              "function": "take_first"
            }
          ]
        }
      ],
      "should_decontaminate": false,
      "metadata": {
        "version": 3.0,
        "model": "/models/moonshotai/Kimi-K3",
        "base_url": "http://127.0.0.1:8264/v1/completions",
        "num_concurrent": 64,
        "max_retries": 3,
        "tokenized_requests": false,
        "trust_remote_code": true,
        "config_source": "/opt/venv/lib/python3.12/site-packages/lm_eval/tasks/gsm8k/gsm8k.yaml"
      }
    }
  },
  "versions": {
    "gsm8k": 3.0
  },
  "n-shot": {
    "gsm8k": 5
  },
  "higher_is_better": {
    "gsm8k": {
      "exact_match": true
    }
  },
  "n-samples": {
    "gsm8k": {
      "original": 1319,
      "effective": 1319
    }
  },
  "config": {
    "model": "local-completions",
    "model_args": {
      "model": "/models/moonshotai/Kimi-K3",
      "base_url": "http://127.0.0.1:8264/v1/completions",
      "num_concurrent": 64,
      "max_retries": 3,
      "tokenized_requests": false,
      "trust_remote_code": true
    },
    "batch_size": 1,
    "batch_sizes": [],
    "device": "cuda:0",
    "use_cache": null,
    "limit": null,
    "bootstrap_iters": 100000,
    "gen_kwargs": {},
    "random_seed": 0,
    "numpy_seed": 1234,
    "torch_seed": 1234,
    "fewshot_seed": 1234
  },
  "git_hash": null,
  "date": 1790047148.4595535,
  "pretty_env_info": "PyTorch version: 2.10.0+rocm7.2.4.git3d3aa833\nIs debug build: False\nCUDA used to build PyTorch: N/A\nROCM used to build PyTorch: 7.2.53211\n\nOS: Ubuntu 24.04.4 LTS (x86_64)\nGCC version: (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0\nClang version: Could not collect\nCMake version: version 4.4.3\nLibc version: glibc-2.39\n\nPython version: 3.12.3 (main, Mar 23 2026, 19:04:32) [GCC 13.3.0] (64-bit runtime)\nPython platform: Linux-6.8.0-65-generic-x86_64-with-glibc2.39\nIs CUDA available: True\nCUDA runtime version: Could not collect\nCUDA_MODULE_LOADING set to: \nGPU models and configuration: AMD Radeon Graphics (gfx950:sramecc+:xnack-)\nNvidia driver version: Could not collect\ncuDNN version: Could not collect\nIs XPU available: False\nHIP runtime version: 7.2.53211\nMIOpen runtime version: 3.5.1\nIs XNNPACK available: True\nCaching allocator config: N/A\n\nCPU:\nArchitecture:                         x86_64\nCPU op-mode(s):                       32-bit, 64-bit\nAddress sizes:                        52 bits physical, 57 bits virtual\nByte Order:                           Little Endian\nCPU(s):                               256\nOn-line CPU(s) list:                  0-255\nVendor ID:                            AuthenticAMD\nModel name:                           AMD EPYC 9575F 64-Core Processor\nCPU family:                           26\nModel:                                2\nThread(s) per core:                   2\nCore(s) per socket:                   64\nSocket(s):                            2\nStepping:                             1\nFrequency boost:                      enabled\nCPU(s) scaling MHz:                   53%\nCPU max MHz:                          3300.0000\nCPU min MHz:                          1500.0000\nBogoMIPS:                             6600.02\nFlags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good amd_lbr_v2 nopl nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt tce topoext perfctr_core perfctr_nb bpext perfctr_llc mwaitx cpb cat_l3 cdp_l3 hw_pstate ssbd mba perfmon_v2 ibrs ibpb stibp ibrs_enhanced vmmcall fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local user_shstk avx_vnni avx512_bf16 clzero irperf xsaveerptr rdpru wbnoinvd amd_ppin cppc amd_ibpb_ret arat npt lbrv svm_lock nrip_save tsc_scale vmcb_clean flushbyasid decodeassists pausefilter pfthreshold avic v_vmsave_vmload vgif x2avic v_spec_ctrl vnmi avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq la57 rdpid bus_lock_detect movdiri movdir64b overflow_recov succor smca fsrm avx512_vp2intersect flush_l1d debug_swap\nVirtualization:                       AMD-V\nL1d cache:                            6 MiB (128 instances)\nL1i cache:                            4 MiB (128 instances)\nL2 cache:                             128 MiB (128 instances)\nL3 cache:                             512 MiB (16 instances)\nNUMA node(s):                         2\nNUMA node0 CPU(s):                    0-63,128-191\nNUMA node1 CPU(s):                    64-127,192-255\nVulnerability Gather data sampling:   Not affected\nVulnerability Itlb multihit:          Not affected\nVulnerability L1tf:                   Not affected\nVulnerability Mds:                    Not affected\nVulnerability Meltdown:               Not affected\nVulnerability Mmio stale data:        Not affected\nVulnerability Reg file data sampling: Not affected\nVulnerability Retbleed:               Not affected\nVulnerability Spec rstack overflow:   Not affected\nVulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl\nVulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization\nVulnerability Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; STIBP always-on; RSB filling; PBRSB-eIBRS Not affected; BHI Not affected\nVulnerability Srbds:                  Not affected\nVulnerability Tsx async abort:        Not affected\n\nVersions of relevant libraries:\n[pip3] conch-triton-kernels==1.2.1\n[pip3] lmcache==0.5.5rc3+rocm7.2.4.torch2.10.git3d3aa833.cxx11abi1\n[pip3] numpy==2.3.5\n[pip3] onnx==1.22.0\n[pip3] onnx-ir==1.0.0\n[pip3] onnxscript==0.7.2\n[pip3] onnxslim==0.1.96\n[pip3] torch==2.10.0+rocm7.2.4.lw.git3d3aa833\n[pip3] torch_c_dlpack_ext==0.1.5\n[pip3] torchaudio==2.10.0+rocm7.2.4.git5047768f\n[pip3] torchvision==0.25.0+rocm7.2.4.git82df5f59\n[pip3] triton==3.7.0+amd.rocm7.2.0.git89002410\n[pip3] triton_kernels==1.0.0+amd.rocm7.2.0.git89002410\n[conda] Could not collect",
  "transformers_version": "5.16.1",
  "lm_eval_version": "0.4.13",
  "upper_git_hash": null,
  "tokenizer_pad_token": [
    "[PAD]",
    "163839"
  ],
  "tokenizer_eos_token": [
    "[EOS]",
    "163585"
  ],
  "tokenizer_bos_token": [
    "[BOS]",
    "163584"
  ],
  "eot_token_id": 163585,
  "max_length": 2047,
  "task_hashes": {},
  "model_source": "local-completions",
  "model_name": "/models/moonshotai/Kimi-K3",
  "model_name_sanitized": "__models__moonshotai__Kimi-K3",
  "system_instruction": null,
  "system_instruction_sha": null,
  "fewshot_as_multiturn": null,
  "chat_template": null,
  "chat_template_sha": null,
  "total_evaluation_time_seconds": "728.2678453065455"
}```

</details>

<details>
<summary>Server-side facts checked after the run</summary>

```
init engine (profile, create kv cache, warmup model) took 245.85 s
Application startup complete.
grep -cE "HSA_STATUS_ERROR|Memory access fault|EngineDeadError" vllm_oot.log  ->  0
aiter/utility/graph_alloc.py: MemPool present, persistent_alloc present   (ROCm/aiter#5430)
vllm 0.28.1.dev0+g2cf0a6915.d20260916.rocm724 / torch 2.10.0+rocm7.2.4.lw.git3d3aa833
ATOM: editable install of the branch worktree, git dirty count 0, base e4f5ff2d7
```

</details>

Caveat, so the number is read for what it is: **n = 1** for this revision — one run, not a repeat set.

valarLip (approved):

PR #2161, opened 2026-09-08
Reduce frontend stalls by coalescing tokens before detokenization

junhaha666:
## Problem and change

With 4096 streaming requests and a concurrency limit of 2048, the model can finish the first wave while the API process is still draining its output. The benchmark holds each concurrency slot until the full response completes, so the next requests cannot enter the engine. A DeepSeek-V4-Pro run on the main baseline showed a roughly 30.76-second gap in CPU operator/runtime activity on all four DP ranks between waves.

`StreamOutputCollector` already merges unread chunks, but `StreamBatchDispatcher.flush()` previously detokenized every delta before that merge. This change passes raw token deltas and their per-stream detokenizer state to the collector, then decodes the accumulated tokens when the consumer calls `get()`. Backlogged streams therefore avoid repeated decode calls and output-thread GIL contention. There is no timer or minimum batch size; available chunks remain immediately readable. The existing queue compatibility path retains eager decoding; production OpenAI streams use collectors. Per-stream/fan-out state, token order, finish metadata, and incremental Unicode handling are preserved.

The output path also calls `_mark_seq_prefill_complete()` after prefill is already accounted for. An early read now skips `_lb_lock` when the immutable charge tuple is absent or its prefill charge is zero. Counter mutations still re-read and validate the entry under the lock to handle concurrent finish/abort.

Deferred decode exceptions are logged with the request ID and fan-out tag. The consumer receives empty text for the failed update while retaining token accounting and terminal metadata, so one bad decode cannot terminate sibling streams. A later successful update can recover buffered tokens without appending them twice; text from a failing terminal update is not guaranteed. Documentation now states the single-mutator invariant, pending-state lifetime, and the fact that synchronous event-loop decoding does not guarantee unchanged TTFT/ITL.

## Validation

Latest revision (`e7393994`) was validated on CPU only, with GPU visibility disabled:

- `python -m pytest -q tests/entrypoints tests/test_dp_load_balance.py --ignore=tests/entrypoints/test_openai_server.py`: **1952 passed, 31 skipped, 3 xfailed**.
- Regression tests inject exceptions in both prefix and new-text decoding, checking recovery without duplicate tokens, terminal metadata, logging, and fan-out isolation.
- Dispatcher tests now exercise the production collector path for per-loop batching, independent stream state, and immediate reads. Synthetic-output tests assert zero decode calls before consumption and exactly two after merging.
- 36 committed offline checks use a locally trained byte-level BPE tokenizer through the real HF wrapper to compare merged deltas against single-token updates, including full and mid-stream backlogs, multilingual text, emoji-ZWJ, whitespace, code, accents, and special tokens. No model download is needed.
- Earlier validation at `4dafab6d` included 25 checks with the actual DeepSeek tokenizer. Those checks and the GPU A/B below were not rerun for this revision.
- Black and `git diff --check` passed. Ruff findings were unchanged from the baseline (three BLE001 and one S110 in `engine_core_mgr.py`).

## DeepSeek-V4-Pro A/B (measured at `4dafab6d`)

These are the original performance measurements, not a GPU rerun of `e7393994`.

Baseline: `2458289a58d5cfbb3542d232023c58b8fc81345f` on `origin/main`. Both runs used the same gfx1250 model/server configuration, four DP ranks, profiling enabled, 4096 random requests, concurrency 2048, 1024 input tokens, 1024 output tokens, infinite request rate, and ignored EOS. The two production changes above were applied together.

| Metric | Main baseline | `4dafab6d` |
| --- | ---: | ---: |
| Successful requests | 4096 / 4096 | 4096 / 4096 |
| Output tokens per request | 1024 | 1024 |
| Benchmark duration | 221.05 s | 155.80 s |
| Output throughput | 18,974 tokens/s | 26,920 tokens/s |
| Mean TTFT | 12,016 ms | 11,672 ms |
| Mean TPOT | 96.00 ms | 61.55 ms |
| Maximum internal CPU operator/runtime gap, per rank | 30.7591–30.7598 s | 0.2699–0.2963 s |
| Internal CPU operator/runtime gaps over 1 s | One per rank | Zero on all four ranks |

This single A/B measurement reduced duration by **29.5%** and increased output throughput by **41.9%**. Benchmark duration excludes profiler shutdown/export. Gap measurements use the union of CPU operator/runtime intervals between the first and last event; model-step annotations independently had no internal gap above 0.4 seconds after the change.

GPU kernel traces ended before the complete CPU model-step timeline in both runs, so these results do not establish a whole-run GPU utilization percentage or strictly gap-free GPU execution. The measured improvement is removal of the long inter-wave scheduling stall and faster end-to-end completion. Multiple-run variance and the individual contribution of each change have not been measured.

## Follow-up scope

The review's larger collector-state/API refactor (including removal of the queue compatibility path) and the separate atomesh streaming path remain follow-up work. The state identity invariant is documented in this change. Event-loop occupancy and high-concurrency consumers that keep up still need dedicated latency measurement; there is no claim of an ITL bound. The `ci:atom` label was removed, and native/vLLM/SGLang GPU accuracy jobs were skipped. The repository also starts an ungated GPU offline smoke job on every PR update; it entered the offline-inference step before this separate trigger was noticed. Run 34217907352 was cancelled and its cleanup step completed. Local validation used CPU only, and all remote Pre Checkin checks (including non-GPU unit tests) passed. No new GPU performance result is claimed for this revision.

### DIFF
--- a/atom/entrypoints/openai/streaming_dispatch.py
+++ b/atom/entrypoints/openai/streaming_dispatch.py
@@ -116,10 +116,10 @@ class StreamOutputCollector:
     and socket writing. Detokenizer state is advanced only by ``get()``, on the
     event loop; the engine output threads never decode collector-bound chunks.
 
-    Nothing is ever held back. With a consumer that keeps up nothing ever
-    merges, and delivery is identical to the queue this replaces. Merging only
-    covers chunks that were already waiting, so a token is never delivered later
-    than it would have been.
+    There is no timer or minimum batch size: get() can consume any pending
+    delta immediately. Merging covers only unread chunks, but decoding runs
+    synchronously on the event loop, so this is not a delivery-latency bound.
+    Large backlogs can delay other consumers while their text is decoded.
 
     ``tag`` is the fan-out sibling index (``SamplingParams.n>1``) or ``None`` for
     a plain single-sequence stream. Chunks merge per tag, so siblings never mix.
@@ -156,9 +156,18 @@ class StreamOutputCollector:
             self._ready.clear()
         state = chunk.pop("_detokenizer", None)
         if state is not None:
-            chunk["text"] = state.update(
-                chunk["token_ids"], bool(chunk.get("finished"))
-            )
+            try:
+                chunk["text"] = state.update(
+                    chunk["token_ids"], bool(chunk.get("finished"))
+                )
+            except Exception:
+                logger.exception(
+                    "Error detokenizing stream %s (tag=%s)", self.request_id, tag
+                )
+                # Keep token accounting and terminal metadata, and let fan-out
+                # siblings continue. Failed tokens remain in the detokenizer;
+                # a later update can decode them without appending them twice.
+                chunk["text"] = ""
         return chunk if tag is None else (tag, chunk)
 
 
@@ -238,17 +247,18 @@ class _BufferedChunk(NamedTuple):
 class StreamBatchDispatcher:
     """Collect one engine step per output thread and dispatch it by event loop.
 
-    Holds no per-stream state. Each stream's detokenizer belongs to the engine
-    callback that feeds it, and rides along on every chunk, so nothing here is
-    shared between the output threads and the event loop and nothing has to be
-    cleaned up: when the engine drops a finished stream's callback the
-    detokenizer goes with it.
+    The dispatcher has no persistent per-stream registry. Each engine callback
+    creates one detokenizer and must reuse it for every chunk of its
+    (collector, tag); every collector-bound chunk carries that same state.
+    Merging keeps the first pending chunk's state, so it must not be replaced
+    partway through a stream.
 
-    It used to live in a dict here, which cost a lock -- 27% of the API
-    server's CPU, since every buffered chunk looked its state up with all
-    output threads contending -- and then, without the lock, cost an entry
-    that two threads had to keep in agreement and that teardown had to
-    remember to remove.
+    For StreamOutputCollector, references cross threads but only get() on the
+    event loop mutates the state. Output threads just pass it through. An unread
+    chunk or queued delivery keeps the state (including its token history) alive
+    after the engine drops the finished callback; consuming or discarding those
+    references allows it to be reclaimed. Queue consumers use eager decoding
+    on their output thread instead.
     """
 
     def __init__(self, tokenizer: Any, synthetic_text: str | None = None):
--- a/atom/entrypoints/openai/streaming_dispatch.py
+++ b/atom/entrypoints/openai/streaming_dispatch.py
@@ -4,9 +4,9 @@
 """Cross-thread dispatch and per-request delivery for streaming model output.
 
 Two halves of one hand-off. :class:`StreamBatchDispatcher` runs on the engine
-output threads: it buffers a whole engine step, detokenizes it, and schedules a
-single callback per event loop. :class:`StreamOutputCollector` is the loop-side
-landing point each stream's SSE generator reads from.
+output threads and schedules a whole engine step per event loop.
+:class:`StreamOutputCollector` merges pending token deltas before decoding them
+when the SSE consumer reads, so a slow consumer does not accumulate decode work.
 """
 
 import array
@@ -111,12 +111,10 @@ def merge_chunk(into: dict, new: dict) -> None:
 class StreamOutputCollector:
     """Per-request delivery point that merges chunks when the consumer lags.
 
-    Replaces the unbounded ``asyncio.Queue`` that used to sit between the engine
-    output threads and the SSE response generators. A queue hands over one item
-    per ``get()``, so when the frontend cannot keep up with the GPU the backlog
-    grows without bound and every queued item still costs its own coroutine
-    wakeup, JSON encode and socket write. Here a stream holds at most one chunk:
-    anything arriving behind an unread one merges into it.
+    A stream holds at most one pending chunk per tag. Raw token deltas from the
+    dispatcher merge here before detokenization, coroutine wakeup, JSON encoding
+    and socket writing. Detokenizer state is advanced only by ``get()``, on the
+    event loop; the engine output threads never decode collector-bound chunks.
 
     Nothing is ever held back. With a consumer that keeps up nothing ever
     merges, and delivery is identical to the queue this replaces. Merging only
@@ -156,6 +154,11 @@ class StreamOutputCollector:
         del self._pending[tag]
         if not self._pending:
             self._ready.clear()
+        state = chunk.pop("_detokenizer", None)
+        if state is not None:
+            chunk["text"] = state.update(
+                chunk["token_ids"], bool(chunk.get("finished"))
+            )
         return chunk if tag is None else (tag, chunk)
 
 
@@ -275,7 +278,7 @@ class StreamBatchDispatcher:
         buf.append(_BufferedChunk(loop, collector, state, chunk, tag))
 
     def flush(self) -> None:
-        """Detokenize buffered chunks and schedule one delivery per event loop."""
+        """Schedule raw chunks, letting each consumer coalesce before decoding."""
         tl = self._thread_local
         buf = getattr(tl, "buf", None)
         if not buf:
@@ -284,10 +287,18 @@ class StreamBatchDispatcher:
 
         by_loop: dict[AbstractEventLoop, list[tuple[Any, Any]]] = {}
         for item in buf:
-            item.chunk["text"] = item.state.update(
-                item.chunk.get("token_ids") or [],
-                bool(item.chunk.get("finished")),
-            )
+            if isinstance(item.collector, StreamOutputCollector):
+                # Keep this state with the pending chunk until get(). Moving
+                # only the JSON/socket work downstream still made four output
+                # threads decode every token while contending for the GIL.
+                item.chunk["_detokenizer"] = item.state
+            else:
+                # Queue consumers cannot decode on read and still receive a
+                # prepared chunk, as before.
+                item.chunk["text"] = item.state.update(
+                    item.chunk.get("token_ids") or [],
+                    bool(item.chunk.get("finished")),
+                )
             payload = item.chunk if item.tag is None else (item.tag, item.chunk)
             by_loop.setdefault(item.loop, []).append((item.collector, payload))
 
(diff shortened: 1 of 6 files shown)


valarLip:
Reviewed at `4dafab6df`. **No blocking issue** — nothing here fires on today's code paths. I checked the two things that could have turned a latent problem into a live one, and both are closed by construction:

- `state` is never `None` — both enqueue sites receive `detokenizer = _stream_batch_dispatcher.new_state()`, minted once per closure (`api_server.py:1250`, `make_callback:1461`).
- Nothing else feeds a `StreamOutputCollector`. Every other `put_nowait` in the tree targets an engine-internal queue; `_deliver` is the collector's only producer.

Also worth recording, since three separate angles raised it: the "moving decode to the event loop loses real parallelism" concern is **wrong**. `tokenizer.decode` does not release the GIL (0.96–1.03x on 4 threads, against a 3.79x numpy positive control), so the output threads were already GIL-bound. The PR's stated justification holds.

One thing I'd fix before merge, two doc corrections, and the rest is follow-up.

---

# Fix before merge: `get()`'s decode is unguarded, which is a strict regression in error containment

```python
del self._pending[tag]
if not self._pending:
    self._ready.clear()
state = chunk.pop("_detokenizer", None)
if state is not None:
    chunk["text"] = state.update(chunk["token_ids"], bool(chunk.get("finished")))
```

The chunk is removed from `_pending` *before* the decode, so a raise loses it — possibly the `finished=True` one — and leaves the detokenizer half-advanced (`_decode` already did `self.tokens.extend(token_ids)` at `:71` without moving `prefix_offset` / `read_offset`).

Before this PR the same raise happened inside `flush()` on the output thread and was caught and logged at `engine_core_mgr.py:646-652`. Now it propagates out of `await stream_collector.get()` into `serving_chat.py:289/596`, `serving_completion.py:91/252` and `api_server.py:2164` — **none of which has an `except`**. The response dies mid-stream with no log line at any level.

Fan-out is worse than single-stream: all `n` siblings share one collector and one generator, so one bad decode takes down all of them.

A `try/except` that logs and returns `""` makes this a non-event. Three lines.

# Docs that now assert the opposite of the code

**`StreamBatchDispatcher`'s class docstring (`streaming_dispatch.py:241-245`) was not updated:**

> ...rides along on every chunk, so nothing here is shared between the output threads and the event loop and nothing has to be cleaned up: when the engine drops a finished stream's callback the detokenizer goes with it.

After `:294` the detokenizer *is* shared with the event loop and mutated there (`:159`), and an unread pending chunk keeps it — and its unbounded `tokens` array — alive after the callback is dropped at `engine_core_mgr.py:633`. The code is still safe, but for a different reason: single mutator, not non-sharing.

This matters because it is the only place the threading invariant is written down, and it is exactly what a future reader consults before deciding whether the detokenizer needs a lock.

**`docs/serving_benchmarking_guide.md:97-106`** still lists the fixed per-chunk cost as "(detokenize, coroutine wakeup, JSON encode, socket write)" and describes backlog merging without mentioning that detokenization is now the thing being coalesced. That section is where a reader lands when asking what the collector actually saves. `git diff --name-only` touches four files, none under `docs/`.

# Follow-up: three findings that collapse into one refactor

The detokenizer is handed over in-band under a magic dict key, and `merge_chunk` has no notion of it:

```python
def merge_chunk(into, new):
    into["token_ids"].extend(...)
    text = into.pop("text", "")
    ...
    into["finished"] = ...
    for key in _LATEST_WINS: ...
    # no _detokenizer anywhere
```

So a merge keeps the *first* chunk's state object and drops the incoming one's. That is correct today only because (a) every collector-bound chunk carries a `_detokenizer` and (b) exactly one state is minted per `(collector, tag)` and never re-created. Neither is stated or asserted, and `enqueue`'s `collector` parameter is typed `Any`. If (b) ever breaks — a reconnect path minting a fresh `new_state()` mid-stream — merged tokens decode against the wrong offsets and the client gets duplicated or truncated text, with no exception. If (a) breaks, `merge_chunk` fabricates `text=""`, `get()` finds `state is None`, and the consumer reads `""` while still counting the tokens: a silently empty stream.

Two related costs hang off the same design:

- **The `isinstance` type-switch in `flush()` (`:290`) selects between the shipped path and a dead one.** `StreamBatchDispatcher.enqueue` has exactly two callers (`api_server.py:811`, `:844`), both declaring `stream_collector: StreamOutputCollector` and both fed collector instances. `generate_async`'s queue paths bypass the dispatcher entirely. So the branch is unconditionally true, the `else` at `:296-301` is dead, and its comment describes a consumer that does not exist. The cost is a per-chunk-per-step type test forever, plus two decode paths with different prefix-window semantics that have to stay in agreement — kept alive only by test fixtures.
- **`merge_chunk`'s text block is dead work on the only production path.** `_build_stream_chunk` (`api_server.py:786-799`) emits no `"text"` key, so every production merge does a missing-key pop, a missing-key get, a try/finally, an empty-string concat, and a dict insert of `""` that `get()` overwrites four lines later. The comment about CPython growing the string in place now documents an optimization for `""`.

Storing the state beside the pending chunk — `self._pending[tag] = (chunk, state)` — removes the magic key, the pop, the `isinstance`, the dead branch and the dead text block in one move, and makes both failure modes unrepresentable.

# Follow-up: `engine_core_mgr.py`

`:1238` indexes positionally where the same function unpacks by name eleven lines above and six lines below:

```python
if entry is None or entry[2] == 0:      # :1238
...
dp_rank, req_cost, tok_cost = entry     # :1227 and :1244
```

Extending the tuple is loud — the named unpacks raise `ValueError`. A **reorder** is silent: if `entry[2]` ever lands on a field that is commonly 0, `_mark_seq_prefill_complete` returns early forever, the prefill charge is never released, `_rank_tokens[rank]` drifts monotonically upward, and the `least_tokens` router permanently avoids that rank — with no error anywhere. `_, _, tok_cost = entry` costs nothing.

Minor, same block: the comment says "the immutable charge tuple changes to zero on the first output", but `_charge_seq_load_locked` can store `tok_cost == 0` at charge time — `tok_cost = prompt_tokens` with zero/absent `num_prompt_tokens` (`:1161-1162`), and the DP session-affinity branch `max(0, prompt_tokens - previous_prompt_tokens)` (`:1170`) for an agentic turn whose prompt did not grow. Behaviour is right (the in-lock `if tok_cost == 0: return` covers it); the comment sends a reader looking for a release that never happened.

# Follow-up: test coverage

**The property the whole PR rests on has no test.** Nothing in the suite asserts that one `update()` over K merged tokens yields the same text as K single-token updates. I checked it independently and it holds — 630 comparisons across 7 real tokenizers (Qwen3.5-4B/27B/35B, DeepSeek-V4-Pro, DeepSeek-V4-Flash-Vision, GLM-5.2, gpt-oss-120b), 9 merge patterns each including full-merge and mid-stream backlog, covering CJK, emoji-ZWJ, whitespace, code and accents: 0 divergences. But nothing guards it.

**Three of the four new streaming tests pass unchanged on `origin/main`.** Running the PR's tests against main's `streaming_dispatch`, only `test_backlogged_tokens_are_merged_before_decoding` fails (`assert 960 == 0`). The other three are legitimate equivalence guards, but the PR body presents four tests for one behaviour. `test_deferred_synthetic_stream_preserves_token_count` in particular asserts `tokenizer.calls > 0` — true on both sides, and it cannot distinguish 2 calls from 6, which is the quantity it is named for.

**Five pre-existing dispatcher tests now exercise only the dead branch.** `test_dispatcher_batches_direct_and_tagged_chunks_per_loop` (`:71`), `test_dispatcher_keeps_fanout_detokenizer_state_separate` (`:99`), `test_a_fresh_stream_does_not_inherit_a_half_decoded_character` (`:137`), `test_dispatcher_keeps_no_per_stream_state` (`:410`) and `test_each_stream_gets_its_own_detokenizer` (`:440`) all pass `collector=asyncio.Queue()`, so after this diff every one takes `flush()`'s `else`. The new fan-out test re-establishes fan-out and partial-UTF-8 on the deferred path, but two invariants lose deferred-path coverage: "a fresh stream does not inherit a half-decoded character" with two fresh states flushed in one batch, and the `len(loop.calls) == 1` one-callback-per-step property that `_deliver`'s own docstring calls stream-corrupting to violate.

Related: `test_collector_hands_over_a_lone_chunk_untouched` (`:166`, docstring "A consumer that keeps up must see exactly what the queue used to give") asserts `_resolve(collector.get()) is chunk` and still passes only because it hand-builds a chunk that never went through the dispatcher. `get()` now pops `_detokenizer` and assigns `text` on every real chunk, so the collector docstring's "delivery is identical to the queue this replaces" (`:119-122`) is no longer literally true, and the test guarding that sentence covers no production shape.

`tests/test_dp_load_balance.py:524` also has a function-local `from threading import Event, Thread` while line 13 of the same file is `from threading import Lock`; and the new DP test has no ARMED pre-assertion that routing charged anything (`sum(mgr._rank_tokens) == 20` before the first `_mark_seq_prefill_complete`), so if `_charge_seq_load_locked` ever charges 0 the sequence is born in the fast-path state, the test passes, and the charged→released transition it exists to cover is never exercised.

# Follow-up: Fix-then-sweep

`atom/entrypoints/atomesh/atom_standalone_service.py:645` still has the identical pattern this PR removes — `StreamRequestState.record()` runs on the engine output thread, takes a per-request `threading.Lock`, and calls `self.tokenizer.decode(output_tokens, skip_special_tokens=True)` per chunk. Same GIL contention, made worse by the lock, so the inter-wave stall is still reachable through atomesh streaming. (It also ignores `synthetic_token_text` and decodes each chunk with no prefix window, so a multi-byte character split across chunks yields U+FFFD — both pre-existing, but a sweep would have surfaced them.)

# Watch item, not a blocker: the decode is concentrated, not eliminated

Total work goes down; where it lands changes. `get()` decodes the whole accumulated backlog inline with no `await` between the pop at `:154` and the return at `:162`. Measured on the DeepSeek-V4-Pro tokenizer, per decode call: 1024 tokens = 0.22 ms, 8192 = 1.80 ms, 32768 = 5.95 ms — and `update()` makes two. Single-stream worst case is ~12 ms; a 2048-stream wave each draining ~1024-token backlogs is roughly 0.9 s of contiguous loop time, on the loop whose stall sensitivity `api_server.py:1286-1294` documents at 3.3 s per incident.

Strictly better than main in total work, so this is a number to own rather than a problem to fix. But the A/B would be more convincing with a **keep-up client at high concurrency** included — that is the case where merging never fires and the decode simply relocates onto the loop.

valarLip (approved):

PR #2163, opened 2026-09-08
[Perf][MiniMax-M3] Bound the decode index-score split by blocks-per-chunk

yitingw1:
## Motivation

The sparse-indexer score kernel (`_decode_index_score_tiled_kernel`) is the largest single item in a conc=1 decode step — **1.93 ms/step, 20.5%** — and runs on a quarter of the GPU.

```python
DECODE_SCORE_TARGET_GRID = 2048
DECODE_SCORE_MAX_CHUNKS  = 64          # binds for every batch <= 32
```

At 315K context and batch=1 that is 64 workgroups — **25% of 256 CUs**, each walking 39 blocks serially, while the heuristic itself aims for 2048. Exactly the regime where we trail vLLM on conc=1/2 interactivity.

Result: **+17~19% interactivity** end to end, output bit-identical.

## Technical Details

A flat chunk ceiling cannot work, because the best chunk count follows the context length while the best *blocks per chunk* does not:

| context | blocks | best chunks (batch≥8) | blocks/chunk |
| --- | ---: | ---: | ---: |
| 32K | 256 | 64 | **4** |
| 131K | 1024 | 512 | **2** |
| 315K | 2461 | 821 | **3** |

Every measured regression is a chunk down to **one block**: the query tile is loaded once *outside* the block loop, so at one block that fixed cost is the whole cost.

But a floor alone ignores batch — it yields 2731 chunks at every batch and the grid grows without limit (275 μs at batch 128 against 202 capped). So the two constants split the batch range:

```python
DECODE_SCORE_TARGET_GRID = 1 << 14     # ceiling on total workgroups; binds from batch 8 up
DECODE_SCORE_MIN_BLOCKS  = 3           # floor on blocks per chunk; binds below that
# DECODE_SCORE_MAX_CHUNKS deleted -- provably dead under the floor
```

The grid stays a pure function of `(batch, max_block)`, so **CUDA graph replays the grid it captured**.

**Bundled fix** (`fix(attn): keep decode off the gluon kernel past its query-length and group limits`): the gluon decode kernel asserts on `query_length <= 4` and `max_seqlen_q * (q_heads / kv_heads) <= 64`. Drafting multiplies both, and a model can reach either first — M3 sits exactly on the group limit today (16 × 4 draft positions), a gqa=8 model hits the length limit with its group still at 40. Backend selection now accounts for both, and refuses outright when the layer is per-token quantized: the unified fallback carries one descale for the whole tensor, so running it there would be wrong numbers rather than an error.

Two consequences worth stating:

- **`kv_scale` moves to the device.** The new query-group term is the first env-independent way into the unified branch (previously it needed `ATOM_USE_UNIFIED_ATTN` or `use_flash_layout`, neither on by default). `k_descale` is forwarded unchanged to `k_descale_ptr` at every launch site — no `.to(device)`, no `None` substitution — so a host tensor would be dereferenced in the kernel.
- **`PA_ASM_MAX_QUERY_GROUP_SIZE = 16` reroutes existing models** from `paged_attention_asm` to gluon whenever `max_qlen * (q_heads / kv_heads) > 16`; M3 at TP4 with a 4-token draft is itself in that set. The bound has upstream backing (`asm_pa.cu:113-116`: `// # mtp * gqa <= 16`). **The reroute's performance delta was not measured** — it is accepted because past the envelope `get_heuristic_kernel` silently re-runs with `mtp=1`, a kernel built for another query length.

## Test

MI355X (gfx950), TP4, MXFP4 weights + fp8 KV/index cache.

**Kernel**, capture semantics (μs). Decode runs under CUDA graph and capture takes `max_seqlen_k` from `max_model_len` (`aiter_attention.py:1334`), so production sizes the grid from the model limit, not the live context — measured with the two decoupled (`--model-len 1048576`), real context per row, 2 repeats:

| batch | 1 | 2 | 8 | 16 | 24 | 32 | 64 | 128 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| before | 110.6 | 110.6 | 123.1 | — | — | 193.7 | 244.0 | 374.9 |
| after | **27.9** | **34.0** | **72.9** | **132.8** | **189** | **195.7** | **195.9** | **202.0** |

Rows 32–128 use the longest context each batch's KV pool admits. The old rule shrank the split as batch grew (64 → 32 → 16), bounding total CTAs but starving parallelism — 374.9 μs at batch 128 is the worst point measured anywhere.

This also reconciles the profile: the kernel measured 129 μs/call in production, 2.6× an eager-semantics microbench, and the capture number accounts for it.

**End-to-end** (`tp4_eagle3`, real agentic corpus at 315K, 1800 s/point):

| | tok/s/chip | intvty_p90 |
| --- | --- | --- |
| conc=1 | 4,403 → **4,601** (**+4.50%**) | 233.1 → **273.3** (**+17.23%**) |
| conc=2 | 4,751 → **4,856** (**+2.21%**) | 226.1 → **270.0** (**+19.41%**) |

The gap is expected: the change shortens one kernel inside the decode step, which lands on interactivity, while throughput also carries prefill and TTFT.

**Exactness** — 3 contexts × 4 batches × 2 query lengths = 24 shapes; `topk_idx`, `sparse_bt`, `sparse_ctx` all bit-identical to baseline, while the chunk count over the same shapes does change.

**Accuracy** — gsm8k under the CI recipe verbatim (TP8, 32K), flexible-extract **0.9409** ± 0.0065 against a 0.9363 baseline and 0.93 threshold.

**Unit tests** — `tests/test_paged_attention_dispatch.py`, 28 cases over the envelope and the routing table it feeds: the named shapes, `next_pow2` rounding, non-positive query lengths, the constants against the kernel sources, and each arm of `_dispatch_decode` driven through a fake layer. Two worth calling out — `test_every_reachable_group_has_an_arm` asserts every shape reported safe lands on a layout arm aiter actually defines, which nothing else watches; `test_sliding_window_honours_unified_env` covers the arm that returns before the `ATOM_USE_UNIFIED_ATTN` block and so has to carry the flag itself. Revert power was checked, not assumed: every routing case turns red under a named revert of the line it guards. **These skip on the CPU CI runner**, which installs neither triton nor aiter; every other attention test in `tests/` skips there for the same reason.

## Known Limitations

- **A per-token-quantized MHA layer under `ATOM_USE_UNIFIED_ATTN=1` now refuses to start.** `unified_attention` carries one descale for the whole tensor, so that combination previously served traffic with the wrong numbers; it raises `NotImplementedError` at the first prefill instead. No known config is hit — the only `ATOM_USE_UNIFIED_ATTN=1` entries in `.github/benchmark/` and `recipes/` are sglang kimi_k3 (MLA, never reaches `PagedAttentionImpl`) and two `--kv_cache_dtype bf16` recipes, which produce no scale tensor — but it is a behaviour change, not only a new guard.
- **Neither constant is derived.** The measured best `batch × chunks` runs 8k–44k over batch 16–128, so no single `TARGET_GRID` hits every point; `1<<14` lands within 5% of the per-batch optimum there. Landing exactly would mean segmenting on target blocks-per-chunk, a larger change.
- **batch 256 still regresses** (394 μs floor-only, 214 with the ceiling, 235 for the old rule). Reaching it needs every request under 30K tokens: M3 does not compress KV, and a decoding sequence's KV must stay GPU-resident, so the TP4 pool of 7.66M tokens bounds `batch × context` whatever the offload setup — CPU offload raises how many idle sessions stay warm, not how many sequences decode at once. At the 315–375K contexts this workload runs, that is conc 24. Batch 128 is the tuning target as headroom, not because 256 is reachable.
- Under CUDA graph the split is sized from `max_model_len`, so a short request in a long deployment launches chunks that return immediately. Measured at the extreme (4K context, grid sized for 1M): still net positive through batch 32.
- `MIN_BLOCKS=3` was measured at 32K / 131K / 315K; shorter contexts should be re-verified, not assumed.
- Measured on TP4 only. The rule takes no TP argument and index-K is MQA, so per-rank traffic is TP-invariant; the optimum would move only if `idx_heads × query positions > 16` pushes `BLOCK_SIZE_N` to 32 (TP1 with ≥7 draft tokens, TP2 with ≥15).
- The kernel microbench uses a synthetic contiguous block table; production tables are fragmented, which is why the end-to-end A/B decides.

## Submission Checklist

### DIFF
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -34,6 +34,15 @@ def use_pa_decode_bf16_asm() -> bool:
     )
 
 
+# Largest query group the gluon decode kernel can take: its register-layout
+# table has arms for 16/32/64 only, and past 64 nothing binds `register_bases`
+# -- the failure is a Triton compile error that never mentions speculation.
+# The group is max_seqlen_q * (q_heads / kv_heads), so drafting multiplies it:
+# M3 sits exactly on the limit today (16 x 4 draft positions), and one more
+# draft token would overflow it. vLLM guards the same bound.
+PA_GLUON_MAX_QUERY_GROUP_SIZE = 64
+
+
 class PagedAttentionImpl(nn.Module):
     """
     Attention paged implementation
@@ -489,7 +498,14 @@ class PagedAttentionImpl(nn.Module):
 
         num_seqs = attn_metadata.context_lens.shape[0]
 
-        if envs.ATOM_USE_UNIFIED_ATTN or self.use_flash_layout:
+        # Group-size term last: k_cache only carries the 5D shuffle layout this
+        # indexing assumes when the gluon path is otherwise the choice.
+        if (
+            envs.ATOM_USE_UNIFIED_ATTN
+            or self.use_flash_layout
+            or attn_metadata.max_seqlen_q * (q.shape[1] // k_cache.shape[1])
+            > PA_GLUON_MAX_QUERY_GROUP_SIZE
+        ):
             # print(q.shape, k_cache.shape, v_cache.shape)
             sliding_window = (
                 (self.sliding_window - 1, 0) if self.sliding_window > 0 else (-1, -1)
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -54,10 +54,7 @@ def gluon_decode_over_limit(max_qlen: int, num_heads: int, num_kv_heads: int) ->
     group_p2 = qlen_p2 * max(
         16 // qlen_p2, 1 << (num_heads // num_kv_heads - 1).bit_length()
     )
-    return (
-        max_qlen > PA_GLUON_MAX_QUERY_LEN
-        or group_p2 > PA_GLUON_MAX_QUERY_GROUP_SIZE
-    )
+    return max_qlen > PA_GLUON_MAX_QUERY_LEN or group_p2 > PA_GLUON_MAX_QUERY_GROUP_SIZE
 
 
 class PagedAttentionImpl(nn.Module):
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -565,9 +565,7 @@ class PagedAttentionImpl(nn.Module):
             max_context_partition_num,
             query_group_size,
         )
-        exp_sums = torch.empty(
-            intermediate_shape, dtype=torch.float32, device=q.device
-        )
+        exp_sums = torch.empty(intermediate_shape, dtype=torch.float32, device=q.device)
         max_logits = torch.empty(
             intermediate_shape, dtype=torch.float32, device=q.device
         )
@@ -885,7 +883,9 @@ class PagedAttentionImpl(nn.Module):
         get_heuristic_kernel silently re-runs with mtp=1, a kernel built for
         another query length, and computes a wrong answer instead of refusing.
         """
-        over_gluon = gluon_decode_over_limit(max_qlen, self.num_heads, self.num_kv_heads)
+        over_gluon = gluon_decode_over_limit(
+            max_qlen, self.num_heads, self.num_kv_heads
+        )
         over_asm = (
             int(max_qlen) * (self.num_heads // self.num_kv_heads)
             > PA_ASM_MAX_QUERY_GROUP_SIZE
--- a/atom/model_ops/attention_mha.py
+++ b/atom/model_ops/attention_mha.py
@@ -167,6 +167,21 @@ class PagedAttentionImpl(nn.Module):
             return False
         return q.shape[1] % k.shape[1] == 0
 
+    def _reject_per_token_scales_on_unified(self, k_scale, v_scale, phase: str):
+        """unified_attention carries one descale for the whole tensor.
+
+        Feeding it a per-token cache is not an error downstream, just wrong
+        numbers, so it has to be refused here. Both phases route into unified on
+        the same conditions, so both check.
+        """
+        if (k_scale is not None and k_scale.numel() > 1) or (
+            v_scale is not None and v_scale.numel() > 1
+        ):
+            raise NotImplementedError(
+                f"layer {self.layer_num} takes unified_attention for {phase}, "
+                "which cannot carry its per-token KV scales"
+            )
+
     def forward_impl(
         self,
         q: torch.Tensor,
@@ -490,16 +505,7 @@ class PagedAttentionImpl(nn.Module):
         of the reasons for it.
         """
         attn_metadata = fwd_ctx.attn_metadata
-
-        # unified takes a single descale for the whole tensor. Feeding it a
-        # per-token cache is not an error downstream, just wrong numbers.
-        if (k_scale is not None and k_scale.numel() > 1) or (
-            v_scale is not None and v_scale.numel() > 1
-        ):
-            raise NotImplementedError(
-                f"query length {attn_metadata.max_seqlen_q} routes this layer to "
-                "unified_attention, which cannot carry its per-token KV scales"
-            )
+        self._reject_per_token_scales_on_unified(k_scale, v_scale, "decode")
 
         if envs.ATOM_USE_UNIFIED_ATTN and self.kv_cache_dtype.startswith("fp8"):
             o = torch.empty(*q.shape, dtype=torch.bfloat16, device=q.device)
@@ -787,7 +793,6 @@ class PagedAttentionImpl(nn.Module):
     def prefill_attention_triton(
         self, q, k, v, k_cache, v_cache, k_scale, v_scale, fwd_ctx: ForwardContext
     ):
-
         # unified_attention supports both prefill and decode, over either the 4D
         # flash layout (shuffled_kv_cache=False) or the 5D SHUFFLE layout
         # (shuffled_kv_cache=True):
@@ -825,6 +830,9 @@ class PagedAttentionImpl(nn.Module):
         # are read straight from `k_cache`/`v_cache`, identical to the
         # prefix-cache-hit path.
         if envs.ATOM_USE_UNIFIED_ATTN or attn_metadata.has_cached:
+            # Only this branch reads the quantized cache; the else below passes
+            # raw bf16 K/V, where a per-token scale never reaches the kernel.
+            self._reject_per_token_scales_on_unified(k_scale, v_scale, "prefill")
             k_for_attn = k_cache
             v_for_attn = v_cache
             # Reads the paged KV cache, which is 5D SHUFFLE unless the (default)
@@ -876,26 +884,38 @@ class PagedAttentionImpl(nn.Module):
     def _dispatch_decode(self, max_qlen: int = 1):
         """The single place that answers which decode kernel runs.
 
-        The runners do not re-decide; the vLLM bridge mirrors this order. Both
+        The runners do not re-decide. The vLLM bridge shares the constants and
+        the predicate, not this order -- its arms differ (no unified fallback,
+        no sliding-window arm). Both
         paged kernels stop short of unified, at different points, and drafting
         multiplies the query group into both limits. ASM is the one that must be
         checked here rather than inside its runner: past its envelope
         get_heuristic_kernel silently re-runs with mtp=1, a kernel built for
         another query length, and computes a wrong answer instead of refusing.
         """
+        # Clamped once here so both gates see the same value; the predicate
+        # clamps internally too, and a sentinel would otherwise split them.
+        max_qlen = max(1, int(max_qlen))
         over_gluon = gluon_decode_over_limit(
             max_qlen, self.num_heads, self.num_kv_heads
         )
         over_asm = (
-            int(max_qlen) * (self.num_heads // self.num_kv_heads)
+            max_qlen * (self.num_heads // self.num_kv_heads)
             > PA_ASM_MAX_QUERY_GROUP_SIZE
+        )
+
+        # unified takes any shape; the two paged kernels do not. Kept as one
+        # expression because a sliding-window layer needs the same answer, and
+        # it returns before the env checks below.
+        wants_unified = (
+            envs.ATOM_USE_UNIFIED_ATTN or self.use_flash_layout or over_gluon
         )
 
         # Sliding-window layers must use triton (ASM paths don't support it)
         if self.sliding_window != -1:
             return (
                 self.paged_attention_unified
-                if over_gluon
+                if wants_unified
                 else self.paged_attention_triton
             )
 
@@ -904,17 +924,20 @@ class PagedAttentionImpl(nn.Module):
         if envs.ATOM_USE_UNIFIED_ATTN:
             if envs.ATOM_FORCE_ATTN_TRITON:
                 return self.paged_attention_unified
-            if atom_config.kv_cache_block_size == 256 and not over_asm:
+            if atom_config.kv_cache_block_size == 256:
                 return self.paged_attention_persistent_asm
             return self.paged_attention_unified
 
-        if self.use_flash_layout or over_gluon:
+        if wants_unified:
             return self.paged_attention_unified
         if self.use_triton_attn:
             return self.paged_attention_triton
 
-        if use_pa_decode_bf16_asm() and not over_asm:
-            return self.paged_attention_persistent_asm
+        # use_pa_decode_bf16_asm() requires ATOM_USE_UNIFIED_ATTN, which the
+        # block above has already returned on, so it cannot be reached here.
+        # Only run_pa_fwd_asm is bounded here. The persistent paths above call
+        # pa_persistent_fwd / pa_decode_bf16_asm, different kernels with their
+        # own tables, so this envelope does not describe them.
         if over_asm:
             return self.paged_attention_triton
         return self.paged_attention_asm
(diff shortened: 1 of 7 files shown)


valarLip:
Reviewed at `9403dc9bb`, focused on cleanliness, performance, and where each piece of logic belongs.

Checked and cleared, so they don't get re-raised: the pow2 model in `gluon_decode_over_limit` is exactly equivalent to aiter's `define_layout` derivation (`pa_decode_gluon.py:382-390`), is never laxer than aiter's raw `equivalent_query_group_size <= 64` assert, and catches shapes that assert misses (ratio 17–21 at qlen 3 genuinely blow up the layout table); ASM using the raw product while gluon uses pow2 is correct, not an inconsistency. `_decode_score_chunks` is monotonic and there is no grid-dim overflow. Both new tests have real revert power. ruff and black are clean, and the modified test file passes 111/111.

The comment introducing the constants is also **correct**, and I want to record the arithmetic since it is not obvious: MiniMax-M3 is `num_attention_heads=64`, `num_key_value_heads=4`, so at tp4 the ratio is `16/1 = 16`; `gluon_decode_over_limit(4, 16, 1)` gives `group_p2 = 64`, exactly at `PA_GLUON_MAX_QUERY_GROUP_SIZE` and not over, while `(5, 16, 1)` gives 128 and trips. "M3 sits exactly on the group one today (16 × 4 draft positions)" holds, and so does the implication that one more draft token would trip it. Note this applies to M3's **first three layers** — `sparse_attention_freq` starts `[0, 0, 0, 1, ...]`, so those use `MiniMaxM3Attention` (`minimax_m3.py:377`, no `impl_cls`) and the base `PagedAttentionImpl`, while the other 57 use `MiniMaxM3SparseAttention` and never reach this path.

---

# Boundaries

## A dispatch decision was pushed down into a kernel runner

`dispatch_backend` delegates decode to `_dispatch_decode` (`attention_mha.py:925`), which is the single place that answers "which decode kernel runs". After this PR its answer is no longer authoritative — `paged_attention_triton` decides internally whether to run `run_pa_decode_gluon`, `unified_attention`, or raise:

```python
if gluon_over_limit and k_scale is not None and k_scale.numel() > 1:
    raise NotImplementedError(...)
if envs.ATOM_USE_UNIFIED_ATTN or self.use_flash_layout or gluon_over_limit:
    unified_attention(...)
else:
    run_pa_decode_gluon(...)
```

This module already has a clean pattern for "this impl dispatches differently": `SparseMHAPagedAttentionImpl` overrides `dispatch_backend` **wholesale**, replacing the entire backend selection rather than half of it. The new condition goes the other way, so the inconsistency is with the module's own existing convention, not with a style preference.

Two concrete consequences:

- The plugin could not reuse the rule, so `_dispatch_decode_backend` re-derives the same decision separately at `layer_mha.py:726-745`. One fact, two implementations — and they have already diverged (see the ASM guard below).
- `gluon_decode_over_limit(max_seqlen_q, num_heads, num_kv_heads)` is a static property of the layer and the spec-token count, now recomputed per layer per decode step on a host path this repo has spent PRs shortening.

Putting the `gluon_over_limit` term in `_dispatch_decode` keeps the answer in one place and lets the plugin share it. (A function named `paged_attention_triton` that may not run the triton paged kernel is a secondary problem that goes away with it.)

## The plugin imports kernel-envelope constants from a server-mode implementation module

```python
# atom/plugin/vllm/attention/layer_mha.py:12
from atom.model_ops.attention_mha import (
    PA_ASM_MAX_QUERY_GROUP_SIZE,
    gluon_decode_over_limit,
)
```

`run_pa_decode_gluon` and `run_pa_fwd_asm` both live in `base_attention.py`, which **both files already import**. Constants describing those kernels' envelopes belong next to the wrappers they describe; the plugin would then depend only on the shared layer.

As written it is more pointed than that: `PA_ASM_MAX_QUERY_GROUP_SIZE` appears exactly once in `attention_mha.py` — its own definition on line 43. It is never read in the module that defines it, only across the plugin boundary.

## The PR body should state the dependency between its two halves

The diff carries the index-score chunk split, the gluon query-group fallback, and `device=self.device` on `kv_scale`. The third is not mentioned in the body, and it is not cosmetic — it is a **prerequisite** for the second.

`self.kv_scale` feeds `k_descale` / `v_descale` at `:552-553` and `:880-881`, and the new `gluon_over_limit` term at `:528` is the first env-independent way into that branch. Before this PR, reaching `unified_attention` required `ATOM_USE_UNIFIED_ATTN` or `use_flash_layout`, neither on by default. I traced `k_descale` through aiter: it is forwarded unchanged from `unified_attention`'s signature to `k_descale_ptr=params.k_descale` at four kernel launch sites, with no `.to(device)`, no `.item()`, no None substitution. So the device fix is what keeps the new fallback from handing a host pointer to a Triton kernel.

One sentence in the body records that coupling. Without it, the next reader sees an unrelated one-line cleanup and may revert it on its own.

---

# Fix-then-sweep: two identical patterns fixed on one side only

## The ASM query-group guard exists only in the plugin

The plugin got it:

```python
asm_group = int(max_qlen) * (self.num_heads // self.num_kv_heads)
if asm_group > PA_ASM_MAX_QUERY_GROUP_SIZE:
    return self.paged_attention_triton
return self.paged_attention_asm
```

Server-mode `_dispatch_decode` still ends with a bare `return self.paged_attention_asm`. The failure mode is real: in `csrc/py_itfs_cu/asm_pa.cu:118`, `get_heuristic_kernel` searches

```cpp
const std::vector<int> mtp_flags = (mtp > 0) ? std::vector<int>{mtp, 1} : std::vector<int>{0};
```

so an unmatched mtp silently re-runs with `mtp=1` — a kernel built for a different query length. Wrong attention output, no error. That is exactly what the PR's own comment at `layer_mha.py:727-729` describes.

By geometry, M3's three dense layers land in this range themselves (tp4 ratio 16 × 4 draft positions = 64 > 16, and `head_dim == 128` keeps `use_triton_attn` False), which would make them a named reachable case rather than a hypothetical one — **please confirm**, since I cannot tell from the outside which env the reference runs used.

## `kv_scale` was moved to the device in one file and not the other

```
attention_mha.py:104   torch.tensor(f, dtype=torch.float32, device=self.device)   ← fixed here
layer_mha.py:149       torch.tensor(f, dtype=torch.float32)                       ← same idiom, untouched
```

The plugin copy is live. At `:807` it is aliased into `self.per_tensor_scale` (just before `self.kv_scale` itself is overwritten by a device tensor), read back at `:344` as `k_scale = v_scale = getattr(self, "per_tensor_scale", None)`, and passed to `fused_qk_rope_reshape_and_cache` with `apply_scale=self.kv_cache_dtype.startswith("fp8")` — true for fp8, so `HAVE_K_SCALE` is set and the Triton kernel dereferences a host pointer.

`self.device` is already set at `:142`, seven lines earlier, so this is a one-line addition. (`minimax_m3_attnetion.py:202` has the same idiom but is overwritten at `:331`, so that one really is cosmetic.)

---

# The new `NotImplementedError` is keyed on the wrong predicate

```python
if gluon_over_limit and k_scale is not None and k_scale.numel() > 1:
    raise NotImplementedError(...)
if envs.ATOM_USE_UNIFIED_ATTN or self.use_flash_layout or gluon_over_limit:
    ...  # unified_attention, k_descale = self.kv_scale (one per-tensor value)
```

The hazard the guard describes — per-token KV scales silently replaced by a single per-tensor descale — belongs to *taking the unified branch*, and there are three ways in. The guard checks one:

- **False negative.** `ATOM_USE_UNIFIED_ATTN=1`, fp8 per-token KV, `max_qlen <= 4`: the unified branch runs anyway, the per-token scales are replaced, the guard does not fire. Precisely the "not an error downstream, just wrong numbers" case the comment names — and M3's three dense layers sit at `max_qlen <= 4` under that env.
- **False positive.** Same env with `max_qlen > 4` now raises, where before the PR it ran.

The condition wants to be `(ATOM_USE_UNIFIED_ATTN or use_flash_layout or gluon_over_limit) and k_scale.numel() > 1` — keyed on the branch actually taken. It also checks only `k_scale`, while the gluon branch below (`:583-585`) requires both `k_scale` and `v_scale` to be per-token.

---

# Performance: removing the chunk ceiling changes the shape of the grid, not just its size

I re-derived `_decode_score_chunks` under both the old and new constants and swept it:

```
ctx      batch  old  new    x     old CTAs   new CTAs
32768    1      64   86     1×    64         86          ← the PR body's example
32768    256    8    86     11×   2048       22016
131072   256    8    342    43×   2048       87552
315000   64     32   822    26×   2048       52608
315000   256    8    822    103×  2048       210432
```

The number to look at is the middle column: **the new chunk count is 86 / 342 / 822 regardless of batch.** The old rule shrank the split as the batch grew (64 → 32 → 8), which is what kept total CTAs bounded; the new rule is batch-blind across the whole practical range, so the grid now scales linearly with concurrency with no ceiling. That is the property `MAX_CHUNKS=64` was providing.

The PR body's counter-example ("32K/batch=1 yields 86 chunks") lands on the one cell where new and old agree.

There is a second multiplier. `index_topk.py`'s own comment states that decode capture is handed `max_seqlen_k = max_model_len` so the grid covers every replay — so under CUDA graph `max_block` is always the max-model-len value, never the live context. A 32K request in a 315K deployment still launches 822 chunks, of which roughly 735 hit the `if chunk_start >= chunk_end: return` early-out at `:695`.

Worth a `conc >= 64` sweep and one short-request-in-a-long-deployment point before the cap comes off.

---

# The new constant comment states something that is false

```python
# Bounds on the decode score split (_decode_score_chunks). MIN_BLOCKS is the
# only one that binds; the other two are kept as escape hatches.
```

Exhaustively over the legal domain (`max_block < 0xFFFF`, per `_require_packable`), across batches {1, 2, 8, 64, 256, 1024}:

```
DECODE_SCORE_MAX_CHUNKS  changes the result in       0 / 393,204 cases
DECODE_SCORE_TARGET_GRID changes the result in 132,090 / 393,204 cases
```

Both halves are wrong. `MAX_CHUNKS` is not an escape hatch, it is provably dead — with `MIN_BLOCKS = 3`, `chunks <= cdiv(max_block, 3) <= 21845 < 65536` always — and should be deleted rather than kept as a knob. `TARGET_GRID` is the one still doing work: `_decode_score_chunks(256, 65534)` is 4096 with it and 21845 without.

Two doc sites went stale with it:

- `_decode_score_chunks`' docstring still reads "capped so a large batch does not multiply into a pointless grid". That cap is now dead, and the blocks-per-chunk floor that actually governs appears only in an inline comment three lines below.
- `docs/model_ops_guide.md:146` maps `use_triton_attn=True` → `paged_attention_triton` → `torch.ops.aiter.pa_decode_gluon`. That function may now run `unified_attention` or raise instead, and the new query-length / query-group envelope is documented nowhere. The `pa_decode_gluon` rows at `:15` and `:48` were not touched either.

---

# Tests: none, for logic that needs no GPU

`grep -rn 'gluon_decode_over_limit|_dispatch_decode_backend|PA_GLUON_MAX|PA_ASM_MAX' tests/` returns zero hits, and no test in `tests/` instantiates `PagedAttentionImpl`. A test under `tests/plugin` would not substitute: `.github/scripts/run_unit_tests.sh:23` carries `--ignore=tests/plugin`, so it never runs in CI.

`gluon_decode_over_limit` is a three-line pure function. Four cases cost nothing, and I ran them:

```
(4, 16, 1)  M3 dense at tp4, MTP-4 → False   exactly at the limit
(5, 16, 1)  one more draft token   → True
(5,  8, 1)  gqa=8                  → True
(3, 17, 1)  ratio 17               → True    layout table has no arm for it
```

Two edge cases in the same run are worth fixing while the function is open:

```
(-5, 8, 1)  → False               a negative or sentinel qlen is silently let through
( 0, 8, 1)  → False               treated as 2
( 4,16, 0)  → ZeroDivisionError
```

`(-6).bit_length() == 3`, so a negative `max_qlen` yields `qlen_p2 = 8`, `group_p2 = 64`, and passes both limits. `max(1, int(max_qlen))` on the first line fixes the first two. The ratio divide needs no guard only because callers never pass `num_kv_heads == 0`, which is not stated anywhere.

Finally, `PA_ASM_MAX_QUERY_GROUP_SIZE = 16` reroutes existing models from ASM to gluon: any plugin MHA model with `max_qlen * (num_heads // num_kv_heads) > 16` now gets `paged_attention_triton` where it previously got `paged_attention_asm`. M3 at tp4 with a 4-token draft is itself in that set (`4 × 16 = 64`), so the PR's M3 testing is relevant coverage here. The value has upstream backing — `asm_pa.cu:113-116` carries `// # mtp * gqa <= 16` as a source comment — so the constant is not in question; what is missing is whether the reroute's **performance** delta was measured, or is simply accepted because the alternative is silently wrong output. Either is fine; the PR should say which.

---

# Minor: two contracts for the same value

`_dispatch_decode_backend` is now called with `attn_metadata.decode_metadata.max_query_len` (`:933`), which requires non-None, while `paged_attention_triton` (`:428`) and `paged_attention_asm` (`:516`) still compute `max_qlen = decode_metadata.max_query_len if decode_metadata is not None else 1` for the same value on the same call. Either the defensive default is dead and can go, or the dispatcher needs it — and as it stands `max_qlen` is read twice per decode step from the same source.

yitingw1:
All ten items addressed. Fixed: §1/§2 (envelope moved into `_dispatch_decode`, constants to `base_attention.py`), §4.1 and §4.2 (both real bugs — confirmed), §5 (guard now keyed on the branch taken, checks both scales), §6 (grid ceiling restored; swept batch 1–128 plus the short-request case), §7 (`MAX_CHUNKS` deleted, docs updated), §8 (28 dispatch tests), §3/§9/§10.

Not fixed: the per-step envelope recompute. It *is* hoistable — the predicate is monotone in `max_qlen`, so it inverts to a per-layer threshold set at `__init__` (verified exact over ~800 layer shapes). Skipped because only M3's 3 dense layers reach this path, ~3 calls/step against a GPU already ~100% busy, and it would add a derived attribute to keep in sync on both sides of the plugin boundary.

valarLip (approved):

PR #2165, opened 2026-09-08
Expose streaming ITL and API TTFT histograms with latency reports

Jasen2201:
[atomesh-latency-glm-5.2-mxfp4-glm-52-mxfp4-1p1d-tp4-agentic-1m-c8-nightly-1](https://github.com/ROCm/ATOM/actions/runs/34226255170/job/102061094480)
image

image


valarLip:
Reviewed commit `b24b85e2c7a19b9259619574769811e520b7a883`, with particular attention to hot-path cost, cohesion, and separation of responsibilities. I recommend addressing the following before merging.

### 1. [P1] Replace per-token histogram observations with a weighted update

[metrics.py:471–481](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/atom/entrypoints/openai/metrics.py#L471-L481)

`observe_inter_token_latency()` computes one value, then calls `Histogram.observe()` N times for N new tokens. This repeats bucket lookup, locking, and counter updates for identical observations, synchronously inside `_deliver()` on the frontend event loop.

I measured the actual `_deliver()` implementation in a Python 3.12.3 container, using a no-op collector to isolate delivery and instrumentation cost. Median callback times for a batch of 4,096 streams were:

| Tokens per stream | Observer disabled | Observer enabled |
|---|---:|---:|
| 1 | 0.30 ms | 2.94 ms |
| 8 | 0.30 ms | 12.83 ms |
| 64 | 0.30 ms | 88.81 ms |

These are CPU microbenchmarks, not end-to-end serving measurements. However, the callback cannot yield during this work, so the added time blocks other event-loop work. The instrumentation should not introduce a token-count-dependent pause into the path it measures.

Please give the metrics layer a weighted-observation abstraction: identify the bucket once, add N to its count, and add `interval` to the sum. Keep Prometheus implementation details out of the dispatcher and avoid directly mutating private histogram fields there.

### 2. [P2] Separate report collection, finalization, and rendering; preserve the benchmark exit code

[collect_metrics.py:266–326](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/.github/scripts/atomesh/observability/collect_metrics.py#L266-L326) / [export_report.py:281–314](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/.github/scripts/atomesh/observability/export_report.py#L281-L314)

`generate_report()` collects data **and writes HTML**. The caller then adds notes/status and calls `write_report()` again. This renders the same report twice and mixes data acquisition with publication side effects. The final write is also outside the export exception handler.

I reproduced the failure boundary with an isolated benchmark that exits 17 and an injected `OSError` from the HTML renderer: the benchmark ran, but `run()` raised the renderer exception instead of returning 17. At the CLI this replaces the benchmark's exit status with a reporting failure.

Please make collection return data, finalize report metadata/status in one place, and render once. Report publication errors should be surfaced separately while retaining the benchmark's original exit semantics. A single owner for diagnostics would also avoid copying the same query errors back and forth between `meta.notes` and `status.errors`.

### 3. [P2] Move Mesh construction out of observability and router startup

[pd_server_atom.sh:630–636](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/.github/scripts/atomesh/pd_server_atom.sh#L630-L636) / [build_mesh.sh:17–28](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/.github/scripts/atomesh/observability/build_mesh.sh#L17-L28)

For agentic benchmarks, adding latency reporting now also triggers a source copy and Cargo build inside `start_router()`, replacing the existing `release` binary with a `ci` profile binary. Those profiles have different optimization levels, LTO settings, and codegen-unit counts. The default temporary target directory also prevents reuse of previous build outputs.

Building the reviewed source is useful, but it belongs in an explicit CI build/setup stage. That stage should produce a binary with a recorded commit and profile, and `start_router()` should consume its path. Observability should not implicitly select how the measured service is built. Currently a build failure blocks service startup, and enabling this reporting path also changes the benchmark subject.

### 4. Structural change: keep delivery timing out of the detokenizer

[streaming_dispatch.py:62–63](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/atom/entrypoints/openai/streaming_dispatch.py#L62-L63) / [streaming_dispatch.py:338–350](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/atom/entrypoints/openai/streaming_dispatch.py#L338-L350)

`IncrementalStreamDetokenizer.last_output_at` is not decoding state; its own comment says only frontend delivery uses it. The dispatcher now mutates instrumentation state stored inside a component responsible for token-to-text conversion.

A small composed stream state containing the detokenizer and a separate timing state would keep ownership clear. The dispatcher can report delivery events through the existing callback abstraction, while timing owns its clock and metrics owns aggregation. This does not require a new file for every class, but the responsibilities should be explicit.

### 5. [P2] Make fragmented SSE detection linear in input size

[request_timing.py:60–80](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/atom/entrypoints/openai/request_timing.py#L60-L80) / [ttft.rs:146–164](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/atom/mesh/src/observability/ttft.rs#L146-L164)

The Python detector repeatedly copies the accumulated immutable buffer and searches it from the beginning. The Rust detector avoids the append copy, but also rescans the full accumulated prefix on each fragment. Both need a retained search cursor; changing Python's buffer to `bytearray` alone does not fix the repeated scans.

For the Python implementation, a roughly 1 MiB event took 1.36 ms when delivered whole versus 108.82 ms in 1 KiB fragments. This is a fragmented-input stress case, not the latency of a typical short SSE event. It demonstrates the quadratic path within the current buffer limit. I have not benchmarked the Rust implementation.

Please scan only newly appended bytes plus the small overlap needed for delimiters spanning fragments, and consume frames using offsets rather than repeatedly shifting/copying the remainder. Keep SSE framing and generated-output classification behind clear interfaces to the timing code.

### 6. Maintainability: separate editable report sources from the standalone output artifact

[report.html](https://github.com/ROCm/ATOM/blob/b24b85e2c7a19b9259619574769811e520b7a883/.github/scripts/atomesh/observability/report.html)

The report template is approximately 27 KB in 117 lines, including a CSS line of 8,680 characters. Styles, interaction state, canvas rendering, table rendering, and CSV export are compressed together. That makes ordinary changes difficult to review and gives unrelated UI behavior the same maintenance surface.

A single-file offline **output** is appropriate. Please keep the editable HTML/CSS/JavaScript formatted and separated by responsibility, then inline them during export. There is no need to add a large frontend framework to achieve this.

I would organize this work into three independently reviewable pieces: runtime latency instrumentation, CI build/collection lifecycle, and offline report rendering. The API-server wiring and request-local `ContextVar` approach are reasonable foundations; the boundaries above need tightening.

**Validation:** 96 focused Python tests passed; one real-Prometheus integration test was skipped because the container does not have Prometheus installed. Cargo is also unavailable, so the Rust changes received static review only. No GPU end-to-end performance benchmark was run.

valarLip (approved):

valarLip (approved):

PR #2166, opened 2026-09-08
kv_transfer: release M3 offload blocks incrementally

yhl-amd:
## Summary

- transfer ownership only for exact M3/dense PAGE blocks still read by an offload save, allowing decode and unrelated blocks to be released at request completion
- release leased GPU blocks after CUDA-event-backed source-safe TP/PP quorum, independently from backend store commitment
- schedule save-side GPU staging tail-to-head while preserving MemoryObj/range identity; load remains head-to-tail
- handle pending final saves, stale generations, request-ID reuse, failures, cancellation, timeout, and abnormal reclamation
- keep the feature behind ATOM_OFFLOAD_EARLY_BLOCK_RELEASE, disabled by default and explicitly excluded from Kimi-K3/DSV4 state paths

## Validation

- 617 early-release and connector regression tests passed
- 81 BlockManager tests passed; 1 environment-dependent test deselected
- 2 ROCm LMCache GPU disk round-trip tests passed
- focused tail-to-head scheduling and mapping tests passed
- Black, Ruff, and git diff --check passed

## LMCache boundary

GPU-to-staging and source-safe release run tail-to-head. LMCache 0.4.5 keeps backend keys private inside CacheEngine.store, so storage_manager.batched_put remains one opaque original-order batch; this change does not introduce per-chunk synchronous engine.store calls.


atom/kv_transfer/offload/_offload_common.py
```diff
@@ -690,3 +740,20 @@ def offload_save_abandon_timeout_s() -> float:
...
+    still in flight keeps its *entire* block table pinned via
+    `deferred_free_blocks` until the save reports. Enabling
+    `ATOM_OFFLOAD_EARLY_BLOCK_RELEASE=1` narrows that to only the block IDs the
+    pending save actually reads (see `DenseOffloadScheduler.protected_block_ids`
+    / `BlockManager.deallocate_partial`); every other block (decode blocks,
+    unaligned prompt tail, already-saved ranges) is freed at request-finish
+    time instead of waiting on the save. Scoped to dense/M3 PAGE-only for now
+    -- DSV4/K3 keep the pre-existing whole-table defer.
+    """
+    raw = os.environ.get("ATOM_OFFLOAD_EARLY_BLOCK_RELEASE", "0").strip().lower()
```
yhl-amd: default open delete this param

valarLip:
Reviewed commit `c98af780226a786f7eef0fd72526788f4e00e073`, focusing on performance, code cleanliness, cohesion, and ownership boundaries. I recommend addressing the following before merging.

Separating source safety from store completion is useful, and the exact save-generation identities, TP quorum, explicit BlockPool reference claims, and preservation of MemoryObj/range identity are justified. The lifecycle and bookkeeping around them need further work.

**1. [P2] Retire partial source-safe quorums when their save operation completes**

[_block_gpu_connector.py:548–582](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/_block_gpu_connector.py#L548-L582) / [dense/connector.py:1054–1077](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/dense/connector.py#L1054-L1077)

Source-safe reports are emitted only for chunks actually staged on each rank. `_store_finished()` explicitly accommodates a successful store that skipped staging because LMCache already had the chunk. With asymmetric cache hits across TP ranks, one rank can report source safety while another reports only the store terminal.

The successful store quorum releases the lease, but its partial source-safe quorum remains indefinitely in `KVOutputAggregator._connector_completions._reports`. There is no operation-level retirement linking the terminal event to its intermediate events. Missing callbacks after an event-query failure can leave the same kind of residue.

I reproduced this through the production scheduler/aggregator APIs: rank 0 reports source safety and successful store completion; rank 1 reports successful store completion only. Both leased blocks are released and the scheduler has no pending work, but the aggregator's `pending_count` remains `(0, 1)`. After 8,192 completed operations, 8,192 partial source quorums remain. The 4,096-entry tombstone limit does not bound pending reports, and an otherwise empty aggregate takes approximately 0.584 ms because `drain()` scans them all.

Please add operation-level retirement and late-event suppression. A successful store terminal can retire its source milestones; abandonment must invalidate tracking without treating a failed store as proof of source safety. Add coverage for asymmetric rank cache hits and missing source callbacks.

**2. [P2] Avoid global operation scans in request-local scheduler decisions**

[dense/connector.py:807–827](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/dense/connector.py#L807-L827) / [dense/connector.py:912–939](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/dense/connector.py#L912-L939)

`should_defer_free(seq)` scans all operation owners before checking the cheap `_save_inflight` membership condition. Checking N requests with N outstanding saves therefore becomes quadratic. This predicate is used by scheduler preemption and request teardown. `protected_block_ids()` also scans all operations, while `reclaim_stale_leases()` repeats global scans for each expired lease.

I measured the actual scheduler methods in Python 3.12.3, with one outstanding save per request. Query sweep timings are medians of five runs; the previous-predicate column evaluates the equivalent pre-change conditions:

| Outstanding saves | New query sweep | Equivalent previous predicate | Reclaim all expired leases |
|---:|---:|---:|---:|
| 512 | 2.702 ms | 0.071 ms | 12.959 ms |
| 2,048 | 38.011 ms | 0.297 ms | 199.279 ms |
| 4,096 | 149.351 ms | 0.786 ms | 800.622 ms |

These are synthetic CPU scaling measurements. They do not establish end-to-end serving latency, but they expose avoidable work on the scheduler thread. Dense/M3 does not bound emitted saves through `_may_emit_save()`, so the outstanding operation count is not limited to the save executor's thread count.

Please maintain a direct request-owner-to-operations index, short-circuit cheap state checks first, and make timeout cleanup proportional to the entries being reclaimed.

**3. [P2] Avoid cross-stream head-of-line blocking and make poller shutdown deterministic**

[_SourceSafeEventPoller:120–164](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/_block_gpu_connector.py#L120-L164)

The poller queries only `_pending[0]`. With multiple `OFFLOAD_COPY_WORKERS`, events come from independent CUDA streams and have no global completion order. An incomplete event from one worker prevents ready groups from another worker from reporting source safety, delaying their early reclamation.

In a controlled fake-event reproduction, the unready head was queried 48 times over 50 ms while the ready second event was never queried or delivered. `close()` then returned after approximately one second with the polling thread still alive: the non-ready branch never checks `_closed`. This shutdown case requires an event that remains incomplete; healthy workers whose streams have synchronized normally avoid it.

Please poll independent stream heads fairly, or otherwise process ready events without a global FIFO dependency. Shutdown should exit deterministically without marking unready events safe or leaving a daemon retaining the callbacks and queue.

**4. Structural change: give lease lifecycle management a cohesive owner**

[dense/connector.py:439–454](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/dense/connector.py#L439-L454)

The new lifecycle is represented by seven parallel dictionaries and a pending-release list inside `DenseOffloadScheduler`. Save emission, source completion, store completion, timeout, and abandonment each update different subsets. Their invariants are difficult to audit because ownership, release eligibility, cleanup, and statistics are maintained across multiple methods.

Please extract a focused lease tracker with explicit per-operation and per-request records, centralized transitions, and the indexes needed for request-local queries. Keep the connector responsible for adapting metadata/completions and the BlockManager responsible for reference counts. Freeze the finished-request snapshot in an explicit lifecycle transition; `protected_block_ids()` currently also mutates `Sequence`, which obscures its query contract.

**5. Structural change: express block leasing as a complete capability**

[scheduler.py:2918–2929](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/model_engine/scheduler.py#L2918-L2929) / [offload/connector.py:205–230](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/connector.py#L205-L230)

The scheduler, offload shell, and MultiConnector forward related lease methods through independent `getattr` probes. The scheduler transfers BlockPool ownership and then treats `activate_block_leases()` as optional. Current supported implementations provide these methods, but the interface does not enforce the ownership invariant.

Please define a complete lease capability, following the existing `StateOffloadFace` precedent, and validate it before transferring ownership. Protection, activation, completion draining, and reclamation need to form one supported contract.

**6. Scope boundary: make the staging-order policy explicit**

[_block_gpu_connector.py:635–658](https://github.com/ROCm/ATOM/blob/c98af780226a786f7eef0fd72526788f4e00e073/atom/kv_transfer/offload/_block_gpu_connector.py#L635-L658)

`batched_from_gpu()` unconditionally requests tail-to-head sorting and installs the source-safe hook, including when early release is disabled. DSV4 also uses this shared connector, so the staging-order change extends beyond the enabled dense/M3 lease path.

Please either scope the policy to the capability or document and validate it as an independent behavior change. Disabling early release currently does not restore the previous staging order. No GPU throughput regression from that ordering was measured in this review.

**Validation:** 648 tests passed in `ljin_dev`: 531 connector/aggregator/BlockManager tests and 117 scheduler tests. One existing scheduler test module was skipped because its imports need migration to the moriio package. Both GPU/disk integration tests were skipped because `lmcache` is not installed. ROCm Torch and MI355X GPUs are available, but real LMCache GPU/disk execution and end-to-end throughput remain unverified. The quorum and scheduler findings above were reproduced through production Python APIs; the poller reproduction used controlled fake events.

ZhangLirong-amd (dismissed):

PR #2175, opened 2026-09-09
M3 cpu offload cache optimization

ZhangLirong-amd:
## Motivation

Improve MiniMax-M3 prefix-cache reuse under 48-concurrency, CPU-only KV offload.

The previous configuration forced LMCache lookup to rank 0. Under capacity pressure, rank 0 could report a hit while another TP rank had already evicted the corresponding KV shard. An incomplete cross-rank restore then forced the request to fall back to prefill.

## Technical Details

- Honor explicit LMCache lookup-worker configuration instead of unconditionally overriding it with rank 0. The tested M3 configuration sets `LMCACHE_LOOKUP_SERVER_WORKER_IDS=0,1,2,3` so all TP shards participate in lookup, recency updates, and pinning. The minimum hit length across ranks represents a prefix that can be restored by every rank.
- Fix lookup-pin lifecycle handling for misses, exceptions, HBM-satisfied requests, skipped loads, and cancellation. Reuse pending lookup results to avoid duplicate pin acquisition, while retaining pins for actual loads until retrieval completes.
- Add opt-in segmented LRU (SLRU) policies for HBM and CPU caches to protect reused prefixes from one-off prefill scans. CPU SLRU uses separate ordered probationary/protected queues, avoiding repeated scans of the protected segment during probationary eviction.
- Cancel lookup ownership when a waiting request is aborted before HBM allocation; dispatch worker cleanup and release request bookkeeping. Pre-allocation lookups retain pins without advertising non-dispatchable idle work.
- Reject `LMCACHE_ENABLE_ASYNC_LOADING=true` (including connector extra-config overrides) at startup. This connector supports synchronous lookup only; it does not implement the async polling/cancellation protocol.
- Normalize HBM policy names and the `ATOM_SLRU` option for case and surrounding whitespace.
- Add regression coverage and documentation for the cache policies and 48C/1800s benchmark comparison.

The shared defaults remain unchanged: HBM/CPU eviction stays LRU and lookup defaults to rank 0 unless explicitly configured. The pin-lifecycle fix applies to the shared DenseOffloadScheduler path independently of the SLRU switches.

## Test Plan

- Run the block-pool, block-manager, dense-offload, LMCache configuration/connector, cache-policy, offload-shell, Scheduler, and MultiConnector regression suites on the main-based branch.
- Compare rank-0-only lookup with all-rank lookup using MiniMax-M3 FP4, TP4, FP8 KV, 48 concurrent requests, 1800 seconds of profiling, CPU cache capacity of 256 GiB per rank, no NVMe, and `SYNTH_ACC_RATE=0.5933`.
- Keep both SLRU policies, the pin-lifecycle fix, and the workload unchanged between the two benchmark runs; change only lookup scope.

### AIPerf comparison

| Metric | Rank 0 only | All-rank |
|---|---:|---:|
| Total prompt cache hit rate | 82.86% | **95.48%** |
| Total throughput (tokens/s) | 86,770.56 | **171,373.84** |
| P90 ITL (ms) | 66.94 | **24.35** |

All-rank lookup improved the total prompt cache hit rate by **12.62 percentage points** and increased total throughput by approximately **1.98x**. With TP4, the all-rank result corresponds to **42,843.46 total tokens/s/GPU**. The final AIPerf submission was valid (`submission_valid=true`).

This comparison isolates the lookup-scope change within the tested configuration. It does **not** establish the independent benefit of SLRU, since both runs enabled SLRU and the pin-lifecycle fix.

### Regression validation and measurement scope

**604 tests passed; 1 known environment-dependent test was deselected.** The excluded test assumes LMCache is absent and also fails on the original checkout in this production container, where LMCache is installed. Black, Ruff, and diff checks passed.

Regression coverage now includes the real lookup-hit → HBM-allocation-wait → abort → worker-unpin path, async-mode rejection through environment and extra config, and deterministic eviction-scan bounds at 8 / 1,024 / 10,000 resident chunks. Candidate retries, pin/unpin transitions, reinsertion, and stale queue cleanup are also covered.

The performance comparison was measured before rebasing onto main and before these review fixes. The updated branch passed regression tests; GPU performance has not been remeasured.

## Submission Checklist

### DIFF
--- a/atom/kv_transfer/offload/cache_policy.py
+++ b/atom/kv_transfer/offload/cache_policy.py
@@ -1,0 +1,83 @@
+# SPDX-License-Identifier: MIT
+# Copyright (C) 2024-2026, Advanced Micro Devices, Inc. All rights reserved.
+
+"""Opt-in scan-resistant eviction for LMCache's CPU tier."""
+
+from collections import OrderedDict
+from typing import Any
+
+from lmcache.v1.storage_backend.cache_policy.lru import LRUCachePolicy
+
+
+class SLRUCachePolicy(LRUCachePolicy):
+    """Protect reused chunks; single-use chunks compete in probationary LRU.
+
+    Protection covers at most half the resident chunks. ATOM M3 PAGE chunks
+    have a fixed byte size, so this also bounds the protected byte budget.
+    The CPU backend owns synchronization and calls these methods under its
+    cache lock. A pinned or referenced chunk is never an eviction candidate.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._probationary: OrderedDict[Any, None] = OrderedDict()
+        self._protected: OrderedDict[Any, None] = OrderedDict()
+
+    def update_on_hit(self, key: Any, cache_dict: OrderedDict) -> None:
+        """Promote a reused chunk and age out excess protection."""
+        super().update_on_hit(key, cache_dict)
+        self._probationary.pop(key, None)
+        self._protected[key] = None
+        self._protected.move_to_end(key)
+        while len(self._protected) > len(cache_dict) // 2:
+            demoted, _ = self._protected.popitem(last=False)
+            self._probationary[demoted] = None
+
+    def update_on_put(self, key: Any) -> None:
+        """Admit new data without displacing protected data's priority."""
+        super().update_on_put(key)
+        self._protected.pop(key, None)
+        self._probationary[key] = None
+        self._probationary.move_to_end(key)
+
+    def update_on_force_evict(self, key: Any) -> None:
+        """Forget a chunk explicitly removed by the storage backend."""
+        self._probationary.pop(key, None)
+        self._protected.pop(key, None)
+
+    def get_evict_candidates(
+        self, cache_dict: OrderedDict, num_candidates: int = 1
+    ) -> list:
+        """Choose from probation without scanning the protected segment.
+
+        Normal pressure removals do not issue a policy callback. Drop stale
+        keys lazily, but retain selected candidates until they are actually
+        removed: a caller may retry or decide not to evict them. With evictable
+        probationary entries this is amortized O(1) per victim; pinned entries
+        still require scanning because pin changes have no policy callback.
+        """
+        if num_candidates <= 0:
+            return []
+        candidates = []
+        for queue in (self._probationary, self._protected):
+            stale = []
+            for key in queue:
+                cache = cache_dict.get(key)
+                if cache is None:
+                    stale.append(key)
+                elif cache.can_evict:
+                    candidates.append(key)
+                    if len(candidates) == num_candidates:
+                        break
+            for key in stale:
+                queue.pop(key, None)
+            if len(candidates) == num_candidates:
+                break
+        return candidates
+
+
+def register_slru_policy() -> None:
+    """Register ATOM_SLRU through LMCache's public policy registry."""
+    from lmcache.v1.storage_backend.cache_policy import POLICY_MAPPING
+
+    POLICY_MAPPING["ATOM_SLRU"] = SLRUCachePolicy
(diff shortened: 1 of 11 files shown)


atom/kv_transfer/offload/dense/connector.py
```diff
@@ -725,7 +739,11 @@ def has_pending_work(self) -> bool:
         scheduler's ``deferred_free_blocks``, which ``should_defer_free``
         keeps populated for exactly those requests.
         """
-        return bool(self._reqs_need_recv) or bool(self._save_inflight)
+        return (
+            bool(self._reqs_need_recv)
+            or bool(self._save_inflight)
+            or bool(self._lookup_in_step)
```
yhl-amd: **[P1] Clear pre-allocation lookups when a waiting request is aborted.** A positive lookup adds the ID to both `_load_specs` and `_lookup_in_step` before allocation. If allocation is delayed and the waiting request is then aborted, `Scheduler._reject_aborted_waiting()` returns without calling `cancel_pending_load()` because the request was never counted as an inflight load. The ID is then permanently non-dispatchable in `build_connector_meta()` (`sid in self._load_specs`), while this method keeps `EngineCore.has_pending_kv_work()` true forever. The engine repeatedly runs the idle drain with empty metadata, and the worker pin survives until LMCache's timeout. Please cancel/clear connector lookup ownership on the pre-allocation abort path and add a regression test for lookup-hit -> allocation wait -> abort.

atom/kv_transfer/offload/dense/connector.py
```diff
@@ -428,8 +429,24 @@ def get_num_new_matched_tokens(self, seq) -> tuple[int, bool]:
...
+            # must be dispatched before the ID can acquire a new lease.
+            return 0, False
         try:
-            hit = self._lookup_client.lookup(token_ids, lookup_id=str(seq.id))
+            if pending is None:
+                if sid not in self._lookup_in_step:
+                    self._lookup_in_step.append(sid)
+                self._lookup_results[sid] = (seq, 0)
+                hit = self._lookup_client.lookup(token_ids, lookup_id=sid)
+                if hit is None:
```
yhl-amd, line 444: **[P2] Preserve the pending state for LMCache async lookup.** In LMCache v0.4.5, `LMCacheAsyncLookupClient.lookup()` returns `None` while the request is in progress, and subsequent probes are required to call `lookup_cache()`. Popping `_lookup_results` here makes the next probe issue another lookup; meanwhile `_lookup_in_step` has no `_load_specs` entry, so `build_connector_meta()` can dispatch cleanup before the async lookup finishes. This can produce duplicate requests/late pins and means `LMCACHE_ENABLE_ASYNC_LOADING` never yields a usable hit. Please retain an explicit pending state and poll `lookup_cache()`, or reject async loading for this connector at configuration time.

atom/kv_transfer/offload/cache_policy.py
```diff
@@ -0,0 +1,73 @@
...
+        self._protected.pop(key, None)
+
+    def get_evict_candidates(
+        self, cache_dict: OrderedDict, num_candidates: int = 1
+    ) -> list:
+        """Choose evictable probationary chunks before protected chunks."""
+        if num_candidates <= 0:
+            return []
+        candidates = []
+        for key, cache in cache_dict.items():
```
yhl-amd: **[P2] Avoid a full-cache scan for every probationary eviction.** After the original probationary entries have been consumed, protected entries can occupy the front half of `cache_dict` while each newly inserted probationary entry sits behind them. Every allocation under pressure then scans roughly the whole protected segment under `LocalCPUBackend.cpu_lock` before finding one candidate. On a large CPU cache this turns steady-state eviction into O(cache size) serialized work per chunk. Please maintain an ordered probationary set/queue, similar to the HBM `BlockPool`, so selecting its oldest evictable entry is amortized O(1).

yhl-amd (commented):
Reviewed the lookup/pin lifecycle and both SLRU implementations against the scheduler flow and LMCache v0.4.5 contracts. I found three issues that should be addressed before relying on this path under cancellation, async lookup, or sustained cache pressure.

ZhangLirong-amd:
@yhl-amd Thanks for catching these. Fixed in 3e4687b2:

- Pre-allocation aborts now release lookup ownership and dispatch pin cleanup.
- Async loading is explicitly rejected at startup until its polling/cancellation protocol is supported.
- CPU SLRU now uses separate probationary/protected queues to avoid scanning the protected segment on each eviction.

Added regressions: 604 passed, 1 known environment-dependent test deselected. Black/Ruff passed.

yhl-amd (approved):

PR #2186, opened 2026-09-10
Enhance metrics with scheduler, cache, and decoding improvements

Jasen2201:
## Motivation

  Existing TTFT and ITL metrics provide limited visibility into scheduling delays, cache reuse, workload shape, and GPU execution time. This PR adds these measurements and integrates them into the offline agentic benchmark report to help diagnose latency and throughput
  behavior across prefill and decode services.

  ## Technical Details

  - Scheduler metrics: Add queue time, successful PD KV transfer wait time, actual decode batch size, prefill token counts, and prefill/decode context-length distributions. Expose request queue states and KV block utilization, with block-state counts maintained
    incrementally.

  - GPU timing: Add optional per-worker target-model forward timing and cumulative prefill timing across each request’s initial chunks. Timing uses nonblocking device-event polling, bounded buffering, and event reuse. Request-level measurements include the full duration
    of each participating batch; PP measurements cover each local stage.

  - Cache accounting: Expose supplemental LMCache reuse and record the decode consumer’s local prefix reuse after successful remote KV receipt. Reports display total reuse, GPU reuse, and LMCache reuse against a shared input-token denominator.
  - Metrics organization: Centralize GPU timing, scheduler instruments, histogram helpers, Prometheus initialization/HTTP export, snapshot collection/rendering, and request/stream instruments in `atom/metrics/`. Keep API composition, request middleware, streaming dispatch, and `EngineStats` with their owning components. The API injects its live stream-silence callback into the exporter, which reads it on the event loop before rendering in a worker thread. Preserve metric names, buckets, labels, multiprocess initialization order, and observation semantics; update imports, tests, and the metrics guide.
  - Streaming instrumentation: Share SSE parsing between request logging and streaming TTFT detection, while preserving first-generated-output semantics.
  - Reports and CI: Add workload, cache, queue, and GPU panels; role/category filters; independent prefill/decode instance selectors; and corresponding table/CSV data. Aggregate distributions from histogram buckets and ratios from raw counts. Increase engine refresh and
    CI scrape cadence to one second.

  GPU timing is disabled by default and can be enabled with ATOM_ENABLE_METRICS_DEVICE_TIMER=1. Agentic CI enables it for both service roles, with environment overrides supported.
[[glm-5.2-mxfp4-glm-52-mxfp4-1p1d-cpp4-dcp4-agentic-lmcache-1m-c48-nightly](https://github.com/ROCm/ATOM/actions/runs/34489056541/job/102928136600#logs)](https://github.com/ROCm/ATOM/actions/runs/34489056541/job/102928136600)
image
image

  ## Test Plan

  Focused tests cover:

  - Scheduler timing lifecycle, cache accounting, KV block counts, and PP scheduler deduplication.
  - GPU event polling/reuse, chunk accumulation, incomplete measurements, and worker snapshot routing.
  - Histogram consistency, collector registration, concurrent scrapes, and SSE first-output detection.
  - Report generation, PromQL aggregation, per-instance data, and CI collection configuration.

  ## Test Result

  - Focused metrics, request timing, streaming, and report tests after the shared infrastructure migration: **288 passed, 2 skipped** in freshly pulled `rocm/atom-dev:latest`. Skips require a GPU or an explicitly configured Prometheus binary. Prometheus bootstrap import-order and package-independence checks also passed.
  - CI-report and GC regression checks after restoring readiness scripts: **83 passed, 1 skipped** (overlaps the focused suite). All four affected shell scripts pass `bash -n`.
  - Black and diff checks passed. Ruff introduced no findings; existing diagnostics in `engine_core.py` and `api_server.py` are unchanged.
  - Real inference with the migrated code in freshly pulled `rocm/atom-dev:latest` (2026-09-15 build, digest `sha256:38714402a1b5a408d2a57d8f9d5900afe31b33302375b0981ee34509e36c3773`): GLM-5.2 MXFP4, CPP4 prefill + TP4/DCP4 decode, Mooncake, MTP 3, prefix caching, LMCache CPU offload, and GPU timing enabled. The container's Prometheus client was upgraded from 0.23.1 to the PR-required 0.25.0.
  - **12/12 requests succeeded**, covering streaming, non-streaming, an 8,466-token prompt split into three prefill chunks, repeated-prefix reuse, and eight concurrent requests. Input/output totals were 26,028/1,216 tokens. **36 metric checks and 13 migration checks passed** with 33 live scrape rounds and eight concurrent scrapes per endpoint. Migration checks compare metric family names/types/help, sample names and label names against the prior live run, and verify GC and live stream-silence reporting.
  - Verified request/token accounting, all eight GPU workers, once-per-request prefill timing, 12 PD transfer observations, TTFT/ITL, histogram consistency, KV block partitions, idle stability, MTP counters, and zero exporter refresh errors. Warm-prefix reuse matched the response's 8,464 cached tokens; LMCache saved 8,448 tokens.
  - Mesh logged failed HuggingFace tokenizer lookups for the served-model alias; all backend requests and metric checks succeeded. The image also retains optional dependency upper-bound conflicts after installing the PR-required Prometheus client.
  - CPU-cache reload was not exercised because warm requests hit HBM. No full model-accuracy or comparative performance benchmark was run. Independent PD services currently emit `engine_role="default"`; the CI report distinguishes prefill/decode using the scrape target's `role` label.

  GPU telemetry is asynchronous. The live test verified settled idle observations; ensuring final observations are included before every report collection stops remains a follow-up.


atom/utils/histogram.py
```diff
@@ -0,0 +1,89 @@
...
+            self._sum.inc(total)
+            self._buckets[index].inc(weight)
+
+    def observe(self, amount: float, exemplar=None) -> None:
+        self._raise_if_not_observable()
+        with self._observation_lock:
+            super().observe(amount, exemplar)
+
+    def _child_samples(self):
+        with self._observation_lock:
```
jgong5: I guess the consistency problem only occurs due to the added `observe_weighted`. Adding the lock there is enough, but leave other methods unchanged (reusing parent's methods is good enough).

atom/model_engine/async_proc.py
```diff
@@ -241,6 +241,10 @@ def busy_loop(self):
                 if need_barrier and self.all_ranks_barrier is not None:
                     self.all_ranks_barrier.wait()
                 if out is not None:
+                    if func_name == "collect_forward_metrics":
+                        if self.kv_queue is not None:
+                            self.kv_queue.put_nowait(("FORWARD_METRICS", out))
```
jgong5: What does "forward" mean in the "FORWARD_METRICS"? Would there be other metrics that are not "forward" metrics? If we just want to handle "metrics" here, name it bare "metrics".

atom/model_engine/engine_core.py
```diff
@@ -47,7 +48,7 @@
 # How often each EngineCore publishes its metrics snapshot. Kept at the API
 # server's scrape interval: the exporter reads a cache, so this bounds how
 # stale a Prometheus sample can be.
-METRICS_PUSH_INTERVAL_S = 5.0
+METRICS_PUSH_INTERVAL_S = 1.0
```
jgong5: I feel it is better to make it configurable. But you can do that in a separate PR.

.github/scripts/atomesh/observability/collect_metrics.py
```diff
@@ -44,7 +44,7 @@ def target(address):
         return address
 
     return {
-        "global": {"scrape_interval": "5s", "scrape_timeout": "4s"},
+        "global": {"scrape_interval": "1s", "scrape_timeout": "1s"},
```
jgong5: by AI agent

Raising the scrape to 1s multiplies a cost paid on the API server's asyncio event loop. `/metrics` is served by `async def metrics()` in `api_server.py`, which calls `_metrics_exporter.render()` synchronously — nothing awaits, so the loop is blocked for the full render.

Measured on this branch (prometheus_client 0.25.0), a registry shaped like the pre-PR one renders in 0.75 ms / 15.8 KiB. With this PR's scheduler and GPU families populated:

```
DP=1, TP=8     4.45 ms     61.9 KiB
DP=4, TP=8    15.48 ms    192.2 KiB
DP=8, TP=8    29.62 ms    365.8 KiB
```

At DP=8 that is a contiguous ~30 ms stall once per second, on the same loop that delivers SSE chunks, on the server whose TTFT and ITL this report exists to measure. Event-loop occupancy for metrics goes from roughly 0.015% (0.75 ms every 5 s) to roughly 3%. `atom:inter_token_latency_seconds` has buckets starting at 2 ms, so one stall can move a whole wave of in-flight streams up several buckets.

atom/model_engine/gpu_metrics.py
```diff
@@ -0,0 +1,170 @@
...
+        start, end = (
+            self.free.pop()
+            if self.free
+            else (self.event_factory(), self.event_factory())
+        )
+        start.record()
+        try:
+            yield
+            end.record()
+        except BaseException:
```
jgong5, line 147: by AI agent

This branch drops the event pair. `start, end` were either popped from `self.free` or freshly allocated a few lines up; on the exception path they are neither returned to `self.free` nor appended to `self.pending`, so both `torch.cuda.Event` objects are unreachable after the `raise`. A repeatedly failing forward allocates events without bound.

Reusing a recorded-but-unpaired event is safe — the next `measure()` calls `start.record()` again — so returning the pair is enough:

```python
except BaseException:
    for state in requests:
        self._discard_request(state.req_id)
    self.free.append((start, end))
    raise
```

Only reachable with `ATOM_ENABLE_METRICS_DEVICE_TIMER=1`, which the agentic PD dashboard job sets for both roles.

(Copilot raised this in a top-level review body; it is still present at this head, so repeating it inline where it can be actioned.)

atom/model_engine/model_runner.py
```diff
@@ -812,6 +813,24 @@ def __init__(self, rank: int, config: Config):
                     self.drafter.model, fullgraph=True, backend="eager"
                 )
 
+        # Install after startup profiling/warmup/capture, which are not traffic.
```
jgong5: by AI agent

The comment names an ordering that does not hold. CUDA-graph capture is not part of "startup … capture" relative to this line: `capture_cudagraph` is an RPC issued from `engine_core.py` after `ModelRunner.__init__` has already run, so `self.gpu_forward_metrics` is installed *before* capture happens, not after.

The code is nonetheless safe, for a different reason: `capture_cudagraph` calls `self.model(input_ids[:num_tokens], model_positions)` directly rather than `run_model`, so `@record_gpu_forward` never fires inside a capture region. I verified both halves of this.

Worth correcting because the comment would reassure a future reader that ordering protects them. If capture were ever routed through `run_model`, the `start.record()` in `measure()` would execute under stream capture and break it. Suggested wording: "capture bypasses `run_model`, so the decorator never runs under stream capture."

atom/entrypoints/openai/api_server.py
```diff
@@ -374,9 +379,9 @@ def anthropic_stop_reason(finish_reason: Any) -> str:
 # 35 discarded bytes. The keepalive only has to beat proxy and SDK idle-read
 # timeouts, which are tens of seconds.
 _ANTHROPIC_PING_INTERVAL_SECONDS = 5.0
-_metrics_exporter = AtomMetricsExporter()
+_metrics_exporter, _request_metrics, _stream_metrics = create_metrics_exporter()
 _background_tasks: list[asyncio.Task] = []
-_METRICS_REFRESH_INTERVAL_SECONDS = 5.0
+_METRICS_REFRESH_INTERVAL_SECONDS = 1.0
```
jgong5: nit: The same `1.0` number appears multiple times in the code. Make them configurable and if they are related, suggest to avoid multiple variables/configurations.

atom/model_engine/scheduler.py
```diff
@@ -373,6 +374,33 @@ def __init__(
                 seq.multimodal_data = None
         self.external_request_ids = [seq.external_request_id for seq in seqs.values()]
 
+        # Freeze request chunk boundaries before schedule-time advancement.
+        # Workers receive copies of the batch, not the mutable Sequence objects.
+        self.prefill_gpu_requests = []
+        if (
+            not is_dummy_run
+            and total_seqs_num_prefill
+            and envs.ATOM_ENABLE_METRICS_DEVICE_TIMER
```
jgong5, line 383: why is this logic metrics' specific?

atom/utils/histogram.py
```diff
@@ -0,0 +1,89 @@
...
+    20,
+    30,
+    60,
+    120,
+    300,
+    600,
+)
+
+
+class CumulativeHistogram:
```
jgong5: by AI agent

Two questions about this class — one structural, one narrow.

**Why does the API process aggregate these at all?**

`CumulativeHistogram` plus the snapshot transport, `prometheus_buckets()` and the `collect_*` contract is roughly 500 lines that reimplement what `prometheus_client`'s multiprocess mode provides natively. Run against two spawned processes it produces the same shape this layer builds by hand:

```
atom_queue_seconds_sum{dp_rank="0"} 0.25
atom_queue_seconds_bucket{dp_rank="0",le="0.1"} 5.0
atom_queue_seconds_count{dp_rank="0"} 5.0
atom_queue_seconds_sum{dp_rank="1"} 0.5
```

Engine code would be `h.labels(dp_rank=...).observe(x)` — no snapshot, no ZMQ tag, no bucket conversion. Its costs are real though: per-PID `.db` files that need `multiprocess.mark_process_dead()` reaping, no `_created`, explicit `multiprocess_mode` on every gauge, and a shared filesystem for all contributing processes.

The narrower question matters more. `collect_metrics.py` already scrapes prefill and decode as separate Prometheus targets. Given that, is per-`dp_rank` in-process aggregation needed at all, rather than exposing each engine's series and letting `sum by (le)` and `histogram_quantile` do it in PromQL? That is the standard Prometheus split — aggregate in the TSDB, not the application — and it\
 would remove most of this layer.

Not asking for a rewrite in this PR. Asking whether the trade was considered, because the answer belongs in `metrics_guide.md` alongside the Path A / Path B contract: a reader deciding where to put a new metric should know why the custom transport exists.

**The non-finite guard looks like it is in the wrong place.**

`observe()` drops non-finite and negative values. Tracing the callers:

- Scheduler paths (9 of the 11 histograms): `now - timing.received_at`, both `time.perf_counter()` in the same process, so non-negative by construction and never NaN. Token and context values are `int()` of lengths. The guard cannot fire here.
- `GPUForwardMetrics.poll()`: `start.elapsed_time(end) / 1000` is the one caller that can plausibly produce a nonsense value.

So every observation pays a branch for a condition only one caller can create, and in the primitive the guard can only do the least useful thing — drop silently. That turns a loud failure (a `NaN` series, noticed immediately) into a quiet one: samples that never arrive, indistinguishable from an idle engine.

Suggestion: validate in `poll()`, where a bad `elapsed_time` is diagnosable and can be logged with context. Keeping a backstop here is still defensible — these sums are never reset, so one NaN poisons the series for the lifetime of the process — but then it should record that it fired. A `dropped` counter in `snapshot()` is one integer, and turns "why is this histogram empty" into an answe\
rable question.

**Minor.** Bucket semantics are now implemented twice: `bisect_left` here, a linear `<=` scan upstream. I checked that they agree today, but nothing pins them. `test_histogram.py` already uses the right pattern for `WeightedHistogram` — assert against a real `Histogram` as an independent oracle — and the same test for `CumulativeHistogram` would stop them drifting.

Also, every Prometheus histogram is cumulative; the distinguishing property here is that this one is unregistered and transportable. `SnapshotHistogram` would say that.

atom/model_engine/block_pool.py
```diff
@@ -149,9 +150,10 @@ def num_reusable_free(self) -> int:
         vacant blocks remain this is slack, and once they are gone every
         allocation spends one of these. `num_free - num_reusable_free` is the
         vacant count, which is the number that actually has to reach zero
-        before `blocks_evicted` can start moving.
+        before `blocks_evicted` can start moving. Maintained as ownership and
+        hashes change so scheduler metrics never scan the free list.
         """
-        return sum(1 for b in self._free if self.blocks[b].hash != -1)
+        return self._num_reusable_free
```
jgong5: by AI agent

This count is already derivable from state the pool maintains, so the six update sites are avoidable:

```python
@property
def num_reusable_free(self) -> int:
    return len(self._cached) + len(self._protected)
```

That would let `self._num_reusable_free` go, along with its updates in `publish`, `clear_index`, `_unindex`, both `popitem` branches, `_take_named` and `free`.

**Why it holds.** `free()` already routes by exactly this predicate — `hash != -1` goes to `_cached` or `_protected`, otherwise to `_vacant`. `_take_named` pops from both dicts. And `_trim_protected`, the one path that looks like it could drop a block from both, re-inserts it:

```python
while len(self._protected) > limit:
    block_id, _ = self._protected.popitem(last=False)
    self._reused.discard(block_id)
    self._cached[block_id] = None
```

So `_cached ∪ _protected` is the set of free reusable blocks. Running the derivation against this PR's own fuzz harness — 1000 steps, both cache policies, including `clear_index`, `retire_top`, `extend` and raw-unit churn:

```
policy=lru    len(_cached)+len(_protected)   EXACT — matches truth at every step
policy=slru   len(_cached)+len(_protected)   EXACT — matches truth at every step
```

(For contrast, a `num_free - vacant` formulation drifts 92 times and goes negative at step 68, so the containers are the right basis rather than the heap.)

**Why it is better than the counter, not just shorter.** It is still O(1), so the motivation for this change survives and `test_reusable_count_does_not_traverse_free_blocks` still passes — that test only forbids iterating `_free`.

The real gain is that `_cached` and `_protected` are load-bearing: they drive eviction order, so if they are wrong the pool evicts the wrong block and cache hit rate degrades. A metrics-only counter has no such pressure behind it — it can be subtly wrong for a long time, and the only symptom is a dashboard number nobody cross-checks. Deriving the metric from state the allocator itself depe\
nds on means it cannot be wrong unless something louder is already wrong.

The class docstring asserts this invariant already ("which container an id is in is only ever decided from its hash"), so the property states an existing contract rather than inventing one.

The caveat is coupling: if a later change adds a third container for free-reusable blocks, or lets one sit in none, the derivation breaks. But so would eviction, so it would not be silent.

If the coupling is unwanted, the alternative that still avoids six sites is chokepoints — funnel every `_free` membership change and every hash change through two private helpers and update the counter there, so it lives in two places rather than six.

Either way, please keep the fuzz test's `num_reusable_free == sum(...)` assertion; with the property it becomes the check that the derivation stays true.

jgong5 (changes requested):

atom/model_engine/scheduler.py
Jasen2201: Because decode forward time is very fast, we need to add events before and after the forward. This may impact performance.

.github/scripts/atomesh/observability/collect_metrics.py
Jasen2201: Thanks for flagging this. I’ve updated the implementation in two ways:

- `/metrics` now renders on demand in a worker thread using `asyncio.to_thread()`. Concurrent scrapes share an in-flight render, so overlapping requests do not launch duplicate rendering work.
- GPU metrics now use two histograms per worker: one for forward steps and one for cumulative per-request prefill time. The separate prefill/decode/mixed step histograms are pooled, matching how the HTML report already aggregated them. The independent per-request accumulator, histogram buckets, and P95/P99 remain intact.

In a matched DP=8, TP=8 benchmark with `prometheus_client 0.25.0`, the histogram simplification reduced:

| Measurement | Before | After |
|---|---:|---:|
| Render median | 35.83 ms | 20.64 ms |
| Payload | 804.7 KiB | 462.1 KiB |
| GPU time series | 5,632 | 2,816 |
| Synthetic SSE interval P99 | 20.63 ms | 15.56 ms |

atom/entrypoints/openai/api_server.py
Jasen2201: Addressed in 9fb49c1c7. Engine metrics pushes (ordinary/DP/PP) and API snapshot refresh now share a single ATOM_METRICS_UPDATE_INTERVAL_S setting, defaulting to 1 second

atom/model_engine/block_pool.py
Jasen2201: Updated in eb0661a06. `num_reusable_free` now returns `len(self._cached) + len(self._protected)`, and the separate `_num_reusable_free` counter and all its update sites have been removed. This keeps the
  lookup O(1) and derives the metric directly from the queues used by the allocator.

  I also removed the counter-specific test additions to keep the PR diff small

atom/model_engine/model_runner.py
Jasen2201: Fixed in ab9f482a4: the exception handler now returns (start, end) to self.free before re-raising. The regression test covers repeated failures with both RuntimeError and KeyboardInterrupt, verifies event reuse, and confirms that a successful
  retry is measured correctly.

atom/model_engine/gpu_metrics.py
Jasen2201: > by AI agent
> 
> This branch drops the event pair. `start, end` were either popped from `self.free` or freshly allocated a few lines up; on the exception path they are neither returned to `self.free` nor appended to `self.pending`, so both `torch.cuda.Event` objects are unreachable after the `raise`. A repeatedly failing forward allocates events without bound.
> 
> Reusing a recorded-but-unpaired event is safe — the next `measure()` calls `start.record()` again — so returning the pair is enough:
> 
> except BaseException:
>     for state in requests:
>         self._discard_request(state.req_id)
>     self.free.append((start, end))
>     raise
> Only reachable with `ATOM_ENABLE_METRICS_DEVICE_TIMER=1`, which the agentic PD dashboard job sets for both roles.
> 
> (Copilot raised this in a top-level review body; it is still present at this head, so repeating it inline where it can be actioned.)

Fixed in ab9f482a4: the exception handler now returns (start, end) to self.free before re-raising. The regression test covers repeated failures with both RuntimeError and KeyboardInterrupt, verifies event reuse, and confirms that a successful
  retry is measured correctly.

atom/model_engine/scheduler.py
Jasen2201: > why is this logic metrics' specific?

Because decode forward time is very fast, we need to add events before and after the forward. This may impact performance

atom/utils/histogram.py
Jasen2201: > Thanks for the detailed review. After reconsidering the tradeoff, I switched the scheduler and GPU histograms to prometheus_client’s native multiprocess mode in a96b5be7d. This removes CumulativeHistogram, histogram snapshots and bucket conversion, GPU metrics
  > replies, and the generic snapshot-collector registration layer.
  >
  > Each node now exports its local instruments with rank/role labels preserved, and Prometheus handles aggregation across targets. The guide documents per-node scraping and directory ownership/cleanup. Pre-existing engine state metrics retain their original snapshot
  > path.
  >
  > Invalid GPU durations are now validated in GPUForwardMetrics.poll() and logged as warnings before the affected timing samples are discarded. Event pairs are recycled, and invalid values cannot poison the cumulative sums.
  >
  > Removing CumulativeHistogram also eliminates its duplicate bucket implementation and the naming concern. Existing WeightedHistogram oracle coverage remains. Final focused validation: 80 tests passed; one test requiring a GPU was skipped. The changes are pushed to
  > the PR.

atom/utils/histogram.py
Jasen2201: Updated in a6a821197. I removed the observe() and _child_samples() overrides so both inherit the parent’s implementation. The lock remains scoped to observe_weighted(), which preserves our token-weighted ITL semantics without a per-token loop.

atom/model_engine/engine_core.py
Jasen2201: > I feel it is better to make it configurable. But you can do that in a separate PR.

done

atom/model_engine/async_proc.py
Jasen2201: have removed  in latest design

valarLip:
Reviewed at `aaba4f7c5` against merge-base `c93b75d1a` (4682 insertions across 44 files).

**Provenance:** findings marked **[verified]** are ones where I read the code at the cited line myself, and for #1/#2 ran the measurement or read the dependency's source. Findings marked **[reported]** come from a finder pass whose mechanism is traced through the diff but where I did not open the far end of the call chain.

**Cleared, so nobody re-litigates them:** the `block_pool.num_reusable_free` rewrite is genuinely equivalent to the old O(n) scan (two independent randomized differential harnesses, 700 and 400 trials across both cache policies, zero divergence, plus a trace of every mutator) — `kv_blocks_vacant` cannot go negative today. All four `StreamingResponse` sites go through `_client_stream`, so moving streaming TTFT out of the middleware loses no endpoint. `poll_forward_metrics` uses `wait_out=False`, so the known worker-RPC-returns-`None` deadlock does not apply. CUDA-graph capture bypasses `run_model` entirely. `perf_counter` is comparable between the input thread and the scheduler (same process). The `ContextVar`/`to_thread` pairing is correct.

The 529 affected unit tests pass, `black --check` is clean, and the new test modules all import under a simulated CI runner with aiter/triton/GPU absent — no collection-abort risk.

---

# The repo's mandated readiness probe is broken from both directions, in one PR

`scripts/wait_server_ready.sh` is the gate CLAUDE.md requires before any workload is launched. Two independent changes here defeat it — one makes it pass too early, the other makes it time out with no diagnosis.

## 1. [verified] A DP>0 node reports READY before loading a single weight

`api_server.py:2955`:

```python
if dp_rank > 0:
    from atom.entrypoints.metrics import start_metrics_server
    server, _ = start_metrics_server(args.host, args.server_port)
    try:
        engine_args.create_engine(tokenizer=tokenizer)
    finally:
        ...
```

That binds `prometheus_client`'s WSGI app on the **API serving port**, and that app answers every GET path. From the pinned `prometheus_client`'s `make_wsgi_app`:

```python
elif environ['PATH_INFO'] == '/favicon.ico':
    ...
else:
    # Note: For backwards compatibility, the URI path for GET is not
    # constrained to the documented /metrics, but any path is allowed.
    status, headers, output = _bake_output(...)
```

Only `/favicon.ico` is special-cased. So `GET /v1/models` on a DP>0 node returns 200 with Prometheus text, and `wait_server_ready.sh:30/55` —

```bash
READY=$(curl -s -m 3 "http://localhost:${PORT}/v1/models" 2>/dev/null | head -c 60)
...
if [ -n "$READY" ]; then echo "Server READY"; exit 0
```

— treats any non-empty body as ready. A DP>0 node passes the gate the instant the metrics thread binds, minutes before the model exists. Before this PR the port was never bound on DP>0 (`create_engine` blocked in `proc.join()`), so the probe correctly got connection-refused. `/health` is fooled identically.

Separately, `addr=args.host` exposes an unauthenticated metrics endpoint on every worker node's serving port.

## 2. [verified] `ATOM_METRICS_UPDATE_INTERVAL_S` raises *after* the model is loaded, producing exactly the signature the probe cannot report

`_positive_float_env` (`envs.py:27-34`) raises `ValueError` for `0`, negatives, `inf`, `nan` and anything unparseable. The five read sites are all lazy env reads inside a busy loop:

```
atom/model_engine/engine_core.py:338
atom/model_engine/engine_core.py:764
atom/model_engine/pp_engine_core.py:68
atom/model_engine/pp_engine_core.py:388
atom/entrypoints/openai/api_server.py:1595
```

Set `ATOM_METRICS_UPDATE_INTERVAL_S=0` — a natural reading of "update as often as possible" — and the parent parses args, loads the tokenizer, spawns engines, loads weights and captures graphs (minutes on a 671B model), and only then does every child die on the first line of `busy_loop`. The parent never gets the ready handshake, so this presents as the known "hangs between `load model runner success` and `ready`" signature. `wait_server_ready.sh` matches none of its error patterns (`cluster_dims` / `InductorError` / `SHUTDOWN` / `proc died`), so it polls to timeout with the actual `ValueError` only in a child's stderr.

These two were constants at import time before; a module-level constant cannot fail this way. Validate once in `main()` / `EngineArgs`, next to the other argument handling.

---

# 3. [verified] One Prometheus `observe()` per request row per forward, on the EngineCore busy loop, with no enable gate

`scheduler_metrics.py:170-172` and the prefill twin at `:181-186`:

```python
for length in context_lens[: batch.total_seqs_num_decode]:
    tokens = int(length)
    self.decode_request_context_tokens.observe(tokens)
    total_context += tokens
```

`record_forward` is called unconditionally from `engine_core.py:413/1064/1337` and `pp_engine_core.py:127` — unlike the GPU timer, there is no `ATOM_ENABLE_METRICS_DEVICE_TIMER`-style gate.

In the intended deployment `PROMETHEUS_MULTIPROC_DIR` is set, so each `observe()` is an mmap read-modify-write. Measured on this box:

```
multiprocess observe(): 1.00 us/call
bs=256  ->  255 us of synchronous host work between two forwards
bs=512  ->  510 us
```

(Two finder passes measured 1.12 and 1.30 µs independently; same order, same conclusion.)

This lands on the inter-forward host path that #1989, #1990 and #1980 were spent shaving, and it costs the most exactly where throughput matters, scaling linearly with batch size. The aggregate histograms — one `observe` per step — are fine; only the two per-row loops are the problem.

This PR already ships the primitive that fixes it: `WeightedHistogram.observe_weighted` in `atom/utils/histogram.py`. Alternatively `np.searchsorted` + `np.bincount` measured 7 µs for the same histogram.

# 4. [verified] Every gauge panel's headline number is now the least informative series

`report.html:50` appended gauge states to the percentile list:

```js
const GLOBAL_STAT_KEYS=["mean","p50","p90","p95","p99"];
const STAT_KEYS=[...GLOBAL_STAT_KEYS,"running","waiting","waiting_kv","used","evictable","vacant","hit","reuse","lmcache","gpu"].filter(...)
```

while `:182` still picks the headline with an idiom that was written when `STAT_KEYS` held only the five percentiles, where it meant "highest percentile":

```js
const key = s.p.cache_breakdown ? "reuse" : [...STAT_KEYS].reverse().find(k => s.visible.has(k));
```

Reversed, the array now starts `[gpu, lmcache, reuse, hit, vacant, evictable, used, waiting_kv, ...]`. So a `kind:"blocks"` panel (visible = used/evictable/vacant) resolves to **`vacant`**, and a `kind:"queues"` panel resolves to **`waiting_kv`**.

The result contradicts each panel's own title — "KV block utilization" reads out the vacant percentage while the block-counts strip above it shows Used / Total. `grep -c STAT_KEYS tests/test_observability_report.py` returns **0**; nothing catches it.

---

The following are **[reported]**: mechanism traced through the diff, far end of the call chain not opened.

# 5. [reported] `prefill_gpu_chunks` is never reset on preemption, so the metric reports more work than the request cost

`sequence.py:175` declares the counters "kept across preemption/recomputation". The only exclusion in `ScheduledBatch.__init__` is `if seq.num_tokens > seq.num_prompt_tokens: continue`, which catches re-prefill *after* generation started. A sequence preempted **during** prefill still has `num_tokens == num_prompt_tokens`, so it passes; `seq.prefill_gpu_chunks += 1` resumes where it stopped (1,2 → preempt → 3,4,5) and the worker's `chunk == state.chunks + 1` check stays satisfied, so `state.total` silently accumulates the discarded chunks 1-2 plus the recomputed 3-5.

`Scheduler.preempt` (`scheduler.py:2422-2450`) resets a dozen fields — `num_rejected`, `num_bonus_tokens`, `num_placeholder_tokens`, `spec_token_ids`, `is_first_decode`, `is_partial_prefill` — but not these two, and `sequence.py:175-176` are their only writes outside `ScheduledBatch`.

The help string says "sum of participating batch device durations across a request's initial local prefill chunks", which a reader takes as the cost of prefilling that request once. Reset on preempt and let `_request_chunks`' `chunk==1` branch restart the state cleanly.

# 6. [reported] `observe_weighted` has no finiteness guard, and NaN is both permanent and silently misfiled

Verified empirically on the pinned `prometheus_client`: after one NaN observation the exposition reads `atom:w_sum NaN`, and the weight lands in `le="0.01"` — the *first* bucket — because `bisect_left`'s `a[mid] < x` is False for every NaN comparison, collapsing `lo` to 0.

A Prometheus counter can never be un-NaN'd, so every `rate(..._sum[5m])` and every `_sum/_count` mean for that series is NaN for the process's life; `export_report.py`'s mean query is exactly `sum(rate(_sum))/sum(rate(_count))`. `interval` reaches `observe_weighted` from `StreamDeliveryTiming` with no validation.

The sibling code already has the guard — `gpu_metrics.py:104` does `if not math.isfinite(seconds) or seconds < 0: logger.warning(...); continue` — it was just not carried over.

# 7. [reported] `registry=None` silently drops the entire engine half of the exposition on any non-`main()` launch

`MetricWrapperBase.__init__` ends with `if not self._labelvalues: if registry: registry.register(self)`, and `SchedulerMetrics` (`scheduler_metrics.py:51`), `GPUForwardMetrics` (`gpu_metrics.py:38`), `RequestMetrics` and `StreamMetrics` are all constructed without one. In multiprocess mode the mmap files carry the data, so this is correct by design — but the guard

```python
if os.environ.get("PROMETHEUS_MULTIPROC_DIR"):
    if not values.ValueClass._multiprocess:
        raise RuntimeError(...)
```

only fires when the directory **is** set; the directory being absent takes the silent path.

`initialize_metrics()` has exactly three call sites, all behind `if __name__ == "__main__"` or `main()`. Any other launch — `uvicorn atom.entrypoints.openai.api_server:app`, an embedded `LLMEngine`, the vLLM/SGLang plugin, or `python -m atom.entrypoints.atomesh.server` directly (its own `__main__` at `server.py:326` never calls it) — yields a `/metrics` that serves the snapshot gauges normally while `atom:request_queue_time_seconds`, `atom:decode_batch_size`, `atom:gpu_forward_seconds` and all the new token histograms are permanently absent, **indistinguishable from "no traffic yet"**. With the device timer on you also pay a CUDA event record + query per forward for samples nobody can read.

Latent companion: because `registry=None` disables registration, the Duplicated-timeseries check never fires — a second construction of the same name+labels in one process silently undercounts instead of raising (verified: two observations, `_count` 1.0).

# 8. [reported] Dropping the scrape interval to 1 s collapses `scrape_timeout` to 1 s, and `render_async`'s shield-coalescing makes one slow render fail every subsequent scrape

`collect_metrics.py:76`: `"scrape_timeout": duration if interval < 1 else "1s"`. The old config was 5 s / 4 s — four seconds of headroom; the new default is 1 s / 1 s — none.

`/metrics` is rendered on demand (`api_server.py:2699` → `render_async` → `generate_latest` plus a full `copy.deepcopy` of the engine snapshot), and in multiprocess mode `MultiProcessCollector` opens and parses one mmap file per (process × metric) across TP8 × DP for ~12 new histogram families. Because `render_async` returns `await asyncio.shield(task)`, Prometheus disconnecting at the 1 s timeout does not cancel the thread, and the next scrape shares the same still-running task — so once a render exceeds 1 s, **every following scrape times out together**. The target goes `up==0`, `run()` records "One or more metrics targets failed a scrape", status flips to partial, and all histogram panels gap.

Even short of timing out, a shared in-flight render returns byte-identical content at a new timestamp, which Prometheus records as a flat step followed by a jump — a staircase in exactly the `rate()`/`increase()` queries the report is built on. No timing of `generate_latest` against the multiprocess collector appears anywhere in the PR.

# 9. [reported] The LMCache coverage guard deletes the panel's headline curve, not just the LMCache curve

`export_report.py:345`: `if state == "reused": ... return f"({gpu} + {offload})"`, where each operand independently carries `result = f"({result} and ({count}({values}) == {count}({inputs})))"` (line 361).

`atom:prefix_cache_offload_tokens` is emitted conditionally, and `llm_engine.py:558` only populates the key when all cache ranks report it. When the guard fires, `offload` is an empty vector, and PromQL vector addition `(gpu_expr + offload_expr)` drops the whole result — so the `reuse` series goes empty for the entire role while the `gpu` curve still draws. `finalize_report`'s `missing` check requires ALL series null, so the panel is not flagged: the metric the panel is named after silently disappears. The README claims "Disabled LMCache reports zero contribution", which this does not deliver.

Compounding it, the per-instance form of the same guard (`count by (instance)(X) == count by (instance)(Y)`) is a tautology on these label-free counters, so drilling into an individual instance shows values while the pooled view shows a gap — two views of the same data disagreeing.

# 10. [reported] `SchedulerMetrics._loads` leaks a strong Sequence reference on every aborted disaggregated-decode request

`scheduler_metrics.py:142`: `self._loads[key] = (seq, time.perf_counter())`, entered from `DecodeScheduler.allocate_waiting` (`scheduler.py:3754`) for every sequence moved into `prefill_waiting`. `prefill_waiting` has exactly one `.pop` (`scheduler.py:3769`, `on_prefill_done`), which is the only `finish_kv_wait`.

`engine_utility.py:227 _handle_abort_request` iterates `list(self.scheduler.running) + list(self.scheduler.waiting)` — a sequence parked in `prefill_waiting` is in neither, so a client disconnect never releases it. Same for a prefill process that dies before sending `PrefillDone`. Each lost request leaks a `(Sequence, float)` pinning that request's `token_ids` and block table for the process lifetime, with no cap (contrast `GPUForwardMetrics.max_requests=4096`).

`prefill_waiting` itself already leaks the same seq, so the incremental damage is a second, separately-keyed reference — meaning a future fix that drains `prefill_waiting` will not fix this one.

# 11. [reported] `ModelRunnerExt.run_model` skips the decorator on the extract-mode prefill branch

`model_runner_ext.py:240`: `if not self._extract_mode: return super().run_model(...)`, then `if context.is_prefill: ... return logits, hidden_states` — the `super()` fallbacks are only at lines 235 and 266. So `@record_gpu_forward` never runs on that branch, while `scheduler.py:398-402` has already run `seq.prefill_gpu_chunks += 1` and appended to `prefill_gpu_requests` for that batch.

With `ATOM_ENABLE_METRICS_DEVICE_TIMER=1` in a rollout job, `atom:gpu_forward_seconds` loses every prefill step, and because `_request_chunks` never sees chunk 1 for those requests, `self.requests.get(req_id)` returns None and every later chunk hits `if state is None: continue` — so `atom:prefill_request_gpu_forward_seconds` reports nothing at all. Silent, no warning.

The same gap reopens for any future `run_model` override; measuring at a seam the subclass cannot skip (or asserting on it) would close the class of bug.

# 12. [reported] `_discard_request` pops by request id rather than object identity

`gpu_metrics.py:67`: `state = self.requests.pop(req_id, None); if state is not None: state.valid = False`. The call at line 111 (inside the `not math.isfinite(seconds)` branch) takes `state` from a pending tuple that may be arbitrarily old.

Request A's chunk-1 event is queued holding `stateA`; A finishes and is deleted; id A is reused and its chunk 1 installs `stateA2`; A's old event then returns a non-finite duration, and `_discard_request('A')` pops and invalidates `stateA2` — the new request is silently dropped from `atom:prefill_request_gpu_forward_seconds` while the offending one was already gone.

The class recognizes this hazard in the mirror direction (line 108 `if not state.valid: continue`, with the comment "Request IDs may be reused; pending events keep the old state object and must never finish a newly admitted request") but only half-handles it. `if self.requests.get(req_id) is state:` closes it; the two `measure()` call sites pass freshly-created states and are already safe.

# 13. [reported] `dp_rank` means different things in different metric families

`scheduler.py:516` (duplicated verbatim at 3464) constructs `SchedulerMetrics` with the *flattened* rank:

```python
getattr(pc, "data_parallel_rank", 0) * getattr(config, "pipeline_parallel_size", 1)
    + getattr(pc, "pipeline_parallel_rank", 0)
```

against `model_runner.py:1437` (`dp_rank=...data_parallel_rank, pp_rank=...pipeline_parallel_rank` — three separate labels) and `llm_engine.py:599` (`"dp_rank": rank`, keyed by the true DP rank). `docs/metrics_guide.md:93` says only "exported per DP rank. All carry `dp_rank` and `engine_role`".

With DP=2, PP=2, engine (dp1, pp0) reports `dp_rank="2"` in the histograms and `"1"` in the gauges. The natural query for the new dashboard — join queue wait against GPU time per rank — silently pairs the wrong engines, and `sum by (dp_rank)` reports 8 DP ranks where there are 2. PP downstream ranks also construct a `SchedulerMetrics` and publish zero-valued children under `dp_rank` values no gauge ever emits.

(The `getattr(config, "pipeline_parallel_size", 1)` lookup is correct — the field is on `Config` — but four `getattr` fallbacks here are unreachable.)

# 14. [reported] `atom:scheduler_requests{state="running"}` counts `prefill_done` on the decode scheduler

`engine_utility.py:425` forwards `running` straight from `get_request_counts`, and `DecodeScheduler.get_request_counts` (`scheduler.py:3680-3683`) returns `(len(self.running) + len(self.prefill_done), len(self.waiting) + len(self.prefill_waiting))`. The waiting side is then corrected — `external = parked + len(prefill_waiting)`, `"waiting": max(0, waiting - external)` — but `prefill_done` is neither subtracted from `running` nor placed in `waiting_kv`.

Under a burst, `prefill_done` can hold a full `max_num_seqs` of requests that have not entered a decode batch. The new `kind:"queues"` panel shows them as running while zero forwards are issued — the exact state an operator opens that panel to diagnose. The HELP string ("Requests by scheduler state; waiting excludes KV waits") and `docs/metrics_guide.md:107` give no hint of an admitted-but-not-scheduled tier.

(The waiting double-count hypothesis is **disproved** — `get_request_counts` already folds `prefill_waiting` into `waiting`, so the subtraction is correct and the `max(0, ...)` clamp never fires here.)

# 15. [reported] `prefill_context_tokens` and `decode_context_tokens` are documented as a pair but sampled at different points in the step

Both read `batch.context_lens`, which `scheduler.py:266-276` builds as `(num_cached_tokens[i] + num_scheduled_tokens[i] if seq.type == PREFILL else seq.num_tokens)`. So a prefill row's value runs through the **end** of the current chunk while a decode row's value is the length **before** this step's `mtp_k+1` tokens.

`docs/metrics_guide.md:103-106` presents them as the same quantity per phase, and the report draws them as paired panels. Anyone computing attention FLOPs or KV-read bytes from `prefill_context_tokens + decode_context_tokens` — which is what the README's roofline framing invites — gets a consistent bias: decode low by `batch_size * (mtp_k+1)` per step, which at bs=256 / K=3 is 1024 tokens every step.

Either sample both at the same boundary, or state the asymmetry in both help strings and the doc.

---

# Smaller items

- `tests/test_multiprocess_metrics.py:123` spawns `[sys.executable, __file__]` with no `cwd` / `PYTHONPATH`, so it imports the **installed** `atom` rather than the tree under test — reproducible as `ModuleNotFoundError` in a worktree.
- `block_pool.py`'s class docstring and `_take_free`'s docstring both still assert the staleness that the new line 154 depends on not existing. One of the two must be fixed.
- `atom:scheduler_kv_cache_blocks{state="used"}` includes raw PAGE unit reservations, so the "KV block utilization" panel mis-attributes state-checkpoint pressure to KV.
- `render_async` uses the default executor — the same pool the six `run_in_executor(None, do_preprocess)` calls use — which is exactly what `_gc_census_executor`'s comment 400 lines above says it created a private pool to avoid.
- `multiprocess.mark_process_dead` is never called.
- The `clock` fixture and two `json.loads` patches monkeypatch stdlib modules process-wide.
- Six `if metrics := getattr(self, "metrics", None):` guards are dead (it is always set); the metrics-push pacing block is copy-pasted into four busy loops; bucket definitions now live in four separate files.

atom/model_engine/scheduler.py
jgong5: OK, then please add a comment to explain so, and also note `self.prefill_gpu_requests` is for metrics report only.

jgong5 (approved):
nit on adding code comments, others LGTM.

Jasen2201:
## 1.[verified] A DP>0 node reports READY before loading a single weight
Fixed in 8deff6ddc: readiness probes now require a valid, non-empty model list from /v1/models, so metrics responses no longer trigger READY. Multi-node DP probes should target the coordinator.
## 2. [verified] `ATOM_METRICS_UPDATE_INTERVAL_S` raises *after* the model is loaded, producing exactly the signature the probe cannot report
Fixed in cf95b2bac: invalid ATOM_METRICS_UPDATE_INTERVAL_S values now log a warning and fall back to 1 second instead of failing startup.
## 3. [verified] One Prometheus `observe()` per request row per forward, on the EngineCore busy loop, with no enable gate
Fixed by recording context length once per request instead of on every forward, reducing Prometheus overhead in the engine loop.
## 4. [verified] Every gauge panel's headline number is now the least informative series
 Fixed in 6d05a20e2. Gauge headlines now use explicit panel-specific priorities: Used for KV blocks and Running for queues. Percentile and cache-reuse behavior is preserved.
## 5. [reported] `prefill_gpu_chunks` is never reset on preemption, so the metric reports more work than the request cost
This is intentional and reasonable: GPU work performed before preemption is still part of the request’s prefill cost, so the metric retains it across recomputation.
## 6. [reported] `observe_weighted` has no finiteness guard, and NaN is both permanent and silently misfiled
 The production path uses elapsed time from time.perf_counter() and positive integer token counts, so NaN is not a reachable input under normal operation. We’ll keep this unchanged; defensive checks for this case are unnecessary.
 ## 7. [reported] `registry=None` silently drops the entire engine half of the exposition on any non-`main()` launch
> registry=None is intentional. Our production launch path initializes multiprocess metrics correctly, and custom launchers must follow the documented setup requirements. No change is needed for the supported usage.
## 8. [reported] Dropping the scrape interval to 1 s collapses `scrape_timeout` to 1 s, and `render_async`'s shield-coalescing makes one slow render fail every subsequent scrape
 > The 1-second scrape interval matches our production deployment. Coalescing only shares an in-flight render; once it finishes, the task is cleared and subsequent scrapes start a fresh render. A single slow render does not cause all subsequent scrapes to fail.
## 9. [reported] The LMCache coverage guard deletes the panel's headline curve, not just the LMCache curve
> This is intentional: total reuse is GPU reuse plus LMCache reuse, so missing LMCache telemetry makes the total unknown. Disabled LMCache exports zero and preserves the total curve. A fully covered instance can still show a valid total while the pooled view remains
  > unknown due to another instance’s missing data. Existing PromQL tests verify both cases.

## 10. [reported] `SchedulerMetrics._loads` leaks a strong Sequence reference on every aborted disaggregated-decode request
Fixed in 9a2650403. Removed the unnecessary KV-wait tracking for shared-cache prefill, eliminating the additional Sequence reference introduced by this PR. Queue metrics remain intact. The pre-existing RapidServe cancellation issue is separate.
## 11. [reported] `ModelRunnerExt.run_model` skips the decorator on the extract-mode prefill branch
 Fixed in 88c77a1de. The extraction-mode prefill branch now records GPU timing. Paths that call the parent implementation retain a single timing observation.
 ## 12. [reported] `_discard_request` pops by request id rather than object identity
 Fixed in b37f5032a. Invalid-event cleanup now checks object identity before discarding the timing state, so stale events cannot remove a newer request with the same ID.
 ## 13. [reported] `dp_rank` means different things in different metric families
  DP>1 combined with PP>1 is explicitly unsupported, so the DP=2, PP=2 example cannot occur in supported deployments. With PP enabled, only the head stage records scheduler observations, and its dp_rank=0 matches the gauges. Downstream stages register zero-valued
  scheduler histograms under flattened ranks; the labeling is imperfect, but actual request observations remain correctly attributed. We’ll leave this unchanged for now.
  ## 14. [reported] `atom:scheduler_requests{state="running"}` counts `prefill_done` on the decode scheduler
  This is intentional: running includes requests whose prefill has completed and are ready for the next decode scheduling step; it does not represent the current GPU batch size. These requests are no longer waiting for KV. This follows the existing scheduler accounting, predates this PR, and is covered by existing tests. We’ll retain this behavior.
  ## 15. [reported] `prefill_context_tokens` and `decode_context_tokens` are documented as a pair but sampled at different points in the step
  seq.num_tokens already includes the current decode input token and, for speculative decoding, the draft tokens/placeholders. It is not the pre-step KV length, so the claimed fixed undercount of mtp_k+1 does not apply. Both metrics describe scheduler-side logical context lengths, not exact FLOPs or KV traffic. We’ll retain the current calculation.

valarLip (approved):

PR #2188, opened 2026-09-10
fix(openai): validate token limits and preserve thinking history

ZLkanyo009:
## Motivation

This change addresses two serving correctness issues:

1. ATOM currently accepts non-positive max_tokens and max_completion_tokens. Such requests return HTTP 200 with an empty completion instead of being rejected as invalid input. All OpenAI and Anthropic completion endpoints should consistently return HTTP 400 for generation limits below 1.
2. The Anthropic-to-OpenAI message adapter drops assistant thinking blocks from multi-turn history. Reasoning-model templates such as GLM-5.3 rely on reasoning_content to reconstruct previous assistant reasoning. Dropping it produces incomplete conversation context and may affect subsequent turns.

This change adds consistent token-limit validation and preserves Anthropic thinking history through the internal message conversion path.

### DIFF
--- a/atom/entrypoints/openai/protocol.py
+++ b/atom/entrypoints/openai/protocol.py
@@ -28,18 +28,11 @@ STREAM_DONE_MESSAGE = "data: [DONE]\n\n"
 TOOL_CHOICE_VALUES = frozenset({"auto", "none", "required"})
 
 
-def validated_max_tokens(value: int, field: str) -> int:
-    """The generation cap, rejecting values the engine cannot honour.
-
-    A cap below 1 leaves nothing to generate, and the engine returns an empty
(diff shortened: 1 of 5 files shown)


atom/entrypoints/openai/protocol.py
```diff
@@ -28,6 +28,20 @@
 TOOL_CHOICE_VALUES = frozenset({"auto", "none", "required"})
 
 
+def validated_max_tokens(value: int, field: str) -> int:
```
valarLip: over design? validated_max_tokens with field"max_tokens"

valarLip (changes requested):

valarLip (approved):

PR #2196, opened 2026-09-11
scheduler: make the prefill coalescer work on the TP/DCP path

ZhiweiYan-96:
## Summary

The prefill coalescer (`PrefillDelayer`) has never been active on a TP/DCP server. Two things stand between it and a held batch: it is not constructed on that path, and the signal it would measure against is computed for the DP case. This PR addresses both.

Gated behind `ATOM_PREFILL_DELAYER_SINGLE_RANK`, **default off** — this changes prefill batching on every non-DP deployment, so it is opt-in until measured on more than one operating point.

## 1. It is not constructed on this path

`PrefillDelayer` has always documented a `cpu_group=None` mode for `dp_size=1`, but construction only ever lived in `DPEngineCoreProc`. A TP-only or TP/DCP server therefore runs completely uncoalesced: median prefill fill **12.8%** of the 16,384-token budget, **1 request per prefill batch**.

Prefill cannot share a forward with decode, so each of those forwards stalls every decoding request. This PR builds the delayer in `EngineCore` as well.

## 2. Once constructed, the fill signal never reaches the target

`_waiting_new_token_count()` produces the "queued work" signal that the fill target is measured against. A sequence that has not been admitted yet has not been probed against the prefix cache, so its `num_cached_tokens` is still `0` and `num_tokens - num_cached_tokens` is the **full prompt**.

The docstring calls this a best-effort upper bound, which it is. On a prefix-cache-heavy workload the bound is loose enough to dominate the signal:

> GLM-5.2 MXFP4, MI355X, TP4/DCP4, C56, agentic trace (`cc-traces-weka-062126`) — median prompt **79k tokens**, prefix cache hit rate **95.8%**. One waiting request alone pushes the raw count past the 16,384-token cap, so `fill` pins at `1.0` and the delayer releases on every tick. **hold_rate 1.58% over 29,000 decisions.** `TARGET_FILL` had no effect at any value between 0.1 and 0.95.

This PR discounts an un-probed sequence by the observed hit rate:

```python
hit_rate = getattr(self.engine_stats, "recent_cache_hit_rate", None)
discount = 1.0 - hit_rate if hit_rate is not None else 1.0
...
if seq.num_cached_tokens == 0:
    num_new_tokens = int(num_new_tokens * discount)
```

**This is an estimator, not a measurement** — the population hit rate applied to a sequence we have not probed. Deliberately conservative in two ways: a sequence that *has* been probed (`num_cached_tokens > 0`) keeps its exact value, and before any request completes the rate is `None` and we fall back to the raw count — the old behaviour, which only ever fires *early*, never holds too long.

## Measured effect

C56, 1800s, aiperf, `semianalysisai/cc-traces-weka-062126`, everything else identical between the two runs.

| metric | before | after | Δ |
|---|---:|---:|---:|
| prefill batches | 1,481 | 579 | **−61%** |
| prefill fill | 13.6% | 16.6% | +22% |
| TTFT p99 | 86,201 ms | 29,544 ms | **−65.7%** |
| TTFT p90 | 4,801 ms | 4,022 ms | −16.2% |
| TTFT p50 | 683 ms | 843 ms | +23.5% |
| ITL p90 | 113.8 ms | 102.7 ms | **−9.7%** |
| ITL p99 | 238.9 ms | 221.2 ms | −7.4% |
| ITL p50 | 65.6 ms | 65.2 ms | −0.7% (inside ±1.5% noise) |
| interactivity (norm) | 8.30 | 9.23 | **+11.2%** |
| throughput / GPU | 21,349 | 21,984 | +3.0% |

**The gains are entirely in the tail, which is what the mechanism predicts.** A prefill forward costs ~135 ms of fixed overhead and occupies a whole tick, so 61% fewer of them means 61% fewer decode interruptions. TTFT p50 rises because requests genuinely wait to be batched — but p90 and p99 fall hard, because the old path serialized the queue behind single-request prefills.

## What is not covered

- Only measured on one workload (agentic / Claude Code trace). The discount helps in proportion to the prefix cache hit rate, so a low-hit-rate workload should see the estimator approach the old raw count and behave as before — but that has not been measured.
- No change to the DP path; `DPEngineCoreProc` construction is untouched.

## Test plan

- `TestWaitingNewTokenCountHitRateDiscount` — 9 new cases covering: no hit-rate data (fallback), 0% / 50% / 95.8% hit rates, probed vs un-probed sequences, and that unschedulable / `WAITING_FOR_REMOTE_KVS` sequences stay excluded. These drive the real sliding-window counters rather than stubbing the read-only `recent_cache_hit_rate` property.
- Behaviour with the flag unset is unchanged; `test_prefill_scheduler.py` covers that path and is green.
- `tests/test_scheduler.py` + `tests/test_prefill_delayer.py` + `tests/test_prefill_scheduler.py`: **169 passed**.
- Full suite: 5,620 passed. The 8 pre-existing failures in `test_fused_compress_ragged.py` were verified unrelated by reverting `scheduler.py` and re-running.

## Risk

Default-off. With the flag unset, `EngineCore` takes the same path it does today and `_waiting_new_token_count()` is the only behaviour change — and that one is inert until a `PrefillDelayer` exists to consume the signal.

### DIFF
--- a/atom/model_engine/scheduler.py
+++ b/atom/model_engine/scheduler.py
@@ -722,18 +722,36 @@ class Scheduler:
         so the "queued work" signal counts only tokens this rank could actually
         prefill this step. Counting remote-KV / unschedulable tokens here would
         inflate the cross-rank aggregate and reach the fill target before a real
-        batch has accumulated. The `num_cached_tokens` discount is best-effort:
-        an un-admitted seq has not been probed against the prefix cache yet, so
-        this is an upper bound on new tokens for cache-hit prompts.
+        An un-admitted seq has not been probed against the prefix cache yet, so
+        its `num_cached_tokens` is still 0 and `num_tokens - num_cached_tokens`
+        is the FULL prompt. On a prefix-cache-heavy workload that is not a small
+        error -- it is fatal to the coalescer:
+
+            agentic trace, 09-09: median prompt 79k tokens, 95.8% hit rate.
+            One waiting request alone pushed the raw count past the 16,384-token
+            cap, so `fill` pinned at 1.0 and the delayer fired every time.
+            Measured: hold_rate 1.58% over 29,000 decisions. `TARGET_FILL` had
+            no effect at any value.
+
+        So discount an un-probed seq by the observed hit rate. This is an
+        estimator, not a measurement: the population rate applied to a seq we
+        have not probed. A seq that HAS been probed (`num_cached_tokens > 0`)
+        keeps its exact value. Before any request completes the rate is None and
+        we fall back to the raw count -- the old behaviour, which only ever
+        fires early, never holds too long.
         """
         cap = self.max_num_batched_tokens
         total = 0
+        hit_rate = getattr(self.engine_stats, "recent_cache_hit_rate", None)
+        discount = 1.0 - hit_rate if hit_rate is not None else 1.0
         for seq in self.waiting:
             if self._unschedulable_reason(seq) is not None:
                 continue
             if seq.status == SequenceStatus.WAITING_FOR_REMOTE_KVS:
                 continue
             num_new_tokens = seq.num_tokens - seq.num_cached_tokens
+            if seq.num_cached_tokens == 0:
+                num_new_tokens = int(num_new_tokens * discount)
             if (
                 not self.enable_chunked_prefill
                 and num_new_tokens > self.max_num_batched_tokens
--- a/atom/model_engine/engine_core.py
+++ b/atom/model_engine/engine_core.py
@@ -172,6 +172,43 @@ class EngineCore:
                 state_runtime=self.state_runtime,
             )
 
+        # Prefill coalescer on the single-rank (TP-only) path. PrefillDelayer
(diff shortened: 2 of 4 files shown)


atom/model_engine/scheduler.py
```diff
@@ -722,18 +722,36 @@ def _waiting_new_token_count(self) -> int:
...
+        hit_rate = getattr(self.engine_stats, "recent_cache_hit_rate", None)
+        discount = 1.0 - hit_rate if hit_rate is not None else 1.0
         for seq in self.waiting:
             if self._unschedulable_reason(seq) is not None:
                 continue
             if seq.status == SequenceStatus.WAITING_FOR_REMOTE_KVS:
                 continue
             num_new_tokens = seq.num_tokens - seq.num_cached_tokens
+            if seq.num_cached_tokens == 0:
+                num_new_tokens = int(num_new_tokens * discount)
```
ganyi1996ppo, line 754: why use recent hit rate stats in engine times the num_new_tokens?

PR #2197, opened 2026-09-11
release: align ATOM's wheel release with aiter

zufayu:
# release: align ATOM's wheel release with aiter

ATOM's last stable release is **v0.1.5, from 22 June**; `v0.1.6-rc0` has sat
unreleased since 4 July. Releases happened when someone remembered to make
one, which is why they stopped happening. Meanwhile the ROCm 10 line means one
commit now yields two wheels, which the current flow cannot tell apart.

This lines ATOM's release up with aiter's, piece by piece.

## Wheel version carries the ROCm line

The version came straight from `git describe`, which says nothing about which
ROCm the wheel was built against — so the ROCm 7 and ROCm 10 builds of one
commit produced the same filename and overwrote each other on the release.

The ROCm line now goes in the PEP 440 local segment, the way torch already
spells it inside this very image (`2.11.0+rocm10.0.0`):

    atom-0.1.7+rocm10.0.0-cp312-cp312-linux_x86_64.whl
    atom-0.1.7+rocm7.2.4-cp312-cp312-linux_x86_64.whl

Both hang off a single `v0.1.7` tag. The version step emits the release tag
separately with the local segment stripped, and the release step already falls
back to uploading into an existing release, so whichever flavor finishes
second just adds its asset. A local segment does not change which release a
version belongs to, so `pip install atom==0.1.7` keeps resolving.

The ROCm version is read out of the build container rather than guessed from
the image name — which is why the version step moved after the container
starts. If it cannot be read, the build fails rather than publishing a wheel
whose provenance is unknown.

`git describe` output is also normalised to PEP 440 before use: ATOM tags
pre-releases as `v0.1.6-rc0`, which pip cannot parse, and it now becomes
`0.1.6rc0`. aiter-shaped `.postN` / `.devN` tags pass through untouched.

## One dispatch builds every line

Releasing both lines meant dispatching twice and remembering which image went
with which line — a step easy to skip, and skipping it silently ships half the
set. The flavors are a matrix now. They run one at a time (they share a runner
anyway), with `fail-fast` off, because a break in one ROCm line says nothing
about the other — that independence is the point of building both.

`flavors` narrows a dispatch to one line; `build_image` still overrides the
image outright for a one-off.

## Cadence, split out the way aiter splits it

aiter keeps `aiter-release.yaml` (builds) apart from `release-automation.yaml`
(decides whether and what to release). ATOM now does the same: the build
workflow gained `workflow_call`, and a new automation workflow holds the
schedule.

A weekly cron gated on a 14-day grid picks the release days — GitHub cron
cannot express "every 14 days", the same reason aiter's carries that gate. The
job resolves the next tag from the last stable one (`v0.1.5` → `v0.1.6`),
creates `release/<tag>` when missing, and calls the builder.

The anchor sits a week off aiter's, so an ATOM release always lands in the
week following an aiter one — ATOM ships aiter inside its image, and a release
that cannot carry the aiter it was meant to is not worth scheduling.

Some deliberate refusals:

- only `vX.Y.Z` counts when deriving the next version, so `v0.1.6-rc0` cannot
  make the automation skip a number;
- a `.postN` release is a fix on something already shipped, so it refuses to
  invent the branch or tag it is meant to be fixing;
- preview `.devN` tags stay manual, as aiter's are.

The verify job exists because a half-published release is the failure worth
catching: it looks complete from the outside, and whoever pulls the missing
flavor silently gets the other line's wheel instead.

## Credentials

`ATOM_CI_APP_ID` and `ATOM_CI_APP_PRIVATE_KEY` are in place as of 2026-09-11,
matching aiter's app credentials. The automation needs them because it pushes
tags, and a tag pushed with the default `GITHUB_TOKEN` does not retrigger
workflows.

## Checked

- version derivation across five tag shapes × two ROCm lines, and the
  `packaging` library accepts every version produced;
- the version step run end to end in a container against a real repo, plus its
  failure path — an unreadable ROCm version aborts with an empty
  `GITHUB_OUTPUT` rather than a mislabelled wheel;
- the automation's resolve step run end to end against a clone of this repo
  with its real tags, pushing to a throwaway origin, across seven cases:
  a scheduled release derives `v0.1.6` from `v0.1.5` and creates
  `release/v0.1.6` and the tag; the gate skips (exit 0, with a reason) off
  cadence and proceeds on it; a manual dispatch bypasses the gate; and it
  refuses a `.postN` whose branch and tag do not exist, an ill-formed tag,
  and a `.devN` preview. Repeated runs walk v0.1.6 → v0.1.7 → v0.1.8 without
  skipping or repeating a number;
- the 14-day gate checked against aiter's anchor across six Mondays, to
  confirm an ATOM release always lands the week after an aiter one.

Not checked: an end-to-end run. `docker-release.yaml` and this workflow have
no `pull_request` trigger, so neither can execute on a PR.

## Depends on

#2150 — `atom-release.yaml` is introduced there. Rebase onto main once it
lands.


zufayu:
CI is green — 9 checks pass, nothing fails, including `Check GitHub Actions
workflows`. The single pending entry is the external readthedocs status.

That actionlint check caught a real mistake earlier and it is fixed in
`49861934d`: narrowing a dispatch to one ROCm line had been written as a
job-level `if` comparing `inputs.flavors` to `matrix.flavor`, which cannot
work — a job-level condition is evaluated before the matrix exists. The matrix
is now built from the input instead. Same commit fixes a duplicate
`upload-artifact` name that would have failed the second flavor outright,
moves publishing into its own job so a half-complete release is never exposed,
and keeps the ROCm version segment at two components to match the
`atom-<version>+rocm7.0` wheels v0.1.5 already shipped.

Ready for review, @gyohuangxin.

One thing that needs a decision outside this PR: the automation workflow
pushes tags, so it needs `ATOM_CI_APP_ID` / `ATOM_CI_APP_PRIVATE_KEY` the way
aiter's uses `AITER_CI_APP_*`. Until those exist it fails at its first step
and nothing else is affected — manual `workflow_dispatch` releases keep
working. This also depends on #2150, which introduces `atom-release.yaml`;
once that lands I will rebase and the diff here drops to the four release
commits.

PR #2199, opened 2026-09-11
feat(v4): enable asm sparse prefill on gfx1250

amd-ruitang3:
## Summary
- dispatch the aiter FP8 sparse prefill ASM kernel on gfx1250 when the local head count is 128
- keep OPUS as the fallback for unsupported configurations
- add `ATOM_FORCE_V4_PREFILL_OPUS` as an escape hatch

## Validation
- `git diff --check`
- `python -m compileall -q atom/model_ops/v4_kernels/paged_prefill.py atom/utils/envs.py`

### DIFF
--- a/atom/model_ops/v4_kernels/paged_prefill.py
+++ b/atom/model_ops/v4_kernels/paged_prefill.py
@@ -62,8 +62,20 @@ except ImportError:
     pa_sparse_prefill_opus = None
     _HAS_OPUS = False
 
+try:
+    from aiter.ops.mla_sparse_prefill import mla_sparse_prefill_fp8_asm
+
+    _HAS_PREFILL_ASM = True
+except ImportError:
+    mla_sparse_prefill_fp8_asm = None
+    _HAS_PREFILL_ASM = False
+
 from aiter.jit.utils.chip_info import get_gfx
 from aiter.ops.triton.attention.pa_prefill_sparse import pa_prefill_sparse
+
+# Local head count the aiter asm fp8 prefill kernel ships for; anything else
+# raises there rather than dispatching, so we filter on it up front.
+_PREFILL_ASM_HEADS = 128
 
 # Fixed best-known Triton meta for DeepSeek-V4 TP=8 / local H=8 prefill.
 _BLOCK_H = 16
@@ -638,7 +650,20 @@ def sparse_attn_v4_paged_prefill(
         # and extend K fed directly. No dequant of the prefix, no torch quant.
         from aiter.ops.pa_sparse_prefill_opus import pa_sparse_prefill_fp8_opus
 
-        return pa_sparse_prefill_fp8_opus(
+        # aiter's asm kernel is signature-compatible with the OPUS one, so the
+        # argument list below is shared. It is gfx1250-only and hard-requires
+        # H == 128 (one workgroup serves one query token x all heads), which a
+        # TP shard below 128 local heads cannot satisfy -- those keep OPUS.
+        fp8_prefill = pa_sparse_prefill_fp8_opus
+        if (
+            _HAS_PREFILL_ASM
+            and not envs.ATOM_FORCE_V4_PREFILL_OPUS
+            and get_gfx() == "gfx1250"
+            and q_packed.shape[1] == _PREFILL_ASM_HEADS
+        ):
+            fp8_prefill = mla_sparse_prefill_fp8_asm
+
+        return fp8_prefill(
             q_packed,
             q_rope,
             unified_kv,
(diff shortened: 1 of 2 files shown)


junhaha666 (approved):

junhaha666 (dismissed):

valarLip:
Reviewed the dispatch change. The signature-compatibility claim checks out —
`mla_sparse_prefill_fp8_asm` and `pa_sparse_prefill_fp8_opus` take the same 12
positional arguments plus `out=None`, so sharing the call site is sound. Three
things below, one of which I could reproduce.

### The arch gate asks about the build target, not the live GPU

`get_gfx()` honours `GPU_ARCHS`. aiter's own `chip_info.py` is explicit about
which one to use here:

> Unlike get_gfx(), ignores GPU_ARCHS -- always detects the actual running GPU.
> **Use for runtime dispatch decisions** (selecting tuned kernels, picking code
> paths). Use get_gfx() for build-time codegen paths.

And `mla_sparse_prefill_fp8_asm` itself gates on `get_gfx_runtime()`
(`aiter/ops/mla_sparse_prefill.py:98`), so the two sides disagree. Reproduced on
a gfx950 box:

```
$ GPU_ARCHS=gfx1250 python3 -c "..."
get_gfx()=gfx1250   get_gfx_runtime()=gfx950
  PR dispatches to asm: True
  aiter kernel accepts: False
  => uncaught RuntimeError
```

`GPU_ARCHS` is exactly what a cross-targeting build sets, so this is reachable
rather than theoretical. One-word fix: `get_gfx()` -> `get_gfx_runtime()`
(already imported that way in `atom/model_ops/fused_moe/comm_fused_moe.py:15`).

### No error containment, against this file's own convention

The OPUS path fifteen lines below wraps its call:

```python
try:
    return pa_sparse_prefill_opus(...)
except RuntimeError:
    pass
```

with a comment naming "runtime error (e.g. unsupported GPU)" as an expected
case. The new asm path has no such guard. The aiter kernel raises on five
conditions — arch, NoPE dtype mismatch across the three tensors, `q_rope` not
bf16, `H != 128`, and `out` shape/dtype — and this PR pre-filters two of them.
The other three become a crash where the established behaviour here is a
fallback.

### No measurement behind the switch

`git diff --check` and `compileall` are syntax checks. This routes production
prefill to a different kernel, and the PR carries no accuracy run, no test that
the fallback path works, and — the part I would want most — **no number showing
the asm kernel is faster than OPUS**. The justification for the change isn't
stated anywhere, so there's nothing to review it against. Worth a
before/after on the target config before this lands.

### Suggested shape

Drops the `fp8_prefill` temporary, shares the argument list once, and puts the
`try` only around the optional path — the same shape as the `_HAS_OPUS` block
below it:

```python
        # Signature-compatible with the OPUS kernel, so the argument list is
        # shared. gfx1250-only and hard-requires H == 128 -- one workgroup serves
        # one query token across all heads -- so a smaller TP shard keeps OPUS.
        args = (
            q_packed, q_rope, unified_kv, unified_kv_rope,
            kv_indices_prefix, kv_indptr_prefix,
            k_packed.view(k_packed.shape[0], -1),
            k_rope.view(k_rope.shape[0], -1),
            kv_indices_extend, kv_indptr_extend, attn_sink, softmax_scale,
        )
        if (
            _HAS_PREFILL_ASM
            and not envs.ATOM_FORCE_V4_PREFILL_OPUS
            and get_gfx_runtime() == "gfx1250"
            and q_packed.shape[1] == _PREFILL_ASM_HEADS
        ):
            try:
                return mla_sparse_prefill_fp8_asm(*args)
            except RuntimeError:
                pass  # a dtype or shape it does not serve; OPUS covers all of them
        return pa_sparse_prefill_fp8_opus(*args)  # [S, H, head_dim] bf16
```

### Two follow-ups, neither blocking and neither this PR's doing

- **`_PREFILL_ASM_HEADS = 128` is copied from aiter with nothing tying it back.**
  aiter's `__all__` exposes only the kernel and its `_fwd`, so copying is
  currently the only option, and the comment does explain it. The durable fix is
  a predicate on the aiter side — `mla_sparse_prefill_fp8_asm_serves(...) ->
  str | None`, returning the reason it cannot serve or `None` — so the constraint
  lives once and the two sides cannot drift.
- **`envs.__getattr__` re-evaluates its lambda on every access**, so
  `envs.ATOM_FORCE_V4_PREFILL_OPUS` is an `os.getenv` per call on a per-layer
  path. `envs.py`'s own docstring says values are "evaluated lazily on first
  access", which the implementation does not do. Pre-existing, and the
  `ATOM_FORCE_ATTN_TRITON` check right beside it behaves the same way.

The comments themselves read well — they explain why rather than restating the
code, and the `(one workgroup serves one query token x all heads)` aside is
worth keeping.

valarLip (approved):

PR #2203, opened 2026-09-11
[KV-events] match vLLM/SGLang ZMQ framing and per-DP-rank port offset

vMaroon:
## Motivation

`ZmqEventPublisher` (#869) sends one frame with the default empty topic, or `[topic, payload]` with a topic set. vLLM's publisher sends `[topic, seq, payload]` with an 8-byte big-endian sequence number, and vLLM-compatible subscribers reject anything else: llm-d-router's KV-events pool (`pkg/kvevents/zmq_subscriber.go`, `parseEventFrame`) requires exactly three frames, so with `ATOM_KV_EVENTS_ENABLE=1` every ATOM batch is logged as malformed and dropped, and precise prefix-cache routing never sees ATOM blocks. The msgpack payload was already slot-for-slot compatible; only the framing was not.

Every DP rank builds its own `Scheduler` and therefore its own publisher, and all of them bound the same `ATOM_KV_EVENTS_ENDPOINT`. vLLM offsets the bound port by `data_parallel_rank`, and llm-d-router's per-pod discovery dials `socketPort + rank` on the same assumption.

Subset of #1316, which also adds `token_offset` and replay and has been conflicting with `main` since July. This PR is the framing and port pieces only, so #1316 stays additive on top.

## Technical Details

- `atom/distributed/kv_events.py`
  - Every message is `[topic, seq, payload]`. `seq` is a monotonic uint64 assigned at enqueue, so a batch dropped on queue overflow shows up as a gap; the all-0xFF value is kept out of the data range because vLLM reserves it as the replay terminator.
  - `offset_endpoint_port(endpoint, data_parallel_rank)` mirrors vLLM's helper: tcp ports are offset by the rank, other transports get a `_dp<rank>` suffix, rank 0 is unchanged. `ZmqEventPublisher.__init__` binds the offset endpoint.
- `tests/test_kv_events.py`: round-trip test reads `recv_multipart()`; new tests cover the frame layout and monotonic `seq`, `seq` consumption on drop, `offset_endpoint_port` cases, and two ranks binding one base endpoint.

Wire change for existing consumers: single-frame `recv()` readers must switch to `recv_multipart()`. Nothing in-tree consumes the stream.

For llm-d, the engine side then needs `ATOM_KV_EVENTS_ENABLE=1`, `ATOM_KV_EVENTS_ENDPOINT=tcp://0.0.0.0:5557`, and `ATOM_KV_EVENTS_TOPIC=kv@<pod-ip>@<served-model-name>`; the router derives the model name from the topic and folds it into the block hash, so it has to match the request's model.

## Test Plan

- [x] `python -m pytest tests/test_kv_events.py` (35 passed)
- [x] `black --check` and `ruff check` on both files, findings unchanged from `main`
- [x] Interop against llm-d-router `main`: the patched publisher (Python) bound at base port + 1 for rank 1, llm-d-router's `kvevents` subscriber, pool, and vLLM adapter dialed it; two chained `BlockStored` runs (64 tokens, block size 16) resolved all four request keys to the pod, and a `BlockRemoved` evicted the second run while the first stayed indexed

## Test Result

macOS, Python 3.11, pyzmq 27.2, msgspec 0.21. No GPU run; the change does not touch the engine path.

```
$ python -m pytest tests/test_kv_events.py -q
35 passed in 0.06s
$ python -m pytest tests/test_lmcache_offload_connector.py tests/test_token_ids_storage.py tests/test_scheduler.py tests/test_kv_connector_scheduler.py tests/test_prefill_scheduler.py -q
372 passed, 1 skipped in 1.97s
$ black --check --target-version py310 --target-version py311 atom/distributed/kv_events.py tests/test_kv_events.py
2 files would be left unchanged.
$ ruff check atom/distributed/kv_events.py tests/test_kv_events.py
5 findings, all pre-existing on main (S110/BLE001 in shutdown() and an existing test, C408 in a test helper)
$ go test ./pkg/kvevents/ -run TestATOMPublisherInterop -v      # llm-d-router main + patched ATOM publisher
atom_publish: bound tcp://127.0.0.1:15658 (rank 1)
    all 4 request keys resolved to 10.1.2.3:8000 via chained BlockStored runs
    BlockRemoved evicted the second run; first run still indexed
--- PASS: TestATOMPublisherInterop (16.38s)
```

## Submission Checklist

### DIFF
--- a/atom/distributed/kv_events.py
+++ b/atom/distributed/kv_events.py
@@ -6,6 +6,7 @@ plus ATOM extensions (`BlockTransferred`, CPU/DISK/REMOTE medium constants)."""
 
 from __future__ import annotations
 
+import itertools
 import logging
 import queue
 import threading
@@ -134,10 +135,18 @@ class ZmqEventPublisher(EventPublisher):
     batch is dropped — KV events are advisory and a missed eviction is
     cheaper than stalling inference.
 
-    With `topic=""` (default) each message is a single msgpack frame; with a
-    non-empty `topic` the publisher sends two-frame multipart messages
-    (`[topic, payload]`), so consumers must use `recv_multipart()` to read
-    the payload.
+    Every message is a three-frame multipart `[topic, seq, payload]`, the
+    layout vLLM's `ZmqEventPublisher` emits: `topic` is the (possibly empty)
+    subscription key, `seq` is a monotonic uint64 big-endian batch counter,
+    and `payload` is the msgpack-encoded `EventBatch`. Consumers must use
+    `recv_multipart()`. `seq` is assigned at enqueue, so a batch dropped on
+    queue overflow shows up as a gap in the stream instead of vanishing
+    silently. The all-0xFF value is never used as a data seq; vLLM reserves
+    it as the replay terminator.
+
+    With `data_parallel_rank` set, the bound endpoint is moved to that rank
+    (see `offset_endpoint_port`) so co-located DP ranks do not collide on one
+    socket, and subscribers reach rank N at base port + N.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


ZhangLirong-amd:
try to fix the `high` comment from copilot?

PR #2209, opened 2026-09-12
perf(moe): use fixed-slot MegaMoE for small decode graphs

yhl-amd:
MegaMoE currently uses `max_num_batched_tokens` as its capacity for every forward. With a typical capacity of 16384, even a 64-token decode graph uses AITER's compact path. This change preallocates a second, 128-token instance so eligible small forwards can use fixed-slot dispatch.

The fast path is enabled by default for native ATOM with EP8 and 48 experts per rank. Selection uses the current target/draft pass's group-agreed, padded token rows. For example, 64 requests with MTP3 require 256 target-verification rows and retain the original capacity, while a subsequent 64-row draft pass can use 128. Both instances are built in the same order across ranks before graph capture; an unwarmed cache miss during capture raises before symmetric allocation. Existing per-layer weight rebinding and EPLB aliases are preserved.

TBO workers, plugin bridges, non-unified shapes, and unsupported layouts retain the configured capacity. Set `ATOM_MEGA_DECODE_FAST_PATH=0` before startup to disable the optimization.

Validation:

- 133 passed, 3 skipped in the focused CPU regression suite on the rebased PR branch. Skips require AITER/CUDA. Coverage includes target/draft row counts, graph padding, mixed-rank metadata, plugin contexts, TBO, allocation order, capture cache misses, and weight rebinding.
- Black, Ruff, and `git diff --check` pass.
- Full-model serving on 8×MI355X: DeepSeek-V4-Pro, nominal 8192 input / 1024 output, random range ratio 0.8, concurrency 512, seed 0, 1024 warmups and 5120 measured requests, ignore EOS, no MTP. Both variants completed 5120/5120 with identical totals: 37,755,686 input and 4,718,313 output tokens.
- Mega startup logs confirm capacities 16384/128 with EP8/EPR48; 64/128-row graphs select 128 and 256/512-row graphs retain 16384. No server/client errors were found.

| Serving metric | DPA + TBO | DPA + EP8 + EPLB r0 + Mega | Change |
|---|---:|---:|---:|
| Output throughput (token/s) | 4914.75 | 5024.06 | +2.22% |
| Duration (s) | 960.03 | 939.14 | -2.18% |
| Mean TTFT (s) | 7.41 | 7.36 | -0.65% |
| Mean TPOT (ms) | 92.20 | 91.00 | -1.31% |
| Median ITL (ms) | 37.02 | 38.87 | +5.00% |
| P99 ITL (ms) | 1557.09 | 1662.72 | +6.78% |

This is one sequential comparison of complete serving configurations; repeat-run variance, an isolated capacity-toggle ablation, and full-model MTP accuracy remain unmeasured. EPLB migration costs are included. The workload and EPLB interval/window of 200 follow the performance catalog; the linked accuracy case uses 10, and GSM8K was not evaluated.

<details>
<summary>Reproduction and version details</summary>

CPU regression command:

```bash
python -m pytest \
  tests/test_mega_decode_capacity.py tests/test_mega_mxfp4_method.py \
  tests/test_forward_mode.py tests/test_dp_sync_layout.py \
  tests/test_dp_metadata.py tests/test_draft_graph.py \
  tests/test_scheduler_mtp_max_tokens.py tests/test_cudagraph_capture_bounds.py \
  tests/model_ops/test_moe_dp_token_capacity.py tests/test_envs.py -q
```

Serving was measured on ATOM `13a2fc427d02cf8778d9f0542e7988ee52669eff` plus the identical three-file patch, before rebasing onto current main. AITER: `b103cf92b07e26f1a1b7013fe90229408d3795c3`; image: `rocm/atom-dev@sha256:e67a284197327e113e8c0844b77dbaf1964057c74d7ec49713e82a929a8d4cd1`; Torch 2.10.0 / ROCm 7.2.4.

Common server settings: TP8, DPA, FP8 KV cache, FP4 index cache, prefix caching off, `max-num-batched-tokens=16384`, `max-num-seqs=512`, memory utilization 0.9, FULL graphs with sizes `[1,2,4,8,16,32,48,64,128,256,512]`. TBO uses `--enable-tbo prefill`. Mega uses `--enable-expert-parallel --eplb-enable --moe-backend mega` with redundant experts 0, naive placement, interval/window 200 and rebalance threshold 2.0.

Common environment: `AITER_BF16_FP8_MOE_BOUND=0`, `ATOM_MOE_GU_ITLV=1`, `GPU_MAX_HW_QUEUES=5`, `ATOM_NUMA_BIND=1`, `AITER_LOG_LEVEL=WARNING`, `MORI_SHMEM_HEAP_SIZE=16G`, `ATOM_MEGA_DECODE_FAST_PATH=1`.

</details>

### DIFF
--- a/atom/model_ops/fused_moe/flydsl_mega_experts.py
+++ b/atom/model_ops/fused_moe/flydsl_mega_experts.py
@@ -11,9 +11,13 @@ lives inside aiter, so there is no ``ATOM_FLYDSL_KERNELS_PATH`` and no
 MegaMoEV2 auto-selects its execution config from the runtime token count and
 ``max_tok_per_rank`` (MTPR): MTPR<=255 -> low-latency fixed-slot path,
 MTPR>=256 -> compact path (``FIXED_SLOT_MAX_MTPR``). MTPR must be a positive
-power of two. It is passed in as ``max_num_tokens`` (ATOM's
-``max_num_batched_tokens``) so that every rank uses the same value: MTPR also
-selects the p2p wire format, and a rank-dependent MTPR desynchronises it.
+power of two. The configured ``max_num_tokens`` remains the fallback capacity.
+For native ATOM EP8 with 48 experts per rank, DP-unified forwards with at most 128
+padded token rows can use a second, 128-token instance. Target verification
+and draft passes use their own token counts, including MTP and graph padding.
+MTPR also selects the p2p wire format, so the choice must use group-agreed
+metadata, never a rank's local request count. Set ATOM_MEGA_DECODE_FAST_PATH=0
+to retain the configured capacity for all forwards.
 
 Weight prep uses aiter's own shuffles (``aiter.ops.shuffle``):
   w1/w1_scale: ``shuffle_weight_a16w4(., 16, gate_up=True)`` /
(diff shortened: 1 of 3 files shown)


JiaoliangYu:
@GwilliamHu  plz check this

GwilliamHu (approved):

JiaoliangYu (approved):

PR #2223, opened 2026-09-14
fix(offload): reduce MiniMax M3 offload overhead on MI355X

yhl-amd:
This change addresses KV-offload overhead measured while serving **MiniMaxAI/MiniMax-M3-MXFP8 on 8×MI355X (TP8, concurrency 96)**. In the 900-second serving window, the combined configuration completed **218 requests**, compared with **123** for original ATOM P0 offload, **187** for ATOM without offload, and **85** for the tested SGLang direct-offload configuration. The measurements used the full MiniMax model; the synthetic byte probe is separate correctness validation.

KV completion polling could consume early-rank replies and discard them when a later rank timed out. Because worker replies drain completion state, a retry could never recover those reports. Preserve the partial batch and resume at the missing rank without broadcasting another poll.

Add default-off PAGE transfer controls for physically pinned asynchronous host copies and one prepared block-ID upload per transfer. Prepared plans keep the existing validation, stream ordering, source-safe boundaries, and final/recovery fences; failed recovery retains the ID owner with quarantined staging storage. Unsupported codecs/directions and pageable buffers retain their existing paths. Thread-local phase timings identify preparation, submission, and final synchronization without adding per-chunk fences.

MiniMax M3 already uses `M3OffloadConnector`, which passes its 297 K/V/scale/NSA-index PAGE regions to `DSV4PageSlotCodec` with no SLOT state and then to `BlockGPUConnector`. That shared PAGE codec is why transfer changes appear under `hybrid/dsv4`; it is the production M3 copy path. The model forward, attention, and MoE computation are unchanged.

The offload README links [MiniMax M3 offload tuning on MI355X](https://github.com/ROCm/ATOM/blob/codex/offload-completion-batched-ids/atom/kv_transfer/offload/TRANSFER_TUNING.md), a portable benchmark summary, and a manual four-combination M3 byte probe. Save-admission/coalescing prototypes are excluded from the implementation.

Validation:

- Current-main native CPU CI script: **5,350 passed, 381 skipped, 3 xfailed**. Focused offload suite: **150 passed**.
- Whole-repository Black: **774 files unchanged**; Ruff 0.16.7: **0 findings in PR diff context**, matching the CI filter. All accuracy catalogs and `git diff --check` pass.
- All five changed runtime ASTs match the measured recommended implementation; the committed probe differs only in licensing, formatting, import order, executable permission, and equivalent dictionary-literal syntax.
- Eight MI355X devices passed **128 byte-exact roundtrips** across batch IDs off/on × async copies off/on, including host/GPU guards, uneven groups, partial tails, and LMCache pinned allocation.

GPU figures come from the frozen P0 runtime using base `23a99bc5` plus the runtime overlay from #2201 at `cd3ca4513b`, rather than a new GPU run of this main-based branch. All rows below use MiniMaxAI/MiniMax-M3-MXFP8, TP8, c96, 101 warmup requests, 900 seconds sending, 30 seconds grace, and 300 seconds cancellation-credit drain:

| Configuration | Completed in 900 s | TTFT p50 (s) | TPOT p50 (ms) |
| --- | ---: | ---: | ---: |
| SGLang direct offload | 85 | 96.09 | 363.51 |
| ATOM without offload | 187 | 52.23 | 106.74 |
| Original ATOM P0 offload | 123 | 94.40 | 125.56 |
| Completion fix | 187 | 52.92 | 100.96 |
| Completion fix + async copies | 98 | 200.31 | 119.85 |
| Completion fix + async copies + batched IDs | **218** | **31.15** | **89.54** |

The combined run was 16.58% above completion-only throughput. All six final resource gauges were zero despite real KV-output timeouts. Whole-run D2H call p95 fell from 170.489 s to 3.384 s; transfer counts/sizes differ, and event spans include contention. There was one run per configuration; latencies include only requests completed within 900 seconds. Every client run reached its cancellation-credit drain timeout. The document records the exact model/dataset revisions, cross-engine capacity differences, H2D results, paired cohorts, and remaining limitations.


yhl-amd:
### MI355X TP8 / concurrency 96 / 900-second offload A/B

I ran an AIPerf `inferencex-agentx-mvp` comparison on 8x AMD Instinct MI355X with MiniMax M3, TP8, concurrency 96, and a 900-second request-sending window.

> **Version note:** these measurements are for commit `56688e8097179c1598780f4a98a05d1c7c58e558`. The PR head has since advanced to `5d72dc5776a19450f9ca60b6b1ea9b437a00920b`, so this is not a measurement of the current head.

The primary metrics below use a strict window: only successful requests completed before the end of the 900-second sending window are included. Requests completing during grace/drain and partial output from cancelled requests are excluded.

| Metric | Offload off | Offload on | Change |
|---|---:|---:|---:|
| Requests completed in sending window | 215 | 274 | +27.44% |
| Request throughput | 0.2389 req/s | 0.3044 req/s | **+27.44%** |
| Output token throughput | 238.13 token/s | 289.73 token/s | **+21.67%** |
| Input token throughput | 32,970.40 token/s | 41,555.44 token/s | +26.04% |
| Total token throughput | 33,208.53 token/s | 41,845.18 token/s | **+26.01%** |
| TTFT p50 | 57.98 s | 23.77 s | **-59.01%** |
| TTFT p95 | 478.06 s | 473.86 s | -0.88% |
| TPOT p50 | 84.49 ms | 65.53 ms | **-22.44%** |
| TPOT p95 | 322.86 ms | 316.16 ms | -2.07% |
| E2E p50 | 122.61 s | 79.00 s | -35.57% |
| Prompt cache-read ratio | 15.69% | 40.31% | +24.62 pp |

Run details:

| Configuration | Warmup | Sent | Completed at 900 s | Final success | Cancelled | Errors | Full profiling phase |
|---|---:|---:|---:|---:|---:|---:|---:|
| Offload off | 410.31 s | 331 | 215 | 217 | 114 | 0 | 1230 s |
| Offload on | 448.13 s | 392 | 274 | 278 | 114 | 0 | 1230 s |

The offload-on server registered all eight rank-local M3 PAGE workers (297 PAGE regions per rank). The full server log contained approximately 5,848 rank-level save profiles and 928 rank-level load profiles, confirming that this run exercised both save and load rather than only enabling the connector configuration.

For this workload, offload improved request throughput by 27.44%, total token throughput by 26.01%, TTFT p50 by 59.01%, and TPOT p50 by 22.44%. The p95 TTFT/TPOT improvement was small, so the change improved typical latency but did not remove the extreme tail.

Caveats: this is a single A/B run with no confidence interval; both phases reached the 300-second cancel-credit drain limit after the fixed 900-second sending window.

PR #2228, opened 2026-09-14
feat(moe): reach MoRI's low-latency kernels through the all2all manager

JiaoliangYu:
`--all2all-backend low-latency` could not select MoRI's low-latency kernels through aiter's all-to-all manager. ATOM instead carried a second kernel-selection table, and its low-latency branch selected intra-node `AsyncLL` even for a real inter-node process group.

This PR moves the transport decision to ROCm/aiter#5493. ATOM passes the requested latency mode to `MoriAll2AllManager`, which combines it with the measured topology and selects `IntraNode`, `InterNodeV1`, `AsyncLL`, or `InterNodeV1LL`. 
The duplicated `select_mori_kernel_params`, `init_mori_op`, and secondary MoRI config dictionary are removed. This also avoids reconstructing `data_type` from item size, which cannot distinguish every supported format.

When TBO is enabled, ATOM asks the manager for two indexed handles with the same transport configuration, one per concurrent ubatch. Slot 0 also serves regular forwards; slot 1 exists only for the second ubatch. Without TBO, ATOM creates only slot 0. The async call path uses split send/recv for intra-node `AsyncLL` and the existing communication-stream path for monolithic kernels.

MoRI launch geometry is passed by keyword. The previous positional calls placed `warp_per_block` into `rdma_block_num` and left the actual warp argument at its default, which is especially harmful for inter-node kernels.


ZhangLirong-amd (approved):
do we have any perf data from mori low-latency?

JiaoliangYu:
> do we have any perf data from mori low-latency?

i've updated test result in pr message. Now it's only for wideep func, perf optimization is still on the way

PR #2229, opened 2026-09-14
[MiniMax-M3] Indexer-only decode context parallelism (opt-in, off by …

ganyi1996ppo:
## Summary

MiniMax-M3's lightning indexer scores every 128-token block of context to pick
the top-16 that sparse attention will actually read. `index_q` shards like KV
heads, and M3 has `num_key_value_heads == sparse_num_index_heads == 4`. At TP4
each rank therefore gets **one** index head, and at decode `query_len == 1`, so
the score kernel's `tl.dot` sees M = 1 against a hardware MMA floor of 16 —
15 of every 16 columns idle.

This PR adds an opt-in path that keeps sparse attention, KV geometry, the
scheduler and MoE exactly as they are, and changes only the indexer: every rank
projects **all 4** index heads and scores **1/P of the context blocks**
(round-robin, `owner(block) = block % P`). Same multiply count, full MMA
occupancy. An all-to-all then routes each head's top-k *candidates* — not its
scores — to the rank that owns that head, which merges to the global top-k.

**Selection is bit-identical to the existing TP path**, so this is an exact
optimization, not an approximation. The end-to-end bar is identical tokens.

Enable with `--indexer-dcp-only` / `dcp_config.indexer_dcp_only`, or
`ATOM_M3_INDEXER_CP=1`. **Default off.**

## Performance — AgentX agentic replay, conc 32

Real agentic-coding traces (`semianalysis_cc_traces_weka_062126`), 393 entries,
1800 s measured per arm after a 354-request warmup. MiniMax-M3-MXFP4, TP4,
`--block-size 128`, `--kv_cache_dtype fp8`, `--gpu-memory-utilization 0.8`,
prefix caching on, index cache off, CUDAGraph on (default `--level 3`).

Both arms ran **serially on the same GPUs (0–3)**, one server at a time, so
neither the box's 0.52% GPU-set bias nor host CPU/RAM contention can enter.

| Metric | CP | TP | CP vs TP |
|---|---:|---:|---:|
| **Total token throughput** | **109,665 tok/s** | **88,970 tok/s** | **+23.3%** |
| **Output token throughput** | **776.4 tok/s** | **616.2 tok/s** | **+26.0%** |
| Output tok/s per user | 48.43 | 37.68 | +28.5% |
| **ITL p50** | **23.17 ms** | **36.53 ms** | **−36.6%** |
| ITL avg / p90 / p99 | 23.38 / 29.96 / 53.97 ms | 33.68 / 44.22 / 68.24 ms | −30.6% / −32.2% / −20.9% |
| Request latency avg / p50 | 22,149 / 7,124 ms | 30,219 / 9,606 ms | −26.7% / −25.8% |
| Requests completed in 1800 s | 1,643 | 1,398 | +17.5% |
| TTFT avg / p50 / p99 | 1,095 / 316 / 35,459 ms | 1,138 / 304 / 40,904 ms | −3.8% / +3.9% / −13.3% |
| Request error rate | 0.06% (1) | 0.14% (2) | — |
| Prompt cache read | 92.06% | 92.09% | ±0 |

Because 92% of prompt tokens are prefix-cache reads the GPU never computes, here
is the same result counted over **only the tokens actually processed**:

| Tokens the GPU actually computed | CP | TP | CP vs TP |
|---|---:|---:|---:|
| Prefill (prompt − cache read) | 15.92 M | 12.85 M | +23.8% |
| Decode generated | 1.43 M | 1.13 M | +26.0% |
| **Total / sec** | **9,637 tok/s** | **7,771 tok/s** | **+24.0%** |

TTFT is flat because **prefill deliberately stays on the TP path**
(`attention_mha.py:1343`); only decode is context-parallel. The prefill
throughput gain is second-order: faster decode frees scheduler time, so 245 more
requests complete in the same window.

### Confounders checked

**Batch size — it works against CP.** TP ran at *higher* concurrency and still
lost, so the win is not a batching artifact:

| | CP | TP |
|---|---:|---:|
| Effective decode concurrency | 18.80 | 22.10 |
| Tokens in flight (avg) | 3.45 M | 4.03 M |
| Server-side mean `Running:` reqs | 17.45 | 20.58 |
| → per-request decode rate | **41.25 tok/s** | **27.85 tok/s** |

Larger batches normally *help* decode throughput. TP's larger batch is a symptom
of its slowness (requests linger), so a matched-concurrency comparison would, if
anything, widen the gap.

**Input length — eliminated by bucketing.** CP's average ISL is 4.9% longer
(121,946 vs 116,289 tokens) and the indexer win grows with context, so this had
to be removed. Median ITL of successful profiling-phase requests by ISL bucket:

| ISL bucket | n CP | n TP | ITL med CP | ITL med TP | CP vs TP |
|---|---:|---:|---:|---:|---:|
| 0–32k | 124 | 108 | 21.58 ms | 34.32 ms | −37.1% |
| 32k–64k | 361 | 316 | 22.53 ms | 35.65 ms | −36.8% |
| 64k–128k | 611 | 564 | 22.91 ms | 36.24 ms | −36.8% |
| 128k–256k | 414 | 308 | 23.81 ms | 37.66 ms | −36.8% |
| 256k+ | 129 | 100 | 23.51 ms | 38.88 ms | −39.5% |
| **overall** | 1639 | 1396 | **23.18 ms** | **36.54 ms** | **−36.6%** |

Flat at −36.8% ± 1% in every bucket, matching the overall figure — the ISL
difference explains none of the win. (Buckets are flat rather than rising
because decode is batched: a 20k-context request shares its forward pass with a
700k one, so its ITL reflects the batch's total context, not its own.)

**Prefix caching — identical.** 92.06% vs 92.09% read, against theoretical
ceilings of 96.71% / 96.58%.

**Warmup — identical.** 354 requests, 0 errors, 339.6 s (CP) vs 344.9 s (TP).

### Arm verification

Two independent signals, because the `Engine kwargs` dump prints the
*pre-override* value and reads `indexer_dcp_only=False` in **both** arms:

| | CP | TP |
|---|---|---|
| `grep -c "indexer_dcp_only enabled" server.log` | **1** | **0** |
| `peak_torch` | 61.18 GB | 61.05 GB |
| `num_kvcache_blocks` | 52,781 | 52,826 |

The 133 MB `peak_torch` delta equals the 45-block KV delta exactly
(45 × 2,961,408 B = 133.3 MB) — the widened `index_q` projection physically
allocated in CP only. That is a hardware-level arm confirmation independent of
any log line. Cost of the feature when enabled: **0.09% of KV capacity.**

## Accuracy — gsm8k, no regression

`lm_eval --model local-chat-completions --tasks gsm8k --num_fewshot 5
--apply_chat_template --fewshot_as_multiturn`, 1319 problems, greedy
(`temperature=0`), 3 repetitions per arm on the same server config as above.

| rep | CP strict | TP strict |
|---|---:|---:|
| 1 | 0.9393 | 0.9393 |
| 2 | 0.9462 | 0.9477 |
| 3 | 0.9439 | 0.9454 |
| **mean** | **0.9431** | **0.9441** |
| within-arm spread | 0.69 pt | 0.84 pt |

**Delta: −0.100 pt, against a within-arm spread of 0.69–0.84 pt.** The
difference is well inside run-to-run noise — rep 1 is bit-identical across the
two arms. Both arms clear the `recipes/MiniMax-M3.md` MXFP4 reference of 0.9363.
Zero tracebacks, zero retries in all six runs.

This is the expected result: block selection is bit-identical to the TP path, so
accuracy parity is a *check that the wiring is correct*, not a quality tradeoff
being measured.

`max_tokens=8192` is load-bearing for this eval — M3 fills `reasoning_content`
first and returns **empty** `content` if truncated, which scores 0 with no error.

## Design

- **Opt-in flag `dcp_config.indexer_dcp_only`, default off.** It lives in the DCP
  family because it *is* decode context parallelism, applied to the indexer only.
- **It requires `decode_context_parallel_size == 1`.** M3 has no DCP awareness,
  and real DCP shards the main KV pool (`block_manager.py:110,121` —
  `hash_block_size = block_size * dcp_world_size`), which would corrupt M3's KV.
  The flag means "context-parallelize the indexer *instead of* the KV cache".
- **TP4 only in v1**, which collapses the CP group onto the TP group exactly
  (`num_kv_head_replicas == 1`), so **no `new_group` call is needed** — it reuses
  `get_tp_group().device_group`. This removes the whole class of
  "every rank must call new_group for every subgroup" init-hang bugs.
- **Native ATOM only.** The vLLM and SGLang bridges stay on the TP path, and this
  is *enforced* in `indexer_cp_enabled()` rather than documented as a convention.
  Both bridges reshape `index_q` with a `view` on a width they hardcode
  (`plugin/vllm/attention/minimax_m3_attnetion.py:220`,
  `plugin/sglang/attention_backend/minimax_m3_sparse.py:869`), and a 4-head
  tensor viewed as 1 head yields 4x the rows **with no error at all** — silent
  corruption, not a crash.
- **Decode only.** Prefill keeps the existing kernel on a sliced head.


valarLip:
## Review — indexer-only decode context parallelism

Reviewed at `c518d34af` in an isolated worktree. I traced the pack/unpack round trip, the causal math, the forced-block tiering, and the "a shard's discarded block cannot be a global winner" argument against `index_topk.py`, and **the algorithm is correct** — it matches the native selector. The problems are all in the plumbing around it, and the largest one is that most of this plumbing already exists in the repo.

Findings marked **[verified]** I reproduced or read out myself; **[reported]** ones match the code shape but I did not independently confirm.

---

## 1. `atom/model_ops/dcp_ops.py` already has this, at three separate levels

### The algorithm has an existing implementation — `dcp_ops.py:888 dcp_decode_candidate_exchange_fused` **[verified]**

Its flow is stage-for-stage identical: score the local shard → local top-k → all-gather → merge. **Even the lemma is the same sentence:**

```
dcp_ops.py:910   "Why the local top-k suffices: fewer tokens outrank a given token
                  locally than globally, so a token in the global top-K is in its
                  own rank's local top-K..."

this PR, module docstring:6
                 "a rank that keeps k candidates cannot drop a global winner,
                  because a block it discarded lost to k blocks it kept..."
```

That one is already on aiter kernels (`top_k_per_row_decode` + `flydsl_dcp_topk_merge`); this PR hand-writes three Triton kernels for it.

The two are genuinely not interchangeable — that one is token-granular (DSA/V4) and documents a tie-break caveat, this one is block-granular and claims bit-identical. **But the PR never says why a second implementation has to exist.** Please add that argument, or reuse. The repo already carries the cost of one candidate-exchange implementation; a second one means every future change to the pattern has two places to find.

### The transport already has a correct form — and it removes the `except` entirely **[verified]**

`dcp_ops.py:59 dcp_all_gather` + `:38 _ag_custom_view_dtype` solve **precisely** the problem this PR works around: aiter's custom collective dispatches on a float enum and raises on anything else, so non-float payloads must be viewed first. Its answer is a **deterministic, rank-uniform predicate** — no supported view → return `None` → caller falls back to pynccl. **No try/except.**

This PR instead open-codes `keys.view(torch.int32)` plus:

```python
# indexer_cp.py:142
try:
    gathered = group.custom_all_gather(keys.view(torch.int32))
except (AssertionError, RuntimeError, AttributeError):
    return None
```

with twelve lines of comment re-deriving the int64 → float64 dead end that `_ag_custom_view_dtype`'s docstring already explains.

`_ag_custom_view_dtype` currently handles itemsize 4/2/1 and returns `None` for 8. **Add one branch:**

```python
if itemsize == 8 and x.shape[-1] % 2 == 0:
    return torch.float32
```

and `dcp_all_gather` serves this call site directly. That is a one-line change that deletes the try/except *and* the correctness problem in §4.1 below.

### The round-robin arithmetic already has a home — `atom/distributed/dcp_layout.py` **[verified]**

`dcp_global_pos` / `dcp_owner_rank` degenerate to exactly `j*W + r` / `pos % W` at `S == 1`. This PR open-codes `block = local * WORLD + RANK` in three Triton kernels. Triton can't import them, fine — but please **name the relationship in a comment at each site**, otherwise a future change to the interleave rule will not grep to these three.

---

## 2. Module boundaries: all three file names describe something other than their contents **[verified]**

| File | What the name promises | What is in it |
|---|---|---|
| `model_ops/minimax_m3/indexer_candidate_exchange.py` | candidate **exchange** | **zero exchange code** — two Triton *selection* kernels |
| `model_ops/minimax_m3/indexer_context_parallel.py` | context parallelism | one scoring kernel + one **dead** selection function |
| `distributed/indexer_cp.py` | — | the actual exchange (`exchange_candidates`) lives here |

Selection is split across two files, the exchange is in a third, and the file named "exchange" contains none. The natural cut is **scoring / selection / transport** — currently no cut lands on a seam. This is a direct Name-matches-function violation.

**`indexer_cp.py` also mixes three concerns in one 183-line file:** process-group lifecycle, a *model-topology policy gate* (`indexer_cp_unsupported_reason` — a pure function over arch name, kv-head count, sparse_block_size), and a transport with a payload-size heuristic. The consequence: `config.py:2235` now imports from `atom/distributed/` to do **startup-time validation**. The config layer takes a dependency on a distributed runtime module for a predicate that touches no distributed state.

`config.py` also grew 102 lines for one opt-in perf flag, including a block of M3-specific topology reasoning.

---

## 3. Cleanliness

### The packed-key codec is now split across two files **[verified]**

`_pack_score_key` lives at `index_topk.py:270`. This PR is the **first** caller that needs the inverse (native only needs `& 0xFFFF` to recover the id), and it hand-writes that inverse inside a kernel body in a different file:

```python
# indexer_candidate_exchange.py:149-155
block = (key & 0xFFFF).to(tl.int32) - 1
score = ((key >> 16) & 0xFFFFFFFF).to(tl.uint32)
bits  = score ^ tl.where(score >> 31 != 0, 0x80000000, 0xFFFFFFFF)
value = bits.to(tl.float32, bitcast=True)
```

`0xFFFF`, `>>16`, `0x80000000`, `0xFFFFFFFF` all restated. And the decoder was written against only the **last** line of the encoder: `_pack_score_key:293` has a `-0.0` canonicalization step (`bits = tl.where(bits == 0x80000000, 0, bits)`) that the inverse does not model. Harmless today — 0 decodes to +0.0 — but it is a live demonstration of the hazard. **Put `_unpack_score_key` next to `_pack_score_key` in `index_topk.py`** so the two ends of the format stay together.

### Hard-coded values bypassing the shared constants this file already imports **[reported]**

`merge_candidate_keys:297-298` states `NUM_KV_HEADS=1`, `PAGES_PER_BLOCK=8` and a literal `128`, while importing `_alloc_emit` from the module that owns them. `_alloc_emit` sizes each row as `topk * PAGES_PER_SPARSE_BLOCK` and `index_topk.py:60-62` carries an explicit "must match" comment; the native launcher threads it through (`index_topk.py:512`). If `ASM_PAGE_SIZE` changes, `_alloc_emit` allocates `topk*4` while `_merge_topk` still scatters `bt_slot*8 + pj` — an overwrite into the next allocation, **and only on the CP path; TP stays green**.

### `select_global_blocks` is 56 lines of dead code **[verified]**

`grep -rn select_global_blocks` across the whole worktree returns exactly one hit: its own `def`. It is also the sole reason this file imports `DECODE_TOPK_BLOCK_SIZE_K` and `_launch_select`.

Worse, its declared contract is **incompatible** with the only function that could feed it — `indexer_context_scores`, 90 lines above. It validates `block_table.shape == (batch, scores.shape[2])` and documents `[local_heads, tokens, global_blocks]`; `indexer_context_scores` returns `[heads, tokens, cdiv(blocks, world)]` in round-robin **local** numbering. The names read as a pair. Someone wiring them together gets either a geometry `ValueError` or — if the shapes happen to line up — a top-k over shard-local indices interpreted as logical block ids, feeding wrong pages to sparse attention with no error.

### Two dead `constexpr` parameters **[verified]**

`_local_topk`'s `HEADS` and `_merge_topk`'s `TOKENS` are never referenced in their kernel bodies (checked by AST over the full parameter lists). They only widen the JIT specialization key.

`merge_candidate_keys:281` discards `_alloc_emit`'s returned `args` (`output, _ = ...`) and then recomputes `output[0].stride(0)` at `:291`.

---

## 4. Performance

### 0.41 ms/decode-step of host-side validation **[verified — measured]**

I lifted the validation blocks from the three entry functions verbatim and timed them (B=64, topk=512, 128K context):

```
indexer_context_scores      3.00 us/call
local_candidate_keys        1.85 us/call
merge_candidate_keys        2.36 us/call
per sparse layer            7.13 us
x57 sparse layers           0.406 ms / decode step
```

24 multi-clause checks across the three entry points, **all re-run on every call**, and most of what they check (dtype, ndim, contiguity, stride, device) is invariant in steady state. The multiplier depends on the `index_topk_freq` skip pattern — `_load_cached_topk` only serves skip layers — so 57 is an upper bound, but it scales linearly. Given `perf(v4): cut the host work between two forwards` landed recently, this is worth measuring on a real M3 run before merge.

(For the record: the four function-local imports in `_cp_index_topk_decode` are **not** part of this. I measured them at 0.21 µs/call — 0.012 ms/step at 57 layers. Negligible.)

### Runtime quantities promoted to compile time **[reported]**

All three kernels take `TOKENS` / `LOCAL_BLOCKS` / `GLOBAL_BLOCKS` / `CHUNK` / `N` / `SRC_STRIDE` / `ROW_STRIDE` / `REAL_CANDIDATES` as `tl.constexpr`. Native passes `chunk_blocks`, head_dim, all strides and `topk` as runtime args and pins `BLOCK_SIZE_K` to a constant — with the reason spelled out at `index_topk.py:344-351` ("Decode cannot derive it, because it is captured... So decode takes a constant").

Under the default PIECEWISE mode attention runs eager, so `max_seq_len` moves every ~128 generated tokens and `tokens` moves with the running batch. A long-context, high-concurrency serve will compile many variants mid-serve — the `hipModuleLoadData`-growth regression class this repo has hit before.

### Loop bounds derived from `max_model_len` rather than the live length **[reported]**

`_context_score:52` bounds by `min((chunk+1)*CHUNK, LOCAL_BLOCKS)` and `_local_topk:86` uses `tl.range(BLOCK_SIZE_K, LOCAL_BLOCKS, BLOCK_SIZE_K)` (constexpr trip count, fully unrolled). Native computes `num_blocks = cdiv(seq_len, 128)` and `return`s when `chunk_start >= chunk_end` (`index_topk.py:694-698`). At `--max-model-len 1048576` with world 4, `LOCAL_BLOCKS = 2048`, so **a 2K-token request still runs every chunk program**. The unconditional `-inf` stores are also dead work — `_local_topk` already masks those lanes via `block < causal_blocks`, so nothing reads them.

---

## 5. Correctness

### 5.1 The all-gather vs all-to-all choice is made **per rank**, inside a try/except **[verified — code shape]**

`indexer_cp.py:144` swallows `AssertionError`, `RuntimeError` and `AttributeError` and returns `None`, dropping that rank onto `dist.all_to_all_single` while its peers stay in `custom_all_gather`. `torch.OutOfMemoryError` is a `RuntimeError` subclass and is reachable: `all_gather_unreg` allocates a world-times-larger `torch.empty` whose success depends on that rank's fragmentation. One rank diverging is an RCCL hang — per layer, per decode step, with no log line.

**`dcp_all_gather` is the existing fix; see §1.**

### 5.2 The topology gate reads `tensor_parallel_size` before `--enable-dp-attention` rewrites it **[verified]**

`config.py:2237` reads it in `__post_init__`; `engine_core_mgr.py:295` sets `config.tensor_parallel_size = 1` **after** Config is built (documented at `config.py:1793`), and `__post_init__` is not re-run on the worker's unpickled copy.

With `-tp 4 --enable-dp-attention` the gate sees 4, passes, and logs "scores all 4 index heads over 1/4 of the blocks". In the worker, three sites then derive the index-head width from three different quantities:

- `attention_mha.py:1158` → `get_indexer_cp_world_size()` = **1** (and it *discards* the caller-supplied `index_q_size` at `:1161` instead of asserting it matches)
- `linear.py:2007` → `total_num_index_heads` = **4**
- `models/minimax_m3.py:457` → `total_idx_heads` = **4**

`aiter.fused_qknorm_idxrqknorm` locates `index_q` by byte offset inside `qkv`, so this is **silent corruption, not a crash**.

### 5.3 The CP world size is taken from `keys.shape[0]`, not `group.world_size` **[verified]**

`indexer_cp.py:135` does `world, tokens, topk = keys.shape`, then `:150` reshapes the gathered buffer as `(world, world, tokens, topk)`. The true leading dim is `group.world_size`. On a mismatch the element count can still factor (heads=2, group=4 → gathers `4*2*T*K`, reshapes as `(2,2,T,K)`), so the view succeeds, reads the wrong half, and drops the rest — wrong candidate set, no exception. The module docstring at lines 17-19 explicitly plans for the `TP > num_kv_heads` widening that produces this.

One `if world != group.world_size: raise` converts every variant into a first-step error.

### 5.4 `ATOM_M3_INDEXER_CP` is the only bare `os.getenv` in `envs.py` **[verified]**

```python
# envs.py:203
"ATOM_M3_INDEXER_CP": lambda: os.getenv("ATOM_M3_INDEXER_CP"),
# config.py:2230
if envs.ATOM_M3_INDEXER_CP is not None:
    self.dcp_config.indexer_dcp_only = envs.ATOM_M3_INDEXER_CP == "1"
```

`true` / `yes` / `on` / `""` / `"1 "` are all `not None` (the override fires) and all `!= "1"` (it sets **False**), silently, with no warning — the `logger.info` at `:2253` only fires on the ON arm. So each of those **disables** the feature *and overrides* an explicit `--dcp-config '{"indexer_dcp_only": true}'`. `docker run -e ATOM_M3_INDEXER_CP` with the var unset in the parent exports `""` and hits exactly this.

The repo has the correct tri-state idiom already: `ParallelConfig.__post_init__` at `config.py:967` uses `envs.is_set("ATOM_DP_SIZE")` with the lambda doing the coercion.

This is also the root cause of the problem the PR body concedes — the `Engine kwargs` dump printing the pre-override value — which is what forced arm verification onto a log grep and a VRAM delta.

---

## 6. Tests and docs

**No test pins the "bit-identical to the TP path" claim, and no new kernel has a test.** The one new test file covers the gate only, and omits the `"vllm"` arm.

**The diff touches no file under `docs/` or `recipes/`.** Concretely:

- `--dcp-config`'s help enumerates exactly four keys and states "unknown keys raise so a typo fails at startup" — an operator reading it concludes `indexer_dcp_only` is a typo
- `docs/environment_variables.md` documents every other `ATOM_*` flag but not this one
- `docs/distributed_guide.md` and `docs/model_ops_guide.md` maintain file tables the three new modules are absent from
- `recipes/MiniMax-M3.md` is untouched despite a claimed +23% on that exact model
- `DCPConfig`'s own class docstring still enumerates only the original four fields
- **The PR body says "Enable with `--indexer-dcp-only`" — no such argparse option exists.** Following it gives `unrecognized arguments`

---

`black --check` and `ruff check` pass on all changed files; `tests/test_indexer_cp_gate.py` collects and passes without aiter/triton/GPU.

**Suggested order of attack:** the `_ag_custom_view_dtype` one-liner is the highest-leverage single change — it deletes §5.1 and the twelve-line dtype workaround at once. Then §5.2 and §5.4, which an env-toggled A/B will hit immediately. Then the file split in §2, while the module is still new.

valarLip (approved):

PR #2231, opened 2026-09-14
Support LMCache KV offload for GLM-5.2 on the vLLM plugin backend

PerryZhang01:
## Summary

Adds an LMCache-backed KV offload tier for **GLM-5.2 on the vLLM plugin backend**
(`vllm-atom`), and the work needed to make that tier pay for itself rather than
cost throughput.

Main already carries the plugin connector framework
(`atom/plugin/vllm/kv_transfer/{connector,seq_view,offload_config}.py`,
`register.py`) — MiniMax-M3 was the first model through it. This PR is not a new
connector. It is:

1. the layout support GLM-5.2's mixed KV needs to travel through that framework,
2. four bugs that had to be fixed before it would run at all,
3. five performance changes, each with its reason and its measured effect,
4. the removal of four of our own optimisations that main's #2223 superseded.

**Net on the 64-prefix working point: +55.97% output throughput, TTFT p50 −89.8%.**

---

## 1. Layout support — what GLM-5.2 needed, and why

GLM-5.2 (`GlmMoeDsaForCausalLM`) cannot use LMCache's own GPU connector, which
probes for **one** global KV format. GLM-5.2 registers **two**: 78 sparse-MLA
latent layers at 576 B/token and 21 DSA indexer layers at 132 B/token
(47,700 B/token/rank). Driving ATOM's `DenseKVByteCodec` from vLLM's
`KVConnectorBase_V1` instead means LMCache only ever stores **opaque bytes** and
no format probe runs — so stock LMCache 0.4.5 is enough, with no source build of
0.5.x.

`kv_cache_layout.py` (+76) is where the model-specific part lives:

- **Two indexer spellings, folded by name and never by shape.** M3 registers
  `<layer>.index_cache` owned by `<layer>`; GLM-5.2 and DeepSeek-V3.2 register
  `<p>.indexer.k_cache` owned by `<p>.attn`. The GLM pairing is the one
  `AiterMlaSparseIndexerMetadataBuilder` itself uses, so it is the model's
  convention, not a guess here. Folding on shape instead would attach a real
  layer that merely *looked* indexer-shaped to its neighbour, and restore it
  under the wrong key — silently.
- **Only 21 of 78 layers own an indexer** (IndexShare). A layer without one must
  get no `index_cache` slot; an empty slot would change the codec's per-block
  byte stride for every layer after it.
- **No scale hook.** GLM-5.2's fp8 indexer packs its scale *into* the moved row,
  and its MLA layers carry only vLLM's scalar `_k_scale`. Nothing is left behind,
  unlike M3.
- **Hard-fail on ambiguity**: mismatched block counts across registered tensors,
  and more than one KV cache group. Two groups would still divide evenly often
  enough to pass the codec's own `numel // num_blocks` check and then slice the
  smaller tensor at the wrong granularity, with nothing logged.

`atom/kv_transfer/offload/` gains the completion tracking the vLLM plugin path
needs and the ATOM native path did not: an inflight-job fence
(`wait_for_requests`) so a preempted request's copy jobs stop reading HBM before
its blocks are reused; failed-load block draining (`take_load_error_blocks`) so a
partially filled block is re-prefilled rather than read as valid; and by-request
completion on the chunked scheduler, because vLLM's `KVConnectorOutput` carries
only plain request-id strings.

## 2. Four bugs fixed before it would run

| commit | what was wrong, and why it mattered |
|---|---|
| `31f3e77a6` fix(sparse-mla) | The indexer row-chunk bound was **inclusive**, so GLM-5.2's 16384×32768×4 B logits tensor landed on exactly 2 GiB. aiter's `fp8_mqa_logits` then fell to the `USE_BUFFER_STORE=False` gluon specialisation, which fails codegen on gfx950: `Assertion 'Begin <= End'`, all four TP workers dead ~13 s in, no traceback. Crashed with or without a connector attached. The bound is now exclusive, and the 2 GiB buffer-descriptor cap is enforced **even when the soft budget is 0** — it is a hard hardware limit, not a tuning knob. |
| `633bcfad1` fix(vllm-plugin): grow the block table | The block table was set only in `update_state_after_alloc`, which vLLM calls once per request. Blocks allocated later arrive via `scheduled_cached_reqs.new_block_ids`. Any prompt longer than one prefill budget therefore offloaded **only its first chunk**. |
| `69a2c5b01` fix(vllm-plugin): forward completions | ATOM's connector completions were not forwarded from the worker, so block leases were never released. |
| `5a9bac5e4` fix(offload): stop a failed load from livelocking a slot | A request whose whole prompt hits gets `hit -= 1` so the model has a token to run on. When that lands below a chunk boundary, the `LoadSpec` asks for a chunk the tier cannot serve; the load fails — and **nothing remembered that it had**. The request re-parked and re-asked, forever. One request retried 137 times, re-reading ~1.45 GiB each attempt. |

The livelock is why an earlier revision of this PR reported the *opposite*
conclusion (−48.41%, "costs throughput at both working points"). That run was not
transferring, it was retrying; the arithmetic built on it ("1.64× recompute cost",
"break-even needs ~14 reads") was arithmetic over a livelocked run and does not
hold. Same pool, same seed, after the fix: **−48.41% → +35.21%**.

## 3. Performance changes, and why each one

| # | change | reason | measured |
|---|---|---|---|
| a | `191d27458` size the default GPU staging buffer **in bytes, not chunks** | `OFFLOAD_GPU_STAGING_CHUNKS` defaults to `2` but is denominated in *chunks*, and a chunk is `LMCACHE_CHUNK_SIZE / block_size` blocks — not a fixed size. The reference config (chunk 256 / block 32) makes that 8 blocks = 33.5 MiB; GLM-5.2 (chunk 64 / block 64) makes it 5.8 MiB, **5.75× smaller than the geometry the default was tuned against**. The cost that buffer amortises is per-group, so a small-chunk model pays the difference directly. | **+5.20%** |
| b | `a89cc89cb` size the pack grid **by the bytes**, not by the widest segment | The old rectangular grid used `cdiv(nblocks × max_seg_bytes, 1024)` columns, pricing every segment at the widest — so most programs carried little or no payload and per-lane efficiency was hidden behind launch count. A flat tile table launches one program per 1024 payload bytes, all of them full. `_NUM_WARPS` drops 8 → 2 in the same commit because it is the other half of the same change: at `_BLOCK_BYTES = 1024`, 8 warps is 2 B/lane and 2 warps is 8 B/lane, and sizing the grid by bytes is what makes bytes-per-lane the dominant cost instead of program count. | kernel **1.64×** (403.0 → 660.8 GB/s) on the real geometry; see caveat below |
| c | `d20b91e21` build token-id tensors through numpy | `torch.tensor(list_of_ints)` unboxes element by element while holding the GIL, on a path that runs per group. | save-side overhead **−20.7%** |
| d | `10f61649a` make the load pool width tunable | The load pool width was hard-coded; retrieve is the leg that gates TTFT, so it needs to be sweepable without a patch. Adds `OFFLOAD_LOAD_WORKERS` / `OFFLOAD_COPY_WORKERS`; **defaults unchanged**. | enables measurement; no default change |
| e | `89fe335b6` skip the stage handshake when both stages share a stream | When stage A and stage B are on the same stream, the event record/wait pair is four HIP calls that order work already ordered. Each blocking call drops and re-acquires the GIL, which the model's threads are contending for. Gated by `OFFLOAD_SINGLE_STREAM`, **default off**. | 86 lines of tests; off by default pending an end-to-end arm |

**Caveat on (b), stated plainly.** 1.64× is the *kernel in isolation*, measured on
the geometry the server itself logs (296 segments, max 16384 / min 512,
`bytes_per_block` 3,008,512): 37,888 programs at 403.0 GB/s → 23,968 at 660.8 GB/s.
`a89cc89cb`'s message also quotes a 41× taken on a much larger group; **1.64× is the
figure to trust** — the larger ratio is the old grid's launch overhead at a group size
the server does not use. End to end on the **save** side it is worth far less again:
in-server `pack_ms` moved −15.6%, but `pack_ms` is not the kernel (1.78 ms in-server
against 0.060 ms standalone — the kernel is 2–4% of it, the rest is host launch gap),
which puts the throughput upper bound at 0.263% against a 0.109% repeat floor, because
`_do_save_req` runs on a 4-worker executor averaging 0.26 threads busy and never blocks
the forward pass. The leg this is kept for is **retrieve**, where
unpack is on the critical path. The warp-geometry sweep that picked 2
(643/634/632 GB/s at 512×1, 1024×2, 2048×4; 448 at 1024×8; 45 at 8192×1) was run
on the M3 segment geometry and its raw output was not retained — the numbers live
in `a89cc89cb`'s message. Correctness is pinned independently: old and new grids
agree over all **17,294,967 bytes** of a deliberately lopsided geometry (segments
7 / 512 / 1023 / 1024 / 1025 / 16384 / 980992, with an empty chunk among the
counts).

Staging-buffer sweep behind (a) — TP4, 28K prefix, pool 64, 600 s, seed 71502,
0 errors in all five:

| `OFFLOAD_GPU_STAGING_CHUNKS` | per buffer | tok/s | vs 2 | TTFT p50 |
|---|---|---|---|---|
| 2 (old default) | 5.8 MB | 378.73 | — | 586.61 ms |
| 11 | 33.6 MB | 403.24 | +6.47% | 436.58 ms |
| 16 | 48.8 MB | 403.98 | +6.67% | 425.26 ms |
| 32 | 97.7 MB | 403.23 | +6.47% | 454.51 ms |
| unset (new default) | 48.8 MB | 398.42 | **+5.20%** | 446.44 ms |
| unset, rerun 10 h later | 48.8 MB | 401.02 | +5.89% | 438.32 ms |

The new default is `clamp(48 MiB // chunk_bytes, 2, 64)`, applied as a
**floor-raiser**: the reference geometry still resolves to 2 chunks, unchanged, and
no geometry gets a smaller buffer than it has today. An explicit
`OFFLOAD_GPU_STAGING_CHUNKS` still wins, and `OFFLOAD_GPU_STAGING_MAX_BYTES` still
caps the result. The count is clamped **above** because `_ensure_staging_buffer`
allocates at full size, so a very small chunk must not turn a byte target into an
unbounded count; 64 is 4× past the measured plateau. The three large forced arms
span 0.19%, below the 1.4% run-to-run floor, which puts 48 MiB mid-plateau rather
than on its edge; **+5.20%** is the pessimistic member of the three runs that
resolve to the same byte-identical buffer. External hit rate is 81.0–81.2% in every
arm above 2 (79.7% at 2), so this is not "reads less, therefore faster".

## 4. What this branch deliberately does *not* ship

py-spy on a live, unprofiled server put `Event.synchronize` at **46.25%** of the
save thread and **56.12%** of the load thread, all of it in the staging ring's slot
wait. Our answer was to deepen the ring (353.77 → 383.15 → 393.50 tok/s at 4 → 32 →
128 host slots). **Main's #2223 removes the wait instead of hiding it**:
`prepare_chunk_major_groups` uploads every group's pointers, sizes and block IDs
once into a single resident int64 tensor, and each launch is a slice of it — there
are no per-group pinned slots left to wrap on.

So the rebase drops four of our own optimisations, each as an **empty commit**
rather than a silent deletion, so the reasoning stays in the history next to the
code it used to be:

- `16c9dbb55` the host slot ring
- `59d6c60d1` unconditional async slice copies
- `1e7d4f282` the staging plan cache
- `f6e45fea0` the plan-key fast path that served it

`fea451b2b` additionally removes the transfer-timing instrument
(`OFFLOAD_TRANSFER_PROFILE`, `_TransferProfile`, `_TimedStage`, and six PROF
fields, −242 lines). It was diagnostic scaffolding used to produce the numbers
above, not something main should carry; the commit that introduced it stays in this
branch's history for anyone who needs to re-run that measurement.

## 5. Measured — 600 s aiperf, gfx950 ×8, TP=4, 64 prefixes, seed 71502

Adjacent arms on the rebased base, back to back on the same box, both at this
branch's default settings:

| metric | off | on | Δ |
|---|---|---|---|
| output throughput tok/s | 257.11 | **401.02** | **+55.97%** |
| TTFT avg ms | 4,275.45 | 1,309.70 | −69.4% |
| TTFT p50 ms | 4,286.06 | 438.32 | −89.8% |
| ITL avg ms | 22.79 | 17.39 | −23.7% |
| requests completed | 304 | 472 | +55.3% |

Zero errors, `requests_success == 304 / 472`. The noise floor for this grid is
**1.4%**, measured rather than borrowed: three runs whose staging geometry resolved
byte-identically span 403.98 / 401.02 / 398.42 tok/s. (The 0.11–0.145%
adjacent-repeat floor quoted elsewhere in this repo is the wrong ruler here.)

It is not a luckier cache — the HBM-cached prompt fraction is 17.9% (off) vs 17.4%
(on). What moved is where prompt tokens came from:

| prompt token source | off | on |
|---|---|---|
| `local_compute` (re-prefill) | 8,179,299 (**82.1%**) | 2,455,804 (**15.9%**) |
| `local_cache_hit` (HBM) | 1,782,464 (17.9%) | 2,684,032 (17.4%) |
| `external_kv_transfer` | 0 | 10,327,104 (**66.8%**) |
| total prompt tokens | 9,961,763 | 15,466,940 |

A **3.33× cut in absolute prefill work** while serving 55% more requests; external
hit ratio 0.808. Both arms' three sources sum exactly to their
`prefix_cache_queries_total`, which is the check that the server was drained when
the counters were read — admission-time and completion-time counters are recorded at
different points and the identity only closes with nothing in flight.

~97% of the remaining per-chunk cost is **host-side, not transfer**: a 2.91 MiB
chunk takes ~2.1 ms through the tier against ~0.06 ms on the raw link, identically
in both directions and across LMCache versions (0.5.5rc4 measured **−5.72%** against
the 0.4.5 this ships with). The version is not the lever, and neither are the bytes
— GLM-5.2's KV is already the fp8 MLA latent.

## 6. Correctness — two-pass restore check

`tools/kv_offload_twopass_check.py`: pass 1 primes, a flood evicts HBM, pass 2 reads
back. A single pass only ever SAVES and proves nothing about restore. Measured on a
drained server:

| | pass-2 prompt tokens | pass-2 cached | marker recall |
|---|---|---|---|
| off (noise floor) | 160,200 | **0** | 8/8 |
| on | 160,200 | **159,936 (99.84%)** | 8/8 |

The off arm returning 0 cached tokens is what makes the on arm's 99.84%
meaningful. Server side on the on arm: 0 alloc failures, 0 KV load failures, 0
zero-retrieves, 0 block-table errors. The tool's `--max-tokens` defaults to 256, not
a couple of dozen: a reasoning model emits a chain of thought before the answer, and
a small budget truncates the marker — which scores as a restore failure when it is
really a truncation.

## 7. Two defaults that are wrong out of the box

- **`kv_load_failure_policy` defaults to `"fail"`.** A chunk evicted between the
  scheduler-side lookup and the worker-side retrieve becomes a **500 for the user**
  instead of a re-prefill. One 600 s run lost **10.09% of its requests** (47 of 466)
  this way. The documented launch line sets `"recompute"`.
- **`LMCACHE_MAX_LOCAL_CPU_SIZE` must cover the whole run, per TP rank** — not the
  prefix pool. At 40 GiB/rank the tier was exhausted ~150 s into a 600 s run and
  logged 27,695 allocation failures. The recipe works the number out (180 GiB) and
  explains the eviction race that follows.

`PYTHONHASHSEED=0` is also required: the TP ranks must agree on chunk hashes.

## 8. Known gaps

- **The 16-prefix working point has not been re-measured since the livelock fix.**
  Its old −12.31% comes from the same broken build as the −48.41%; treat it as
  unknown, not as a result. The recipe says so in place.
- The staging byte target is swept at one working point on one model. 48 MiB sits
  mid-plateau there; the plateau's lower edge (at or below 11 chunks) is not
  resolved, and no second KV geometry has been swept.
- The pack-grid warp sweep's raw output was not retained, and it was run on the M3
  segment geometry rather than GLM-5.2's. The mechanism is geometry-independent
  (bytes-per-lane is `1024 / (64 × num_warps)`, and `num_warps` does not enter the
  tile count) but the plateau was not re-measured on this model.
- A 255-chunk store costs 5.000 ms/chunk against 1.749 ms at 257 chunks (~15.3% of
  store time). 16,320 vs 16,448 tokens straddles `--max-num-batched-tokens 16384`;
  unexplained.

## 9. Docs and tests

- `recipes/atom_vllm/GLM-5.2-LMCache-Byte-Offload.md` (+345) — launch line, how to
  verify the tier is actually on (**read the counters on a drained server**), sizing,
  the measured result, the corrected analysis, and the gotchas above.
- `atom/kv_transfer/offload/README.md` (+53) — the new staging default and the
  tunables.
- **411 tests across the 11 test files this branch touches, all passing**: layout,
  seq view, config, block growth, completions, defer-release, sparse indexer
  chunking, dense codec, staging tile tables, staging buffer sizing. 361 are inside
  the CI unit job's collection set; the other 50 live in `tests/plugin/`, which
  `.github/scripts/run_unit_tests.sh` excludes, so they were run separately.
- `black` and `ruff` clean on every file this branch touches. The 4 remaining `ruff`
  findings in `atom/plugin/vllm/register.py` were verified pre-existing by running
  `ruff` at the merge-base in a throwaway worktree — same 4 rules, same code, shifted
  line numbers. The 34 `tests/plugin/` sglang/rtpllm failures on this branch
  reproduce **identically** at the merge-base.


atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -618,6 +657,47 @@ def _absorb_worker_meta(self, meta) -> None:
...
+        report clear a newer save generation.
+        """
+        for completion in completions:
+            key = completion.key
+            report = self._completion_reports.get(key)
+            if report is None:
+                report = self._completion_reports[key] = [0, True]
+            report[0] += 1
+            report[1] = report[1] and bool(completion.succeeded)
+            if report[0] < self._world_size:
```
yhl-amd, line 796: [P1] Please bound this quorum wait. If one rank's completion report is lost, this entry never reaches `world_size`, so `should_defer_free()` remains true forever. `request_finished()` has already told vLLM to retain the request's GPU blocks, and `_collect_releases()` has no timeout/reclaim path, so no `finished_sending` is produced and the referenced blocks can exhaust the pool and stall the server at zero running requests. The native offload path already has a stale-lease timeout; the plugin path should use the scheduler's save-abandon timeout and release the stalled save through the normal deferred-free flow.

atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -422,6 +454,12 @@ def build_connector_meta(self, scheduler_output) -> KVConnectorMetadata:
         """
         preempted = self._handle_preempted(scheduler_output)
         block_size = int(self._config.kv_cache_block_size)
+        for req_id, new_blocks, replaces in _scheduled_block_growth(scheduler_output):
+            seq = self._seqs.get(req_id)
+            if seq is None:
+                continue
+            grown = _block_ids(new_blocks)
+            seq.set_block_table(grown if replaces else seq.block_table + grown)
```
yhl-amd, line 483: [P2] This coverage calculation uses the physical cache block size, but under DCP each scheduler block ID covers `kv_cache_block_size * decode_context_parallel_size` tokens. The ATOM scheduler already records that as `virtual_block_size`. With block size 64, DCP=4 and four block IDs, the table covers 1024 tokens, but this caps `num_cached_tokens` at 256 and emits only a `[0, 256)` save. DCP therefore silently stores about `1 / DCP` of the available prefix. Please multiply by the virtual block size, e.g. `self._scheduler.virtual_block_size`.

atom/kv_transfer/offload/_offload_common.py
```diff
@@ -112,26 +141,48 @@ def _init_worker_common(
...
             raise ValueError("offload save worker count must be positive")
+        # A single load thread saturates once the HBM pool is small enough for
+        # the CPU tier to serve real traffic: measured 143s of `retrieve` inside
+        # a 163s window (88% duty cycle) on the radix workload at 7900 blocks,
+        # which turned a +57.8pp hit-rate win into a 4% throughput loss. The
+        # byte-copy path keeps its staging buffers and CUDA streams in
+        # thread-local state (`_BlockGpuConnector._thread_state`), so extra load
+        # threads are independent; each costs one more
+        # `gpu_staging_buffer_bytes` allocation per rank.
+        n_load = (
```
yhl-amd, line 169: [P1] Applying this load-pool width to every offload connector is unsafe for DSV4. Its SLOT path reserves one `_SlotLoadBatchReservation`/staging row for a batch and relies on the load executor being serial. With `OFFLOAD_LOAD_WORKERS=2`, two `_restore_slot_payload()` calls can write the same staging row concurrently and then scatter it to different requests; both operations report success, so recurrent state can be silently corrupted. A CPU stub reproduces two concurrent loads using the same staging ID. Please force `load_workers=1` for DSV4 or allocate a separate staging row per concurrent SLOT load.

yhl-amd (commented):
I found five actionable issues in the offload changes. The inline comments include two paths that can stall or corrupt serving state, one DCP coverage bug, and two observability/recipe issues.

atom/kv_transfer/offload/_offload_common.py
```diff
@@ -583,6 +638,59 @@ def _chunk_floor(self, tokens: int) -> int:
...
+        is how a single failure becomes a permanent one: `load_failed` clears
+        the pending load and the lookup memo, so the next scheduler pass hits,
+        parks the request in WAITING_FOR_REMOTE_KVS again, and fails again --
+        forever, holding the request's KV blocks and its concurrency slot the
+        whole time. Prefilling normally instead is exactly what would have
+        happened with no external tier at all.
+        """
+
+        if self._load_failed_seqs.get(sid) is seq:
+            self.total_suppressed_load_retries += 1
```
yhl-amd, line 678: [P3] This counter is never exposed by `get_statistics()`. The production statistics reader only consumes that returned dictionary, so `total_suppressed_load_retries` remains invisible outside tests even though the PR/recipe relies on it to make the former livelock observable. Please add a `suppressed_load_retries` entry to `get_statistics()` and cover it in the statistics test.

tools/kv_offload_twopass_check.py
```diff
@@ -0,0 +1,162 @@
...
+        "cached_tokens_pass2": ext2,
+        "pass1": p1,
+        "pass2": p2,
+    }
+    print(
+        json.dumps(
+            {k: v for k, v in summary.items() if k not in ("pass1", "pass2")}, indent=2
+        )
+    )
+    out = os.path.join(args.out_dir, f"twopass_{args.label}.json")
```
yhl-amd, line 159: [P3] The documented command passes `--out-dir results/`, but neither the recipe nor this tool creates that directory. On a clean checkout the requests finish and then this `open()` raises `FileNotFoundError`, losing the result file. Please create `args.out_dir` with `os.makedirs(..., exist_ok=True)` before opening the output.

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Valid, fixed in 605a48e73.

Confirmed the failure mode: `_completion_reports[key][0]` only ever increments from `_apply_completions`, and `should_defer_free` stays true until the quorum lands, so one lost report pins that request's blocks in vLLM and its SeqView here for the life of the server, while `has_pending_push_work` keeps the engine stepping over a request that can never finish.

Added `_reconcile_stale_saves`, which mirrors the *request* half of the native engine's reconcile (`atom/model_engine/scheduler.py`) on the same 5 s cadence and the same window — `offload_save_abandon_timeout_s()`, i.e. LMCache's own `LMCACHE_EC_PIN_TIMEOUT_SEC` plus a margin. That derivation is what makes abandoning safe rather than merely convenient: `_guard` reports on both the return and the raise path, so past the window a missing report means the save is parked inside LMCache and LMCache has already force-unpinned its source. A non-positive pin timeout disables reclamation, matching the native path. It runs before `_collect_releases`, so a save abandoned on a step is released on that same step.

One deliberate deviation from the native reconcile: only the request half is mirrored. I first wrote this around `reclaim_stale_leases` and it would have been a no-op — that half reclaims *block ids*, which the native engine hands to `BlockManager.free_leased_blocks`, and a vLLM connector has no such channel (it can only name request ids in `finished_sending`). Consistently, nothing under `atom/plugin/` ever calls `activate_block_leases`, so no lease exists to reclaim. `abandon_save(req_id)` is keyed by the unit vLLM actually frees and clears exactly the state `should_defer_free` reads.

Tests: `test_a_quorum_that_never_completes_is_abandoned`, `test_the_reconcile_is_off_when_lmcache_pinning_is_disabled`, `test_build_connector_meta_runs_the_reconcile_before_releasing` in `tests/plugin/test_vllm_offload_defer_release.py`. ⚠️ `.github/scripts/run_unit_tests.sh` ignores `tests/plugin/`, so CI green does **not** cover this — I ran them manually under Python 3.12 with vLLM installed (54 passed).

atom/plugin/vllm/kv_transfer/connector.py
PerryZhang01: Valid, fixed in 605a48e73.

`self._scheduler.virtual_block_size` is `kv_cache_block_size * decode_context_parallel_size` (`chunked_scheduler.py:75`), and it is the unit that same block table is indexed in at `chunked_scheduler.py:517-518` and `701-702` — so this cap was the one place pricing it physically. Under DCP>1 it under-reported coverage by exactly the DCP factor, clamping away the tail of every prompt. Identity when DCP=1, so no behaviour change on the default deployment.

Test: `test_the_frontier_is_capped_in_virtual_blocks_under_dcp` in `tests/plugin/test_vllm_offload_block_growth.py` (8 block ids at dcp=2 must cover 1024 tokens); the `_adapter` factory is now parameterized by `dcp`. Same CI caveat as the other plugin fix — `tests/plugin/` is not in `run_unit_tests.sh`'s collection set, so this was verified by running it manually.

atom/kv_transfer/offload/_offload_common.py
PerryZhang01: Valid, fixed in 605a48e73 — and worth flagging that this is a regression **this PR introduced**: before it, `_offload_common.py` hardcoded `max_workers=1` for the load pool, so DSV4 was safe by construction. Making the width configurable in one shared helper silently handed DSV4 a knob its SLOT path cannot survive.

The hazard is as described: `start_load_kv` gives every SLOT load in a worker batch the *same* `_SlotLoadBatchReservation` — one staging row, shared — which is only sound because the executor runs them in submission order. A second load thread runs two loads through that row concurrently and corrupts both.

Fixed at the DSV4 connector rather than in the shared helper, since the knob is legitimate on the dense path (each load thread owns its own thread-local staging state): `DSV4OffloadConnector.__init__` now passes `load_workers=1` unconditionally and logs a warning naming the ignored value. Silently dropping it would read as "measured, no effect" to anyone tuning it.

Test: `test_slot_load_executor_stays_serial_regardless_of_env` in `tests/test_lmcache_offload_v4_page_slot.py` — asserts `load_workers == 1` and `_load_executor._max_workers == 1` both by default and with `OFFLOAD_LOAD_WORKERS=4`, plus that the warning names the ignored value. This one *is* in CI's collection set.

atom/kv_transfer/offload/_offload_common.py
PerryZhang01: Valid, fixed in 605a48e73. `get_statistics()` now returns `suppressed_load_retries`. The counter is initialized in the same mixin (`_offload_common.py:485`) and incremented at `:677`, so it is always defined on that dict.

Worth stating why it matters beyond tidiness: a suppressed retry is a *silent* hit-rate loss — the request still serves correctly, out of HBM, having skipped the tier — so there is no error, no latency spike and no log line. The statistics dict is the only channel where it is visible in a running server.

Tests (both in CI's collection set): the exact-dict assertion in `test_scheduler_offload_statistics_are_cumulative`, and a new assertion in `test_dsv4_failed_load_is_not_retried_for_the_same_request` checking the value reaches the reader, not just the attribute.

tools/kv_offload_twopass_check.py
PerryZhang01: Valid, fixed in 605a48e73 — `os.makedirs(args.out_dir, exist_ok=True)` right after `parse_args()`.

Placed up front rather than next to the `open()` on purpose: the failure this avoids is not the exception itself but its timing. At the documented settings both passes plus the eviction flood run first, so a typo'd `--out-dir` costs the whole run before it reports. Now it fails in 0 s.

Verified end-to-end with `--n 0 --flood 0` against a fresh directory, and against the mutant (fix reverted) to confirm the mutant reproduces `FileNotFoundError` by message, not just by exit code.

valarLip (approved):

PR #2233, opened 2026-09-14
fix(runtime): capture CUDA graphs in inference mode

yhl-amd:
## Summary

- run `ModelRunner.capture_cudagraph()` under `torch.inference_mode()`, matching model warmup and serving forwards
- prevent CUDA graph capture from mutating persistent inference tensors outside InferenceMode
- add a non-GPU source-level regression test for the capture entry-point contract

## Motivation

`ModelRunner.forward()` runs under InferenceMode, so lazy inference backends may create persistent buffers as inference tensors during warmup. `capture_cudagraph()` did not restore that mode before invoking the model.

This can fail when graph capture later reuses and updates one of those buffers. One observed case is the AITER TP communication-fused MoE path with DSpark q7 padding:

```
RuntimeError: Inplace update to inference tensor outside InferenceMode is not allowed
```

The failure surfaced at `aiter/ops/comm_fused_moe_runtime.py` while padding `runner.output`, but the mode mismatch originates at ATOM's inference-only graph-capture entry point. Keeping capture under InferenceMode also makes its execution contract consistent with normal serving.

Related downstream context: https://github.com/SemiAnalysisAI/InferenceX/pull/2912

## Tests

- `pytest -q tests/test_model_runner_staging_fence.py tests/test_forward_mode.py` (26 passed)
- `ruff check atom/model_engine/model_runner.py tests/test_model_runner_staging_fence.py`
- `git diff --check`


yhl-amd:
Local MI355X validation passed against the original failing stack:

- image: `rocm/atom-dev:nightly_202609121454` (`sha256:972e37f6e0f7eb12de83188b07709ed0578a2eb4aebc20fbdcfcec562ca8d60c`)
- ATOM base: `c61fbfb0e5fabd478978ba04801fb80028240176`, with only this PR's `capture_cudagraph()` decorator applied
- AITER: unmodified `4676838b5607542585a9a04d19f61b6055d341fa`
- AITER runtime SHA256: `131a3e0be4c39b2d47474c8a06572e83adce1806fc0d4201ddeaef13f6019f49` (the pre-hotfix source)
- config: TP8, DP1, DSpark K6/q7, FULL graph, `max_num_seqs=32`

All q7 graphs captured successfully for batch sizes `32, 16, 8, 4, 2, 1`, including the `bs=2` padded bucket where the original run failed. The server reached ready state with no `InferenceMode` error.

A two-turn concurrency-16 smoke then completed `32/32` correct requests with zero errors. DSpark stats after the smoke reported 1,482 draft tokens, 1,333 accepted tokens, and average 6.397 tokens/forward.

valarLip (approved):

PR #2242, opened 2026-09-15
feat: integrate AITER FlyDSL GDN for Qwen3.8-Flash-Next-FP8 standalone

junna2016:
## Motivation

Accelerate Qwen3.8-Flash-Next-FP8 standalone inference on MI308 by integrating
AITER FlyDSL GDN kernels for both prefill and decode, while preserving the
original ATOM Triton implementation as a fallback and explicit rollback option.

Qwen3.8-Flash-Next is implemented by the `Qwen4Exp` backend in ATOM;
`Qwen4Exp` below refers to that internal implementation, not a different model.

## Technical Details

- Prefill uses FlyDSL fused K1–K4 and FlyDSL K5, followed by ATOM Triton K6
  specialized for head-major VK inputs. Q/K normalization and gating remain
  consistent with the original pipeline.
- Decode uses AITER's fused normalization, gating and recurrent-update kernel.
  State storage is physically VK but exposed as a logical KV view, eliminating
  per-token state packing and pool-wide transposes. Triton fallback and
  checkpoint stores honor the view's inner strides.
- Preserve the baseline dtype policy: FP32 K5 accumulation/raw final state,
  BF16 snapshots with round-to-nearest conversion, and configurable persistent
  state dtype.
- Build prefill scheduling metadata once per scheduler step from host lengths,
  shared across layers, without per-layer GPU-to-CPU length synchronization.
- Preserve deterministic zero output for padding: use the kernel's advertised
  `zeroes_invalid_output` capability, otherwise allocate zeroed output.
- Enable this integration for Qwen3.8-Flash-Next through ATOM's `Qwen4Exp`
  backend only. Unsupported inputs
  retain Triton; expected AITER import/initialization failures warn and fall back.
  Kernel execution errors are not silently retried after possible state updates.
- Decode support validated here: gfx942, BF16 activations, K=V=128, BF16/FP32
  persistent state, and one token per request with multiple requests per batch.

Both stages default to automatic dispatch. To restore the original baseline,
set the following variables **before starting the service**:

```bash
export ATOM_GDN_PREFILL_BACKEND=triton
export ATOM_GDN_DECODE_BACKEND=triton
```

Use `auto` to re-enable either stage independently. No AITER source patch or
wheel replacement was required in the tested environment.

## Test Plan

Test environment: MI308, AITER commit
`c213dea9163254a8e1085c73e40771712fb583db`, FlyDSL 0.3.2, Triton 3.7.

From the ATOM source directory, with writable AITER/FlyDSL caches configured:

```bash
HIP_VISIBLE_DEVICES=0 python3 -m pytest \
  tests/test_aiter_flydsl_gdn.py tests/test_gdn_midstep_state_gpu.py -q
```

Coverage includes:

- BF16/FP32 persistent state; output, final-state and intermediate-snapshot
  comparisons against the original ATOM Triton implementation.
- Prefill lengths 64/129/1024/8192 and ragged batches [3808,4352]/[7648,512].
- Decode batch sizes 1/3/4/16/64/128, production strided gate views and separate
  state read/write slots.
- Graph replay with changing live/padded rows and exact zero padding outputs.
- Checkpoint state copies, Triton fallback on VK-backed views, and the actual
  Qwen3.8-Flash-Next cache-builder entry point (`Qwen4ExpMetadataBuilder`).
- Model-level fixed 2K/8K/12K prompts, greedy generation with normal EOS.
- Separate unprofiled performance runs and mixed prefill/decode traces.

## Test Result

**32 tests passed in 50.80 seconds.** Ruff lint and diff whitespace checks pass.

Observed operator output errors relative to ATOM Triton: prefill maximum absolute
error approximately 0.000977 and relative RMS approximately 0.38%; decode maximum
absolute error approximately 0.000244. Results are numerically close, not claimed
to be bit-identical.

Model checks: all **6/6 final answers** were correct and matched baseline;
**2/6 complete outputs** matched exactly. Reasoning text differed in the remaining
cases. These focused checks do not establish broad benchmark-quality equivalence.

### Operator performance

Per-rank Qwen3.8-Flash-Next TP=2 shape: Q/K heads=8, V heads=24, K=V=128; BF16 activations
and persistent state, BF16 snapshots. Prefill Q/K=[1,T,8,128], V=[1,T,24,128].
Decode AITER layout is [batch,1,heads,128]; logical state is
[slots,24,128,128]. Each result is the median of three measurements.

Decode uses Graph timing for **both** implementations; the ATOM baseline includes
gating and recurrent kernels. Prefill includes normalization and K1–K6, with
gating outside both timed regions.

| Workload | ATOM Triton (µs) | FlyDSL (µs) | Latency reduction |
|---|---:|---:|---:|
| Decode B=1 | 9.511 | 4.670 | 50.90% |
| Decode B=3 | 10.943 | 7.178 | 34.41% |
| Decode B=4 | 12.769 | 9.204 | 27.92% |
| Decode B=16 | 26.204 | 16.998 | 35.13% |
| Decode B=64 | 68.239 | 51.674 | 24.28% |
| Prefill T=1024 | 331.334 | 209.365 | 36.81% |
| Prefill T=8192 | 2066.550 | 1318.123 | 36.22% |
| Prefill lengths=[3808,4352] | 2064.407 | 1344.500 | 34.87% |

### End-to-end performance

Qwen3.8-Flash-Next-FP8; same two MI308 GPUs; TP=2, EP=2, DP=1;
input=12000, output=350, concurrency=3; chunked-prefill budget=8192;
BF16 SSM/KV cache; Graph enabled; prefix cache disabled; no MTP.
Stop the previous service before switching modes. Each mode uses four warmup
requests and 30 measured requests. Profiler timings are excluded.

| Metric | ATOM Triton baseline | FlyDSL | Improvement |
|---|---:|---:|---:|
| QPS | 0.281406 | 0.289038 | +2.71% |
| Mean TTFT | 3906.57 ms | 3817.72 ms | 2.27% lower |
| Mean TPOT | 19.3500 ms | 18.7983 ms | 2.85% lower |
| P99 TPOT | 23.8734 ms | 23.2161 ms | 2.75% lower |

These are single paired-run observations, not multi-run confidence intervals or
a global maximum-QPS claim. Mean TPOT is below 21 ms; P99 TPOT is not.

Both TP-rank traces confirm actual execution of `gdn_prepare_kernel`,
`chunk_gdn_fwd_h_flydsl_opt`, `chunk_fwd_kernel_o`, and
`gdr_decode_bf16_kh8x128_vh24x128_q1_4w8x8_vs1_0`; the original recurrent decode
kernel is absent from the captured optimized workload.

## Submission Checklist

- [x] Added focused numerical and Graph replay tests.
- [x] Preserved explicit Triton rollback and unsupported-input fallback.
- [x] Documented state layout, dtype policy and supported scope.
- [x] Measured operator and unprofiled model performance against ATOM Triton.
- [ ] Run the repository's full CI suite on the target branch.


HaonanWang98:
A few issues to address:
- ReplaySSM correctness: The cache is transposed to VK storage even when ReplaySSM is enabled, but its kernels still assume KV layout. This can silently produce incorrect results. Please exclude ReplaySSM from the layout switch or adapt its state reads/writes.
- Backend selection: The model constructor, cache builder, and forward path make separate backend decisions with different conditions. These need to be consistent, especially around state layout.
- Naming: Please rename aiter_flydsl.py to something GDN-specific, such as gdn_flydsl.py. The current name suggests a general AITER/FlyDSL integration layer.
- Test coverage: Decode only supports gfx942, but the tests run on any ROCm GPU and assert support. Please add architecture guards and test the unsupported-device fallback.

junna2016:
> A few issues to address:
> 
> * ReplaySSM correctness: The cache is transposed to VK storage even when ReplaySSM is enabled, but its kernels still assume KV layout. This can silently produce incorrect results. Please exclude ReplaySSM from the layout switch or adapt its state reads/writes.
> * Backend selection: The model constructor, cache builder, and forward path make separate backend decisions with different conditions. These need to be consistent, especially around state layout.
> * Naming: Please rename aiter_flydsl.py to something GDN-specific, such as gdn_flydsl.py. The current name suggests a general AITER/FlyDSL integration layer.
> * Test coverage: Decode only supports gfx942, but the tests run on any ROCm GPU and assert support. Please add architecture guards and test the unsupported-device fallback.

Thanks for the review. All four points have been addressed in the latest commit:
- ReplaySSM: Preserve the original contiguous KV state layout and existing execution paths when ReplaySSM is enabled.
- Backend selection: Resolve a shared policy during cache binding. Cache layout and forward dispatch now use the same policy, with additional per-call input checks.
- Naming: Renamed aiter_flydsl.py to gdn_flydsl.py and updated imports.
- Tests: Added gfx942 guards for FlyDSL decode numerical tests, unsupported-architecture fallback coverage, and cache-builder-to-forward tests covering ReplaySSM, explicit Triton selection, and FlyDSL dispatch.
Validation in the MI308 development container: 86 tests passed, including numerical comparisons, CUDA Graph replay, state relocation, and dispatch regression tests. Black and Ruff checks also passed. Unsupported-architecture behavior was tested through mocking; no gfx950 hardware validation is claimed.

PR #2243, opened 2026-09-15
Size the staging pack grid by the bytes it moves, not the widest segment (kernel 1.87x, end-to-end not measurable)

PerryZhang01:
## What

The dense chunk-major staging kernels launched on a rectangular grid,
`(num_chunks × num_segments, ceil(max_tile_bytes / 1024))`. A job is one
`(chunk, segment)` pair and owns `count × segment_bytes` bytes, but every job got
the **widest** job's tile count and masked the surplus off.

The dense offload worker registers K and V per layer alongside their scales, so
the segment list spans 16384 B down to 512 B — a factor of 32. On the geometry it
logs (296 segments, 8 chunks × 1 block, 23.0 MiB) that grid is **37,888 programs
for 23,968 tiles of real work**: 37% of the launch moved nothing.

This PR:

1. Replaces the grid with a flat 1-D grid of exactly the tiles being staged, plus
   a memoised `_tile_table` mapping each program id to its `(job, tile-within-job)`.
2. Drops `num_warps` from 8 to 2. A program moves `BLOCK_BYTES` of `uint8`, so the
   warp count only chooses bytes per lane; nothing here reduces across lanes, so
   this is semantically inert.
3. Exposes `DenseKVByteCodec.segment_block_bytes` and logs its length and extremes
   in the worker's startup line.

## Numbers

Kernel **1.87×** on the geometry the server actually runs. **End-to-end is not
measurable** — see the caveat below, which is the more important half of this PR.

One MI300X, packing 296 segments / 23.0 MiB, against `origin/main`:

| | ms | GB/s |
|---|---|---|
| rectangular grid, 8 warps (`origin/main`) | 0.0498 | 483.0 |
| rectangular grid, 2 warps | 0.0346 | 696.0 |
| flat grid, 8 warps | 0.0452 | 532.1 |
| **flat grid, 2 warps (this PR)** | **0.0267** | **902.2** |

The two factors are not independent: the grid is worth 1.10× at eight warps and
1.30× at two; the warp count is worth 1.44× on the rectangular grid and 1.70× on
the flat one.

Other shapes, same harness, all bit-identical to main's output:

| geometry | main | this PR | |
|---|---|---|---|
| 296 segments × 4 blocks, 91.8 MiB | 0.1908 ms | 0.0910 ms | **2.10×** |
| 64 × 16 KiB uniform, 8.0 MiB | 0.0208 ms | 0.0220 ms | 0.94× |
| 64 × 512 B uniform, 0.2 MiB | 0.0209 ms | 0.0224 ms | 0.93× |
| Kimi-K3 4 segments, 0.1 MiB | 0.0205 ms | 0.0220 ms | 0.93× |
| Kimi-K3 64 segments, 0.1 MiB | 0.0208 ms | 0.0218 ms | 0.95× |

Shapes small enough to be launch-bound sit at ~21 µs either way and pay ~1.5 µs
for the table load. The win scales with how uneven the segment list is and how
many bytes are in flight; **uniform segment lists get nothing**, which is the
honest answer for the hybrid Kimi-K3 staging paths that share these kernels.

## The caveat

Take this for the geometry independence, not for throughput. In the server the
pack kernel is 2–4% of the 1.78 ms a group spends in
`gpu_to_chunk_major_device_buffer` — the rest is host launch gap — and pack is
1.68% of wall clock. The whole ceiling here is **~0.26%**, below the 0.109%
run-to-run noise measured on that configuration. This is a kernel change with a
kernel-sized result.

## The table's cost, which is the real design question

Rebuilt per launch the table would have been a net regression: 0.24 ms against a
0.027 ms kernel. It is a pure function of the block counts and the segment sizes,
and both repeat across transfers, so tables are memoised (LRU, 16 entries, 374 KiB
for the geometry above). A hit is 0.0024 ms, 9% of the kernel; a miss costs one
extra H2D upload, both columns travelling as one `2 × tiles` tensor so the plan's
single-upload accounting only gains one per new geometry, none on repeats.

Keeping the hit cheap is why `_tile_table` requires already-normalized ints —
coercing 296 segment sizes on the way in measured 11 µs, four times the lookup.

## Why the logging is in the same PR

`max_segment_bytes`, not `bytes_per_block`, set the cost of this kernel, and it
was not logged. Twice a benchmark reconstructed the segment list from the model
config instead of reading it: once one filler segment 30× too wide (which reported
a 41× speedup), once fusing K and V into a single 32 KiB segment, missing that
`layout=kv-split` registers them separately. Both were wrong in the direction that
flatters the change.

## Testing

- Bit-identical to `origin/main` on all seven geometries above; on 40 randomized
  ragged plans (chunks of differing block counts, zero-block chunks, multiple
  groups per plan); on a pack→unpack round trip; and through the legacy
  non-prepared entry points that `hybrid/kimi_k3/staging.py` calls.
- CPU suite (`.github/scripts/run_unit_tests.sh` scope): same 18 failures and 4
  errors as `origin/main` on this host, +4 new tests passing.
- New tests: flat-grid shape and `num_warps` in the dense contract test, tile
  table sizing per job, zero-byte jobs getting no programs, cache bound, and the
  codec's segment geometry.
- `black . && ruff check` clean on every touched file.


JiaoliangYu (commented):
Should the global tile-table LRU be protected by a lock?
Dense offload can prepare groups concurrently from the separate load and save ThreadPoolExecutors. Although the GIL protects individual OrderedDict operations, the get() → move_to_end() and insert → eviction sequences are not atomic. For example, a cached key could be evicted by another thread between get() and move_to_end(), resulting in a KeyError; concurrent misses could also build and upload the same table more than once. Would it make sense to guard the cache bookkeeping with a small lock, while keeping the NumPy construction and H2D upload outside the critical section and double-checking the cache before insertion?

Could we avoid materializing tile_pos for every tile?
Tiles belonging to the same job are contiguous in the flattened grid, so tile_pos can be derived as pid - job_start[job]. We could store tile_job[T] and the much smaller job_start[J], then compute:
pid = tl.program_id(0)
job = tl.load(tile_job + pid)
tile = pid - tl.load(job_start + job)
Both arrays could likely use int32, with explicit bounds checks before conversion. For the geometry in the PR (T=23,968 tiles and J=2,368 jobs), this would reduce the table from roughly 374 KiB (2 × T × int64) to 103 KiB ((T + J) × int32), while also avoiding construction of the full pos_in_job array.

PR #2250, opened 2026-09-16
feat(offload): support native DSv4 checkpoints with LMCache MP

yhl-amd:
Native DSv4 requests on ATOM need both PAGE KV and a matching compact checkpoint to reuse an external prefix. This adds DSv4 support to the existing `lmcache_mp` connector, saving and restoring PAGE/STATE through a standalone LMCache server while running requests keep their fixed Active SLOTs.

The connector pins existing immutable READY checkpoint PAGE units after admission. It registers their strided native image views directly and releases each lease only after the matching operation completes across TP ranks. Restore uses fresh PAGE units, then the native image-to-SLOT codec; requests wake only after local restore completes. Save admission is round-robin, bounded by count and pinned bytes, and serialized per request.

Sparse STATE lookup requires a complete checkpoint at the same boundary as PAGE KV. Full-prompt lookup truncates tokens before querying. Per-group null markers (PAGE 0 is valid; STATE -1 is absent) make matching LMCache versions automatically separate PAGE and STATE with the default server configuration; no `--separate-object-groups` launch flag or capability RPC is required. Native layout/model revision namespaces and exact generation tracking reject incompatible or unsafe transfers. Uncertain DMA completion retains its lease.

### Dependencies and review order

- Stacked on #2246 (`fix/dsv4-save-admission`, `a8c377a5`), which provides bounded save admission. This PR's diff contains the MP integration. Rebase and retarget to `main` after #2246 merges.
- Requires the matching LMCache protocol PR: https://github.com/LMCache/LMCache/pull/5132.
- Checked that this branch merges cleanly with ATOM `main` at `5c947cbf`.
- Start with `atom/kv_transfer/offload/mp/README.md` for launch configuration, lifetime rules, and supported scope; the main implementation is in `dsv4_scheduler.py`, `dsv4_worker.py`, and `dsv4_layout.py`.

### Validation

- Earlier ATOM validation: **686 passed, 1 skipped** across checkpoint/lease, MP layout/worker/scheduler, existing offload, TP completion aggregation, and MultiConnector regressions. The skip is the existing scheduler test module that still imports the old `kv_transfer_engine` path; new real Scheduler integration tests pass.
- Current focused validation: **58 ATOM CPU tests**, **135 matching LMCache CPU tests**, and **2 real standalone MP GPU round-trip tests** passed with the server's default object-group configuration on MI355X / ROCm 7.2 / PyTorch 2.14.
- Ruff formatting/checks and `git diff --check` pass on the changed files.

### Draft acceptance scope

Full DSv4-Pro output/logit equivalence, TP8 serving, and performance measurements remain outstanding. External restore currently requires an actual post-allocation HBM hit of zero. Finished requests retain ordinary PAGE/SLOT allocations while a save is pending. Local stream fences protect the native restore descriptor buffer; their cost needs model-level measurement. Equal-shaped native views with incompatible physical strides are rejected.


kvnloo:
While tracing the native MP path, could we add `--eviction-policy LRU` to the existing Running the MP server example?

In the pinned LMCache 05fc77a, `ServerCommand.add_arguments()` includes `add_storage_manager_args()`, which requires `--eviction-policy`; the command currently omits it. This is a source-level finding, I haven’t run the full CLI or GPU path.

I also noticed the PR description still points to mp/README.md, while the guide is now in the parent offload README. Would a one-line recipe correction plus updating that pointer be the right scope? I’d keep the topology, checkpoint and lifetime design unchanged.


_AI-assisted source review and drafting; no GPU results claimed._

PR #2253, opened 2026-09-16
plugin/vllm: add Kimi-K3 (hybrid) LMCache KV offload

PerryZhang01:
> **Stacked on #2236** (`perzhang/vllm-0.29.0-upgrade`). Please review/merge that
> one first; this PR's base is that branch, not `main`. Review only the single
> commit `f06bb0528` — 7 files, +1928/-21.

## Why this cannot target `main`

The KDA (mamba) leg is fed by vLLM APIs that exist only in 0.29, which #2236
brings in and `main` (pinned to 0.28) does not have:

- `SchedulerOutput.kv_connector_block_state` — `.block_ids` (the authoritative
  per-group block tables) and `.boundary_state_offloads`
- `KVConnectorBase_V1.bind_gpu_block_pool()` — pinning a handed-off block across
  the asynchronous D2H
- `SupportsHMA` / `request_finished_all_groups()`

Verified absent in 0.28: `kv_connector_block_state` is not in
`git show 2cf0a6915:vllm/v1/core/sched/output.py`.

## What was broken

`AtomLMCacheOffloadConnector` could not serve Kimi-K3 at all. Two guards
rejected it at startup (`kv_cache_layout.py` block-count check;
`connector.py:_block_ids` multi-group check), and behind them sat a correctness
problem rather than a missing case: K3 is hybrid, so vLLM builds **two** KV
cache groups (MLA full attention + KDA recurrent state), and a restored MLA
prefix is only valid if the KDA state at the *same token boundary* is restored
with it. Half a restore does not crash and does not log — it is wrong output.

Separately, `VllmConfig` auto-disables the hybrid memory allocator for a
connector that does not declare `SupportsHMA`, and hybrid SSM models require it.
So before this change, K3 + `--kv-transfer-config` did not mis-save; it did not
boot.

## Design: two groups, two mechanisms, one gate on lookup

**MLA group** keeps the existing `DenseKVByteCodec` path untouched.

**KDA group** cannot be read positionally at all. Under `--mamba-cache-mode
align` a mamba block table is not append-only (a superseded state block is freed
and nulled; speculative blocks relocate in place), so indexing it by
`token // block_size` can persist a null, freed, or live speculative block under
a valid prefix hash. The only safe source is vLLM's explicit hand-off,
`boundary_state_offloads`, which names the exact block holding a committed
boundary state — and which is nulled as soon as `build_connector_meta` returns,
so it is consumed in the step it arrives. Each accepted boundary is pinned via
the bound GPU block pool and released on rank quorum.

**Correctness of the pair is enforced on lookup, not on save.** The reported
external hit is capped at the largest chunk boundary the state index still
claims, so a state that was never stored or was evicted *shortens* a prefix
instead of corrupting it. That removes the need for any cross-group save
parking (the native path's `_JointPark`). A KDA miss additionally invalidates
the attention blocks the dense leg filled, so vLLM cannot cache an MLA prefix
whose state never arrived.

Load is positionally safe, unlike save: vLLM allocates the boundary state block
for the external hit, so the page is written through the mamba group's own block
table.

## Requirements this adds for K3

- `--mamba-cache-mode align` is **mandatory** (asserted at construction), not
  just for DSpark.
- `"kv_load_failure_policy":"recompute"` must be set in `--kv-transfer-config` —
  the default `fail` turns an offload-tier miss into a user-visible 500.
- `LMCACHE_CHUNK_SIZE % mamba_block_size == 0`, so chunk ends land on boundaries.

Both are documented in the recipe section this PR adds.

## LMCache version

Stays on the pinned version. The 0.5.5 upgrade was permitted, not required, and
nothing here needs it; measured on the GLM-5.2 arc, 0.5.5rc4 lost to 0.4.5 by
**5.72%**, with ~97% of chunk cost host-side, so the version change does not buy
that back. If it is taken later it should be its own A/B, not a rider on this.

## Verified

- `tests/test_vllm_kda_state_offload.py` — **34 new tests, all passing**. Placed
  at the `tests/` root deliberately: `run_unit_tests.sh` passes
  `--ignore=tests/plugin`, so tests under `tests/plugin/` are not run by CI.
- The new tests were mutation-checked: ten real degradations (uncapped hit, each
  `collect_stores` filter removed, quorum of one, indexing despite a failed rank,
  `boundary_hash` off-by-one, dropping an unresolvable load, silently dropping an
  unmapped layer) each turn them red, while two equivalent mutations leave them
  green.
- All 45 non-plugin test files touching the changed modules, on this PR's base:
  **1464 passed, 13 skipped**.
- `black --check .` clean (807 files unchanged). `ruff check` clean on all 6
  changed source files. (A repo-wide `ruff check` reports 704 errors, but base
  `ac7dbb39d` reports the identical 704 — pre-existing, not from this change.)

## NOT verified — no hardware run yet

**Nothing here has been run on a GPU.** No TP8 window has been available: cards
0-3 are on a GLM-5.2 job with no ETA and 4-7 on another engineer's sweep. The
recipe section states this explicitly rather than implying measured numbers.

Still outstanding, and the intended acceptance criteria:

1. **Boot** — K3 TP8 with the LMCache `--kv-transfer-config`; confirm HMA stays
   on and no "Turning off hybrid kv cache manager" warning appears; verify with
   `rocm-smi --showmemuse` inside the container, not `curl /health`.
2. **Accuracy, two-pass** — a SAVE-only single pass measures nothing. Force
   read-back by sizing the HBM pool down, take the noise floor from the OFF arm's
   own two-pass delta, and check whether the delta scales with read-back volume.
   Baseline to beat, already measured on #2236's branch for K3 without LMCache:
   GSM8K 1319 / 5-shot / 64-concurrent, flex **0.9507** / strict **0.9500**.
3. **Hit rate** — short priming tail, fresh seed each round,
   `--use-server-token-count`; compute HBM headroom first, or an oversized GPU
   pool leaves the offload tier with nothing to do and the measurement reads as
   a null result.

I will post these numbers on this PR once 8 GPUs free up. Reviewers may prefer
to hold the merge until then.

## Files

| file | |
|---|---|
| `atom/plugin/vllm/kv_transfer/kda_state.py` | new, 706 — boundary planner + state tier |
| `atom/plugin/vllm/kv_transfer/connector.py` | +485/-21 — `SupportsHMA`, two-group dispatch, pool binding |
| `atom/plugin/vllm/kv_transfer/kv_cache_layout.py` | +89 — per-group split; block-count guard scoped *within* a group |
| `atom/kv_transfer/offload/chunked_scheduler.py` | +35 — `install_hit_cap_hook` |
| `atom/kv_transfer/offload/dense/connector.py` | +7 — retain LMCache metadata |
| `tests/test_vllm_kda_state_offload.py` | new, 504 — 34 tests |
| `recipes/atom_vllm/Kimi-K3.md` | +102 — LMCache section, with a *Status* paragraph saying the numbers are requirements, not measurements |


PerryZhang01:
Superseded by #2268, which targets `main`.

The premise of this PR — that the recurrent (KDA) leg needs vLLM 0.29 — was wrong. Three of the four APIs are in the pinned 0.28 (`2cf0a6915` = `v0.28.0`) verbatim: `SupportsHMA`/`request_finished_all_groups` (`kv_connector/v1/base.py:85,93`), `bind_gpu_block_pool` (`base.py:464`, bound at `sched/scheduler.py:296`), and the boundary-state hand-off as `SchedulerOutput.partial_tail_offloads` (`sched/output.py:265`) — the same `{req_id: [(group_id, block_id, boundary_tokens)]}` payload 0.29 moved under `kv_connector_block_state.boundary_state_offloads`. The check behind this PR's claim grepped only for the 0.29 spelling and read a zero hit as non-existence.

#2268 reads either spelling, so it works on the pinned 0.28 and still works once #2236 lands. The branch here is kept for reference; no code is lost.

PR #2261, opened 2026-09-16
fix(plugin): bound the deferred free of a save that never reports

PerryZhang01:
## What this fixes

On the vLLM **plugin** path (`AtomLMCacheOffloadConnector`), `request_finished()`
returns `True` for as long as `should_defer_free(seq)` holds. vLLM honours that by
keeping the finished request's GPU blocks until the connector names the id in
`finished_sending`. `_collect_releases()` re-tests the condition every step, but
**nothing bounds how long it may keep answering "still saving"**.

`should_defer_free` clears only when the save is reported, and a report can be lost --
most cheaply when the all-TP-rank quorum in `_absorb_worker_meta` never completes.
After that the deferral has nothing left to wait for and the blocks are held for the
lifetime of the process.

The native engine already has a ceiling for exactly this case,
`reclaim_stale_leases(save_abandon_timeout_s())`. The plugin connector implements
none of that family (grep count 0), so on this path there is no ceiling at all.

## Why it is not a slow leak

Those blocks are **referenced**, not cached. The prefix cache cannot be dropped to
recover them and preemption cannot take them back.

Measured on MiniMax-M3 agentic replay, TP4, concurrency 40, 400 GiB/rank CPU tier --
i.e. with the tier nowhere near full and zero allocation failures:

| probe | value |
|---|---|
| `vllm:num_requests_running` | 0 |
| `vllm:num_requests_waiting` | 50, `reason="capacity"` |
| `vllm:kv_cache_usage_perc` | 96.75% |
| `vllm:num_preemptions_total` | 0 |
| `POST /reset_prefix_cache` | `{"success": false}` |

The last row is the discriminating one: vLLM is saying the blocks still carry
references, so this is not "the cache is full", it is "someone has not let go".
The engine emits nothing from then on.

## This failure is intermittent -- read the table above accordingly

A later run of the **same** configuration (same image, same 400 GiB/rank tier, same
concurrency 40, with this patch loaded) completed normally: 695 profiling requests,
aiperf rc=0, and at the end `Running=0` with **`Waiting=0`** rather than `Waiting=50`,
`num_preemptions_total=2` (the engine could still take blocks back), and **zero**
reclaim warnings from the code added here. The bound added by this PR did not fire in
that run and was not needed by it.

So please do not read "reproduces at 400 GiB/rank" into the table above. Read it as
"was observed there once, and the state it was observed in is one the engine cannot
leave on its own". The argument for this PR is the structural gap -- the native path
bounds this wait and the plugin path does not -- not a reproduction recipe.

## The fix

Reclaim after `save_abandon_timeout_s()`, **asked of the scheduler** rather than
re-derived in the connector. The value is LMCache knowledge -- it sums LMCache's own
`LMCACHE_EC_PIN_TIMEOUT_SEC` with a margin, and that ordering *is* the safety
argument -- and the native path already reads it from the same place. A second env
var owned by this connector could be set the wrong way round relative to the pin
window with nothing in the system to say so.

Reclamation takes the normal release path (`scheduler.request_finished(seq)`, then the
id into `finished_sending`), so ATOM's save tracker is popped exactly as it is for a
healthy save and the release is still emitted exactly once.

Every reclaim logs a warning naming the request and how long it was held. A bounded
wait that actually fires is still a defect somewhere upstream of here and must not
pass silently.

## Compatibility

* `save_abandon_timeout_s() <= 0` disables reclamation.
* A scheduler that predates the hook reports nothing (`getattr` returns `None`).

Both keep today's behaviour exactly, so this cannot change a healthy deployment.

## Tests

Five tests in `tests/plugin/test_vllm_offload_completions.py`, all driving a
monkeypatched `time.monotonic`:

* a never-reported save does not hold the blocks forever;
* reclamation is off when the scheduler reports no window;
* a save landing inside the window is not counted as reclaimed;
* the window is per-request, not per-connector;
* a scheduler without the hook never reclaims.

Discriminating power checked with a 4-mutant matrix (drop the timeout compare, drop
the `_deferred_since` stamp, make the window per-connector, ignore a missing hook);
all four are caught.

Offload suites: **403 passed, 1 failed**. The failure,
`test_lmcache_connector_fused_chunk_fastpath_uses_chunk_major`
(`'_FakeStream' object has no attribute 'device'`), is **pre-existing**: it fails
identically with these two files checked out at `origin/main`, and it exercises
`atom.kv_transfer.*`, which this PR does not touch.


PerryZhang01:
CI note, so the red square is not read as this PR:

- **Check Code Style with Black** — was mine (hand-wrapped lambda assignment that fits on one line at the 88-column default). Fixed in 875cf202b.
- **Run non-GPU unit tests** — `tests/test_scheduler_metrics.py::test_decode_request_context_gauge_records_first_dispatch_once`, `AttributeError: 'types.SimpleNamespace' object has no attribute 'num_placeholder_tokens'` at `atom/model_engine/scheduler.py:2503`. **Pre-existing on `main`**, not introduced here: the identical single failure (`1 failed, 5761 passed`) is on main runs [35101975032](https://github.com/ROCm/ATOM/actions/runs/35101975032) (709c25be0) and [35089552171](https://github.com/ROCm/ATOM/actions/runs/35089552171) (b3cccda1c). Neither `tests/test_scheduler_metrics.py` nor `atom/model_engine/scheduler.py` is in this diff — it touches only `atom/plugin/vllm/kv_transfer/connector.py` and `tests/plugin/test_vllm_offload_completions.py`. Leaving it to whoever owns the scheduler change rather than fixing it here.

PerryZhang01:
Closing: this is already in `main`. #2231 merged `AtomLMCacheOffloadConnector._reconcile_stale_saves`, which bounds the deferred free against `offload_save_abandon_timeout_s()` — the same fix, in a more developed form (it also drops the stale quorum entries and is interval-gated). This branch was written against a tree that predated that merge, and rebasing it produces only conflicts against its own duplicate.

The follow-on work that was stacked on this branch has moved to #2284, rebased on `main` and written against `_reconcile_stale_saves`.

PR #2280, opened 2026-09-18
[sgl+atom] Upgrade SGLang to v0.5.19

zhangxinyuanliuhengyu:
## Summary
- Bump SGLang docker pins from v0.5.17 to **v0.5.19** in `docker-release.yaml`, `docker/sglang_release.dockerfile`, and `docker/Dockerfile`.
- Minimal plugin-only adapt: retarget Kimi-K3 pool patch from removed `ModelRunnerKVCacheMixin` onto `KVCacheConfigurator` + `hybrid_arch.kimi_linear_config`; MiniMax-M3 pool patch is a no-op when upstream `MiniMaxSparseKVPool` exists; capture-mode import prefers `runner_utils.get_is_capture_mode`.
- Everything else is left as-is. The 0.5.19 image is expected to be built by CI from these pins.

## Precheckin
- Black (`--target-version py311`) on the 3 Python files: pass
- Ruff on the 3 Python files: pass
- Accuracy catalog schema validation: pass
- Host GPU-less pytest cannot load this machine's CUDA torch; plugin tests are also out of Pre Checkin scope (`run_unit_tests.sh` ignores `tests/plugin/`)

## Local smoke (MI308X, overlay `sglang-v0.5.19` + this branch on `rocm/atom-dev:sglang-latest`)
- Qwen3.5-35B-A3B-FP8 TP2 (PR CI args): completions OK
- Qwen3.5-397B-A17B-FP8 TP4 (nightly/MI308 catalog args): completions OK
- MiniMax-M3-PTPC-FP8 TP4 (PR MiniMax extra_args; local MXFP4 weights not present): completions OK, upstream `MiniMaxSparse` pool in use
- Not smoked here: GLM-5.1 / DeepSeek-V4-Flash TP8 (GPUs 0–1 occupied), Kimi-K3 / DeepSeek-R1-FP4 / MiniMax-M3-MXFP4 (no local weights)

## Test plan
- [ ] Pre Checkin (black / ruff / catalog / unit tests)
- [ ] Nightly docker release builds `sglang-v0.5.19` image from these pins
- [ ] ATOM SGLang Test PR matrix (GLM-5.1-FP8 TP8, DeepSeek-R1-FP4 MTP3, MiniMax-M3-MXFP4 TP4, Qwen3.5-35B-A3B-FP8 TP2, DeepSeek-V4-Flash TP8, DeepSeek-V4-Pro MTP3 TP8)


zhangxinyuanliuhengyu:
## Review notes (plugin + 0.5.19 pin)

Plugin retarget onto `KVCacheConfigurator` / `hybrid_arch.kimi_linear_config` / `runner_utils.get_is_capture_mode` looks directionally right for 0.5.19. Local smoke on Qwen3.5 + MiniMax-PTPC is useful; the PR matrix still needs K3 / GLM / DSV4.

Please check before relying on K3 CI green:

1. `KVCacheConfigurator` is `@dataclass(slots=True)`. `_restore_kimi_k3_mem_fraction` does `owner._atom_kimi_k3_mem_fraction_restored = True`. On a slotted instance that is AttributeError. Store the flag on the class, on `model_config`, or use a module-level set — not a new instance attr.
2. K3 was not smoked locally. This is the only path that monkeypatches pool construction. MiniMax is now a no-op when `MiniMaxSparseKVPool` exists; that is fine for 0.5.19 if ATOM still reads index-K / FP8 scales off the upstream pool (the deleted shim used to wrap `ATOMMiniMaxM3SGLangKVPool`). Confirm `maybe_get_minimax_m3_pools_from_sglang_batch` still finds those buffers without the wrapper.
3. CI will rebuild SGLang. `sglang-latest` is still labeled v0.5.17, so `ci:sglang` will take the slow path in `atom-sglang-test.yaml` (build `docker/sglang_release.dockerfile` at v0.5.19) rather than overlay-only. Expect longer queue/runtime, not a fast overlay.
4. 0.5.19 breaking notes worth a one-line watch: unified radix tree is default; Lean attention auto-enables on ROCm Triton (`SGLANG_DISABLE_LEAN_ATTENTION=1` if decode graph looks off); `ServerArgs` no longer auto-resolves (launcher is fine; custom constructors are not). FlashInfer 0.6.18 is mainly CUDA — ROCm/AITER path should stay on aiter.

`ci:sglang` added so the ATOM SGLang Test matrix can start after Pre Checkin.

zhuyuhua-v:
Scope changed since the last pass: the PR rebased onto main and retargeted 0.5.19 → **0.5.20**, adding six commits. I re-verified everything against the real `v0.5.20` tree (`94602c9c2b`).

**Verdict: request changes.** One new deterministic breakage (shipped script + three recipes pass a flag 0.5.20 no longer accepts), one new silent-divergence in the K3 mem-fraction predicate, and the two open items from the previous round are untouched.

---

### What's new and verified correct

- `--cuda-graph-bs-decode` is the right new name (`arg_groups/cuda_graph_hook.py:81`, `field_order.py:231`).
- `configs/qwen4_exp.py` ships in 0.5.20, so deleting the 498-line recognition shim is justified. **And the multimodal path is safe** — I checked: `get_mm_processor_cls` matches on `model_cls.__name__ not in hf_config.architectures`, i.e. by name string, and ATOM's `EntryClass` is also named `Qwen4ExpForConditionalGeneration`, so upstream's `Qwen4ExpMultimodalProcessor` (`multimodal/processors/qwen_vl.py:348`) still resolves for the ATOM class. Good call, not a regression.
- The `_unified_kv` ABI publication is correct, including the part I expected to be a gap. `disaggregation/utils.py:1598` reads `_unified_kv` unguarded on **both** target and draft pools (so both proxies need it — they get it from `__init__`), and `unified_swa_window` / `unified_swa_ring_size` / `unified_swa_pages` are read only inside `if token_to_kv_pool._unified_kv:`, so publishing `False` correctly short-circuits them. `compression_ratios` is read unguarded above and the proxy already has it.
- The `get_sglang_server_args` retarget covers all five call sites; no direct `get_global_server_args` import is left in `atom/` outside the shim itself.

---

### P0-A (new, blocking) — `--cuda-graph-bs` is gone in 0.5.20, but a shipped script and three recipes still pass it

0.5.20 has **no `add_argument` for `--cuda-graph-bs` and no deprecation alias**. argparse will hard-fail on it.

The CI files were renamed (`sglang_benchmark_models.json`, `sglang_models_accuracy.json`, `atom-sglang-test.yaml`, `atom-sglang-accuracy-validation.yaml`), but `ded283da "Drop mesh/recipe CLI renames from 0.5.20 fix"` deliberately reverted the user-facing ones. What ships on this branch:

```
atom/mesh/scripts/start_decode.sh:66     --cuda-graph-bs $(seq ...)     ← a shipped script; the decode server will not start
recipes/atom_sglang/DeepSeek-R1.md:219,262,303
recipes/mesh/multi-node-sglang.md:119,132
recipes/mesh/single-node-sglang.md:104
```

This is the worst shape for a version bump: **CI goes green and every documented command breaks.** `start_decode.sh` in particular is executable infrastructure, not prose.

If the revert was because the mesh/recipe rename touched unrelated diff noise, please split it into its own commit rather than dropping it — the pin and the CLI are the same change.

### P0-B (new) — the K3 mem-fraction predicate diverges from 0.5.20's, twice, silently

Upstream `arg_groups/attention_hook.py:175-204`:

```python
if resolved_view(server_args).attention_backend == "aiter":          # RESOLVED
    if model_config.context_len > 8192:
        explicit_mem_fraction = (
            getattr(server_args, "_raw_input", None) or {}
        ).get("mem_fraction_static") is not None
        if explicit_mem_fraction and envs.SGLANG_AITER_HONOR_EXPLICIT_MEM_FRACTION.get():
            ...   # 0.85 NOT applied
        else:
            declare_resolution(..., mem_fraction_static=cfg.mem_fraction_static * 0.85)
```

`_restore_kimi_k3_mem_fraction` in this PR:

```python
attention_backend = str(getattr(get_context().server_args, "attention_backend", ""))   # RAW
...
if attention_backend == "aiter" and context_len > 8192:
    restored = current / 0.85
```

1. **Raw vs resolved backend.** If the operator did not pass `--attention-backend` and resolution selected `aiter` (the ordinary ROCm path), the raw record holds `None`/`"auto"`, the predicate is False, and K3 silently keeps the 0.85 reserve. This is precisely the failure mode `get_global_server_args()` was retired for, reproduced inside the code written to accommodate that retirement. The resolved value is `get_exec().kernel.attention_backend` — the retirement message names it explicitly, and ATOM currently reads it nowhere.

2. **The `SGLANG_AITER_HONOR_EXPLICIT_MEM_FRACTION` escape hatch.** When that env is set **and** the operator passed `--mem-fraction-static` explicitly, upstream does *not* apply 0.85. ATOM divides by 0.85 regardless, handing K3 `mem_fraction / 0.85` — a **17.6% over-allocation**. Every ATOM CI config passes `--mem-fraction-static` explicitly, so flipping that env anywhere OOMs K3. Nothing in the bridge checks it.

Please mirror upstream's predicate exactly, or (better) stop reconstructing it — read the resolved value and the raw input the same way `attention_hook` does.

For the record, the rest of this path does check out on 0.5.20: `get_schedule().mem_fraction_static` exists (`kv_cache_configurator.py:2165`, `:2333`) and the 0.85 reserve still exists, having moved to `arg_groups/attention_hook.py:202`.

### P0-C (new) — `get_sglang_server_args()` restores the API but not the invariant it was retired for

0.5.20's `get_global_server_args()` is annotated `NoReturn` and raises, and its docstring states the reason:

> the record answers with the operator's *input*, so a caller reading a field resolution decided got a stale value and no error

`_generate_atom_config_from_sglang_config` reads ~30 `server_args.*` fields, several of which SGLang resolves later. The sharpest instance is in this very PR:

| reader | source | value on ROCm + aiter + long ctx |
|---|---|---|
| `atom/plugin/config.py:611` → `gpu_memory_utilization` | `server_args.mem_fraction_static` | the operator's input (e.g. 0.85) |
| `kimi_k3_bridge._restore_kimi_k3_mem_fraction` | `get_schedule().mem_fraction_static` | the resolved value (e.g. 0.7225) |

ATOM now reads one setting from two places with two different values in a single process. Same question applies to `max_running_requests` (resolved by `resolve_max_num_reqs` against the token budget), `context_length`, and `attention_backend`.

I'm not asking to migrate all thirty fields in this PR — but the shim should carry a comment naming the ones known to be resolution-decided, and `mem_fraction_static` should probably move to `get_schedule()` here, since the PR already proves it knows the difference.

Two smaller things about the shim itself:

```python
try:
    from sglang.srt.runtime_context import get_server_args
    return get_server_args()
except Exception:                                   # catches both cases
    from sglang.srt.server_args import get_global_server_args
    return get_global_server_args()                 # raises "retired" on 0.5.20
```

- **The fallback is dead.** 0.5.19 already has `runtime_context.get_server_args` (verified), and its `get_global_server_args` is just a shim over the same object. So the fallback can only fire on ≤0.5.18 — which `kimi_k3_bridge.py` ("0.5.19+") already refuses to support.
- **It masks the real error.** A live failure inside `get_server_args()` (context not yet published) falls into the fallback, which raises `RuntimeError("get_global_server_args() is retired...")`, and the caller re-wraps that as "Failed to retrieve SGLang global ServerArgs". Three layers, none of them the actual cause. Split the version probe from the call:

```python
try:
    from sglang.srt.runtime_context import get_server_args
except ImportError:
    from sglang.srt.server_args import get_global_server_args as get_server_args
return get_server_args()
```

### P1 — `_unified_kv = False` alongside a working `get_unified_kv`

The proxy advertises `_unified_kv = False` and `unified_kv_pool = None`, then implements `get_unified_kv` / `get_unified_kv_rope` returning the ATOM view. Upstream's own `get_unified_kv` (`deepseek_v4_memory_pool.py:845`) dereferences `unified_kv_pool` and is therefore only reachable when the flag is True — so the pair is self-contradicting: the flag says "no unified layout", the accessor serves one.

The docstring says "callers that still ask for the plane (env-gated backends, diagnostics) get the proxy view". Which caller is that, concretely? I couldn't find one in 0.5.20 that calls it without checking the flag first. If there isn't one, this is speculative API — drop it, or name the caller in the comment so the next reader can tell whether it's still needed.

---

### Carried over from the previous round — still open, re-verified on 0.5.20

**P0 — MiniMax index-dtype swap is applied to the pool build but not to `cell_size`.** The bridge is byte-identical to the last review, and both upstream halves are unchanged in 0.5.20:

- `kv_cache_configurator.py:1806-1809` still picks `index_dtype = self.model_dtype` unless `m3_fp8_attn_gemm_enabled`, which still requires `trtllm_mha` + `is_sm100` → still always False on ROCm.
- `pool_configurator.py:389` still bills the indexer at `torch._utils._element_size(kvc.model_dtype)`, and it runs in `_resolve_memory_pool_config`, i.e. **before** `_init_pools` reaches the patched builder.

So the budget charges index-K at 2 B/elem while the pool allocates at 1. On the M3 TP4 numbers we measured (main 15,360 B/token, bf16 index 14,592 B/token) that leaves **~24% of the profiled KV budget unallocated** — `max_total_num_tokens` could be 1.32× larger. Safe but expensive, and it will read as "capacity dropped after the upgrade".

`model_dtype` is read in exactly three places across the two files, so extending the swap to wrap `configure()` (or `_resolve_memory_pool_config`) has a fully enumerable blast radius.

**P0 — `_KIMI_K3_MEM_FRACTION_RESTORED` keyed by `id(owner)`.** Unchanged. `id()` is reused after GC; spec-decode builds two configurators, so the second can inherit the first's address and skip the restore, leaving K3's draft pool at 0.85× with no log line. The `override` is already name-keyed on `"atom-kimi-k3-mem-fraction"` — ask the context instead. Also still needs a `logger.warning` on the `except Exception: return` path.

**P1 — the `kimi_linear_config` rebind is a `sys.modules` snapshot.** The three-module list is correct and complete for 0.5.20, but `sys.modules.get()` silently skips a module not yet imported, and `kv_cache_configurator` (the one this function forces to import) pulls in neither `kv_cache_builder` nor `attention_registry`. At minimum make the new `logger.info` report which modules were *actually* rebound — as written it claims all three unconditionally, which is misleading exactly when it matters. The durable fix is to install a real `KimiLinearConfig` onto `model_config.hf_config.text_config` so upstream's unpatched function returns it everywhere.

**P1 — `empty_kv_indptr` is still fixed at the wrong layer.** `AttentionMetaData_DSV4` already declares the field with a `None` default and the consumer already guards `is not None`; having the three builders construct that class instead of the base `AttentionMetaData` fixes this and the whole class of bolted-on field. As written, every decode and verify step pays an `np.zeros` alloc + host memcpy + H2D for a buffer `CpuGpuBuffer` already zeroed, feeding a path that needs `ATOM_USE_V4_PREFILL_ASM_FOR_DECODE=1` **and** gfx1250.

**P1 — dead code.** `ATOMMiniMaxM3SGLangKVPool` still has zero instantiations (and `69e6b388` edited it); `_dtype_size` is unreferenced; `_m3_index_dim` / `_m3_sparse_layer_ids` are reachable only from the dead class.

**P1 — `install_minimax_m3_pool_patch` is called from two places**, and the one inside `bind_minimax_m3_sparse_cache_views` now cannot do anything useful: the patch wraps *pool construction*, which has already happened by bind time.

**P2 — deleted MM test.** Dropping `register_qwen4_exp_processor` is right, but `tests/plugin/test_sglang_qwen4_exp_mm.py` was the only thing asserting that a Qwen4Exp config resolves to a multimodal processor. Keep it, retargeted at `get_mm_processor_cls` — that's a three-line test and it pins the name-matching behaviour this PR now depends on.

---

**Blocking: P0-A, P0-B, and the two carried-over P0s.** P0-A is the cheapest and the most user-visible — restore the `--cuda-graph-bs-decode` rename in `atom/mesh/scripts/start_decode.sh` and the three recipes, in a separate commit if the diff noise was the objection. P0-B and the `id()` key are a few lines each. The MiniMax budget/allocation desync is the one worth a measured `max_total_num_tokens` in the PR body.

Also, please add `MiniMax-M3-MXFP8 EAGLE3 TP4 MTP3` to the test matrix: checking the catalog, the listed `MiniMax-M3-MXFP4 TP4` entry passes no `--kv-cache-dtype`, so it runs bf16 KV and cannot exercise the fp8 index path this PR's MiniMax fix exists for.

zhangxinyuanliuhengyu:
@zhuyuhua-v Thanks for the re-verify against v0.5.20 — agreeing on the P0s.

**This push (`3b60c987`) is plugin-only** (mesh/recipes CLI left for a follow-up so this pin PR stays narrow):

### Done in plugin
- **P0-B K3 mem-fraction**: restore now mirrors `attention_hook` — resolved `get_exec().kernel.attention_backend`, skips undo when `SGLANG_AITER_HONOR_EXPLICIT_MEM_FRACTION` + explicit `--mem-fraction-static`, dedupes via override source `"atom-kimi-k3-mem-fraction"` (no `id(owner)`), and warns if schedule/exec is unread.
- **MiniMax index budget/alloc**: `model_dtype` swap moved from `_build_minimax_sparse_kv_pool` alone to wrapping `KVCacheConfigurator.configure()`, so `_resolve_memory_pool_config` cell_size and pool `index_dtype` share the same FP8 billing.
- **P0-C shim**: `get_sglang_server_args` only falls back on `ImportError`; docstring names resolution-decided fields; `gpu_memory_utilization` notes it is still raw input (ATOM pool owned by SGLang).
- Small: `kimi_linear_config` rebind log reports which modules were actually rebound.

### Intentionally not in this commit
- **P0-A mesh/recipes `--cuda-graph-bs-decode`**: deferred on purpose (keep pin PR to sgl plugin). CI yaml/json already renamed. Follow-up will restore `start_decode.sh` + the three recipes.
- **P1** (`_unified_kv` / `get_unified_kv`, empty_kv_indptr layer, MiniMax dead code, MM test retarget, MXFP8 EAGLE3 in PR matrix): acknowledged; follow-up unless you want any of them blocked here.

Pre Checkin: `black` + `ruff` clean on the three touched plugin files.

zhangxinyuanliuhengyu:
@zhuyuhua-v Follow-up on the two API mistakes in `3b60c987` (called out in review):

**Fixed in `434ef26b` / `$(git rev-parse --short HEAD)`** (plugin only, still no mesh):
1. **`overrides_log()`** — 0.5.20 exposes a method; we now call it (with `_overrides_log` fallback). Dedup check is *outside* the publish `except`, so a TypeError can no longer be mistaken for “schedule unread”.
2. **HONOR env** — now `from sglang.srt.environ import envs` then `envs.SGLANG_AITER_HONOR_EXPLICIT_MEM_FRACTION.get()` (ImportError-only fallback). No broad `except Exception` around the env read.
3. Added `tests/plugin/test_sglang_kimi_k3_mem_fraction.py`: method-call dedupe, restore-once-then-skip, and HONOR skip (CPU / mocked SGLang).

Mesh/`--cuda-graph-bs-decode` and P1 items remain deferred as before.

zhangxinyuanliuhengyu:
Correction: the tip is `742d3d63` (test expect fix on top of `434ef26b`).

zhangxinyuanliuhengyu:
@zhuyuhua-v Tightened the empty HONOR test (`4755c2ea`):

- HONOR skip now uses **schedule=0.85** (upstream did not *0.85). If the env lookup is wrong, restore would override to 1.0 and the assertion fails.
- `overrides_log=lambda: []` instead of `list`.
- Added `attention_backend=""` / `"auto"` cases with raw CLI still `aiter` and schedule already scaled — restore must not fire.

Plugin P0s for this pin are done. P0-A mesh/recipes `--cuda-graph-bs` and remaining P1 stay as follow-up.

PR #2281, opened 2026-09-18
perf(plugin): hold only the blocks a save still reads, not the whole request

PerryZhang01:
Stacked on #2261 (`perzhang/plugin-deferred-free-reclaim-20260916`), which is the base branch of this PR. Review that one first; this PR's diff is only the second commit.

## The problem

`request_finished` is all-or-nothing. Returning `True` makes vLLM keep the finished request's **entire** block table until the connector names the id in `finished_sending`.

The native engine never pays that. `atom/model_engine/scheduler.py:3085` asks `_connector_protected_block_ids` (`:896`) and frees everything except the exact blocks an in-flight save is still reading. The plugin connector degraded to whole-request deferral — `protected_block_ids` appears nowhere in it — even though everything needed was already present on both sides:

| piece | where it already exists |
|---|---|
| `_supports_early_block_release = True` | `atom/kv_transfer/offload/dense/connector.py:77` |
| `protected_block_ids` / `activate_block_leases` | `atom/kv_transfer/offload/chunked_scheduler.py:481` / `:521` |
| `take_source_safe_releases` / `reclaim_stale_leases` | same file, `:535` / `:541` |
| engine keeps stepping while leases are out | `has_pending_work()` already counts `_save_lease_blocks` |
| `bind_gpu_block_pool(pool)` | called unconditionally on the scheduler half before any request runs |
| `BlockPool.touch` / `BlockPool.free_blocks` | vLLM 0.28 |

The connector simply never accepted the pool.

## The change

Ask ATOM for the exact blocks. If it can name them: `touch` those, record the lease, drop the SeqView, and return `False` — vLLM frees the rest of the table immediately. If it answers `None` (no early-release support, or a load in flight, since a load also touches HBM blocks that are not tracked per-range), nothing changes: the request takes the whole-request deferral path exactly as before, still bounded by #2261's reclaim.

Leases are handed back in `update_connector_output`: the source-safe drain first, then `reclaim_stale_leases(self._save_abandon_timeout_s())` as the bounded fallback. Giving it the *same* window as the deferral is deliberate — a lease must not be able to outlive the deferral it replaced. Freeing is tail-first within each set, matching vLLM's own convention so a shared prefix is the last thing eligible for eviction.

Shape borrowed from the already-merged Mooncake store connector (`pool.touch` per emitted store job, quorum → `pool.free_blocks(..., reversed)`, `request_finished` returning `False`).

## Measured

MiniMax-M3 MXFP4 agentic replay on the vLLM plugin backend, TP4, concurrency 40, 900 s profiling window, LMCache CPU tier 64 GiB/rank. Two arms differing **only** by this commit — same image, same script, same seed and request set.

| | before | after |
|---|---:|---:|
| GPU KV cache usage p50 | 97.3 % | **92.0 %** |
| whole-request frees reclaimed at the abandon ceiling | 530 | **0** (narrow path taken instead) |
| `still_deferred` p50 | 216 | 0 |
| Waiting reqs max | 34 | 25 |
| Running reqs p50 | 40.0 | 34.5 |
| requests completed in the window | 572 | 795 |

`Running` p50 falling while completions rise is reported as measured, not explained. Fewer requests resident at once and more of them finishing is consistent with each clearing faster, but this pair was not designed to separate that from batch-composition effects.

## What this PR does not claim

- **No throughput claim from the table above.** By that pair's own pre-registration it is void for throughput: the *before* arm failed the harness's LMCache value gate at this tier (external hits = 0).
- A separate matched-working-point arm puts the patched plugin at **21,956.1** tok/s/GPU against ATOM native's **22,853.2** (−3.93 %) at the same tier, concurrency and window. That is **not** a backend difference: native's tier supplied 3.0720 % of prompt tokens there against the plugin's 0.0104 %, a 295× asymmetry, so it compares a plugin with an inert tier to a native engine with a live one. n=1, no noise floor, and 3.93 % is barely above the 3 % "cannot measure" line that arm was pre-registered with.
- **This changes _what_ is held, not _how long_.** The source-safe completions that would return a lease early never cross the worker → scheduler boundary on this path: `AtomOffloadWorkerMetadata` carries only `saved` and `load_failed`, and `connector_completions` has no transport here at all. So `take_source_safe_releases()` drains an empty list today and every lease is returned by `reclaim_stale_leases` at the same window the deferral used. Carrying the source-safe identity across that boundary is separate work, not in this PR.

## Tests

Seven cases added to `tests/plugin/test_vllm_offload_completions.py`, covering the narrow path, both fallbacks (`protected_block_ids` → `None`, and no pool bound at all — the worker half is never given one), the empty-protection case where there is no share to give back, tail-first release ordering, the force-release warning, and that leases are left recorded rather than dropped when there is no pool to return them to. `33 passed` in `rocm/atom-dev:vllm-latest`.


PerryZhang01:
Superseded by #2284, which carries this commit rebased on `main` together with the functional fixes it needs (the fp8 index-cache dtype from #2122 and the LMCache `world_size` fix) in a single PR against `main` rather than a stack.

Closing rather than retargeting because the commit changed on the way: this branch was based on #2261, which turned out to duplicate work already merged in #2231. In #2284 the lease path is written against `main`'s `_reconcile_stale_saves` instead, and it also corrects that method's docstring, which asserted `reclaim_stale_leases` was "unimplementable" on a vLLM connector.

PR #2282, opened 2026-09-18
fix(offload): read PP/TP from parallel_config for LMCache world size

PerryZhang01:
`lmcache_replica_world_size()` collapses to 1 on the vLLM plugin path, so an all-rank lookup client never builds and the slow tier is written but never read. This PR only fixes that config-shape bug. Native ATOM is bit-for-bit unchanged.

Applies cleanly on `1c4caef48`. Depends on nothing outstanding (#2231 is merged).

---

## Problem

It reads `pipeline_parallel_size` / `tensor_parallel_size` off the **top level** of the config object:

```python
pp_size = int(getattr(config, "pipeline_parallel_size", 1) or 1)
tp_size = int(getattr(config, "tensor_parallel_size", 1) or 1)
```

That is right for ATOM's own config, which carries them directly. On the vLLM plugin path `config` is a `VllmConfig`, where both live under `config.parallel_config`. Neither `getattr` finds anything, both default to 1, and a TP4 replica reports **world=1**.

## Why that is not a cosmetic off-by-one

`world=1` puts the all-rank lookup scope out of range. With the scope the MiniMax-M3 agentic recipe prescribes at 48c:

```
LMCACHE_LOOKUP_SERVER_WORKER_IDS=0,1,2,3
```

the four per-rank lookup *servers* start normally, but the scheduler-side lookup *client* fails to build. `dense/connector.py` catches that with a bare `except Exception` and continues with `lookup_client=None`:

```
[atom] LMCache offload scheduler: lookup client unavailable:
```

— with an empty exception message. The server then comes up healthy and nothing about the run looks wrong, but `ChunkedOffloadSchedulerBase` can never learn `lmcache_cached_tokens`, never builds a `LoadSpec`, and `_reqs_need_recv` stays empty. **The offload tier is written and never read**, for the whole run.

Measured on MiniMax-M3 MXFP4, TP4, plugin backend, concurrency 40, 1 h window:

| | plugin (buggy) | ATOM native, same build/config |
| --- | ---: | ---: |
| `Retrieved` | 0 | 1760 |
| `Double unpin` | 0 | 878384 |
| `Stored` | 3508 | — |
| `vllm:external_prefix_cache_hits_total` | 0.0 | — |
| `vllm:external_prefix_cache_queries_total` | 1.3e+07 | — |

An "LMCache on" arm paying the full store cost for zero reuse. The direct proof that the config *shape* is what differs is the log line itself — same build, two backends:

```
native : LMCache offload scheduler: lookup client on atom-offload-dp0 (world=4)
plugin : LMCache offload scheduler: lookup client on atom-offload-dp0 (world=1)
```

`Stored` proceeding normally is **not** evidence the tier works: the save path does not use the lookup client. The discriminators are `Retrieved`, `Double unpin` and `external_prefix_cache_hits_total`.

## Why it stayed hidden

The default lookup scope is `[0]`, which is in range even when world is wrongly 1. Every arm that left `LMCACHE_LOOKUP_SERVER_WORKER_IDS` unset built a rank-0-only client and did retrieve. The bug only fires in the cell "plugin backend + all-rank lookup" — which is exactly what the MiniMax-M3 agentic recipe prescribes for 48c, where it reports the all-rank scope taking the prompt cache hit rate from 82.86% to 95.48%. That improvement is not reproducible on the plugin path without this fix.

## Fix

Read the top level first and fall back to `parallel_config`. The ordering matters: ATOM's native config always has correct top-level values, and `VllmConfig` has nothing there, so this is strictly additive — the native path is bit-for-bit unchanged, including the case where `parallel_config` exists but only carries ranks (ATOM's own shape), which must not shadow the top level.

The helper lives in `atom/kv_transfer/offload/config.py`. Call sites in `dense/`, `hybrid/dsv4/` and `mp/backend.py` pick up the new value with no further change.

## Verification

- New regression test `test_replica_world_size_reads_vllm_parallel_config` covers the vLLM shape (TP4 → 4, PP2×TP4 → 8) and re-asserts that an ATOM-shaped `parallel_config` carrying only ranks does not shadow the top level. It **fails on main** with `assert 1 == 4` and passes with the fix.
- End to end on the same MiniMax-M3 arm as the table above, after the fix: the boot logs `(world=4)`, all four lookup servers come up, and the three discriminators go non-zero — `Retrieved` 4 / `Double unpin` 476 / `external_prefix_cache_hits_total` 30464.0, the last matching the first `Retrieved 30464 out of 30464 required tokens` log line exactly.
- Independently confirmed on a **second model and a second connector**: GLM-5.2-MXFP4 TP4 on the vLLM plugin backend with ATOM's own `AtomLMCacheOffloadConnector`, AgentX replay at concurrency 32, 64 GiB/rank tier, `LMCACHE_LOOKUP_SERVER_WORKER_IDS=0,1,2,3`. Boot logs `lookup client on atom-offload-dp0 (world=4)` — the line the old code printed as `(world=1)` — and the tier is read on all four ranks: 50,048 tokens / 2.22 GiB per rank at 7.0–12.1 GB/s, with `vllm:external_prefix_cache_hits_total` ending at 50,048, matching the `Retrieved` log lines exactly.

On that GLM-5.2 arm the tier supplied only **0.0529%** of prompt tokens. That is a structural limit of the plugin path (vLLM asks the connector only about the HBM-miss tail) and is not something this fix addresses. Confirmed bit-exact in that run by `prefix_cache_queries - prefix_cache_hits = 95,172,276 - 16,337,856 = 78,834,420 = external_prefix_cache_queries`. This fix makes the tier reachable at all; how much it can then supply on the plugin path is a separate problem.

## Not in this PR

- The bare `except Exception` in `dense/connector.py` that turned a fatal misconfiguration into a silent write-only tier. Out-of-range lookup scope should arguably fail the launch rather than warn with an empty message.
- Slow-tier save-admission / lookup-skip gates (`OFFLOAD_SAVE_ADMISSION`, `OFFLOAD_LOOKUP_SKIP_UNLOADABLE`). Those skip lookups whose answer cannot change the *current* request, but they leave the save floor at the HBM frontier and therefore starve the slow tier when the HBM pool is tighter than the working set (MiniMax-M3, 16612 blocks: −3.8 throughput points). If that tuning is still wanted as a default-off experiment, it belongs in #2283, not here.

## Tests

- `tests/test_lmcache_offload_connector.py`: new `test_replica_world_size_reads_vllm_parallel_config`.
- Plugin-path evidence is the two end-to-end arms above. `.github/scripts/run_unit_tests.sh` passes `--ignore=tests/plugin`, so `tests/plugin/` is not covered by CI unit tests.


PerryZhang01:
Related: #2283 ports the slow-tier admission / lookup-skip gates from the same MiniMax-M3 line. Independent of this fix — different files, no dependency either way.

yhl-amd (approved):

yhl-amd:
Post-merge review: I did not find an additional correctness issue in this change. The helper preserves the native ATOM top-level PP/TP values and only falls back to `parallel_config` for the vLLM-shaped config; the TP4, PP2×TP4, and native-config precedence cases all produce the intended replica-local world size. The call sites in dense, DSV4, and MP all consume this helper consistently.

PR #2283, opened 2026-09-18
perf(offload): port the slow-tier admission and lookup-skip gates to main

PerryZhang01:
## What this is

Nine commits from the MiniMax-M3 offload line, rebased onto current `main`.
They add two **optional, default-off** gates to the LMCache offload path, plus
the telemetry that decides whether either is worth turning on, plus an
`offload/README.md` section documenting both.

Depends on nothing outstanding — #2231 is already merged, and this applies
cleanly on top of `1c4caef48`.

## Why

Saves and lookups on the offload path are unconditional today. That is the
right behaviour when the slow tier has customers, and pure cost when it does
not. Nothing in the connector currently measures which regime it is in, so
there is also no way to answer the question for a given workload without
guessing.

The first thing these commits add is therefore not a gate but a **verdict
tally**: every reload decision is already classified (`aligned_large_hit`,
`slow_tier_unused`, `too_small`, …), and `[OFFLOAD-ADMISSION]` now reports the
running payoff **whether or not either gate is enabled**. That part is free and
always on, and it is what makes the two gates decidable rather than a matter of
taste.

## The two gates

**`OFFLOAD_SAVE_ADMISSION=1`** — treat each reload verdict as one sample
(`aligned_large_hit` = the tier paid off, anything else = it did not) and run an
EMA over them. At or above `OFFLOAD_ADMISSION_TARGET` (0.10) every save is
admitted; below it, the admitted share falls linearly to
`OFFLOAD_ADMISSION_FLOOR` (0.25), carried by a credit accumulator so the
realized share tracks the computed one without randomness.

**`OFFLOAD_LOOKUP_SKIP_UNLOADABLE=1`** — skip a lookup whose answer cannot
change the outcome. When even a total hit (`num_prompt - 1`) would leave the
loadable gap `num_prompt - 1 - hbm` below `OFFLOAD_MIN_LOAD_TOKENS` (8192), the
verdict is `too_small` no matter what the tier says, so the round trip is pure
cost.

Skipping is **not** free, and the commits are explicit about the two ways it
bites:

- It leaves the save floor at the HBM frontier, which *asserts* the tier already
  holds this prefix — an assertion nothing checked, because nothing asked.
  Dropping the floor instead does not mean "write the prefix once", it means
  rewriting the whole prompt every request: measured, stored bandwidth
  0.272 → 2.448 GB/s and throughput −44.7%.
- It **biases the evidence the other gate reads.** The skipped requests are
  exactly the ones HBM already covers, so what survives over-reports payoff:
  at 16612 blocks `slow_tier_unused` falls 1779 → 226 while `aligned_large_hit`
  holds at 1293 → 1313, roughly doubling the estimate the controller sees. The
  skip therefore records the verdict it provably would have reached
  (`note_slow_tier_verdict(False)`) instead of dropping the sample, stands down
  whenever the tier is measurably load-bearing (actionable-rate EMA ≥ 1% after
  64 lookups), and keeps one forced probe every 32 skippable lookups so the
  estimate stays fed after the gate closes.

## What is measured, and where it does nothing

**On MiniMax-M3 the regime decides the sign.** At 46647 blocks the tier answered
0.03% of queried tokens and the skip won 6.1% throughput; at 16612 blocks the
same switch **lost 3.8 points**, because there it starved a tier that was being
read.

**On GLM-5.2 both gates are identity transforms, and I could not make either of
them do anything.** Concurrency 16, 600 s, pool 64, the full 2×2 plus an
interleaved repeat of the admission cell:

- The skip **never fires**. With 32k prompts and an 18% HBM prefix hit the bound
  is ~26.9k tokens against an 8192 floor, so no lookup is ever skippable.
  `OFFLOAD-LOOKUP-SKIP` is zero in every cell.
- Admission **refused not one save** in any of the three runs that enabled it
  (356 / 512 / 522 admitted, `skipped=0` in all three).
- All cells land inside run-to-run spread: eight behaviourally equivalent runs
  span 2.7% end to end, so anything under ±2.7% here is "not measurable", not
  "no effect".

The same code is also live (default-off) in a GLM-5.2 AgentX plugin-backend run
at concurrency 32 finished today, where the tier supplied 0.0529% of prompt
tokens and neither gate triggered either.

**So this PR does not claim a throughput win.** It claims: the gates are correct,
they are off by default, they are documented together with the regime that makes
each one right or wrong, and the telemetry that tells you which regime you are in
is now always available. On the one workload where they help, they help; on
GLM-5.2 they correctly do nothing.

## Two cautions that are in the README, not just here

- `warmup`, `alpha` and the probe cadence are defined in **event counts**, so
  their wall-clock constants drift with the verdict rate — a change elsewhere
  that changes the rate silently retunes them.
- The payoff dips far below target during ramp (4e-05 in GLM-5.2's first report,
  before settling near 0.9), so a threshold sitting between the transient and
  the steady state throttles during ramp rather than at the working point.

One thing not to misread in the telemetry: `save_share` is the **realized** share
`saved/(saved+skipped)`, so `skipped=0` pins it to 1.0. It is not independent
evidence that the gate computed 1.0.

## Overlap with in-flight PRs — please note the merge order

Related, and independent of this one: #2282 fixes the LMCache world-size
collapse on the plugin path. That fix is what makes the slow tier readable at
all there; the gates in this PR only decide whether a save is worth doing.
Neither depends on the other and they touch different files.

"Admission" already meant two things in this subsystem before this branch, and
this adds a third. The glossary entry in the README commit spells all three out:
*staging-row admission* bounds DSV4 temp rows; *worker save admission* is the
`OFFLOAD_MAX_PENDING_SAVES` running-plus-queued bound; *slow-tier admission*
(`OFFLOAD_SAVE_ADMISSION`, this PR) decides whether a save is worth doing at all.
The first two bound concurrency and always run; the third declines work outright
and is off by default.

That matters here because #2201 (`fix(offload): bound Dense/M3 saves with safe
retirement`) and #2246 (`fix(offload): bound DSV4 save admission`) are open on
the *second* meaning and touch three of the same files —
`atom/kv_transfer/offload/_offload_common.py`,
`atom/kv_transfer/offload/README.md` and
`tests/test_dense_offload_connector.py` — and #2201 adds
`atom/kv_transfer/offload/dense/save_admission.py`, which is **not** this gate.
The changes are independent in behaviour but will conflict textually. I am happy
to rebase this one behind either of them; say which order you want.

## Tests

`tests/test_slow_tier_admission.py` (new, 255 lines) and additions to
`tests/test_dense_offload_connector.py` (+233) cover the EMA and credit
accumulator, the skip's precondition and stand-down, the forced probe cadence,
the verdict-recording on the skip path, and the save-floor behaviour.

`.github/scripts/run_unit_tests.sh` was run on `origin/main` (`1c4caef48`) and on
this branch, in the same container. The failure sets were compared test-by-test
from the JUnit reports, not by count: **identical** — the same 28 failures and
the same 4 collection errors on both, with no test newly failing and none newly
passing. main: 28 failed / 7015 passed / 103 skipped / 4 errors. This branch:
28 failed / 7047 passed / 103 skipped / 4 errors — the +32 are the new tests in
`tests/test_slow_tier_admission.py` and `tests/test_dense_offload_connector.py`.
The pre-existing 28/4 are main's own and are untouched by this PR.


PerryZhang01:
Superseded by #2282 — folding the fix and the optimization into a single PR as requested. All nine commits from this branch were cherry-picked onto #2282 unchanged (no conflicts; the two parts touch disjoint files), and #2282's description now covers both. Closing here to avoid a duplicate; the branch is left in place.

PR #2284, opened 2026-09-18
MiniMax-M3 on the vLLM plugin backend with LMCache offload: fp8 index cache, spec-decode uniformity, and exact block leases

PerryZhang01:
Everything needed to run **MiniMax-M3 on the vLLM plugin backend with LMCache KV offload**, collected into one PR: two functional fixes that stop it crashing, and one performance change on top.

The plugin-path LMCache world-size collapse (`lmcache_replica_world_size()` reading TP/PP off the top level of a `VllmConfig` and reporting `world=1`) is **not** in this PR. That fix lives in #2282. Merge #2282 first if the slow tier must be readable under `LMCACHE_LOOKUP_SERVER_WORKER_IDS=0,1,2,3`; this branch does not touch `atom/kv_transfer/offload/config.py`.

Rebased on `main` (`1c4caef48`). Supersedes #2261 and #2281, and carries the functional fixes from #2122.

## What is in it

| # | Commit | Why |
|---|---|---|
| 1 | `fix(minimax-m3): keep the vLLM-plugin index cache in a real fp8 dtype` | `--kv-cache-dtype fp8` kills EngineCore on the first request. vLLM maps every fp8 kv-cache-dtype to `torch.uint8` (a byte buffer its own kernels reinterpret), but ATOM's `_index_block_score_kernel` dispatches on `k.dtype.is_fp8()`, so a uint8-labelled index cache goes down the bf16 branch and `tl.dot` refuses `bf16 × uint8`. |
| 2 | `fix(minimax-m3): give the index-cache dtype exception one named home` | Review feedback on #1: the exception is layout-specific, so it lives where the layout is known (`_index_cache_torch_dtype`) instead of in vLLM's shared `kv_cache_dtype_str_to_dtype` table, which every attention layer of every model reads. |
| 3 | `[MiniMax-M3][Spec decode] Check that a decode segment is uniform before using it` | M3 + EAGLE3 dies part way into a long run with `total_q 121 not divisible by max_query_len 4`. Three decode paths recover a request from a flat query row by dividing by `max_query_len`; that is only valid when the segment is uniform. Now checked rather than assumed. |
| 4 | `perf(plugin): hold only the blocks a save still reads, not the whole request` | `request_finished` returning True makes vLLM hold a finished request's *entire* block table until the connector names it in `finished_sending`. The native engine frees everything except the exact blocks an in-flight save is still reading. Details below. |

## The performance change (commit 4)

The lease machinery already existed on both sides and was simply never connected:

* ATOM already implements `protected_block_ids`, `activate_block_leases`, `take_source_safe_releases`, `reclaim_stale_leases`; the dense offload connector already sets `_supports_early_block_release = True`; `has_pending_work()` already counts `_save_lease_blocks`.
* vLLM already calls `bind_gpu_block_pool` unconditionally on the scheduler half before any request runs, and `BlockPool.touch` / `BlockPool.free_blocks` are the two calls that take and return a refcount share.

The connector just never accepted the pool. Now it does: ask ATOM for the exact blocks, touch those, record the lease, return **False** so vLLM frees the rest of the table on the spot. When ATOM answers `None` (no early-release support, or a load in flight) nothing changes — the request takes the whole-request deferral path exactly as before, still bounded by `_reconcile_stale_saves`. The same shape is merged upstream in Mooncake's store connector, down to freeing tail-first so a shared prefix evicts last.

This also **corrects a claim in `_reconcile_stale_saves`'s own docstring** (added in #2231): it said `reclaim_stale_leases` was "not merely inert here but unimplementable" on a vLLM connector, because it reclaims block ids and "a vLLM connector has no such channel". `bind_gpu_block_pool` *is* that channel. The paragraph now says so, rather than being quietly deleted.

### Measured

MiniMax-M3 MXFP4 agentic replay, vLLM plugin backend, TP4, concurrency 40, 900 s profiling window, LMCache CPU tier 64 GiB/rank. Two arms differing only by the lease commit — same image, same script, same seed, same request set:

| | before | after |
|---|---|---|
| GPU KV cache usage p50 | 97.3% | 92.0% |
| whole-request frees reclaimed at the abandon ceiling | 530 | 0 (narrow path instead) |
| `still_deferred` p50 | 216 | 0 |
| Waiting reqs max | 34 | 25 |
| Running reqs p50 | 40.0 | 34.5 |
| requests completed in the window | 572 | 795 |

`Running` p50 falling while completions rise is reported as measured, not explained: fewer requests resident and more of them finishing is consistent with each clearing faster, but this pair was not designed to separate that from batch-composition effects.

### Three things this PR does **not** claim

1. **No throughput claim from the table above.** The *before* arm failed the harness's LMCache value gate at this tier (external hits = 0), so its tokens/s is void and was pre-registered as such before the pair ran.
2. **The cross-backend number is not a backend difference.** A separate matched-working-point arm puts the patched plugin at **21,956.1** tok/s/GPU against ATOM native's **22,853.2** (−3.93%) at the same tier, concurrency and window — but native's tier supplied **3.0720%** of prompt tokens there against the plugin's **0.0104%**, a 295× asymmetry. That compares a plugin with an inert tier to a native engine with a live one. n=1, no noise floor, and 3.93% is barely above the 3% "cannot measure" line that arm was pre-registered with.
3. **Commit 4 changes *what* is held, not *how long*.** The source-safe completions that would return a lease early never cross the worker → scheduler boundary on this path — `AtomOffloadWorkerMetadata` carries only `saved` and `load_failed`, and `connector_completions` has no transport here at all — so `take_source_safe_releases()` drains an empty list today and every lease is handed back by `reclaim_stale_leases` at the same window the deferral used. That is why it is given `offload_save_abandon_timeout_s()`: a lease must not be able to outlive the deferral it replaced. Carrying the source-safe identity across that boundary is separate work.

## Tests

`tests/plugin/` was run in a throwaway `rocm/atom-dev:vllm-latest` container on an idle GPU, against this branch **and against clean `main`**, and the two failure sets compared line by line. The failures were **the same 34**, all in `test_sglang_*` / `test_rtpllm_*`, none in a file this PR touches.

(`tests/plugin/test_sglang_qwen35_attention.py` fails to collect on both sides and was excluded from both runs.)

New coverage:

* `tests/plugin/test_vllm_offload_completions.py` — seven cases for the lease path: exact blocks touched and leased; `None` falls back to whole-request deferral; no pool falls back too; an empty protection set takes no share at all; a lease handed back tail-first once source-safe; a lease whose save never reports is force-released on the abandon window; leases left alone when there is no pool to give them back to.
* `tests/plugin/test_minimax_m3_decode_uniformity.py` — the non-uniform decode segment check.


atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -812,6 +871,26 @@ def request_finished(self, request, block_ids) -> tuple[bool, dict | None]:
...
+                    pool = self._gpu_block_pool
+                    if protected:
+                        pool.touch([pool.blocks[block_id] for block_id in protected])
+                    self._scheduler.activate_block_leases(seq, frozenset(protected))
+                    self._seqs.drop(req_id)
+                    # False: vLLM frees the table now, and the refcounts taken
+                    # above keep exactly the blocks the save still reads.
+                    # `has_pending_work()` already counts `_save_lease_blocks`,
+                    # so the engine keeps stepping until they are handed back.
+                    return False, None
```
yhl-amd, line 893: **[P1] Keep vLLM polling while an exact lease or final save is pending.** This path bypasses `_deferred_frees`, but `has_pending_push_work()` still returns only `bool(self._deferred_frees or self._releases_in_flight)`. vLLM's `EngineCore.step()` exits before `schedule()/execute_model()` once that hook is false. Therefore, when the last live request takes this branch, a frozen final save in `_save_tracker` may never be dispatched, and an already-dispatched save's worker completion cannot reach `update_connector_output()`; the touched blocks remain referenced until another request happens to wake the engine. The native scheduler's `has_pending_work()` counting `_save_lease_blocks` does not help because vLLM never calls it directly. Please include the wrapped scheduler's pending state (or at least exact leases/final saves) in `has_pending_push_work()` and cover the last-request/idle case. Upstream Mooncake's equivalent returns `bool(self._pinned_saves)` for this reason.

atom/plugin/vllm/kv_transfer/connector.py
```diff
@@ -804,6 +828,41 @@ def _apply_completions(self, completions) -> None:
...
+        including when two requests protect the same shared block -- which then
+        stays alive until the last of them reports.
+        """
+        pool = self._gpu_block_pool
+        if pool is None:
+            return
+        for block_ids in block_id_sets or ():
+            if not block_ids:
+                continue
+            ordered = sorted(block_ids, reverse=True)
```
yhl-amd, line 863: **[P2] Preserve token order instead of sorting pool IDs.** A `block_id` is a pool identity, not its position in the request. After prefix hits and free-queue reuse, a shared prefix block can have a larger ID than a tail block, so `sorted(..., reverse=True)` is not tail-first and may enqueue the valuable prefix for eviction before the tail. vLLM's normal path explicitly frees `reversed(req_blocks)`, where `req_blocks` is in allocation/token order; its Mooncake connector likewise retains the ordered block-id list. Please keep request/table order in the lease bookkeeping rather than reconstructing it numerically.

yhl-amd (commented):
I found two actionable issues in the exact-block lease path. The first can leave final saves and lease releases unpolled when the engine becomes idle; the second does not preserve vLLM's tail-first eviction order.

valarLip (approved):

PR #2286, opened 2026-09-18
[FlyDSL] [Gfx1250] mega moe stage1 gfx1250 no overlap

XingerZhu:
## Submission Checklist


junhaha666 (approved):

atom/model_ops/fused_moe/mori_v2_prepare_finalize.py
```diff
@@ -326,12 +381,129 @@ def init_mega_transport(
...
         mega._config.dispatch_wire,
+        mega._config.compact_plan,
         # The other half of the pair: logged together so a mismatch is readable.
         os.environ.get("AITER_FORCE_A8W4", "0"),
     )
     return mega
 
 
+def _decode_token_capacity(prefill_capacity: int) -> int:
+    """Per-rank token capacity for the decode transport, or 0 for "do not build".
```
JiaoliangYu, line 392: Derive compact transport shapes from the runner’s finalized capture plan
These helpers duplicate CUDA Graph shape calculation using Config.capture_sizes and speculative_config, but those are not the authoritative runtime values. ModelRunner filters unschedulable capture sizes using the actual token budget and drafter.mtp_k, and may capture multiple query-length buckets for DSpark. It then computes the final running_bs, max_seqlen_q, and running_tokens after DP synchronization.
Consequently, _decode_token_capacity() can size the transport from capture buckets that the runner later drops, while _warmup_decode_compact_plans() only warms bs * fixed_q_len * dp_size and can miss plans for DSpark’s smaller q buckets. If those plans cannot be JIT-compiled during graph capture, the affected graph shapes may fail to capture.
Please defer creation of the decode transport until the runner has finalized capture_sizes and q_buckets, pass the resulting maximum token count into the transport, and warm each compact plan from the actual ForwardContext during the runner’s existing pre-capture warmup. This would also remove the need for _spec_q_len() and avoid maintaining a second, incomplete implementation of the runner’s shape logic.

atom/model_ops/fused_moe/mori_v2_prepare_finalize.py
```diff
@@ -326,12 +381,129 @@ def init_mega_transport(
...
+
+def _warmup_decode_compact_plans(decode: Any, dp_size: int) -> None:
+    """Compile every captured decode bucket's compact plan while still eager.
+
+    The plan's tile alignment follows the recv bound's token bucket and aiter
+    refuses to JIT one inside a capture, so each rung of the ladder is run
+    through here first -- plus the unbounded case, which is what _recv_bound
+    returns when the bound would not shrink the arena.
+    """
+    warmup = getattr(decode, "warmup_compact_plan", None)
```
JiaoliangYu, line 433: no set for _warmup_compact_plan_

PR #2290, opened 2026-09-18
[DCP] Allow query replication (QREP) with MTP

yitingw1:
## 1. Motivation

Two DCP restrictions turn out to be first-cut caution rather than technical limits.

**QREP + speculative decode.** `qrep_unsupported_reason()` disabled query replication whenever `speculative_config` was set, suspecting a conflict with the `cprr` kernel used by MTP's `qlen>1` verify. The original comment says "not yet wired", not "found to break", and no test ever exercised the combination. The two are orthogonal: QREP only changes *how* `q_out` is produced (a load-time replicated `q_proj` instead of a runtime AllGather), while the verify mask keys off the KV side — `cprr` on dense MLA, the sparse indexer's per-token candidate exchange on sparse / DSA.

**bf16 sparse prefill persistent.** The fp8-only restriction on `mla_dcp_sparse_prefill_is_persistent()` traces to one commit ("Fix DSA DCP accuracy", 2026-08-05) that bundled it with two genuinely fp8-only guards (the q/kv quantization scales) while first extending this path from fp8-only to bf16. Nothing ties the *work metadata* to dtype — `get_mla_metadata_info_v1`/`get_mla_metadata_v1` take `dtype_q`/`dtype_kv` unconditionally, so the buffers were already built correctly for bf16 and simply discarded at the call site. The cost was a 2–4x padded query-head width on every bf16 sparse prefill.

## 2. Usage

No new flags. QREP (`--dcp-config '{"enable_query_replication": true}'`, default on) now survives `--method mtp/eagle3/dspark`; `--kv_cache_dtype bf16` now reaches persistent sparse prefill.

**Verifying QREP engaged** — the flag being `true` is not the same as QREP running. The positive log line (`... is on and active`) fires **once per rank with no layer number**; only the fallback line (`... layer N's q_proj was not built with qrep_tp_override ...`) is per-layer. So the check is the *absence* of fallback lines: a fully wired model shows one positive line per rank and no warnings.

### 3.1 QREP under speculative decode

| Aspect | Before | After |
|---|---|---|
| Config gate | `speculative_config is not None` → QREP off | removed; only `dcp<=1` and mxfp4-BMM remain |
| Per-layer decision | global config flag only | also requires `q_proj.weight.shape[0] == qrep_num_heads * qk_head_dim` |
| Draft models' `q_proj` | built plain; QREP unreachable and unsafe if reached | `qrep_tp_override` wired into `kimi_k3_dspark.py` and `eagle3_deepseek_mla.py` |
| A layer that never opted in | protected incidentally by the removed gate | falls back to AllGather, logged per layer |

**Key decision — gate on the actual weight width, not the config flag.** Removing the config gate alone would expose layers that build their own `q_proj` independently of the target. Reinterpreting a plain per-rank `q_proj` as the wide QREP layout either folds the wrong rows into one head group or raises, depending on shape divisibility. Reading the real width means such a layer degrades to AllGather instead. Every model wired today passes; it guards the next one that isn't — including GLM-5.3-Flash, whose `_ZeroRopePad` wrapper exposes no `.weight` at all and is deliberately left unwired.

```
                    q_proj
   QREP off: sharded tp    QREP on: sharded tp/W
     q [T,H/tp,D]            q [T,H*W/tp,D]
           │                        │
     AllGather Q over DCP      (skipped)
           └───────────┬────────────┘
                       ▼
          q_out [T,H*W/tp,D]   identical either way
                       ▼
                 mla_decode_fwd
       dense + q_len>1 + causal → cprr  (mask reads
       sparse            → q_len→1       g_kv_indptr,
       DSpark draft      → non-causal    cp_world/rank,
                                         never q_out)
```

### 3.2 Persistent mode for bf16 sparse prefill

| Aspect | Before | After |
|---|---|---|
| Predicate | `mla_dcp_sparse_prefill_is_persistent(kv_cache_dtype, …)`, `False` unless fp8 | dtype parameter removed |
| `use_work_meta` | `is_fp8 and (dcp<=1 or sparse_dcp_persistent)` | `dcp<=1 or sparse_dcp_persistent` |
| bf16 KV, dcp4 / dcp8 | 32 / 64 gathered heads, padded to **128**, non-persistent | 32 / 64, persistent |

The predicate stays separate from decode's rather than merged — this call site has its own width table, and an `assert` at the gate ties predicate and width so neither drifts alone.

## 4. Test Plan

Accuracy A/B on the same commit with only the flag differing, for each of: QREP + sparse MTP; QREP + the `cprr` path (a separate arm, because sparse MTP flattens `q_len` to 1 and never selects `cprr`); bf16 persistent, scored by **McNemar paired test** — at this setup's noise floor (63 of 1319 answers flip on a same-code rerun) an aggregate delta under ~1.2pp is not resolvable. Plus 8k1k TTFT for the bf16 change, and a control arm per wired draft (override removed) so a zero-fallback result is distinguishable from "those layers never reached the check".

### 5.1 QREP with speculative decode

GLM-5.2-FP8 sparse DSA, tp8/dcp8, fp8 KV, MTP=3, gsm8k nshot=20 — QREP off `0.9492/0.9492`, on `0.9469/0.9477`, within combined stderr (two further on-runs gave 0.9333/0.9310 and 0.9462/0.9462, the same spread, i.e. run-to-run noise at concurrency 64).

Kimi-K3 + DSpark, tp8/dcp8, fp8 KV, `num_speculative_tokens=7`, gsm8k 200 @ nshot=5 — **0.98 ± 0.0099**, accepted-token mean ≈3.69/7. K3's MLA layers are dense and the target verify is causal with `q_len>1`, so this run dispatches `cprr` (`mla_a8w8_qh32_qseqlen4_gqaratio32_lse_cprr_ps` in the log) with QREP active on every layer — the combination the removed gate blocked.

| Per-layer fallback | wiring | positive | fallback |
|---|---|---|---|
| K3 + DSpark | as-is | 8 | **0** |
| K3 + DSpark | override removed | 8 | **40** = 5 layers x 8 ranks, layers 93–97 |
| DSV3.2 + eagle3 (synthetic config, `--load_dummy`) | as-is | 8 | **0** |
| DSV3.2 + eagle3 | override removed | 8 | **8** = layer 61 |

### 5.2 Persistent mode for bf16 sparse prefill

GLM-5.2-FP8 sparse, tp8/dcp8, **bf16 KV**, gsm8k nshot=20 — before `0.9484/0.9469`, after `0.9447/0.9424`. McNemar: flexible 21/18 discordant 39 p=0.749; strict 23/18 discordant 41 p=0.533. No significant asymmetry, and 39–41 is **below** the 63/1319 same-code rerun floor. Non-DCP arm: `0.9416/0.9424` → `0.9416/0.9409`.

8k1k, `--request-rate inf`: median TTFT **43671 → 33156 ms (−24.1%)**, P99 −24.2%, median TPOT −12.5% — matching the mechanism (no longer padding 64 heads to 128). Single load point, not a performance characterisation.

### 5.3 Unit

`tests/test_dcp_merge_ops.py`, 212 passed. Every added test was checked to fail on revert: dropping the `and` at the call site reddens one case, forcing the width predicate true reddens three, re-adding a `speculative_config` parameter reddens the signature guard, removing either draft's override reddens exactly that draft's case, and adding an unclassified MLA model reddens the census that requires each MLA `q_proj` producer to be wired or documented-unwired.

## 6. Known Limitations

1. **eagle3 is construction-level only** — no Kimi-K2.6 weights, so that arm used a synthetic draft config with `--load_dummy`. DSpark covers the equivalent code path end to end.
2. **No throughput data for QREP itself** — the saved AllGather Q is not quantified against QREP's KV cost (~5% of the pool: 235016 → 221020 blocks on DeepSeek-R1 tp8/dcp8).
3. **bf16-persistent TTFT is one load point**, in a regime where queueing amplifies the difference.
4. **MXFP4 + TP4/DCP4 + MTP + QREP untested.**
5. **gfx942 stays unsupported for sparse DCP + MTP** — unchanged here; a measurement on MI300X had all 8 ranks take a GPU memory fault under real concurrency, so that `assert gfx == "gfx950"` is load-bearing.
6. **Two independent changes in one branch** — they share no code beyond `attention_mla.py` and can be split.

## Submission Checklist


ZhangLirong-amd (dismissed):

valarLip:
## Review — QREP by default under DCP, and bf16 sparse DCP prefill on the persistent path

Reviewed at the PR head in a detached worktree; `/app/ATOM` was never switched.

**The core mechanism holds up, and two of my starting concerns were actively refuted** — recording them so nobody re-spends on them. gfx950 does ship dedicated bf16/bf16 persistent lse kernels at config gqa 32/64 (`asm_mla_configs.hpp`: `mla_a16w16_qh32_...lse_ps.co`, `mla_a16w16_qh64_...lse_v3_ps.co`); the "bf16 silently miscomputes gqa=64" note is scoped to the *non-persistent* table only; and aiter requires `q_scale`/`kv_scale` only when Q is fp8, so `q_scale=None` on the bf16 path is safe. The `speculative_config` gate removal is mechanically sound too: `dcp_all_gather_query_heads` is a pure head-dim gather with no `q_len > 1` special-casing, so a QREP-produced `q_out` is layout-identical. Signature and import changes are clean repo-wide (no stale 3-arg callers, no plugin copy), and ruff/black are clean on the diff context.

So nothing below is "this computes the wrong answer today". The findings cluster on one statement:

> **Both of this PR's new decisions are carried by the wrong artifact. QREP eligibility is decided from a row count, which provably cannot distinguish the case it exists to reject at `tp == dcp` — the configuration the PR validates. And the persistence predicate, the thing that genuinely is shared, is now open-coded in four places, while the width table, the thing that genuinely differs, is a plain alias.**

A separate, smaller one, because it lands on people who did not opt in:

> **Deleting the speculative-decode arm flips a default for every existing MTP/eagle3/DSpark + DCP deployment. KV sizing does not account for `qrep_enabled`, and the only migration note lives inside one GLM-5.2 recipe.**

**Provenance:** **[verified]** means the deciding code was read at this head or the behavior was confirmed against aiter/torch on this box. **[reported]** means the shape matches the code but was not traced end to end.

---

# 1. QREP eligibility is decided by shape arithmetic, and the fallback is unsound in its inverse

**The width check cannot detect what it was written to detect** (`config.py:1631`) **[verified]**. At tp8/dcp8, `qrep_tp_override` yields `override_tp_size = tp // dcp = 1`, so `divide(output_size, tp_size)` is a no-op and the "widened" width **is** the full unsharded `num_heads_total * qk_head_dim`. `q_proj_is_qrep_widened` therefore cannot distinguish a QREP-widened q_proj from a never-sharded one at `tp == dcp` — the exact configuration this PR validates. It also says nothing about the rank ordering `_local_q_proj` relies on.

A future model handing `MLAAttention` a `ReplicatedLinear` or a full-width q_proj passes the check, QREP turns on, `_local_q_proj` row-slices at `dcp_rank * rows` on a weight never partitioned by effective TP (`make_row_view` only bounds-checks), and every rank but rank 0 reads the wrong heads — silently wrong attention. Catching exactly this class of model is the guard's stated job.

Provenance is exact where a shape product is not: record the override on the layer, or declare `MLAModules.qrep_widened` next to the existing `rope_is_zero_pad`.

**The per-layer fallback is one-directional** (`attention_mla.py:1281`) **[verified]**. `qrep_tp_override` widens q_proj off the same global flag `wants_qrep` reads, so a `False` width check does **not** restore AllGather — it reshapes an already-wide projection.

If a wired layer's width probe returns False (a wrapper exposing no `.weight`, a quant scheme that rebinds `weight` after `__init__`), `qrep_enabled` goes False and `_q_proj_and_k_up_proj`'s else branch runs

```python
self.q_proj(x, x_scale).view(-1, self.num_heads, self.qk_head_dim)
```

on a `[T, num_heads*dcp*qk_head_dim]` tensor. The element count stays divisible, so this reshapes into `dcp`× more token rows with **no size error**, and the bmm against `W_K` also succeeds — garbage `q` feeding `fused_qk_rope_concat_and_cache_mla` against a `k` of the original length. Same at `:2653`.

Under `wants_qrep` the safe form is an `assert`, not a fallback whose else branch never re-derives the width it just rejected.

# 2. The separation the diff defends in prose is an alias; the thing that is genuinely shared is copied four times

**`_MLA_DCP_SPARSE_PREFILL_WIDTHS_PERSISTENT = _MLA_DCP_KERNEL_WIDTHS`** (`attention_mla.py:258`) **[verified]** is a plain alias of decode's tuple. The diff's central new claim — that this call site keeps "its own" width table so it cannot read one sized for the other mode — is false for the persistent row, **which is the only row this PR newly activates**.

Every persistent DCP sparse prefill (now including the bf16 one unlocked here) reads decode's tuple under a second name. A width added or removed from `_MLA_DCP_KERNEL_WIDTHS` for a decode-kernel reason silently changes the gathered pad width of sparse prefill. And `test_sparse_prefill_persistent_table_would_allow_gqa64` asserts a property of decode's tuple *through* the alias, so it can never fail independently.

Two concrete mismatches hang off this:

**Work-buffer sizing is tied to decode's predicate, and the guard is on the wrong side** (`attentions/aiter_mla.py:570`) **[verified]**. The sparse-prefill work buffers are sized from `persistent_num_heads`, computed with decode's width function, decode's table and decode's predicate — nothing ties the allocation to `mla_dcp_sparse_prefill_num_heads` / `mla_dcp_sparse_prefill_is_persistent`, and this PR makes bf16 consume these buffers for the first time. The module-side `assert q.shape[1] == self.dcp_kernel_num_heads` guards only the rebuild; **the sizing has no guard at all**. At a gathered width where the two width functions disagree — decode's persistent branch returns the unrounded multiple of 16 (48/80/96/112) while sparse prefill rounds up to 64/128 — the work/reduce buffers are allocated for N heads while the kernel dispatches at M > N. Separately, the per-step fill at `aiter_mla.py:1811` passes `self.padded_num_attention_heads` (per-rank, e.g. 16) rather than the gathered width, so correctness depends on an indexer-owning layer rebuilding first in every step.

**The rebuild asserts against a width the caller chose from a different table** (`attention_mla.py:2214`) **[verified, latent]**. `_rebuild_sparse_dcp_persistent_metadata` asserts `q.shape[1] == self.dcp_kernel_num_heads` and passes that as the kernel nhead. Decode returns any multiple of 16 up to 128; sparse prefill snaps to (16, 32, 64, 128). A sparse-indexer MLA model with 96 query heads at tp8/dcp4 gathers 48 → prefill pads to 64 → the rebuild asserts and plans for 48: a hard `AssertionError` mid-prefill, or under `python -O` work descriptors planned for 48 heads while the kernel runs 64 (wrong partials, out-of-range reduce writes). Latent today — current sparse models are 128- or 64-head with power-of-two tp/dcp, so the gathered width is always in the tuple — but this PR newly routes bf16 through this path.

**And the predicate itself is now four copies** (`config.py:1666`) **[verified]**. With the fp8 gate dropped, `mla_dcp_sparse_prefill_is_persistent` is body-equivalent to `mla_dcp_decode_is_persistent(is_sparse=True, ...)`: decode's `dcp_world_size <= 1 or (is_sparse and not sparse_metadata_rebuild)` with `is_sparse=True` is the prefill guard verbatim, the only textual difference being a dead `None -> 1` page-size branch (`envs.ATOM_MLA_PAGE_SIZE` is `int(os.getenv(..., "1"))`, never None). The four copies:

```
config.py:1666            the sparse-prefill predicate
attention_mla.py:2028     the per-forward assert — now a character-for-character restatement,
                          so it can no longer catch the mismatch it was written for
attention_mla.py:816      the log gate
attention_mla.py:264      decode's
```

Adding a condition to one and not the others is silent: the width is chosen for one mode while the kernel runs the other. The genuinely separate thing is the width table, not the predicate — the diff has it exactly backwards.

# 3. Removing the speculative arm flips a default for deployments that did not opt in **[verified]**

`config.py:2088`. Every existing MTP/eagle3/DSpark + DCP deployment now defaults QREP to true, and nothing in KV cache sizing accounts for `qrep_enabled`.

A deployment tuned against the old auto-disable restarts on this commit: q_proj grows dcp-fold per rank and the KV pool loses blocks — the guide's own number is 235016 → 221020 on R1 tp8/dcp8, ~6%. A run pinned to a fixed `--max-model-len` / `--max-num-seqs` at `--gpu-memory-utilization 0.95` now fails the KV-capacity check at startup, or silently loses batch capacity, **with no message naming QREP**. The gating table simply has one fewer row, and there is no log distinguishing "QREP on because you asked" from "on because the auto-disable was deleted".

**Related, and the same kind of error as §1 — the remaining fp4 guard keys off an env var rather than the weight** (`config.py:1612`) **[verified]**. With the speculative gate removed, a q_proj carrying an mxfp4 (`per_1x32`) `weight_scale` now reaches `_local_q_proj` → `make_row_view`, which raises `NotImplementedError` for any per-output-channel scale that is not `per_1x128` or `per_Token`. A full-MXFP4-attention checkpoint (the `is_quark_static_mxfp4` path, attention weights/scales stored quantized on disk) served with `-dcp > 1` and `ATOM_USE_TRITON_MXFP4_BMM` unset: startup succeeds, then the **first** prefill dies inside `make_row_view` with `does not handle a per-output-channel scale for quant_type=per_1x32`.

I verified this is not reachable for the recipe's own `amd/GLM-5.2-MXFP4` — its `config.json` excludes all 79 `q_b_proj` from MXFP4, and the recipe's `ptpc_fp8` online quant makes them `per_Token`. But the recipe's new note argues safety from the env var alone, which is the wrong gate. Fold the layer's `quant_type` into the QREP decision, or handle `per_1x32` in `make_row_view`.

# 4. The documented verification procedure does not work

The new docs make reading two log lines the operator's way to confirm QREP. Both are unreliable.

**The flags are class attributes, never reset** (`attention_mla.py:752`) **[verified]**. `_qrep_enabled_logged` is once-per-process-per-class and shared with `SGLangATOMGLM52MLAAttention(MLAAttention)` and the vllm plugin layer, so a second engine init in one process — plugin re-init, multi-engine, a whole pytest session — emits **no positive line at all**, and the operator concludes QREP is off when it is on. `_qrep_fallback_logged_layers` is keyed on a bare `layer_num`; target/draft numbering does not collide today (`dspark_proposer.py:340` and `eagle_proposer.py:228` pass `layer_offset`), but the sglang eagle3 bridge already defaults `layer_offset` to 0, so the masking is one non-MLA-to-MLA change away. Key both on `(model/class identity, layer_num)`, or stop documenting log-absence as the check.

**Under the vllm plugin the decision is frozen before the real DCP size is known** (`attention_mla.py:744`) **[verified]**. `plugin/vllm/attention/layer_mla.py:253` configures head padding with vLLM's `dcp_size`, then `:258` sets `self.dcp_world_size = -1`, with the real value bound only at first forward (`:1018`/`:1364`). But `MLAAttention.__init__` already computed `wants_qrep` from ATOM's `get_dcp_world_size()` — often 1 — so QREP is permanently off and `qrep_num_heads = num_heads * 1` is stale while the widths next to it describe a different world size. A silent loss of the optimization with no log line, and it makes the new rule ("absence of any fallback line means QREP took effect on every layer") false under the plugin. The QREP decision is now width-dependent; it needs to be recomputed where the widths are.

# 5. The dtype axis was deleted from the only test that covered it, in the commit that activates it **[verified]**

`tests/test_dcp_merge_ops.py:1392`. `test_sparse_prefill_width_matches_its_own_persistence_predicate` lost `@pytest.mark.parametrize("dtype", ["bf16", "fp8"])`, so "non-persistent sparse prefill must not dispatch gqa=64" no longer runs across dtypes. The replacement, `test_sparse_prefill_persistence_does_not_look_at_dtype`, only inspects `inspect.signature` — it asserts nothing about the newly enabled behavior. A later change that re-breaks bf16 sparse DCP prefill (wrong width, work-meta whose buffers describe another nhead — i.e. §2) leaves the whole CPU gate green.

**The census test cannot deliver its advertised guarantee** (`:1124`) **[verified]**. `test_every_mla_model_has_an_explicit_qrep_decision` claims "a new MLA model has to land in one list or the other". It is `_ATOM_MODELS.glob("*.py")` — non-recursive, so an MLA model under `atom/models/<family>/`, the direction this tree is already moving, is invisible — filtered on the literal substring `"MLAModules("`, so an aliased import or a shared factory is missed and a mention in a comment yields a false failure. And `_qrep_override_names` collects any name ever bound to `qrep_tp_override(...)` anywhere in the module, then accepts `**that_name` at any q_proj call in any class on any branch, so a second class rebinding the name breaks the property with the test green. The census also only checks list membership, so **it passes on a full revert of the override wiring**.

This is also the repo's explicitly banned pattern: a test may not grep the code it tests.

# 6. Docs-match-code: the second headline change is documented nowhere, and the guide contradicts the recipe **[verified]**

`docs/context_parallel_guide.md:560`. The entire "Sparse DCP persistent attention and gqa=64" section and the constraints row at `:693` still scope persistent gqa=64 to fp8 KV, so bf16 sparse prefill persistent — this PR's second headline change — appears nowhere in the guide. Line 693 still reads `| gathered head width 64 + **fp8 KV** |`, and `:576-577` still assert "Unsupported paths ... remain non-persistent and round a gathered 64 up to **128**", which is now false for bf16 at dcp8 (64) and dcp4 (32). A reader debugging a bf16 sparse DCP run will conclude the path is non-persistent at width 128 when it is persistent at 64.

`:628`. The rewritten support matrix and the recipe edited in this same PR disagree on the validated speculative depth for sparse/DSA + DCP + QREP. The matrix pins MTP under DCP at `num_speculative_tokens` 1/2/3 and records sparse as validated at exactly 3, dcp8, fp8 KV; `recipes/Agentic-GLM-5.2.md:262-267` prescribes `MTP_K=5` (C1-C8) and `MTP_K=4` (C10-C12) on GLM-5.2 sparse, with the TP4/DCP4 block at MTP_K=3-4 and QREP now on by default. There is no runtime bound on `num_speculative_tokens` on the DCP side, so a reader following the recipe runs a point outside both the matrix's 1-3 range and every arm this PR validated — and it fails, if at all, as accuracy drift rather than rejection.

# 7. Placement: three dispatch predicates moved into `config.py` for a stated testing reason **[verified]**

`config.py:1618`. Three MLA kernel-dispatch / layer-construction predicates now live in `atom/config.py` (2532 lines), one of them moved there from `attention_mla.py`, and the docstrings give the reason outright:

> *"Dependency-free ... so it stays importable, and testable, without triton/aiter"*
> *"Only depends on atom.utils.envs ... so it stays importable, and testable, from here."*

That is test-driven placement, against `/app/CLAUDE.md`'s 工作规范 (*"代码组织结构的边界感，不同功能的代码不应该都写在一起"*) and coding-style's 800-line soft ceiling. `atom/distributed/dcp_utils.py` is 90 lines, already CPU-importable (its aiter import is function-local), already owns this predicate's input `dcp_persistent_supported()` and its sibling `dcp_prefill_merge_bf16_ok()`, and is already imported in the same block this diff edits.

As placed, the dependency arrow inverts (`dcp_utils` imports `config`), so these can never move back without a cycle — and the width table the predicate selects stays in `attention_mla.py`, splitting the coupled pair across two layers. That split is what §2 is.

# 8. Smaller **[verified]**

- The new `use_work_meta` comment is vacuous rather than wrong: it claims the non-DCP call site "is fixed the same way", but `use_decode_kernel = is_fp8 or return_lse` and `return_lse` is only ever passed by DCP callers, so that arm is unreachable for bf16.
- `make_row_view`'s `per_1x128` 128-alignment assert is newly reachable for drafts with an odd local head count; no supported shape hits it today.
- The "Source Files" doc table was not updated for the moved functions.
- The in-code comment says the guard protects "a future model", while `_QREP_UNWIRED_MODELS` already lists `glm5_next.py`.
- Function-local `import inspect` in two new tests, against the repo's no-local-import rule.
- The fallback log rebuilds a frozenset per call, O(n²) over layers.
- `qrep_enabled_for_layer` is a one-line `and` wrapper whose test cannot actually pin the call site.

---

Of the above, §1 is the one I would hold on: it is the only finding whose failure mode is silently wrong attention output, and the fix (provenance instead of a shape product) is smaller than the check it replaces. §2 is the one I would hold on for design reasons rather than severity — the predicate and the width table have been separated in exactly the wrong direction, and §5's deleted dtype axis means the resulting mismatch has no test that would catch it. §3 deserves a line in the release notes regardless of whether anything else changes, because it lands on operators who did not opt in and gives them no message naming the cause.

valarLip (approved):

PR #2292, opened 2026-09-18
[feat] Supports loading Quark NVFP4 and online quantization of NVFP4 to MXFP4 (minimax-m3)

haoyangli0109:
1.Support loading Quark NVFP4 minimaxM3 and nvidia NVFP4 minimaxM3
2.Support online NVFP4-to-MXFP4 quantization

[amd/MiniMax-M3-NVFP4](https://huggingface.co/amd/MiniMax-M3-NVFP4/blob/main/model.safetensors.index.json)
```
cd /home/haoyanli/lhy_atom/ATOM

export ATOM_FORCE_ATTN_TRITON=1
export AITER_QUICK_REDUCE_CAST_BF16_TO_FP16=0
export ATOM_ONLINE_QUANT_STREAMING=1
export PYTHONUNBUFFERED=1

  --model /shareddata/haoyanli/amd/MiniMax-M3-NVFP4 \
  --host 127.0.0.1 \
  --server-port 18101 \
  --port 19101 \
  --tensor-parallel-size 4 \
  --kv-cache-dtype fp8 \
  --block-size 128 \
  --max-model-len 32768 \
  --max-num-seqs 128 \
  --max-num-batched-tokens 32768 \
  --enforce-eager \
  --level 0 \
  --cudagraph-mode NONE \
  --no-enable-prefix-caching \
  --trust-remote-code \
  --online-quant-config '{"global_quant_config":"mxfp4","exclude_layer":["lm_head","model.layers.*.self_attn","model.layers.*.block_sparse_moe.gate"]}'
  
  
MODEL_PATH=/shareddata/haoyanli/amd/MiniMax-M3-NVFP4
OUTPUT_PATH=/home/haoyanli/lhy_atom/runs/gsm8k-m3-online-ci-chat-exclude-attn-20260918
LOG_PATH=/home/haoyanli/lhy_atom/runs/gsm8k-m3-online-ci-chat-exclude-attn-20260918.log

lm_eval \
  --model local-chat-completions \
  --apply_chat_template \
  --fewshot_as_multiturn \
  --model_args "model=${MODEL_PATH},base_url=http://localhost:18101/v1/chat/completions,num_concurrent=32,max_retries=3,max_gen_toks=16384,tokenized_requests=False,trust_remote_code=True" \
  --tasks gsm8k \
  --num_fewshot 5 \
  --output_path "${OUTPUT_PATH}" \
  --log_samples \
  2>&1 | tee "${LOG_PATH}"
  

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9371|±  |0.0067|
|     |       |strict-match    |     5|exact_match|↑  |0.9378|±  |0.0067|
```
[nvidia/MiniMax-M3-NVFP4](https://huggingface.co/nvidia/MiniMax-M3-NVFP4/raw/main/config.json)
```
export ATOM_FORCE_ATTN_TRITON=1
export PYTHONUNBUFFERED=1

  --model /shareddata/haoyanli/nvidia/MiniMax-M3-NVFP4 \
  --host 127.0.0.1 \
  --server-port 18101 \
  --port 19101 \
  --tensor-parallel-size 8 \
  --kv-cache-dtype fp8 \
  --block-size 128 \
  --gpu-memory-utilization 0.8 \
  --max-model-len 32768 \
  --max-num-seqs 128 \
  --max-num-batched-tokens 32768 \
  --no-enable-prefix-caching \
  --trust-remote-code \
  --online-quant-config '{"global_quant_config":"mxfp4","layer_quant_config":{"model.layers.*.self_attn*":"ptpc_fp8","model.layers.*.mlp*":"ptpc_fp8","model.layers.*.block_sparse_moe.shared_experts*":"ptpc_fp8"},"exclude_layer":["lm_head","model.layers.*.block_sparse_moe.gate"]}'

cd /home/haoyanli/lhy_atom/ATOM

MODEL_PATH=/shareddata/haoyanli/nvidia/MiniMax-M3-NVFP4
OUTPUT_PATH=/home/haoyanli/lhy_atom/runs/gsm8k-m3-nvidia-nvfp4-to-mxfp4-tp8
LOG_PATH=/home/haoyanli/lhy_atom/runs/gsm8k-m3-nvidia-nvfp4-to-mxfp4-tp8.log

lm_eval \
  --model local-chat-completions \
  --apply_chat_template \
  --fewshot_as_multiturn \
  --model_args "model=${MODEL_PATH},base_url=http://127.0.0.1:18101/v1/chat/completions,num_concurrent=32,max_retries=3,max_gen_toks=16384,tokenized_requests=False,trust_remote_code=True" \
  --tasks gsm8k \
  --num_fewshot 5 \
  --batch_size 65 \
  --output_path "${OUTPUT_PATH}" \
  --log_samples \
  2>&1 | tee "${LOG_PATH}"

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9462|±  |0.0062|
|     |       |strict-match    |     5|exact_match|↑  |0.9462|±  |0.0062|
```


valarLip:
## Review — NVFP4 source format and online NVFP4→MXFP4 conversion

Reviewed at the PR head. A lot of this is sound — I checked and am **not** flagging: `dict.fromkeys` over `(pattern, LayerQuantConfig)` tuples is safe (the dataclass is `frozen=True`, hence hashable); the MXFP4 target buffers are zero-filled by `create_weights` and `dynamic_mxfp4_quant` emits byte 0 for an all-zero block, so the zero-padded regions are not a NaN source; `uint8 >> 4` / `& 0xF`, the fp8 handling, `_EXPERT_LAYER_RE`, and the `weight_scale * weight_scale_2` multiply direction are all correct.

The headline concern is where the damage lands:

> **Two of the three most serious defects are not on the NVFP4 path at all.** One changes quantization resolution for six model families; one makes every static-FP8 MiniMax-M3 MoE checkpoint run with activation scale 1.0.

**Provenance:** **[verified]** means I read the deciding code myself. **[reported]** means the shape matches the code but I did not trace it end to end.

---

# 1. `apply_exclude_name_mapping` lost its no-op guarantee and now rewrites pattern specs **[verified]**

```diff
-        if not mapping or not self.exclude_layers:
+        if not mapping:
             return
```

Previously this was a **guaranteed no-op** for any config without `exclude_layers`. It now runs for every model that declares `quant_exclude_name_mapping` — `deepseek_v2`/GLM-5, `kimi_k3`, `kimi_k25`, `mistral3`, `qwen3_5`, `minimax_m3` — and it no longer only touches excludes:

```python
for pattern, spec in self.layer_pattern_specs:
    mapped_pattern_specs.append((pattern, spec))          # original
    mapped_pattern = _map_name(pattern)
    if mapped_pattern != pattern:
        mapped_pattern_specs.append((mapped_pattern, spec))   # twin, inserted right after
```

`get_layer_quant_config` is first-match-wins (`config.py:412-418`), so a mapped twin of pattern `p1` is now consulted **before** `p2`. A layer that previously resolved via `p2` can silently pick up `p1`'s `quant_type`/`quant_dtype` — wrong quantization for that layer, on six model families that have nothing to do with NVFP4.

The same block repeats for `online_layer_pattern_specs`.

Two asks:

- Restore an early-out, or scope the pattern-spec rewrite to the case that needs it. Whatever the NVFP4 path requires here, it should not be paid for by every model that happens to declare a name mapping.
- Per CLAUDE.md's Name-matches-function rule: a function that now rewrites `layer_pattern_specs` and `online_layer_pattern_specs` should not be called `apply_exclude_name_mapping`.

# 2. The new `input_scale` alias misroutes non-NVFP4 MiniMax-M3 checkpoints **[reported]**

`make_minimax_m3_expert_params_mapping` gains an `input_scale` → `input_scale_2` alias. `loading_core.py:181` sorts expert prefixes longest-first, so `experts.0.w1.input_scale` is now claimed by the new key instead of falling through to the generic `experts.0.w1.` prefix.

It rewrites to `experts.w13_input_scale_2`, which `Fp8MoEMethod` (`moe.py:1498`) does not register, so `weight_dispatch.py:226` returns "claimed, skip silently". `w13_input_scale` stays at its `torch.ones` init and `moe.py:1520` takes `.max()` of that — **every static-FP8 MiniMax-M3 MoE checkpoint now runs with activation scale 1.0**, silently, with no warning, on a path unrelated to this feature.

The alias also looks redundant: the generic prefix entry already produces the right name. If that is right, deleting the alias is the whole fix.

I did not trace this one end to end, but it is the finding I would check first — the blast radius is larger than the feature being added.

# 3. The w2 all-gather pads at the wrong edge, then re-splits at the padded stride **[verified]**

```python
w2_full = tp_group.all_gather(w2_local, dim=1)                  # [r0's I_local | r1's I_local | ...]
target_cols = self.w2_weight.shape[2] * 2 * self.tp_size        # = I_pad * tp
w2_full = torch.nn.functional.pad(w2_full, (0, target_cols - w2_full.shape[1]))
w2_q, w2_s = _quant_weight(w2_full)
...
local_q_cols = w2_q.shape[1] // self.tp_size                    # = I_pad
w2_q = w2_q.narrow(1, self.tp_rank * local_q_cols, local_q_cols).contiguous()
```

The gathered layout is `[r0's I_local | r1's I_local | … | (I_pad−I_local)·tp zeros]`, but the re-split slices at multiples of `I_pad`. **Rank 0 takes its own `I_local` columns plus the first `I_pad − I_local` columns of rank 1**; rank 1 takes a shifted window; and so on. No shape error — silently wrong `down_proj` weights *and* scales.

`need_gather_w2` is true exactly when `intermediate_size_per_partition % 32 != 0`, which is exactly when the pad fires. So this triggers precisely in the case the padding was added to handle.

The correct form is to pad per-rank region, or — cheaper by a factor of `tp_size` — narrow before quantizing rather than after.

# 4. The NVFP4 global-scale route reached 4 of 12 `weight_loader` overrides **[verified]**

An AST pass over `linear.py`:

```
ROUTED   ColumnParallelLinear         MergedColumnParallelLinear
         RowParallelLinear            MergedReplicatedLinear

 ---     QKVParallelLinear            MinimaxM3QKVParallelLinearWithIndexer
         QKVZBAParallelLinear         QKVZParallelLinear
         BAParallelLinear             QKVGParallelLinear
         LinearBase                   ReplicatedLinear
```

The one that matters is **`QKVParallelLinear`** — fused QKV is most of the model zoo. A Quark two-stage NVFP4 checkpoint sets the global spec to NVFP4, so `qkv_proj` gets a `weight_scale_2` of shape `(3,)` (5 for the M3 indexer variant). Loading `...q_proj.weight_scale_2` reaches `QKVParallelLinear.weight_loader`, which runs `param_data.narrow(0, num_heads*head_size, shard_size)` on a 3-element tensor — `RuntimeError: start+length exceeds dimension size` at load time.

Routing it to the new helper as-is would not fix it either: `_load_nvfp4_global_scale` requires an int shard id and raises `ValueError` on `'q'`/`'k'`/`'v'`. The helper needs a named-shard branch before the sweep can be completed.

Only `MergedColumnParallelLinear` is covered by the new tests.

---

# Also reported, not independently traced **[reported]**

- **`quant_mxfp4_dynamic` returns a column-major, non-contiguous block scale** and never calls `.contiguous()`, unlike `quant_mxfp4_online_even` — the helper whose docstring calls itself the shared entry point for both the Linear and MoE online paths. It also uses a different kernel (Triton vs HIP), a different rounding rule, and fp32 rather than bf16 input. `linear.py:964` stores that transposed tensor straight into `self.weight_scale`; it is normalized only if `e8m0_shuffle` runs, which is skipped when `params_dtype == fp4x2` and `use_fp4_non_shuffle_triton_gemm()` is on. Both new tests monkeypatch `quant_mxfp4_dynamic`, so neither difference is exercised. The rounding divergence also means NVFP4→MXFP4 accuracy is not comparable to the quark MXFP4 baseline.
- **`input_scale_2` is registered on every NVFP4 Linear but no name mapping ever produces that key for linears** (only the MoE path got the alias), so it is never written — yet it is counted in `_stream_expected_numel`. With `ATOM_ONLINE_QUANT_STREAMING=1` (the mode this PR's own repro uses), `_stream_loaded_numel` can never reach it, `_submit_finalize` never runs, the new streaming branch is dead, and `replay_stragglers_and_report` logs *"the model is almost certainly wrong"* on a healthy load — burying real warnings.
- **Group-size validation gap**: construction accepts `K % 16 == 0` (NVFP4), but `quant_mxfp4_dynamic` raises unless `K % 32 == 0`, and the Linear path does no padding. A row-parallel `down_proj` whose per-TP `K` lands on an odd multiple of 16 constructs fine, loads the whole checkpoint, then dies in post-load. `FusedMoE._online_quant` handles exactly this by `F.pad`-ing, so the two converters disagree about which shapes are loadable.
- **`_online_quant` writes `self.params_dtype` but never updates `self.layer_quant_config`**, leaving the layer's source-format state split across two fields that disagree. `FusedMoE.weight_loader` derives `is_nvfp4` from `layer_quant_config` (`moe.py:4880`), which still says nvfp4 after conversion — so any post-load loader call (RLHF/rollout weight sync, EPLB reload, plugin reload) hits `raise RuntimeError('Merged NVFP4 expert loading is not supported')` on a layer that now holds MXFP4. The `params_dtype` write on the row-local batched early-return path also silently changes state for every existing **non**-NVFP4 online-quant MoE layer.
- **The four NVFP4 global-scale parameters are allocated with `torch.empty`** where every sibling MoE method uses `zeros`/`ones` for scales, and nothing verifies a global scale actually arrived before multiplying a weight by it. A slot whose `weight_scale_2` never shows up — an EPLB redundant replica, a fused shared-expert slot, a differently-named tensor — is silently dropped by `weight_dispatch.py:226` and then multiplied by uninitialized float32.
- **An NVFP4 layer excluded by the online-quant config survives load as NVFP4 and only fails at the first forward** (`'direct NVFP4 inference is intentionally blocked'`). The server comes up, `/v1/models` answers, `rocm-smi` shows VRAM. The MoE mirror instead aborts the load with a message blaming call ordering for what is a legal exclusion. Both are knowable at config-parse time.
- **`should_stream_online_quant`'s new NVFP4 branch ignores the `source_quant_type` it was handed** and re-resolves with `get_layer_quant_config(prefix)` *without* `check_children=True`, while `FusedMoE` resolves its own spec *with* it (`moe.py:3343`). For a Quark NVFP4 MoE whose per-expert specs live on children, streaming is silently disabled and both the packed NVFP4 tensors and the MXFP4 target are resident at once. It also compares `QuantType` with `==` rather than `.value ==`, which the rest of this diff is careful to do — the hazard documented at `linear.py:1000`.
- **`_is_moe_layer` now gives `mlp_layer_types` unconditional precedence over `moe_layer_freq`** and hard-raises on a length mismatch or an unrecognized literal. For an existing config carrying both keys the source of truth silently switches; for a VL config whose `num_hidden_layers` is the wrapper's count, startup dies on a checkpoint that worked before. Falling back to `moe_layer_freq` would be safer than aborting.
- **`Nvfp4MoEMethod.create_weights` registers `w13_bias`/`w2_bias`, but `_online_quant` snapshots only weights and scales** before calling the target `create_weights`, which re-registers biases as fresh zeros — and the split-per-expert loader has no bias branch to fill them in the first place. A `has_bias` MoE runs with zero expert bias, no diagnostic.

# Duplication **[verified in part]**

This adds a second copy of infrastructure that already exists: a second 16-entry E2M1 LUT and nibble-unpack beside `quant_v4.py`'s `_FP4_LOOKUP`/`dequant_fp4_e2m1` (and aiter's `mxfp4_to_f32`), a second MXFP4 online quantizer beside `quant_mxfp4_online_even`, and `_NVFP4_BLOCK_SIZE = 16` duplicating the `NVFP4_GROUP_SIZE` added in this same PR.

**The two E2M1 tables already disagree** — index 8 is `0.0` in one and `-0.0` in the other, and their storage-dtype asserts differ. A fix to nibble order or signed-zero handling lands in one copy and the two paths decode the same bytes differently, with nothing comparing them.

NVFP4 is also routed *around* `dequant_weight_online` / `quant_weight_online` — the two helpers whose docstrings declare them the single shared entry point — via seven `source_is_nvfp4` forks. An MXFP4 numerics fix in `quant_weight_online` silently does not reach NVFP4-sourced layers, and the next source format repeats the same seven-site fork.

# Tests and docs **[verified]**

All new NVFP4 coverage is behind `pytest.importorskip('aiter')`, and `.github/scripts/run_unit_tests.sh` states its scope is *"CPU torch, no GPU, no aiter"* — so `Nvfp4MoEMethod`, `_load_nvfp4_global_scale`, the online NVFP4→MXFP4 path and `dequantize_nvfp4` ship with **zero enforced coverage**. `tests/aiter_stub.py` already exists for exactly this.

Docs not updated, against CLAUDE.md's Docs-match-code checklist: `docs/model_ops_guide.md`'s MoE quant dispatch table and `moe.py` module index (no `Nvfp4MoEMethod`); `docs/online_quantization_guide.md`'s supported-source table (no modelopt/NVFP4, and no note that NVFP4 sources reject every target but mxfp4); `docs/configuration_guide.md` (`quant_dtype` still documented as `torch.dtype` only, `quant_method` list has no `'modelopt'`). `atom/config.py:350`'s comment also still omits ModelOpt.

---

Of the above I would hold on **1** and **2**: both change behaviour for models that have nothing to do with NVFP4, and both are silent. **3** is the one I would fix next — it is wrong on the feature's own main path, in exactly the shape the padding was added for. **4** means the feature does not load at all on fused-QKV models, which is at least loud.

valarLip:
## Review (second pass) — NVFP4 source format and online NVFP4→MXFP4 conversion

Re-reviewed at `71429813a` (base `e4f5ff2d7`) in an isolated worktree. This is a second pass over a newer head than my [earlier comment](https://github.com/ROCm/ATOM/pull/2292#issuecomment-5749841293); I am not repeating items from it. Two things below (§5 on the shard map, §6 on `check_need_allgather`) are the same axis as items 4 and 3 there, and I flag them again only because the current head resolves them in a way that moves the failure rather than removing it.

Two candidates from this pass were **refuted** and dropped, so they do not get re-raised: the `weight_scale` uint8 view-flip in `dequantize_nvfp4` is correct, and the `torch.empty` vs zero-fill concern for `Nvfp4MoEMethod.create_weights` does not apply — NVFP4 allocates unpadded buffers, so there is no unwritten region.

The findings have two centres of gravity, and they are worth stating separately because they call for different fixes.

> **NVFP4 is represented as a `str` in a field whose every other value is a `torch.dtype`. `str == torch.dtype` is silently False and never an error, so six existing dispatch sites take the wrong branch with no diagnostic — this is a class of defect, not an instance of one.**

> **The MoE path has no equivalent of the Linear path's `_validate_nvfp4_global_scales`, and the scalar it would have validated sits in *both* of the loader's warning-suppression lists. This PR made `weight_scale_2` load-bearing; the loader is configured not to report its absence.**

**Provenance:** **[verified]** means the deciding code was read at this head or the command was run. **[reported]** means the shape matches the code but was not traced end to end.

---

# 1. `quant_dtype` now holds a string, and the mismatch is unobservable **[verified]**

`quant_spec.py:63` defines `NVFP4_DTYPE = "nvfp4"` and stores it in `LayerQuantConfig.quant_dtype` — a field whose every other value is a `torch.dtype`, and whose accessor in `config.py` is annotated `-> torch.dtype`. A `str == torch.dtype` comparison is False, silently, forever.

Sites that therefore mis-dispatch today, none of which raises:

- `comm_fused_moe.py:59` — `quant_dtype != dtypes.fp4x2` disables the comm-fused MoE backend for an NVFP4 source whose *target* is mxfp4
- `deepseek_v2.py:2602` — the fp4x2 attention weight-loading branch
- `deepseek_v2.py:279 / 402 / 3036+` — the fused input-norm-quant branches
- `kimi_k3.py:175-177`
- `attention_mla.py:487`

The PR escapes only the last one, and only because it had to patch `self.layer_quant_config = online_layer_quant_config` into `linear.py:1042` to do it.

Compounding it: NVFP4 is recorded as `QuantType.per_1x32` while its real block size is 16, and the rest of the tree reads `per_1x32` as "MXFP4, 32-element blocks" — which is exactly what §6 turns into a runtime assertion. And the repo already has a rival vocabulary for this concept: `model_ops/fused_moe/config.py:187` asserts `quant_dtype in {"nvfp4", "mxfp4", ...}`.

Each of the six sites above is a grep the next NVFP4-adjacent PR has to redo by hand, with no way to know it was missed. A `QuantFormat` enum — or a separate `source_format` field, leaving `quant_dtype` a true `torch.dtype` — makes every one of them either match or raise.

# 2. The MoE global scales have no validator, and the loader is configured not to warn about them **[verified]**

`moe.py:1195`. The Linear path has `_validate_nvfp4_global_scales`. The MoE path has no equivalent: `w13_weight_scale_2` / `w2_weight_scale_2` are zero-initialized, `_load_nvfp4_global_scale` records no arrival flags, and `_online_quant` consumes them unchecked.

What makes this more than a missing assert is where `weight_scale_2` already appears:

```
loading_core.py:420   suppressed_suffixes   ← contains weight_scale_2
loading_core.py:450   benign_substrings     ← contains weight_scale_2
```

Both diagnostics are switched off for exactly the tensor this PR just made load-bearing. Any name-mapping gap for `...experts.N.w1.weight_scale_2` — a different exporter naming, a `weights_mapping` rule, an expert-mapping entry that does not cover the suffix — leaves the scalar at `0.0`. `dequantize_nvfp4` then multiplies every block scale by zero, the expert dequantizes to exactly zero, and the MXFP4 target is all-zero. No exception, no warning on either path. The Linear path raises `incomplete NVFP4 global weight scales` for the identical failure.

Worth fixing on both sides at once: the Linear validator has the same blind spot for an explicitly-zero scale, because it tests `~torch.isfinite(values)` and `0.0` is finite. A NaN sentinel initializer plus one `isfinite(v) & (v > 0)` covers Linear and MoE together.

# 3. `ModelOptParser` mishandles two shapes, and the new docs claim otherwise **[verified]**

**A plain `quant_algo: "NVFP4"` export is classified as MXFP4** (`quant_spec.py:426`). `parse` special-cases only `MIXED_PRECISION`; everything else falls through to `GenericParser`, whose `_DTYPE_PATTERNS` substring-matches `fp4|float4|mxfp4` *inside* `"nvfp4"` and yields `dtypes.fp4x2` + `per_1x32`. This path is newly reachable because `config.py:366` adds `"modelopt"` to the online-quant-eligible list in this very PR. Consequence: `weight_scale` is allocated `(N, ceil(K/32))` e8m0 while the checkpoint ships `(N, K/16)` float8_e4m3fn, and `weight_scale_2` is registered as `None` so the global scale has nowhere to go — either a shape error deep into the load, or group-32/E8M0 semantics applied to group-16/E4M3 bytes. If the config also contains `"fp8"` anywhere (e.g. `kv_cache_quant_algo: "FP8"`) it is misclassified as fp8 instead. Wrong either way.

**Fused-projection conflicts are not reconciled** (`quant_spec.py:446`). The parser emits one pattern spec per *source* projection and reconciles only for experts (`expert_specs.setdefault` + `previous != spec`). q/k/v and gate/up are all remapped onto the same fused pattern by `config.py:_remap_layer_name`, and `get_layer_quant_config` returns the **first** match. A genuine MIXED_PRECISION checkpoint with `q_proj=NVFP4` and `k_proj=MXFP8` builds the whole fused `qkv_proj` in whichever spec comes first in `quantized_layers` iteration order; the sibling's packed `(N, K/2)` uint8 weight is then loaded into an `(N, K)` fp8 parameter. Same for one projection in `exclude_modules` while its siblings are not — `_is_excluded` de-quantizes the entire fused layer to bf16 while the checkpoint holds packed uint8.

This one needs a correction in the PR as well as a fix: `docs/online_quantization_guide.md:310`, added here, states these cases are *"rejected during model setup"*. No such code exists. And the new `test_packed_conflict_validation_is_modelopt_mixed_only` asserts the **absence** of the check as though it were the intent, so the documentation and the test currently disagree with each other and the test wins.

# 4. Quark NVFP4 makes `input_tensors` a list, and the activation-dtype probe swallows the result **[verified]**

`moe.py:3513`. `FusedMoE.__init__` derives `a_quant_dtype` via

```python
config.quantization_config.get("global_quant_config", "").get("input_tensors", "").get("dtype", "")
```

inside `except AttributeError`. Making Quark NVFP4 `input_tensors` a two-element **list** is the premise of `_quark_weight_spec` in this PR, so `list.get` raises and is swallowed. `a_quant_dtype = None` → `MoEActivationQuant.from_model_config(None)` returns BF16 → `moe.py:1808` falls to `mxfp4_w4a16_moe_quant_config` and `use_a4w4` at `moe.py:2278` is False.

So an NVFP4→MXFP4 converted MoE silently runs the a16wfp4 kernels while every dict-shaped MXFP4 Quark recipe runs w4a8/a4w4. ModelOpt NVFP4 hits the same swallow, having no `global_quant_config` key at all. Dispatch and throughput rather than wrong numbers — but nothing anywhere says it happened.

# 5. The global-scale loader override is installed on every Linear; the shard map exists on two **[verified]**

Following on from item 4 of the earlier review. `linear.py:773` now unconditionally overrides `weight_scale_2.weight_loader = self._load_nvfp4_global_scale` for **every** `LinearBase` subclass, but the string-shard-id path resolves through `_nvfp4_global_scale_shard_map`, a ClassVar defined only on `QKVParallelLinear` and `MinimaxM3QKVParallelLinearWithIndexer`.

`QKVZBAParallelLinear` (Qwen3-Next / Qwen3.5 in-proj, `linear.py:1818`, asserting shard ids `['qkvz','ba','qkv','z','b','a']`) therefore gets `shard_map = {}`. `qwen3_next.py:1000` maps `in_proj_qkvz -> ('in_proj_qkvzba','qkvz')` and `WeightDispatcher._dispatch_packed` forwards the string verbatim, so `...in_proj_qkvz.weight_scale_2` raises `does not support NVFP4 global scale shard id 'qkvz'; supported ids: none`. Same exposure for `QKVZParallelLinear`, `BAParallelLinear`, `QKVGParallelLinear`, `MergedReplicatedLinear`.

This one is loud rather than silent, so it is not a correctness risk — but it means NVFP4 support is narrower than the PR reads, with no default in the base class and no test covering a third subclass. Either give `LinearBase` a default map keyed on `output_sizes`, or install the override only where a map exists.

**Related, and silent:** `_validate_nvfp4_global_scales` cannot pass under `--load_dummy` (`linear.py:826`). The `_nvfp4_global_scale_loaded` flags flip only on real `weight_loader` arrivals, but `loading_core.py:289` installs a `lambda _: False` wanted-filter and `initialize_dummy_weights` writes `param.data` directly — no loader ever runs. NVFP4 + online quant + `--load_dummy` (startup, profiling, CI smoke) raises on every merged Linear. The precedent is already in-tree: `glm5_next.py:526-532` guards the analogous check with `if not getattr(get_current_atom_config(), "load_dummy", None):`.

# 6. `source_is_nvfp4` returns before the `% 32` rule that existed for exactly this reason **[verified]**

Following on from item 3 of the earlier review, which was about the w2 all-gather padding edge; the current head sidesteps it with an early return, and that moves the failure rather than removing it.

`moe.py:3925` places `if source_is_nvfp4: return False` **before** the pre-existing

```python
elif online_quant_type == QuantType.per_1x32:
    need_gather_w2 = w2_in % 32 != 0
```

— the rule that exists precisely so MXFP4's 32-wide blocks line up with offline quantization. The Linear path added the compensating rejection (`linear.py:913`, K%32). `Nvfp4MoEMethod.create_weights` validates only `% 16`, and on neither axis 32.

A TP config with `intermediate_size_per_partition % 16 == 0` but `% 32 == 16` — for instance I=1408 at TP8 → 176 — passes construction, loads the entire checkpoint, and then dies inside aiter on `quant_mxfp4_hip`'s `cols % group_size == 0` (`aiter/ops/quant.py:989`): an unlabelled `AssertionError` minutes into the load, where the Linear path would have produced a clear message at construction time. The w13/hidden axis has no guard on either side.

# 7. Non-runnability is expressed as a runtime raise instead of a setup-time rejection **[verified]**

`moe.py:1216`. NVFP4 is made non-runnable by raising in `Nvfp4MoEMethod.apply` and `LinearBase.forward`. Two consequences:

**(a)** With no `--online_quant_config`, `process_weights_after_loading` returns cleanly and `init_prepare_finalize` no-ops, so the model loads and reports healthy — it satisfies this repo's own readiness criteria (`/v1/models` answers, `rocm-smi` VRAM% > 0) and then fails every request and every graph capture with a per-layer `RuntimeError`.

**(b)** When a layer *is* excluded from the online target, `_online_quant` returns early via `should_skip_online_quant`, and this method then raises `reached post-load processing before the FusedMoE online conversion ran` — a message asserting the opposite of what happened. The operator's actual situation is "this layer has no NVFP4→MXFP4 target".

The validity condition (an NVFP4 source requires a `per_1x32`/`fp4x2` target) is known at parse time, and is currently re-checked in three scattered places: `quant_spec.py:130`, `linear.py:905`, `moe.py:3862`. Hoisting it to config validation collapses all three and turns both failures above into a startup error naming the recipe.

# 8. Two smaller correctness gaps in `create_weights` and the conversion **[verified]**

**Biases are dropped, and the drop is unreportable.** `Nvfp4MoEMethod.create_weights` does `layer.register_parameter("w13_bias", None)` / `("w2_bias", None)` unconditionally, ignoring `layer.has_bias`, while `Mxfp4MoEMethod` (`moe.py:1418`, `:1455`) and `Fp8MoEMethod` allocate real bias params when the layer declares them; `_online_quant` never carries biases across the method swap either (no `bias` reference anywhere in 3833-4200). An NVFP4 MoE with expert biases has `experts.N.w1.bias` swallowed by `weight_dispatch.py:220`, and because the parameter was never registered it is also absent from `expected_param_names` — so the unloaded-parameter warning cannot fire. `Mxfp4MoEMethod.create_weights` then allocates zero biases and `moe.py:2100/2182` consumes them. No in-tree NVFP4 model sets `has_bias=True` today, so this is latent rather than live (`gpt_oss.py:223` and `glm4_moe.py:169` do).

**The post-conversion state mutations are new and asymmetric, and they affect every online-quant source.** `moe.py:4011`. Base `_online_quant` set neither field; this PR has the row-local batched exit set **only** `self.params_dtype`, while the per-expert tail sets `params_dtype` **and** `self.layer_quant_config = online_quant_config`. The re-entry guard `should_skip_online_quant(source_quant_type, self.params_dtype, online_quant_config)` compares both type and dtype: after the tail exit both match and a second `process_weights_after_loading()` correctly skips, but after the batched exit `layer_quant_config.quant_type` is still the **source** type, the guard does not fire, and a re-entrant call re-quantizes already-quantized buffers. Separately, overwriting `layer_quant_config` with the target spec changes how `weight_loader` (`:4822`), `_copy_expert_shard` (`:4584`) and `expected_batched_arrivals` (`:4672`) interpret any post-load arrival. The RLHF rollout path is currently protected by `weight_updater.py:509` raising on quantized storage, so that misroute is not reachable today — the state asymmetry is.

# 9. Load-time cost: a dequant that can OOM, and a fast path disabled without a reason **[verified]**

`dequantize_nvfp4` (`quantization/quark/utils.py:176`) materializes roughly **9 bytes of transient device memory per logical weight element**: two int64 nibble tensors (`(weight & 0xF).to(torch.int64)`, `(weight >> 4).to(torch.int64)`), two fp32 LUT gathers, an fp32 output, and `scale.repeat_interleave(16, dim=-1)` expanded to full weight size. Both call sites hard-code `out_dtype=torch.float32` while `quant_mxfp4_online_even` immediately does `.to(torch.bfloat16)`. `_FP4_E2M1_LUT` is a module-level CPU tensor re-copied H2D on every call.

For a fused gate_up of 2×18432 rows × 7168 logical K the packed weight is 132 MB and the decode needs ~1.06 GB + ~1.06 GB for the int64 halves plus the fp32 output and the expanded fp32 scale — several GB live at once, on a GPU simultaneously holding the checkpoint shard, and under streaming online quant this runs concurrently on N worker threads. No other online-quant source does this; `dequant_weight_online` returns bf16 directly. This is a load-time HIP OOM on models that fit comfortably once loaded. int32 indices, `view(rows, groups, 16) * scale.unsqueeze(-1)`, and `out_dtype=bf16` cut it roughly 5×; a per-device LUT cache removes ~46k tiny pageable copies on a 256-expert / 60-layer model.

Separately, `moe.py:3991` disables `_online_quant_row_local_batched` for every NVFP4 MoE layer via `not source_is_nvfp4`, with no stated reason. The padding argument given for `check_need_allgather` does not transfer: the batched path already writes into padded targets using the same `target_half_rows` arithmetic the NVFP4 branch reproduces by hand (`moe.py:3797-3830` vs `:4042-4060`). The actual blockers look mechanical — the batched helper calls `dequant_moe_weight_online` rather than `_dequant_func`, and `dequantize_nvfp4` rejects a 3D batch. The cost of leaving it off is `E*3` round trips instead of `ceil(E/8)*2` — 96 vs 8 for 32 local experts, ~12× the launches and ~12× the huge fp32 intermediates above, over ~61 MoE layers. Either fix is fine; a comment saying which one is intended would help.

Three per-layer syncs pile on top: `target.copy_(value[0])` is a 4-byte blocking H2D per (expert, projection); `torch.nonzero(~isfinite(...)).tolist()` in `_validate_nvfp4_global_scales` forces a device sync **on the success path**; and `repeat_interleave(torch.tensor(output_partition_sizes, device=...))` builds a device tensor and cumsums back to host for an at-most-five-element expansion.

# 10. An HF-config schema change rode along, in a file the repo forbids editing **[verified]**

`minimax_m3.py:101`. `_is_moe_layer` now short-circuits on `config.mlp_layer_types` with unconditional precedence over `moe_layer_freq`, with no length check, no vocabulary check and no error branch:

- a list shorter than the layer count raises a bare `IndexError` during construction — MTP/draft layers are built with ids past `num_hidden_layers`, and VL configs may list only the text stack;
- a config spelling the label `"moe"` / `"sparse_moe"` / `"MoE"` rather than `"sparse"` silently builds dense MLPs for every MoE layer. The expert tensors then have no destination and are dropped by the loader's benign-drop filter, producing garbage with no error. The `moe_layer_freq` path it shadows at least defaults to True;
- a checkpoint carrying both fields silently changes which layers are MoE.

This has no connection to NVFP4 loading and belongs in its own accuracy-gated commit — otherwise bisecting that last failure lands on a PR titled "NVFP4 loading".

**Process note:** `MiniMaxM3Model` carries `@support_torch_compile`, and `/app/ATOM/CLAUDE.md` says *"NEVER modify `@support_torch_compile` decorated model files — instrument at call sites instead"*. Both edits to that file land inside it: `_is_moe_layer`, and `self.experts._intermediate_pad_override = 0` in `MiniMaxM3MoE.__init__`. The second has an obvious call-site home in `moe.py`.

# 11. 790 lines of new tests, none of which run in CI **[verified]**

`tests/test_nvfp4_loading.py:11` gates the entire file on `aiter = pytest.importorskip("aiter")`. `tests/conftest.py:92-93` states the consequence outright — *"CI has no aiter, so CI never saw"* — `tests/aiter_stub.py` exists as the supported alternative, and the sibling file this same PR touches (`tests/test_quant_config.py`) correctly uses mocks and does run.

The subset that does execute is also weak:

- the scale-math test picks `block=2.0` and `global=0.5`, whose product is exactly 1.0 — a `dequantize_nvfp4` that dropped **both** multiplications passes;
- `test_nvfp4_moe_online_conversion_uses_local_tp_shards` parametrizes `tp_size` over `{1, 2}`, but NVFP4 returns from `check_need_allgather` before `tp_size` is read, so both cases execute identical code and the `NoGatherGroup` guard is unreachable (`get_tp_group()` is only called when `need_gather_w2`);
- `test_unknown_sequential_quark_format_is_rejected` never reaches the branch it names — its fixture omits `input_tensors`, so it trips the earlier shape check and duplicates another test.

Net effect: 1951 lines of new production logic ship with automated coverage that never executes on CI, against `/app/ATOM/CLAUDE.md`'s `python -m pytest tests/  # all tests (no GPU needed — mocks AITER and torch.cuda)`. And the part that does execute cannot catch a dropped scale multiplication, a reintroduced w2 all-gather, or a relaxed `group_size == 16` predicate — so §2, §6 and §9 would all merge green if they regressed further.

# 12. Smaller, all verified

- `target_half_rows` in the NVFP4 w13 copy is derived from the **weight** tensor and reused as the **scale** row offset. The `_copy_scale` helper it replaced deliberately recomputed the midpoint from the tensor being written.
- `pad_align` is still not replayed across the `_online_quant` method swap while `intermediate_pad` now is. The gap is pre-existing, but the new asymmetry is conspicuous — `kimi_k3.py:353` and `deepseek_v4.py:3543` pass 128 and get 256.
- `register_parameter("input_scale_2", None)` runs for every Linear in every model, NVFP4 or not.
- The `ReplicatedLinear` early return in `online_quantize_weight` bypasses all the new NVFP4 checks.
- Dead code: the `isinstance(mapped_shard_id, tuple)` branch, and the `high_nibble_first` parameter.
- `dequantize_nvfp4` is a **third** copy of the E2M1 decode — `atom/model_ops/quant_v4.py:35-57` and `aiter.utility.fp4_utils.mxfp4_to_f32` are the other two. The "NVFP4 may only target MXFP4" predicate is likewise written out three times (§7).
- Comments this PR itself edited still carry stale source-format enumerations (`linear.py:919`, `moe.py:3870`), and `docs/model_ops_guide.md` and `docs/configuration_guide.md` gain no rows for the new format — `/app/CLAUDE.md` "Docs-match-code" items 3 and 4.
- `FusedMoE._online_quant` is now 325 lines and `moe.py` is 5301, both well past the repo's soft ceilings.

---

Of the above, §2 is the one I would hold on. It is the only finding whose failure mode is *silently wrong output with both diagnostics disabled*, and §11 means nothing would catch it later either. §1 is the one worth fixing even though nothing is visibly broken by it today — it is the reason six unrelated call sites are currently wrong, and every future NVFP4 change inherits the same manual grep. §3's documentation claim should be corrected in this PR regardless of whether the validation lands here, and §10 would be better as a separate commit against the model it actually changes.

valarLip:
## Review (third pass) — NVFP4 source format and online NVFP4→MXFP4 conversion

Re-reviewed at `e95cf4007`, a newer head than my [second pass](https://github.com/ROCm/ATOM/pull/2292#issuecomment-5774934498). I mark below which items are new at this head and which carry over, so nothing here reads as re-litigation.

**First, an acknowledgement:** my second pass said the docs claimed packed-conflict validation happened at setup while no such code existed. `_validate_packed_layer_specs` now exists. §4 is about that new function, not about its absence.

**Second, this pass had ground truth.** I downloaded the real `nvidia/MiniMax-M3-NVFP4` `config.json` and safetensors index, which **refuted two candidate findings** — the ModelOpt parser *is* reached (`quant_method: "modelopt"` is present), and shared-expert fusion is correctly blocked because shared experts are MXFP8 while routed experts are NVFP4. It also confirmed the checkpoint ships `weight_scale_2` per expert per projection. And the NVFP4 decode math — low-nibble-first, *multiply* by `weight_scale_2` — matches SGLang's `nvfp4_online.py` and vLLM. That part is right.

Two statements for what remains:

> **Two defects fail the load outright today, on architectures the PR does not exclude: every NVFP4 DeepSeek-V3-family checkpoint, and NVFP4 Qwen3-Next / GLM-5.**

> **`weight_scale_2` has been load-bearing since the first commit and is still on both of the loader's "benign to miss" suppression lists — and on the MoE side the expert mapping never declares it at all, so it resolves by an accidental longest-prefix substitution whose failure mode is a silent skip that is not even recorded in `dropped_ckpt_keys`.**

**Provenance:** **[verified]** means the deciding code was read at this head or checked against the real checkpoint. **[reported]** means the shape matches the code but was not traced end to end. **[carried]** marks an item also raised on the previous head.

---

# 1. Two hard load failures **[verified, new]**

**`MergedReplicatedLinear` collapses `output_size` to a scalar, so `weight_scale_2` gets one slot while its loader is driven with integer shard ids** (`linear.py:658`). `linear.py:2481` passes `sum(output_size)` to `super().__init__`, so `output_partition_sizes == [sum]` and `weight_scale_2 = torch.zeros(1)`. Meanwhile `config.py:588` installs `"kv_a_proj_with_mqa": ("fused_qkv_a_proj", 1)` for **every** deepseek_v3 / kimi_k2 / glm_moe_dsa / deepseek_mtp config with `q_lora_rank`, and `weight_dispatch.py:142` calls `param.weight_loader(param, tensor, 1)`. `_load_nvfp4_global_scale` then raises:

```
Invalid NVFP4 global scale shard id 1; expected [0, 0]
```

Any NVFP4 DeepSeek-V3-family checkpoint fails at load. The packed-split direction has the same shape (`"gate_up_proj": ["gate_proj", "up_proj"]`, `qwen3_next.py:1004`).

**`_nvfp4_global_scale_shard_map` is defined on 2 of the 6+ Linear subclasses that pass string shard ids** (`linear.py:2184`) **[carried, now enumerated]**. Present on `QKVParallelLinear` (`:2186`) and `MinimaxM3QKVParallelLinearWithIndexer` (`:2281`). Missing on `QKVGParallelLinear` (`:2038` — ids q/k/v but four output partitions), `QKVZParallelLinear` (`:1871`), `QKVZBAParallelLinear` (`:1689` — ids qkvz/ba/qkv/z/b/a) and `BAParallelLinear` (`:1982`). An NVFP4 Qwen3-Next or GLM-5 checkpoint dies at load with `<cls> defines no _nvfp4_global_scale_shard_map`. Loud rather than silent, but nothing documents the supported set — and the ClassVar is annotated `dict[str, int]` while the loader at `linear.py:820` explicitly handles tuple values and the error message tells the reader to add one.

# 2. The load-bearing scalar is still suppressed on both diagnostics **[verified, carried]**

`loading_core.py:420`. `linear.py:658` and `moe.py:1224` allocate `weight_scale_2` / `w13_weight_scale_2` / `w2_weight_scale_2` with `torch.zeros`, and nothing checks they were written — `online_quantize_weight` only checks `numel()`. Meanwhile:

```
loading_core.py:420   suppressed_suffixes = (..., "weight_scale_2")   ← never-loaded-param warning
loading_core.py:450   benign_substrings   = (..., "weight_scale_2")   ← dropped-ckpt-key warning
```

Both entries predate NVFP4, when the tensor was genuinely ignorable. A `weights_mapping` rule that fails to route `...gate_proj.weight_scale_2` to `gate_up_proj.weight_scale_2` leaves it at 0.0 → `dequantize_nvfp4` multiplies every block scale by zero → MXFP4 quantizes an all-zero weight → the server starts clean, emits garbage, and reports nothing on either diagnostic.

**On the MoE side it is worse, and this part is new** (`models/minimax_m3.py:169`) **[verified, new]**. `make_minimax_m3_expert_params_mapping` declares **no `weight_scale_2` entry at all**. `...experts.0.w1.weight_scale_2` resolves only because `expert_weight_prefixes` is sorted longest-first: the `weight_scale` entry wins, `name.replace()` leaves the `_2` tail dangling, and it coincidentally yields `experts.w13_weight_scale_2`. If that ordering or a model's mapping string changes, `weight_dispatch.py:224` hits `param = self.params_dict.get(name); if param is None: return True, name` — a silent skip that is **not recorded in `dropped_ckpt_keys` at all**, so even removing the two suppressions above would not surface it. Add an explicit `weight_scale_2` entry to the expert mapping.

# 3. Two Fix-then-sweep misses **[verified, new]**

**`SiluAndMul` and `RMSNorm` dispatch on `QuantType.per_1x32` with no dtype term** (`activation.py:101`). `quant_spec.py:64` states outright that "NVFP4 shares `QuantType.per_1x32` with MXFP4", and this PR added NVFP4 rejection only to `LinearBase` and `FusedMoE`. `activation.py:74` sets `self.params_dtype = layer_quant_config.quant_dtype`, and `:101-106` selects `mxfp4_act_mul_quant_fuse(x, shuffle=True)` purely on `self.quant_type.value == QuantType.per_1x32.value` — so an NVFP4-resolved prefix produces MXFP4-formatted activations, and `SiluAndMul` has no `online_quantize_activation` hook to correct it. `layernorm.py:296` has the same assignment; `RMSNorm` self-corrects only when `use_fused_quant` is set (`:318`), reachable today via `kimi_k3.py:1533` / `kimi_k3_dspark.py:552`.

**`FusedMoE._online_quant` never resets the layer's quant state, although the same PR added exactly that reset to the Linear path** (`moe.py:4182`). `linear.py:1036-1039` sets `quant_type`, `params_dtype`, `layer_quant_config` and `_stream_online_quant`; `moe.py:4182-4192` only `delattr`s the two `*_weight_scale_2` params. So a converted MXFP4 `FusedMoE` still reports `quant_dtype == "nvfp4"`, and the new guard at `moe.py:4848`

```python
if is_nvfp4 and weight_name == "":
    raise RuntimeError("Merged NVFP4 expert loading is not supported")
```

stays armed forever, shadowing the `mxf4_merged_weight_loader` dispatch on the very next line. An RLHF rollout weight sync (`atom/rollout/weight_updater.py:414`) pushing a merged expert tensor into an already-MXFP4 layer raises instead of loading. And if `_online_quant` were ever re-entered, `moe.py:3984`'s `self.w13_weight_scale_2.data` raises `AttributeError` on the `delattr`'d name.

# 4. The new validation: too narrow, too early, too broad **[verified, new]**

`_validate_packed_layer_specs` closes the gap my second pass reported. Three things about how it landed:

**Too narrow** (`config.py:642`). It is gated on `quant_method == "modelopt"`, but the hazard its own docstring describes — *"a packed (N, K/2) uint8 NVFP4 weight can be loaded into an (N, K) fp8 parameter with nothing said"* — is a property of `_remap_layer_name` plus `get_layer_quant_config`'s first-match rule, not of ModelOpt. Quark NVFP4 exports also record one entry per source projection; that is why `_quark_weight_spec` exists. A Quark checkpoint with q_proj NVFP4 and k_proj FP8, or with only v_proj excluded, still silently builds the packed QKV parameter from whichever spec comes first. And `tests/test_quant_config.py::test_packed_conflict_validation_is_modelopt_mixed_only` explicitly asserts that a quark config with conflicting gate_proj/up_proj specs is **accepted** — pinning the narrowing in place.

**Too broad** (`config.py:686`). The exclusion loop walks every pattern rather than only those that collapsed from disagreeing sources: `seen` is populated unconditionally by `setdefault`, so `for pattern in seen: if self._is_excluded(...)` fires for a single non-fused layer that merely appears in both `quantized_layers` and `exclude_modules` — a hard startup error, with a message claiming a partial packed exclusion, where the rule everywhere else (`config.py:444`) is exclude-wins.

**Too early** (`config.py:642`). The call sits ahead of `apply_exclude_name_mapping` (`:660`), and well ahead of `apply_default_exclude_layers` on both plugin paths (`plugin/vllm/model_wrapper.py:481/485`, `plugin/sglang/prepare.py:27`) and in MTP models (`deepseek_mtp.py:378` appends `["*.eh_proj"]`). Any partial exclusion introduced there escapes the check entirely. I verified the shipped nvidia checkpoint has zero overlap, so today only the false-negative half bites.

# 5. Two silent mis-constructions **[verified, new]**

**`ModelOptParser` returns a bare default `global_spec` (`QuantType.No` / bf16)** (`quant_spec.py:519`), so any module not enumerated in `quantized_layers` silently becomes unquantized rather than erroring. Every other parser derives a global spec from the checkpoint; this one hardcodes `LayerQuantConfig()`. Combined with `get_layer_quant_config`'s fall-through (`config.py:456`) and its substring rule (`if layer_name in pattern`), a module path that is not a substring of any of the 22422 enumerated names gets bf16 with no warning — indistinguishable from a name-mapping failure.

Concretely, against the shipped checkpoint: `text_config.num_mtp_modules == 1`, and the MTP block appears in **neither** `quantized_layers` nor `exclude_modules`. Loading it as a drafter builds bf16 `(N, K)` parameters for packed `(N, K/2)` uint8 checkpoint tensors. None of this PR's NVFP4 guards fire, because `global_spec.quant_dtype != NVFP4_DTYPE`. (Same site: `kv_cache_quant_algo` is parsed by nobody.)

**`_normalize_minimax_m3_mlp_layer_types` is handed only `text_config`** (`config.py:951`), so a top-level `mlp_layer_types` is silently ignored — producing exactly the "every layer built as MoE" outcome its own docstring says must not happen. `_normalize_minimax_m3_text_config` passes `text_config if text_config is not None else hf_config`. If `mlp_layer_types` lives on the root config while a `text_config` exists — and M3 exports are nested, which is why that function exists — `getattr` returns None, the function returns early, no `moe_layer_freq` is synthesized, and `minimax_m3.py:101`'s `_is_moe_layer` returns True for all 60 layers. Dense-layer checkpoint tensors then land in MoE params or are dropped, with no error. The attribute-copy loop at `config.py:975` only propagates text_config → hf_config, never the reverse.

Second defect in the same function (`config.py:944`): `not isinstance(existing, (list, tuple))` rejects the conventional scalar `moe_layer_freq: 1` with a message that is factually wrong for that input ("they disagree on which layers are MoE").

# 6. Uninitialized memory, and an offset derived from the wrong tensor

**`Nvfp4MoEMethod.create_weights` uses bare `torch.empty` for all four expert buffers** (`moe.py:1197`) **[verified, new]**, while the sibling `Mxfp4MoEMethod` deliberately byte-zeroes weights — `moe.py:1430/1467` do `w13_weight.data.view(torch.uint8).zero_()` with the comment *"to avoid garbage in padded rows"* — and uses `torch.zeros` for scales. `local_num_experts` includes EPLB redundant slots and the fused shared slot (`moe.py:3470`), and `_online_quant` iterates all of them (`moe.py:4051`). The only zeroing of unwritten slots is in `flush_staged` (`moe.py:4797`), which never runs when batching is off (`loading_core.py:236`: `batching_enabled = executor is not None`) or when the fused/merged dispatch calls `staging_pool.decline`. Uninitialized E4M3 bytes (0x7F/0xFF) decode to NaN, and `dequantize_nvfp4` → `quant_weight_online` propagates NaN into those MXFP4 blocks.

**The w13 store derives `target_half_rows` from the weight tensor and reuses it to index the scale tensor** (`moe.py:4080`) **[verified, carried]**. `target_half_rows = w13_expert.shape[0] // 2` is applied both to `w13_expert` and, at `moe.py:4095`, to `w13_scale_expert`. The helpers this replaced (`_load_w13`, `moe.py:4393`) derived the half from each destination's own shape via `expert_shard_view` + `narrow()`, which raises on out-of-range. Today weight and scale both have `2*intermediate_padded` rows so the value is coincidentally right; the moment a target's scale rows differ from its weight rows — a per-token-style target, or `pad_align` applied to weights only — w3's scales land at the wrong row offset. And because the copy is `[a:b, :w3_s.shape[1]]` Python slicing, an out-of-range end silently produces a shorter view instead of an error.

# 7. `pad_align` is lost across the method swap, and an inert field makes it look handled **[verified, carried]**

`moe.py:3613` writes the caller's `pad_align` onto `self.quant_method`, with a comment that it must happen before `create_weights`. But `Nvfp4MoEMethod` never reads `pad_align`, and `_online_quant` (`moe.py:3996`) builds a fresh `Mxfp4MoEMethod` whose `__init__` resets `pad_align` to 256 (`moe.py:1277`); only `intermediate_pad` is re-pinned (`moe.py:4011`). `kimi_k3.py:353` and `deepseek_v4.py:3730` pass `pad_align=128`, so the same model gets differently-sized expert buffers from an NVFP4 checkpoint than from a native-MXFP4 one, with no diagnostic — and `quant_method.intermediate_size` is what Mori/FlyDSL size their buffers from (`mori_v2_prepare_finalize.py:442`). `Nvfp4MoEMethod.pad_align = 1` is never read by anything, which is what makes the loss look handled.

Related: the `pinned_intermediate_pad` restore is gated on `source_is_nvfp4`, so MiniMax-M3's `intermediate_pad = 0` pin (`minimax_m3.py:291`, *"avoids backend pad-skip precision issues"*) is still lost for a BF16 or FP8 source with an mxfp4 online target.

# 8. A weight-only Quark NVFP4 export is hard-rejected **[verified, new]**

`quant_spec.py:352`. With `weight` a two-entry list and `input_tensors` None — the normal Quark spelling for weight-only quantization — `weight_is_multi_stage` is True, so the single-stage early return is skipped, and then `not isinstance(input_tensors, list)` raises *"Unsupported Quark multi-stage quantization config... two dictionary stages in both `weight` and `input_tensors`"*.

This contradicts the PR's own design: NVFP4 activations are never executed, and `input_scale`/`input_scale_2` are deliberately unregistered and documented as expected dropped tensors — yet the absence of the activation half is fatal, and the message blames "multi-stage" rather than weight-only.

Same predicate, asymmetric: `not weight_fp4.get("is_dynamic")` tolerates a missing key, but `input_fp4.get("is_dynamic") is True` requires it to be literally present — so an export omitting it, which `_parse_is_dynamic` treats as dynamic everywhere else, falls into the same raise.

# 9. 884 lines of tests, 29 SKIPs in CI **[verified, carried]**

`tests/test_nvfp4_loading.py:11` — `aiter = pytest.importorskip("aiter")` at module scope. The only unit-test job (`.github/workflows/pre-checks.yaml:83`, ubuntu-latest, torch CPU wheel, no aiter) runs `pytest tests/` with only `tests/plugin` and `test_openai_server.py` ignored, so all 29 tests report SKIP.

The same workflow already installs av/pillow/httpx with the comment *"Without them 38 diffusion tests self-skip via importorskip and the subsystem gets no CI coverage at all"* — the established intent here is the opposite of what this file does. And `tests/test_quant_config.py` mocks aiter and does run, so the technique was available in the same directory.

Separately, the `_online_quant` MoE test (`:721`) asserts only on `w2_weight` / `w2_weight_scale` and the fake quantizer's inputs. Writing w3 at `w1_q.shape[0]` instead of `target_half_rows`, or swapping the two `_copy_quant_storage` destinations, passes the suite and ships a silently wrong model — i.e. §6's second item has no test that would catch it.

# 10. Smaller, and one altitude point **[verified]**

- **Load-time memory.** `dequantize_nvfp4` materializes `low`/`high` as **int64** (8 B per *packed byte*), an fp32 `(N, K)` buffer, and a full-size `repeat_interleave`d fp32 scale — roughly 20-24 bytes per logical element, and the Linear path does this on the whole fused qkv/gate_up at once with no chunking. Both callers pass `out_dtype=torch.float32`, which `quant_mxfp4_online_even` immediately downcasts to bf16. `_FP4_E2M1_LUT.to(device=...)` is a fresh 64-byte H2D copy on every call — ~46k copies for an M3-class model. Cheap fixes: int32 indices, `deq.view(rows, -1, 16).mul_(scale.unsqueeze(-1))` instead of `repeat_interleave`, a per-device LUT cache, `out_dtype=bfloat16`.
- **Duplicate decoder.** `atom/model_ops/quant_v4.py:57`'s `dequant_fp4_e2m1` already implements the same LUT + nibble split + `repeat_interleave`, with a parameterized block size, and currently has no callers. The two LUTs have **already drifted** — index 8 is `0.0` in one and `-0.0` in the other — and they assert opposite storage dtypes (int8 vs uint8).
- **Docs-match-code.** `docs/model_ops_guide.md:259-266` (the FusedMoE quant_method table) and `:501` (module contents) do not list `Nvfp4MoEMethod`; `docs/configuration_guide.md:140/151/189` still describe `quant_method` as only quark/compressed-tensors and never mention `modelopt`.
- **Altitude [carried].** `NVFP4_DTYPE = "nvfp4"` is a string smuggled into a `torch.dtype` field, widening `QuantizationConfig.quant_dtype` to `torch.dtype | str` and adding ~13 `== NVFP4_DTYPE` sites. `LayerQuantConfig` already has `weight_block_size`, used exactly this way in `models/deepseek_v41/layers.py:16` — so `quant_dtype=dtypes.fp4x2, weight_block_size=(1, 16)` would keep all ~23 existing `== dtypes.fp4x2` sites correct instead of silently falling to the bf16 branch. The literal `"nvfp4"` also already means something else in `model_ops/fused_moe/config.py:187`.
- Minor: the `% 16` alignment checks at `linear.py:549` and `moe.py:1188` are dead, implied by the `% 32` checks immediately after; `high_nibble_first` has no caller; `register_parameter("input_scale_2", None)` at `linear.py:726` runs for every Linear in every model regardless of format.

---

Of the above, §1 is what I would fix first — it needs no argument and it takes out two model families the PR does not claim to exclude. §2 is the one I would hold on: it has been load-bearing since the first commit, its failure mode is the only one here that produces a clean startup and garbage output, and the MoE half of it resolves by coincidence rather than by declaration. §3's `activation.py` item is worth a grep pass of its own — `QuantType.per_1x32` is now overloaded, and every existing site that reads it as "MXFP4" is a candidate.

valarLip:
## Review (fourth pass) — NVFP4 source format and online NVFP4→MXFP4 conversion

Re-reviewed at `c23571b8c`, newer again than my [third pass](https://github.com/ROCm/ATOM/pull/2292#issuecomment-5795235438). Scoped to **functional defects** this time — wrong output, crash, or a checkpoint that should load and does not. Docs, coverage and structure appear only in the last section, and only where they cause or conceal something in the list.

**First, a correction I owe you.** My third pass asked for `weight_scale_2` to come off the loader's suppression lists. That landed — it is now out of both `suppressed_suffixes` (`loading_core.py:420`) and `benign_substrings` (`~:446`). But every `LinearBase` also registers `input_scale_2 = None` unconditionally, and a None-registered parameter is absent from `params_dict`, so **every** `input_scale_2` tensor in a Quark checkpoint is now reported as an actionable drop — which the PR itself concedes at `linear.py:724`. Any layer excluded to bf16 in a mixed checkpoint reports its `weight_scale_2` the same way. The comments at `loading_core.py:416-419` and `442-443` still describe these lists as "known-OK skips" / "so the warning is signal, not noise".

Net effect: a normal NVFP4 load now emits a wall of "this is a weights_mapping bug" lines, **and that wall buries the one diagnostic that would surface §1 below**. The right shape is for suppression to be a property of the layer — the quant method declares its expected-unused tensor names — not a global suffix list every format shares. That is on my earlier advice, not on the author.

**Second, a narrowing of my own third-pass claim.** I listed four Linear subclasses as missing `_nvfp4_global_scale_shard_map`. That was too broad: `QKVZParallelLinear` and `BAParallelLinear` take no shard id, and `MergedColumnParallelLinear` / `MergedReplicatedLinear` take int ids, so all four are fine. The real gap is two classes — see §6.

**Third, candidates refuted at this head**, so nobody re-spends on them: the comm-fused MoE backend is never built for NVFP4 (double-disqualified at `comm_fused_moe.py:55-75`); `Nvfp4MoEMethod`'s plain `layer.w13_input_scale = None` cannot collide with `register_parameter`, because online targets are always `is_dynamic=True`; `hidden_pad` staleness is unreachable (gpt-oss passes `has_bias=True`, which `Nvfp4MoEMethod.create_weights` rejects first); the `pinned_intermediate_pad` block is **live**, not dead (MiniMax-M3 pins 0 after construction); the scalar-`moe_layer_freq` rejection is **correct**, because M3's own reader subscripts the value; and the nibble order and global-scale multiply direction match the SGLang reference (`/app/sglang/python/sglang/srt/layers/quantization/dequantization.py:72-100`) exactly.

**Provenance:** **[verified]** means the deciding code was read at this head. **[carried]** marks an item also raised on an earlier head.

---

# 1. The new validation misses the case its own docstring defines **[verified, new]**

`atom/config.py:687`. The partial-exclusion guard calls `self._is_excluded(pattern, self.exclude_layers)` **without** `check_children=True`, so it cannot see a leaf exclusion under a fused MoE.

`exclude_modules` names one expert projection — say `...block_sparse_moe.experts.5.w1`. The guard tests the experts-container pattern with plain prefix matching (`config.py:551`), which cannot match a longer child name, so nothing is raised. `FusedMoE.__init__` then resolves its spec **with** `check_children=True` (`moe.py:3341`), `_is_excluded` returns True via the child branch, and the container gets `LayerQuantConfig(quant_dtype=torch_dtype)` — an unquantized bf16 MoE. From there `source_is_nvfp4` is False, `validate_nvfp4_online_target` never arms, `Nvfp4MoEMethod` is never selected, and the loader pours packed `(E, 2I, H/2)` uint8 NVFP4 tensors into bf16 `(E, 2I, H)` buffers.

This is, word for word, the failure the docstring promises to catch: *"a partial exclusion de-quantizes the whole fused layer to bf16 while the checkpoint still holds packed weights."*

# 2. Global-scale completeness is still unchecked, and the Linear side now uses `torch.empty` **[verified, carried]**

**`linear.py:658`** allocates NVFP4 `weight_scale_2` with `torch.empty`, and the only check — `numel() != len(output_partition_sizes)` — is a property of the allocation and can never fail. A merged QKV NVFP4 layer whose checkpoint ships `weight_scale_2` for q but not k/v (a name variant, an excluded projection, a `weights_mapping` miss) leaves two of three slots holding **uninitialized float32**: possibly denormal, huge, inf or NaN. `online_quantize_weight` builds `per_row_scale_2` from it and `dequantize_nvfp4` multiplies those output row-ranges by garbage. The model loads clean and serves silently wrong (or NaN) weights.

**`moe.py:1223`** uses `torch.zeros`, and `_load_nvfp4_global_scale` writes one scalar per arrival with no completeness check. Any (expert, shard) global scale that never arrives — a fused shared-expert slot, which this PR now explicitly allows for NVFP4 at the `uses_all2all_fusion` check; an EPLB redundant replica; a mapping miss — makes that expert's w1/w3/w2 dequantize to an all-zero matrix, re-quantize to all-zero MXFP4, and contribute zeros forever. **No NaN, no raise, clean load log.** Note `weight_loader` short-circuits these params *before* the `ExpertStagingPool` path, so they never appear in `staging_report.incomplete` either.

The only signal in either case is a rank-0 warning from `_report_coverage`, and `ATOM_LOADER_STRICT_COVERAGE` does not cover this path — which is what the correction at the top of this comment is about. Minimum fix: make Linear use `zeros` too, and assert every slot is finite and `> 0` in both consumers before `dequantize_nvfp4`.

# 3. `_commit_online_quant_state` changes post-load state for every online-quant MoE, not just NVFP4 **[verified, new]**

`moe.py:3887`. Before this PR both exit paths of `_online_quant` set only `self._online_quant_info`. Now both call `_commit_online_quant_state`, which assigns `self.layer_quant_config = online_quant_config` and `self.params_dtype = online_quant_dtype` **unconditionally**.

`weight_loader` (`moe.py:4887`) branches on `self.layer_quant_config.quant_dtype == dtypes.fp4x2 and weight_name == ""` and hands off to `mxf4_merged_weight_loader`. So for a bf16- or FP8-source MoE converted online to MXFP4, any expert tensor arriving *after* conversion — reachable on the streaming path, where conversion fires the moment `_stream_loaded_numel` hits `_stream_expected_numel` rather than at end-of-load — now takes the MXFP4 merged loader instead of the source loader it took before.

The NVFP4 half of this change is intentional and tested. The non-NVFP4 half is an unflagged, untested behavior change. Either gate the rewrite on `source_is_nvfp4`, or state why the new routing is correct for all sources.

# 4. The new validation is gated on ModelOpt, so Quark NVFP4 gets no conflict check **[verified, carried]**

`config.py:642`. `QuarkParser` also emits one `layer_pattern_specs` entry per checkpoint layer name and goes through the identical `_remap_layer_name`. A Quark export giving `...q_proj` NVFP4 and `...k_proj` FP8 remaps both onto `...qkv_proj`, and `get_layer_quant_config` returns whichever comes first (`config.py:449-454`). If NVFP4 wins, the fp8 v weight is loaded into an `(N, K/2)` uint8 buffer; if fp8 wins, packed NVFP4 bytes land in an `(N, K)` fp8 parameter — the exact silent corruption the docstring describes, unguarded because `quant_method` is `"quark"`.

The cheap fix is to drop the `quant_method` gate entirely: the conflict loop is a no-op when siblings agree.

# 5. The same validation aborts startup on configurations that are fine **[verified, carried]**

`config.py:686`. The second loop iterates **every** quantized pattern, not just fused ones — `_remap_layer_name` returns non-packed names unchanged. `_matches_exclude` supports `re:` regexes, fnmatch globs, exact names and parent prefixes, so a ModelOpt export whose `quantized_layers` lists e.g. `model.layers.0.mlp.down_proj` while `exclude_modules` carries a glob or parent that also matches it raises:

```
ValueError: Packed layer '...' is quantized by at least one source projection and excluded by another
```

— a claim about fused projections, for a layer that has none. Narrow the check to patterns a packed remap actually collapsed (more than one source, or where `_remap_layer_name` returned something different), and reword the message otherwise.

# 6. The tuple branch is dead code, and the one class that needs it has no shard map **[verified, narrows my third pass]**

`linear.py:2038`. `QKVGParallelLinear.output_partition_sizes` has **four** entries — [Gate, Q, K, V] (`:2069-2074`) — so `weight_scale_2` is allocated with four slots, while its weight_loader takes str ids `'q'`/`'k'`/`'v'`. The single HF `q_proj` carries both Gate and Q, i.e. `'q'` must map to `(0, 1)` — exactly the `isinstance(mapped_shard_id, tuple)` + `values.expand(...)` path the new loader added at `linear.py:814-816`.

But the class defines no `_nvfp4_global_scale_shard_map`, so any NVFP4 checkpoint for `qwen3_next.py:322` (`attn_output_gate`) or `qwen4_exp/qsa_attention.py:224` raises `ValueError: ... QKVGParallelLinear defines no _nvfp4_global_scale_shard_map` on the first tensor. `QKVZBAParallelLinear` has the same gap.

Loud and actionable, and not reachable by any in-tree checkpoint today — but the class the tuple machinery was built for is the one left out, which makes that branch currently unexecutable.

# 7. `mlp_layer_types` on the outer config is silently ignored **[verified, carried]**

`config.py:952`. `_normalize_minimax_m3_text_config` passes `text_config if text_config is not None else hf_config` — only **one** of the two configs is inspected — and the field-copy loop that follows (`:965-973`) moves only `use_index_cache` / `index_topk_freq` / `index_topk_pattern` / `index_skip_topk_offset`, never `mlp_layer_types`.

For a nested MiniMax-M3 export carrying `mlp_layer_types` on the outer config while `text_config` exists, `getattr` returns None and the function returns having done nothing. `minimax_m3.py:102-103` then sees `moe_layer_freq is None` and returns True for every layer — **every layer is built as MoE**, dense-layer checkpoint tensors land in MoE params or are dropped, with no error. That is precisely the silent failure the function's own docstring forbids ("a list that cannot be read exactly must fail here rather than build the wrong layers"). The new tests only ever place the field on `text_config`.

# 8. A weight-only Quark NVFP4 export cannot load at all **[verified, carried]**

`quant_spec.py:378`. With `input_tensors: null` (or a dict), `weight_is_multi_stage` is True and `input_is_multi_stage` is False, so the single-stage early return is skipped and the strict validation raises *"ATOM recognizes only NVFP4 with two dictionary stages in both `weight` and `input_tensors`"*.

ATOM discards the activation stage entirely anyway — `a_quant_dtype` is forced to None for NVFP4, and Linear never reads `input_scale_2` — so the activation spec is irrelevant to whether the weights can be decoded. This is an untested restriction that turns a loadable checkpoint into a startup `ValueError`, with a message that misdescribes the problem. The reverse shape (single-stage weight, two-stage `input_tensors`) hits the same branch with the same misleading message.

# 9. `dequantize_nvfp4` can OOM at load on the path that exists to bound load memory **[verified, carried]**

`atom/quantization/quark/utils.py:177`. The function holds roughly **20 bytes of transient GPU memory per logical weight element**: two int64 nibble index tensors, an fp32 `torch.empty` output, an fp32 `repeat_interleave`'d scale, and the product — against ~2 B/elem for every other online source.

Both call sites pass `out_dtype=torch.float32`, which is thrown away immediately: `quant_mxfp4_online_even` (`quark/utils.py:426-428`) does `if q_in.dtype not in (fp16, bf16): q_in = q_in.to(torch.bfloat16)`, so the fp32 buffer is down-converted on the very next line while still live. For a fused gate_up of N = 2×8192, K = 6144 (~100M logical elements) that is roughly 2 GB alive per layer, and in the MoE loop `w1_bf16` and `w3_bf16` are alive simultaneously.

I am filing this as functional rather than as a perf item because the symptom is a load-time HBM OOM, and it lands on `should_stream_online_quant` — the path whose stated purpose is bounding load-time memory. All the fixes are local: `out_dtype=torch.bfloat16`; int32 indices; `dequantized.view(rows, cols//16, 16).mul_(scale.unsqueeze(-1))` instead of `repeat_interleave` plus an out-of-place multiply; and a per-device LUT cache instead of a 64-byte pageable H2D copy on every call (~3 calls per expert per layer).

---

# Not findings on their own — but they are why the above is hard to see

I am not filing docs, coverage or structure items this pass. These two are exceptions, because each is the cause or the cover for something above.

**The MoE global scale reaches its parameter by an unstated string coincidence** (`models/minimax_m3.py:163`). `make_minimax_m3_expert_params_mapping` emits no `weight_scale_2` entry. It works only because `loading_core.py:181` sorts prefixes longest-first and `weight_dispatch.py:216-219` does `name.replace(wm_name, pm_name)`: `...experts.0.w1.weight_scale_2` matches the longer key `...experts.0.w1.weight_scale`, and the trailing `_2` survives the rewrite. Add a competing longer key, switch the dispatcher to exact matching, or rename the tensor, and the FP32 scalar is silently rerouted into `w13_weight_scale` (the group-scale param) or dropped — **which is §2's all-zero-expert outcome, with no error**. The one test covering this (`tests/test_nvfp4_loading.py:890-903`) re-implements the longest-prefix rule inside the test body instead of driving the dispatcher, so it would keep passing. The clean resolution is to delete the explanatory comment and add the real mapping entry — which also removes this PR's only edit to a `@support_torch_compile` file.

**The new test file runs zero assertions in CI** (`tests/test_nvfp4_loading.py:11`). Module-level `aiter = pytest.importorskip("aiter")`, and `.github/workflows/pre-checks.yaml` installs only torch-cpu/pytest/msgpack/quart/av/pillow/httpx — so all 21 tests are collected and skipped. `tests/test_quant_config.py` follows the repo's documented convention (temporary `sys.modules` mocks) and does run, so the technique was available in the same directory. Concretely: `Nvfp4MoEMethod`, the `FusedMoE._online_quant` NVFP4 branch, `dequantize_nvfp4`'s nibble order and `_load_nvfp4_global_scale` have no CI coverage at all. Separately, `test_nvfp4_moe_online_conversion_uses_local_tp_shards` (`:609`) is inert for the property it names: the fixture hardcodes `intermediate_size_per_partition=source_intermediate` for both `tp_size` values so the two runs are byte-identical, and `check_need_allgather()` returns False for `source_is_nvfp4` before `tp_size` is read, so `NoGatherGroup.all_gather` can never be invoked and its assertion can never fire.

---

Of the above, §1 is what I would fix first: it is a one-argument change, and without it the function does not do the job it was added for two heads ago. §2 is the one I would hold on — it has been load-bearing since the first commit, it is the only item here whose failure mode is a clean startup with garbage output, and its MoE half now resolves by coincidence rather than declaration. §3 is worth a decision rather than a fix: it is the only change in this PR that reaches sources other than NVFP4, and right now nothing says whether that was intended.

valarLip (approved):

PR #2298, opened 2026-09-19
[Lumen-RL] Receive trained weights over RCCL, transactionally

i-chaochen:
## Motivation

Stacked on https://github.com/ROCm/ATOM/pull/2297

ATOM rollout has no direct weight path from a trainer: weights arrive via
safetensors on shared storage or a ZMQ CUDA-IPC hop. This lets each TP worker
join an independent process group alongside the trainer, so weights land
straight in resident GPU memory.

## Technical Details

The wire format is frozen by the sending side and matched byte for byte: a
4x int64 header `[command, metadata_bytes, payload_bytes, version]`, a uint8
JSON metadata list, then a uint8 payload each entry views into. Decoded tensors
are views over the payload rather than copies, since a transfer already sized in
tens of GB cannot afford to double its peak footprint. A test reads the sender's
own constants from source, so drift on either side fails here rather than
decoding tens of GB into garbage.

Two things make this more than a memcpy.

Fused parameters: `qkv_proj` and `gate_up_proj` combine several checkpoint
tensors, and one fused parameter's shards can arrive in different buckets, so
state must survive across them. `update_weights`' application loop is extracted
into `_apply_named_tensors` and shared, leaving finalisation -- expert relayout,
KV clear, accumulator teardown -- with the caller. A bucketed stream does that
once at commit; per bucket it would relayout a half-built parameter.

In-place application: a stream that fails halfway leaves the model a mix of two
versions, and inference would keep serving, quietly wrong. So a stream is a
transaction -- begin/apply/commit/abort -- with monotonic versions so a replayed
or out-of-order stream is refused, and commit-time coverage verification against
`named_parameters()`. `forward` now calls `assert_weight_update_ready()`, which
refuses to serve mid-reload or after a partial one. The cost of a false stop is
a raised error; the cost of not stopping is rollouts from a half-updated model,
surfacing much later as unexplained divergence.

Coverage accounting credits `weight_scale` values that requantisation writes as
a side effect. They are derived, never sent, so without that they look
permanently missing and verification would reject every stream.

`init_independent_process_group` is vendored into `atom/utils` rather than
imported from the RL framework, so a rollout container carrying only ATOM can
still build the group.

Note the coverage check currently gates the RDMA path only; extending it to the
shm and IPC transports is follow-up work.

## Test Plan

11 new CPU-only tests covering the wire-format round trip, view-not-copy, and
rejection of every corrupt-metadata shape, plus a test pinning the sender's
constants. Re-ran #2028's weight-sync tests, which are the real check on the
shared-loop extraction.

## Test Result

- new tests: 11 passed
- #2028 weight-sync tests: 34 passed, 2 skipped
- full suite: 5712 -> 5723 passed; failures, skips and errors unchanged
- `black --check .` clean; `ruff` no findings on touched files

Not yet validated on hardware. The 9-rank run needs two GPU nodes
simultaneously and has not been schedulable yet.

## Submission Checklist


atom/model_engine/engine_core_mgr.py
```diff
@@ -1313,6 +1324,119 @@ def abort_request(self, req_id):
...
+                            f"DP rank {dp_rank} did not answer {method!r} "
+                            f"before the deadline"
+                        ),
+                    )
+                )
+                continue
+            if body.get("error"):
+                # The engine failed before it could reach its runners, so there
+                # is no per-TP detail to report.
+                flat.append(RpcResult(request_id, -1, error=body["error"]))
```
ZhangDanyang-AMD, line 1427: same as copilot's comment. CoreManager appends only one tp_rank=-1 error even though the API promises one result per TP rank, preventing accurate coverage checks.

atom/rollout/weight_updater.py
ZhangDanyang-AMD, line 307: abort_weight_update clears only the packed accumulator. A later reload can inherit expert-relayout state from the failed transaction. See 1177-1187

atom/model_engine/engine_core_mgr.py
ZhangDanyang-AMD, line 1321: engine_utility.py:312-329 writes an unconditional response while it is consumed synchronously so non-RPC responses enter the legacy queue.

ZhangDanyang-AMD (changes requested):
The end goal is to run a 200-step ATOM + Megatron Qwen3 disaggregated case and match vLLM [Disaggregated 2-Node Megatron + vLLM RDMA Deployment](https://github.com/ZhangDanyang-AMD/Lumen-RL/blob/main/examples/docs/07-disaggregated-rdma.md) then run [Running the eight examples from the release image](https://github.com/ZhangDanyang-AMD/Lumen-RL/blob/main/examples/docs/08-release.md#821-overview) cases 1 and 3 for 1 step to confirm other functionality is unaffected.

PR #2299, opened 2026-09-20
Enable GLM-5.3-Flash under vllm serve on the ATOM plugin.

sajandhy:
Register Glm5Next with the hybrid KDA shim, tolerate a missing ATOM Context during vLLM profile runs, and re-align the KDA/MLA KV page so workers no longer die in setStorage.

## Submission Checklist


sajandhy:
Please check https://github.com/ROCm/ATOM/pull/2340

PR #2305, opened 2026-09-20
fix(offload): memoize the external-tier lookup per HBM frontier (native + plugin)

PerryZhang01:
## 场景

两个 scheduler 都在**分配 KV 块之前**问 offload scheduler「外部层还能供多少 prompt token」：

- vLLM plugin：`scheduler.py:811` 只用 `request.num_computed_tokens == 0` 作门，:829 调 `get_num_new_matched_tokens`，然后才 `allocate_slots`。
- ATOM 原生：`atom/model_engine/scheduler.py:1377-1381` 的 `_query_connector_prefill_match`，在 :1461 的 `block_manager.can_allocate` 之前。

分配失败的每一条退出路径都把请求原样退回 waiting 队列头部、frontier 不动，于是下一步再问一遍**完全相同**的问题。

这个问题不便宜：它拷一份整条 prompt、在 scheduler 线程上按 chunk 逐段做哈希（645k token 的 prompt 约 5k 次），再阻塞等层的回复。KV 打满时这变成「每个等待请求每步一次」，步进速率塌陷 → running 请求跑不完 → KV 永远不释放 → **自锁**。现象是 GPU 全部 ~1%、线程全 runnable、`/metrics` 还能答。GLM-5.3 TP4 上用两种池大小复现两次。

`py-spy dump --locals` 每隔几秒比对，`lookup_id` 不变而 `chunk_id` 在动 —— 是活锁不是死锁，热点是 CPU 哈希不是 ZMQ 往返。

## 修复

把**整个答案**memo 在共享的 `ChunkedOffloadSchedulerBase` 里，key 是 `(sequence, HBM frontier)`。

答案是 prompt（不变）、frontier（key）、层内容三者的函数；只有层内容会在底下动，而且只会多出一个前缀 —— 漏掉它的代价是一次重算，而这正是 miss 本来就要付的代价。新鲜答案留下的副作用（load spec、save floor、`offload_loaded_tokens`）是同一个 key 的函数，填 memo 那一步已经做过，重放是幂等的。

memo 在三处丢弃：请求被准入（frontier 要动了）、请求 ID 交给新 sequence、请求结束。**不 memo** 的三种情况：lookup 超时/被拒（`hit is None`，那不是答案）、陈旧 pin 守卫（清它的是一次 metadata dispatch 不是 frontier）、异常路径。

放在共享层而不是 plugin 层，是因为**原生 backend 有同样的暴露**：完全 miss（`hit == 0`）和被拒绝的 load 都不会留下 `_load_specs`，于是它们的 lookup 每步被 dispatch 一次又丢掉一次。plugin connector 这边只留自己的 promise 记账。

被 memo 挡下的重复计入 offload statistics 的 `lookups_skipped_by_memo`。

## 测试

新增 12 个测试，**两条路径各自构造触发**：

- 原生：`tests/test_dense_offload_connector.py::test_native_retry_on_a_full_kv_cache_reuses_the_lookup` 驱动真实的 `Scheduler.schedule()`，`can_allocate` 恒返回 -1，跑 8 步 —— 修复前 8 次 lookup，修复后 1 次 + `lookups_skipped_by_memo == 7`。`test_native_admission_spends_the_memo` 在中途放开分配，断言准入后 memo 被清空。
- plugin：`tests/plugin/test_vllm_offload_lookup_memo.py` 用真实的 `DenseOffloadScheduler` 穿过 plugin adapter，每次循环都跨一个 step 边界（`build_connector_meta()`），否则测的是本来就有的 in-step 结果而不是 memo。

另有 frontier 移动、请求 ID 复用、lookup 失败不入 memo、请求结束不留 memo 等守卫测试。触发类测试在 `origin/main` 上全部失败。

全量套件在容器 `perz_glm53_v028` 内对 `origin/main` 逐行 diff：新增失败 **0**，新增通过 **+12**（117 failed / 12 collection errors 两臂逐字相同，均为既有问题）。

## 范围外

`atom/kv_transfer/offload/hybrid/dsv4/connector.py` 的 `DSV4OffloadScheduler.get_num_new_matched_tokens` 是另一份实现，结构上有同样的暴露（每次调用一次新 lookup，无 memo），但它还带 `_pending_slot_loads` / sidecar hash 状态，验证需要 K3 的 TP8。本 PR 不动它。


atom/kv_transfer/offload/chunked_scheduler.py
```diff
@@ -850,6 +907,7 @@ def cancel_pending_load(self, seq) -> None:
         if self._load_lifecycles.get(sid) is not seq:
             return
         self._clear_pending_load(sid)
+        self._match_memo.pop(sid, None)
```
yhl-amd: [P1] `cancel_pending_load()` is also used by `MultiConnectorScheduler` to disarm every non-winning connector. With the supported `[moriio, lmcache_offload]` ordering, MultiConnector still queries the offload sub on every call, then this line deletes its memo because moriio won. If allocation fails and the request returns to waiting at the same frontier, the next step performs the full offload lookup/hash again. I reproduced five identical scheduler queries producing five offload lookups and an empty `_match_memo`, so the original full-KV livelock remains in the composite configuration. Please preserve the one-lookup invariant at the MultiConnector layer, for example by not querying connectors after the first winner or by separating ‘disarm side effects’ from ‘invalidate the frontier answer’.

atom/kv_transfer/offload/chunked_scheduler.py
```diff
@@ -203,25 +244,41 @@ def get_num_new_matched_tokens(self, seq) -> tuple[int, bool]:
                 hit,
                 _lh,
             )
-        if not hit:
+        if hit is None:
+            # Timed out or refused: not an answer. Ask again next step rather
+            # than freezing a transport failure into the memo.
             return 0, False
+        if not hit:
+            return self._memoize_match(sid, seq, frontier, (0, False))
```
yhl-amd: [P2] `_MPLookupClient.lookup()` returns `0`, not `None`, when its poll deadline expires (`mp/backend.py:477-487`). Therefore this branch memoizes an MP timeout as a real tier miss. I reproduced a timed-out first call followed by an available 8-token result: the second call still returned `(0, False)` without submitting or checking another lookup. This contradicts the stated ‘timeouts are not memoized’ contract. Please propagate timeout as `None`/an explicit non-answer sentinel and add a test through `_MPLookupClient`, rather than only a fake lookup client that already returns `None` for failure.

atom/kv_transfer/offload/chunked_scheduler.py
```diff
@@ -203,25 +244,41 @@ def get_num_new_matched_tokens(self, seq) -> tuple[int, bool]:
...
         self._load_specs[sid] = LoadSpec(
-            hbm_cached_tokens=int(seq.num_cached_tokens),
+            hbm_cached_tokens=frontier,
             lmcache_cached_tokens=hit,
             can_load=False,
         )
-        return need, True  # True => park in WAITING_FOR_REMOTE_KVS
+        # True => park in WAITING_FOR_REMOTE_KVS
+        return self._memoize_match(sid, seq, frontier, (need, True))
```
yhl-amd: [P2] The base memo is committed before `LMCacheMPConnectorScheduler` performs its final full-prompt decision. For `block_size == 1`, the subclass clears the pending load and returns `(0, False)`, but the memo still contains the base `(4, True)` answer. Repeating the call at the same frontier then returns `(4, True)` even though both `_load_specs` and the MP lookup state are gone; I reproduced exactly this false-to-true transition. Please memoize only the final subclass-visible answer, or provide a finalization hook that runs before `_memoize_match`, and add a repeated-call test for the block-size-1 full-prompt case.

yhl-amd (commented):
I found three lifecycle gaps in the new frontier memoization. Details are inline.

atom/kv_transfer/offload/chunked_scheduler.py
PerryZhang01: Confirmed and fixed in da8c39ea3.

`cancel_pending_load()` no longer drops the entry; it demotes it to `(0, False)` at the same frontier (`_disarm_match_memo`). That keeps both properties the composite needs: one lookup per frontier, and no parking answer left standing behind a load spec that cancel took away -- replaying `(need, True)` after a disarm would promise a transfer nobody dispatches. Disarmed is also the truthful answer there: this connector is no longer the one loading the request.

I kept MultiConnector querying every sub, since a sub's lookup is where its save floors and lifecycle come from; the invariant is restored on the offload side instead.

Test: `test_a_multi_connector_disarm_does_not_restart_the_lookup` drives a real `MultiConnectorScheduler` with a moriio-shaped winner and a real `DenseOffloadScheduler`, five passes. Without the fix it produces your five lookups (`assert ['64', '64', '64', '64', '64'] == ['64']`); with it, one.

atom/kv_transfer/offload/chunked_scheduler.py
PerryZhang01: Confirmed and fixed in da8c39ea3. You are right that the contract and the MP client disagreed, and the client was the one that was wrong.

`_MPLookupClient.lookup()` now returns `None` on deadline expiry instead of `0`, with the adapter job deliberately left intact as before (`state.hit` stays `None`, which is exactly what tells `clear_lookup_status()` the job is still outstanding). The sid stays in `_lookup_in_step`, so `build_connector_meta()` still dispatches the unpin. The only other consumer of that return value, `DSV4OffloadScheduler`, already treats a falsy hit and `None` identically, so its behaviour is unchanged.

Test: `test_mp_lookup_timeout_is_not_recorded_as_a_tier_miss` runs through a real `_MPLookupClient` and a real `LMCacheMPConnectorScheduler`, with the adapter replying `[None, None, 8]` and a clock that steps past the deadline. Pre-fix it fails on `assert '7' not in _match_memo` -- the memo holds `(0, False)`; post-fix the second call re-submits and returns `(4, True)`. The two existing timeout tests now assert `is None`.

atom/kv_transfer/offload/chunked_scheduler.py
PerryZhang01: Confirmed and fixed in da8c39ea3, via the finalization hook you suggested.

`_memoize_match` became `_settle_match`, which calls an overridable `_refine_match(seq, sid, answer)` and records what that returns, so the memo holds exactly what the caller was told. `LMCacheMPConnectorScheduler` no longer overrides `get_num_new_matched_tokens`; its whole-prompt decision is now that hook, which runs before the entry is written. The base returns early from the memo without re-entering the method, so any decision taken after the fact could only hold for one step -- which is precisely the false-to-true transition you reproduced.

Test: `test_block_size_one_full_prompt_hit_stays_declined_when_asked_again` calls twice at the same frontier with `block_size == 1` and a chunk-aligned full-prompt hit. Pre-fix the second call fails with `assert (4, True) == (0, False)`; post-fix both calls answer `(0, False)`, `_load_specs` stays empty and the adapter sees one submission.

Full suite on this branch vs `origin/main`: 117 failed on both arms, `comm` in both directions empty, +15 passing.

yhl-amd (approved):

valarLip:
## Review — memoize the offload lookup so a waiting request is not re-queried every step

Reviewed at `da8c39ea3` in a dedicated worktree; `/app/ATOM` was never switched. The PR's own suite is green on its base (7349 passed; the 32 failures are pre-existing GPU/env ones, none in offload/lmcache), and `ruff`/`black` are clean on the changed files.

Two findings were produced by **running code**, not by reading it, so I will lead with those — they need no argument.

Then one statement that most of the rest hangs off:

> **The memo caches the answer *and its side effects* rather than the one expensive input. `hit` is the only costly quantity — prompt copy, chunk hashing, tier round trip; everything after it is arithmetic and dict writes. Memoizing `(seq, hit)` and re-deriving the rest would remove the `_refine_match` hook, `_settle_match`, the frontier key and the disarm mechanism entirely — which is to say, most of §3, §4 and §6 below.**

**Provenance:** **[reproduced]** means I ran it. **[verified]** means the deciding code was read at this head. **[reported]** means the shape matches the code but was not traced end to end.

---

# 1. The memo can be permanently disarmed, on a normal path **[reproduced]**

`chunked_scheduler.py:943`. `_disarm_match_memo` freezes `(0, False)` at the frontier with **no re-arm path**, and moriio's matcher is one-shot: `MoRIIOConnectorScheduler.get_num_new_matched_tokens` (`moriio_connector.py:913`) returns `(len(prompt), True)` only while `not hasattr(seq, "kv_async_tagged")`, and sets that flag itself. Nothing ever clears it.

Under the documented `[moriio, lmcache_offload]` ordering:

- **Step 1** — moriio wins; `MultiConnectorScheduler` cancels the offload loser → memo becomes `(seq, F, (0, False))`. `allocate` fails, the request returns to `waiting` at the same `F`.
- **Step 2+** — moriio answers `(0, False)` (already tagged) and offload replays the disarmed memo, so `winner` is None and the composite returns `(0, False)` **forever**. A multi-hundred-K CPU-tier prefix is silently recomputed.

Reproduced: a probe asserting `(16, True)` after the winning sub-connector stops matching gets `AssertionError: offload tier shut out: (0, False)`. Pre-PR the offload sub-connector re-answered honestly on every pass. `_reject_aborted_waiting` → `cancel_pending_load` is a second entry into the same shape.

This is not an edge case: a one-shot matcher plus MultiConnector cancelling the loser is the documented configuration.

# 2. The merge base is stale, and two new tests go red on rebase **[reproduced]**

`tests/test_dense_offload_connector.py:667`. The two new native tests monkeypatch `block_manager.can_allocate` with single-positional-arg lambdas — `:667` `lambda _seq: -1`, `:696` `lambda value: -1 if kv_full["yes"] else can_allocate(value)`. On current `origin/main`, `scheduler.py:1750` is:

```python
self.block_manager.can_allocate(seq, record=False, block_hashes=block_hashes,
                                reuse_hashes=self._local_prefill_coalescing)
```

introduced by `ee49a460d` ("coalesce prefills on TP"). Running the PR's test file against a clean `origin/main` worktree gives `TypeError: <lambda>() got an unexpected keyword argument 'record'` for both.

The merge base is ~20 commits behind main and **all three changed source files moved in that window** — notably `request_finished`, the exact method this PR edits. CI is green only because it runs on the stale base. Stub as `lambda seq, **kwargs: -1`, and rebase before merge.

# 3. The headline property has three holes **[verified]**

**The timeout path is deliberately unmemoized and unthrottled** (`:249`). `_MPLookupClient.lookup` polls `check_lookup_result` every 10 ms on the scheduler thread until `lmcache.mp.lookup_timeout` — **default 30.0 s** (`mp/backend.py:959`). Each step re-enters, because `build_connector_meta` dispatches and pops the `_lookup_results` entry for a sid with no `_load_specs`. So one waiting request costs a 30 s synchronous stall **per step, forever**. The triggering condition — tier slow because the box is saturated — is exactly what fills the KV cache in the first place, and it presents with the same symptom this PR set out to fix (GPUs idle, `/metrics` answering) while `lookups_skipped_by_memo` reads 0. Not a regression (main re-looked-up too), but it is the hole in the stated property. A bounded retry — memoize the non-answer with a next-retry-step budget — closes it.

**Partial verdicts are frozen for the whole wait, which the soundness argument does not cover** (`:207`). The docstring (`:180-188`) justifies staleness with "missing it costs a recompute, which is what a miss would have cost anyway". That holds for a complete miss. It does not hold for the two common partial verdicts: `need <= 0` (`:258`) and the `too_small` decline (`_decide_load_after_alloc:363`, default floor 8192). A request told "hit 4k, below the floor" at step 1, whose sibling stores 90k of shared prefix at step 5, stays on the step-1 answer for a wait that is minutes long under exactly the full-KV-cache condition this PR is about. A bounded re-ask (re-lookup every N steps, or invalidate on `save_finished`) is cheap relative to what it guards.

**On the vLLM plugin path the frontier half of the key is dead** (`:206`). By the PR's own stated premise — `connector.py:421-425` and the new test module both assert vLLM gates the call "only on `num_computed_tokens == 0`" — `frontier = int(seq.num_cached_tokens)` is 0 on every plugin call, so `memo[1] == frontier` never discriminates. The only content-independent invalidation axis the design has is inert on the path the livelock was actually measured on (GLM-5.3 TP4 through the plugin), and the memo degrades to one answer per request lifetime. The sole test of that guard, `tests/plugin/test_vllm_offload_lookup_memo.py:167`, feeds `num_computed_tokens=CHUNK` — a state the PR says vLLM cannot produce. Either the premise is wrong and the justification block needs rewriting, or the frontier key is plugin-dead and should be documented as a native-only contract.

# 4. Two pieces of state survive where they should not **[verified]**

**`_disarm_match_memo` rewrites rather than pops** (`:940`), so an aborted-while-waiting request keeps a strong reference to its `Sequence` — and its full prompt `token_ids` — for the life of the process. `_reject_aborted_waiting` (`model_engine/scheduler.py:1995`) calls `cancel_pending_load` on a request that was lookup-pinned but never allocated, which is the one state in which a memo entry exists, and `_disarm_match_memo` deliberately retains it. The only unconditional pops are admission (`:302`), lifecycle rotation (`:159`) and `request_finished` (`:969`); the `unschedulable` early-`continue` at `scheduler.py:1360-1365` notifies the connector not at all, and `_begin_load_lifecycle`'s pop fires only when a *different* seq reuses the sid, which never happens for unique request ids. One 645k-token prompt list pinned per such request. The root cause is that the method conflates two callers with opposite needs: the MultiConnector per-step disarm (retention is the point) and abort/teardown (retention is pure garbage).

**`load_failed()` leaves a stale `(need, True)` memo, and correctness rests on an undocumented statement ordering** (`:890`). It calls `_record_failed_load_attempt` + `_clear_pending_load` and never touches `_match_memo`, so the stale parking answer survives with its `LoadSpec` gone. Nothing replays it today only because `_repeat_load_suppressed` happens to sit at `:198`, above the memo read at `:207`. Move, invert or short-circuit that check — or add any earlier return — and a request whose single external-tier attempt is already spent replays `(need, True)` and promises a transfer nobody will dispatch. `cancel_pending_load` disarms; `load_failed` should too, and the ordering dependency should be stated at both sites.

# 5. The promise watchdog now cries wolf on the workload this PR targets **[verified]**

`plugin/vllm/kv_transfer/connector.py:440`. Changing `_promised_loads[rid] = 0` to `setdefault` interacts badly with the memo. vLLM parks a request only after `allocate_slots` succeeds; when it fails the request is left in `waiting`, not `WAITING_FOR_REMOTE_KVS`. With the memo the answer stays `(need, True)` every step and `should_park_for_load_after_alloc` keeps returning True (the `LoadSpec` is intact), so `setdefault` leaves the counter climbing. `_check_promised_loads` (`:683-710`) increments it once per step — the id never appears in `inner.requests`, because no load is queued — and at >50 logs:

> `ERROR ... promised a load for [...] but none was dispatched within 50 steps; those requests are parked in WAITING_FOR_REMOTE_KVS and cannot be released`

for a healthy, merely-queued request. It then `del`s the entry, so the genuine never-dispatched hang it exists to catch is no longer tracked for that request. The per-step reset made this impossible before.

# 6. Altitude: the hook exists only because the wrong thing is cached **[verified]**

`mp/backend.py:978`. Converting the MP override into a `_refine_match` hook narrows an invariant from "always" to "three of eight exits". The deleted `LMCacheMPConnectorScheduler.get_num_new_matched_tokens` ran after **every** `super()` return. `_settle_match` is reached only from `chunked_scheduler.py:254/261/268` — not from `_do_load`/no-client, not `_repeat_load_suppressed`, not the foreign `pending[0] is not seq` branch, not the `lookup()` except, not `hit is None`.

The `block_size == 1` refusal, whose entire job is `_clear_pending_load(sid)` for a hit that one-token blocks cannot host, is now skipped on all five. The foreign-pending branch in particular returns before touching `_load_specs`, so a surviving spec is no longer torn down.

And this is downstream of the thesis above: `hit` is the only costly quantity. Memoizing `(seq, hit)` and re-deriving everything after it removes this hook, `_settle_match`, the frontier key (§3) and the disarm (§1) — four of the findings in this review are the cost of caching a verdict instead of an input.

# 7. Fix-then-sweep: the DSV4 copy was not swept **[verified]**

`hybrid/dsv4/connector.py:1978`. `DSV4OffloadScheduler` is a **sibling, not a subclass**, of `ChunkedOffloadSchedulerBase` (`class DSV4OffloadScheduler(OffloadSchedulerMixin, KVConnectorSchedulerBase)`, `:1855`), carrying a verbatim copy of the fixed method and reached through the same `_query_connector_prefill_match` → `can_allocate` retry shape. So:

- the livelock stands unchanged for every DSv4 hybrid-offload deployment — full prompt copy at `:1986`, blocking `self._lookup_client.lookup(...)` at `:1990`, ~5k hashes per step;
- it has no `_lookup_results` dict at all, so unlike the base it **re-dispatches unconditionally**;
- `:2016`'s `if not hit:` cannot distinguish a non-answer from an empty one — the exact bug `da8c39ea3` fixes in the base.

Per `/app/ATOM/CLAUDE.md` ("after fixing a bug, immediately grep for the same pattern across the codebase and fix all occurrences in one pass"). The fix belongs in `OffloadSchedulerMixin`, which both classes already share and where this PR put only the counter.

# 8. Tests: one passes on main, one never runs what it is named for, six never run in CI **[verified]**

**`tests/plugin/test_vllm_offload_lookup_memo.py:110`** — `test_a_promised_load_stuck_in_waiting_is_looked_up_once` passes **unchanged on `origin/main`**. For a request whose `LoadSpec` stays armed, `build_connector_meta`'s filter `sid in loading_sids or sid not in self._load_specs` (`chunked_scheduler.py:454-458`) is false on both clauses, so `_lookup_results[sid]` is never popped and the next call takes `hit = pending[1]` without calling `lookup()`. Re-running the scenario against `origin/main` sources yields `calls == ['70']` — the same assertion the test makes. It exercises the pre-existing `_lookup_results` carrier, not the memo. The genuinely new coverage is the declined variant at `:86` and `test_a_declined_load_is_not_looked_up_again_every_step`.

**`:131`** — the only test of the `setdefault` change never runs the watchdog. `_end_of_step` (`:74-83`) calls `adapter._scheduler.build_connector_meta()`, the inner ATOM scheduler, bypassing `AtomLMCacheOffloadConnector.build_connector_meta` — the only caller of `_check_promised_loads(inner)` (`connector.py:506`). So it proves only that `dict.setdefault` does not overwrite an existing value: it never ages the counter, never crosses `_PROMISE_GRACE_STEPS`, never reaches the `logger.error` or the `del`. It is structurally incapable of catching §5. Route `_end_of_step` through `adapter.build_connector_meta(scheduler_output)`, or call `_check_promised_loads` 51 times and assert no ERROR.

**All six tests in that file never execute.** `.github/scripts/run_unit_tests.sh:23` passes `--ignore=tests/plugin`, and the file's `importorskip` skips it locally too (verified: `1 skipped`).

# 9. The observability counter never reaches an operator, and lies for one scheduler **[verified]**

`_offload_common.py:596`. `get_statistics()` is consumed by `engine_utility.py:375`, but `llm_engine.py:594-602` aggregates only a seven-key `offload_keys` tuple that excludes both `lookups_skipped_by_memo` and its sibling `suppressed_load_retries` — so the counter added to diagnose this livelock is readable from tests but not by an operator.

Worse, it is declared on the shared `OffloadSchedulerMixin` (`:488`) while the only `+= 1` is `chunked_scheduler.py:209`. So `DSV4OffloadScheduler`'s stats read `lookups_skipped_by_memo: 0`, which says "memo working, nothing repeated" when the truth is "there is no memo here at all" (§7).

# 10. Docs and shape **[verified]**

- **Docs-match-code.** `README.md:236` ("on a new request, queries the worker's LookupServer over ZMQ") and `:290` are now false — `chunked_scheduler.py:207-210` returns `memo[2]` with no query — and the memo's three-site lifetime is undocumented in the 1384-line subsystem design doc. `chunked_scheduler.py:59-62` still claims the MP adapter has "the same public lookup/clear_lookup_status contract" as LMCache's legacy client, which stopped being true when `_MPLookupClient.lookup` became `int | None` while the factory client stayed `int`.
- **The two new adjacent hooks take their arguments in opposite orders**, with `seq` unannotated: `_refine_match(self, seq, sid: str, answer)` at `:270` versus `_settle_match(self, sid: str, seq, frontier, answer)` at `:284`, with the `mp/backend.py:978` override following the first. A transposed call raises nothing at the boundary — `self._load_specs.get(seq)` returns None and the hook silently degrades to "no refinement", or `_match_memo` is keyed by the `Sequence` object and never answers.
- **The livelock narrative is told four times**, each with its own numbers: `chunked_scheduler.py:130-132`, the 32-line docstring at `:161-192`, `connector.py:421-432`, and the test module docstring — 61 comment lines to 36 lines of code. Per this project's review bar ("注释比代码长 ⟹ 改结构", "不加纯说明性注释"), only the "Answering from the memo is sound because…" paragraph is constraint-type content that earns its place.

---

Of the above, §1 and §2 are the two I would hold on, and neither needs to be argued — both were reproduced by execution. §1 in particular is a normal configuration, not an edge case. §6 is the one worth reading even if nothing else changes: four separate findings here are the price of caching a verdict plus its side effects instead of the single expensive input, and the smaller design makes them disappear rather than need fixing.

PerryZhang01:
Thanks for the detailed pass — the §6 pivot reshaped most of the rest, so here is each section against the current head.

**§1 / §3c / §4a / §6 / §10-2 — memoize only the `hit`.**
Adopted. The frontier is not part of the question: how much of the prompt the tier holds is a property of the prompt, and the frontier only decides how much of that hit is still worth transferring. So the memo now stores the one expensive input and the answer is re-derived on every call. That dissolved `_refine_match`, `_settle_match`, the frontier key, and the disarm mechanism entirely.

**§3a / §3b — staleness.**
Handled with two budgets rather than a frozen verdict: `OFFLOAD_LOOKUP_MEMO_STEPS` bounds how many times a remembered hit is replayed, `OFFLOAD_LOOKUP_RETRY_STEPS` backs off a non-answer (each retry costs `lmcache.mp.lookup_timeout`). Both default to 32. A memo is a bounded optimisation, not a cache — nothing tells the connector when the tier's answer changed.

**§4 — pins.**
`meta.lookup_requests_in_step` is the worker-side *unpin* list. A memo-served answer carries no pin, and an answer needs none: the request only waits. A transfer does. Pin re-acquisition therefore lives in the single dispatch choke point — the load-emit loop of `build_connector_meta` in both schedulers — and is confirm-only: `hit >= spec.lmcache_cached_tokens`, else the load is dropped with reason `tier_lost_prefix`. One lookup per load actually dispatched. This also answers @yhl-amd's P1 without a permanent disarm: the MultiConnector cancel keeps the remembered hit (re-answering from it is honest and free) and only dispatch pays the tier.

**§5 — `_promised_loads`.**
Kept as the plain `= 0` assignment, not `setdefault`. The per-step reset is exactly what distinguishes a merely-queued request from a genuinely parked one: vLLM parks in `WAITING_FOR_REMOTE_KVS` only after `allocate_slots` succeeds, and a queued request re-answers (and so re-promises) every step. With `setdefault` the counter would climb past the grace window and log a never-dispatched load for a healthy request — and then drop it, losing the hang the watchdog exists to catch. Covered by `test_a_repeatedly_answered_promise_does_not_cry_wolf` (100 steps, no ERROR).

**§7 — sweep to DSV4.**
Done. `DSV4OffloadScheduler` gets the same memo shape and the same `_ensure_lookup_pin` gate before `slot_load_spec = None`. It releases every pin at the end of the step it was taken in, so it has no `_lookup_results` carrier and always answers the gate with a fresh lookup. Four new tests in `tests/test_lmcache_offload_connector.py`.

**§8 — the plugin tests never run in CI.**
Correct, and it is the pre-existing `--ignore=tests/plugin` in `.github/scripts/run_unit_tests.sh`, not something this PR introduces. I executed all six locally in `rocm/atom-dev:vllm-latest` with a CPU-only `sitecustomize` shim (no GPU, peers are benchmarking) and they pass. Doing so corrected three of my own assumptions:
- the declined-request test is 2 lookups in 64 steps, not 1 — that is the 32-step memo budget, and it is now named and documented as such;
- the promised-load test is served by the pre-existing `_lookup_results` carrier, not by the memo (`total_lookups_skipped_by_memo == 0`) — exactly your point; it passes on `main` too and is kept only because that carrier is also what `_ensure_lookup_pin` consults;
- the watchdog test now asserts the *absence* of a false alarm, per §5.
For the record, the other `tests/plugin` offload files are already red on `origin/main` for unrelated fixture rot (verified by running them from a `git archive origin/main` tree in the same image).

**§9 — statistics.**
`offload_keys` in `atom/model_engine/llm_engine.py` gained `lookups_skipped_by_memo` and `suppressed_load_retries`, so both are visible through the engine's stats path rather than only on the scheduler object.

**§10 — docs and the contract claim.**
The `__init__` docstring no longer claims the adapter's `lookup` always returns an int; it now says it also returns `None`, meaning "the worker did not answer in time", which is not a hit of zero and must not be remembered as one. `mp/backend.py` was changed accordingly (typed `int | None`, returns `None` on deadline expiry) and the scheduler override reads `_last_tier_hit` rather than `hit_tokens`. The README's `get_num_new_matched_tokens` bullet documents the memo, the budgets and the dispatch-time re-query/drop; lifecycle step 1 now says "once per request, not once per step" and step 3 notes the pin re-take.

**§2 — rebase.**
Rebased onto current `origin/main` and squashed to a single commit.

Verification: `black --check` and `ruff check` clean on the ten changed Python files; the CI-shaped native suite is green (6178 passed — the three failures in `tests/entrypoints/test_serving_chat.py` and `tests/test_gc_utils.py` reproduce identically on `origin/main`); the six plugin memo tests pass in the container as described above.

valarLip (approved):

PR #2314, opened 2026-09-20
feat(deepseek-v41): native DeepSeek-V4.1-Flash — text, vision, paged runtime and DSpark

valarLip:
## What

DeepSeek-V4.1-Flash as its own model package rather than a reshape of V4.
48 commits, staged P00–P10b; `recipes/DeepSeek-V4.1-Flash.md` carries the stage
table and states what each stage accepted.

- **Model** — `atom/models/deepseek_v41/`: CSA2 topology, single-pass mHC, MoE,
  Engram, vision, and the DSpark draft.
- **Ops** — `atom/model_ops/deepseek_v41/` and
  `atom/model_ops/attentions/deepseek_v41/`: compressor, paged index scorer,
  packed attention, grouped output projection, rotary. V4's BF16 attention and
  inverse RoPE are reused rather than copied.
- **Runtime** — paged cache and request-state lifecycle, a whole-forward target
  graph, and a decode step that is the draft's replay plus the target's.
- **Protocol** — chat, tools and numeric reasoning effort; native image requests
  with chunked prefill and request-owned embedding lifetime.
- **Quantization** — group32 FP8 / W4A8 with E8M0 scales end to end
  (`atom/model_ops/blockscale*`), latterly on the CDNA4 microscaling matrix core.

## Where it stands

P00–P10b complete. **P11–P14 pending**: candidate-only indexing and Engram
fusion, optional CED decoder replay, optional encoder replay with a persistent
global cache, and validated distributed/deployment combinations. Draft for that
reason.

Two accepted tradeoffs and one open regression, all already documented:

- **DSpark zero-shot raw quality is an open regression, not waived.** Expanded
  fixed-length comparison over 128 questions: 90/128 without speculation versus
  80/128 with, nine gains and 19 losses, paired 95% bootstrap [−0.156, 0.0],
  exact McNemar p = 0.087. The matched five-shot comparison is identical per
  document, which does not cancel it. `docs/deepseek_v41_dspark.md` has the
  full account including what was ruled out.
- Small-chunk multimodal quality (P08) and MoE precision (P09), per the recipe.

## Accuracy

GSM8K, 1319 questions, fewshot 3, TP4 + whole-expert EP, DSpark-5, BF16 KV.
**Six same-code runs span 0.9075–0.9287, σ ≈ 0.8pp.**

A single run inside that band is the bar. Anything finer needs per-question
pairing against a **same-code control**: two runs of identical code disagree on
about 6.5% of questions, and a single pair's exact McNemar reaches p = 0.0026
between two runs of the same commit. The judge is whether cross-arm discordance
leaves the same-code range, not the p-value and not the headline.

## Test plan

- [x] `bash .github/scripts/run_unit_tests.sh` — 8222 passed, 158 skipped,
      3 xfailed
- [x] `tests/models/deepseek_v41/` and `tests/attentions/deepseek_v41/` on GPU
- [x] `tests.attentions.deepseek_v41.validate_runtime`, eager and `--graph` —
      55 scheduler events each; chunked, reordered, forked, replayed from a
      missing graph, preempted, cancelled and slot-recycled routes each produce
      the tokens an uninterrupted run did
- [x] `tests.models.deepseek_v41.validate_dspark_lifecycle --production --graph`
      — reaches the verify step with `accepted_drafts {'5': 10}`. It still stops
      at its own `restores >= 1`; that assertion is pre-existing and tracked
      (the file's checkpoint interval is snapped off against the 256-column
      prefix-cache hash block, and repairing that is necessary but not
      sufficient)
- [x] GSM8K 1319 in the band above
- [ ] Distributed and deployment combinations beyond TP4 + EP on MI355X (P14)

## Notes for review

Arithmetic is judged against the published model by the `ATOM_DSV41_REFERENCE`
unit tests; the runtime validates unsupported combinations before it loads
weights. AITER is pinned in the recipe.

The two most recent commits additionally want ROCm/aiter#5703: without it the
fused dual RMSNorm keys its Triton build on the token count, which costs
113–133 ms of device idle per new shape at serving time.


valarLip:
@ZhangLirong-amd heads up — `cf7e9eaea` moves the whole Engram package, including the two files you added in `01855ed96` / `3fa486f0b`. Nothing in them changed except imports and two attribute reads, but the paths did, so a local branch will conflict on rebase.

| before | after |
|---|---|
| `model_ops/engram.py` | `model_ops/engram/mapping.py` |
| `model_ops/engram_lookup.py` | `model_ops/engram/tables.py` (host half) |
| `model_engine/engram_runtime.py` | `model_ops/engram/host.py` |
| `model_ops/engram_hash.py` | `model_ops/engram/device/hashing.py` |
| `model_engine/engram_staging.py` | `model_ops/engram/device/staging.py` |
| `model_ops/engram_fused.py` | `model_ops/engram/device/gate.py` |
| `model_ops/engram_layer.py` | `model_ops/engram/device/layer.py` |

The rule the directory now carries: **everything at the top level of `model_ops/engram/` imports without Triton; everything under `device/` may not.** That is what the CPU pre-checks job was failing on — `engram_runtime` reached the hash kernels at module scope, so the runner took the whole pytest session down during collection instead of skipping one file. `tests/model_ops/engram/test_import_boundary.py` executes the rule now, in a child interpreter, so it fails as one test rather than as `Interrupted`.

Two things that touch your code directly:

- `EngramStaging.__init__` now takes the device lookup (`EngramUva`) instead of the host; it reads `uva.hash_tables` / `uva.row_ids` and gets the host from `uva.host`. The rest of the class is unchanged.
- `EngramHost` no longer holds `hash_tables` / `_row_ids` / `_ids_staging` / `overlap` as its own attributes — they belong to the lookup. `overlap` and `hash_tables` remain as None-safe properties, so existing readers did not have to change.

Also worth knowing separately from this PR: `b3cb26caa` landed with the Ruff and Black jobs already red (`models/deepseek_v41/config.py` over-long line, `models/deepseek_v41/model.py` import block) and with `test_compilation.py::test_runtime_guard_with_live_steps_and_aux_hooks[3]` failing. I fixed the two formatting ones here because one of those files is in this diff and the other fails a repo-wide job. The test failure is untouched and reproduces on `b3cb26caa` alone — it carries `skipif(not torch.cuda.is_available())`, so the CPU job skips it and only a GPU box shows it.

PR #2327, opened 2026-09-21
fix(eagle3): use scheduler block size for draft KV pool

zufayu:
Fix Kimi-K2.5 Eagle3 startup with FP8 KV cache.

The draft uses scheduler block IDs, but its KV pool was built with the target attention backend's physical block size. For MLA targets that value is 1, which violates the MHA FP8 SHUFFLE layout requirement. Use the scheduler block size instead.

Failure: https://github.com/ROCm/ATOM/actions/runs/35522046579/job/106107926944

Test: `python -m py_compile atom/spec_decode/draft_kv.py`


zufayu:
@HaonanWang98 handing this issue over to you for owner review and the final fix.

Kimi-K2.5-MXFP4 Eagle3 fails deterministically during draft KV-pool initialization. The first actionable exception is:

```
ValueError: SHUFFLE packs 16 elements of torch.float8_e4m3fn per 16B tile, which has to divide both head_dim 128 and block_size 1
```

Failure: https://github.com/ROCm/ATOM/actions/runs/35522046579/job/106107926944

The Kimi target uses MLA with an internal physical page size of 1, while the scheduler block size is 16. After #2147, `DraftKvBuilder` passes `model_runner.attn_metadata_builder.block_size` (1) into the MHA draft pool. Before #2147, the Eagle3 builder used `model_runner.block_size` (16). `EagleProposer` also documents that the draft MHA cache is indexed by `runner.block_size` blocks.

The later TCPStore/NCCL/SHUTDOWN messages are secondary effects after all ModelRunner processes exit on this validation error. This PR was closed so you can confirm the intended block geometry and own the final implementation and regression coverage. The branch remains available for reference.

PR #2332, opened 2026-09-21
feat(kimi-k3): support P/D disaggregation

wanzhenchn:
## Submission Checklist


ZhangLirong-amd:
The fix for the original V4 DPA8 + DSpark hang looks correct. _dp_draft_lockstep_active() now checks is_pipeline_parallel directly, so disabling deferred output no longer causes a producer to skip draft collectives while its DP peers enter them. Keeping deferred output
  enabled for V4 also restores its previous behavior at that decision point.

  However, disabling Scheduler.use_spec for all P/D producers introduces two additional cases that need attention:

  1. EAGLE / serial MTP loses successor-token metadata during chunked prefill.
     Setting self.use_spec=False also makes self.drafter_needs_next_token=False, although the runner still retains and invokes the drafter. Consequently, batch.next_token_ids is absent. For a non-DP middle chunk, EagleProposer.compute_draft_kv() returns without writing draft
     KV; for a DP alignment pass, propose_draft_token_ids() cannot replace the dummy anchor with the actual successor token. Please separate “the producer does not perform speculative verification” from “the drafter still needs successor-token metadata.” This does not affect
     DSpark through this particular path, since its context-KV write does not use successor tokens.

  2. DSpark dynamic q-bucketing can expand a producer’s target-only decode beyond its scheduled width.
     The scheduler now gives the producer a q=1 decode batch, but _dspark_apply_q_bucket() still derives full_q from self.drafter.mtp_k + 1. Executing the existing decision logic with mtp_k=3, ell=1, and buckets 1,2,4 selects q=2 for a batch scheduled with only one token per
     request. This can make the rebuilt token layout and attention metadata exceed what the scheduler reserved. Please skip this transformation for batches without speculative verification, or otherwise ensure it cannot increase any request’s scheduled width.

  I checked the lockstep behavior and the two cases above with CPU-only source/control-flow checks; these are not end-to-end GPU reproductions.

  The passing V4/DSpark accuracy jobs are useful, but the P/D benchmark and Mooncake RDMA smoke jobs are currently skipped. Before re-landing, I would also like to see the current PR head pass the original regression configuration: 1P1D, DPA8, DSpark, chunked prefill, and
  concurrency 128, including warmup and sustained request progress. The original hang fix looks sound, but that distributed regression coverage is still needed for the full re-land

ZhangLirong-amd (approved):

Yuechguo (approved):

atom/model_engine/model_runner.py
```diff
@@ -495,12 +513,29 @@ def prepare_input_ids(
...
+            if self.use_spec and self.is_pipeline_parallel:
                 raise NotImplementedError("pipeline parallel + speculative decode")
 
             self.input_ids.np[:total_tokens_decode] = token_ids
+            # RAGGED needs no overwrite: scheduled_tokens is already the flat
+            # [anchor, drafts...]. The uniform case does, and `scheduled_tokens`
+            # is flat here too -- reshape to one row per request the way the
+            # deferred path below does, rather than indexing it as if it were
+            # already rectangular.
+            if (
```
valarLip, line 531: why need this

atom/model_engine/model_runner.py
```diff
@@ -599,7 +634,11 @@ def prepare_input_ids(
         if fill_to > total_tokens_decode:
             self.input_ids.gpu[total_tokens_decode:fill_to].zero_()
 
-        input_ids = self.input_ids.gpu[:total_tokens]
+        # Slice by the width this path actually staged. Prefill returned above,
+        # so the decode total is the whole batch; a worker-side speculative q
+        # shrink rewrites it, and reading any other total here would leave the
+        # attention metadata wider than input_ids.
+        input_ids = self.input_ids.gpu[:total_tokens_decode]
```
valarLip, line 641: ?

atom/model_engine/scheduler.py
valarLip, line 2208: do not change this

atom/model_engine/model_runner.py
wanzhenchn: `scheduled_spec_decode_tokens` is (num_reqs, q-1). The buffer just filled is flat: each request occupies `max_seqlen_q` slots, and slot 0 is the anchor. The reshape makes one row per request so the drafts land in columns `1:`. Without it the 2-D drafts do not match the 1-D buffer. Ragged DSpark skips this: its drafts are already in the flat `scheduled_tokens`, and the rows are not equal width.

atom/model_engine/model_runner.py
wanzhenchn: This path has already returned for prefill, and it only staged `total_tokens_decode` ids. The old slice used `total_tokens_num`. After the worker shrinks the verify width to a q-bucket, attention metadata follows the shrunk decode count. A wider return includes token ids left in the buffer by the previous step, so the metadata and `input_ids` disagree.

atom/model_engine/scheduler.py
wanzhenchn: `num_placeholder_tokens = len(drafts)` counts only the drafts the producer sent. A K3 producer sends none, so that count is 0. The first decode still reads the last `mtp_k + 1` tokens. With only T0 appended, that slice reaches back into the prompt and verifies prompt tokens as drafts. The missing slots are padded with `eos`, and the count has to be `mtp_k` so `preempt()` strips those pads. Leaving it at `len(drafts)` keeps the` eos` as real context. T0 stays excluded. When the producer did send `mtp_k` drafts, the two counts are equal.

PR #2339, opened 2026-09-21
fix(offload): fence dense LMCache saves

NidhoggD1:
## Motivation

Dense LMCache saves run on a background executor. The save thread packs KV on
its thread-local GPU stream, but current `main` does not order that stream after
the model/RPC stream that produced the KV. A background save can therefore
persist stale or incomplete KV and later return incorrect model output on a
host-tier hit.

## Technical Details

- Record one producer event when a metadata step contains dense save work.
- Share that event with every background save submitted for the step.
- Add `BlockGPUConnector.wait_for_save_source()` and enqueue
  `pack_stream.wait_event(producer_event)` before `LMCacheEngine.store()` reads
  the KV.
- Keep the dependency stream-to-stream so the save worker does not block on
  `Event.synchronize()` in the normal GPU path.
- Fall back to event synchronization only when the connector has no pack
  stream.

The event stays owned by each submitted save closure until that save has
consumed the dependency.

Related: #2332 contains a similar correctness fix inside a much larger Kimi-K3
P/D change. This PR isolates the dense-save fence against `main` and uses a
nonblocking pack-stream wait instead of host synchronization.

## Validation

The same producer-event to pack-stream dependency was validated on an ATOM
dense LMCache stack using MI350X, LMCache 0.4.5, FP8 KV, and TP4:

- Connector-level delayed-write reproduction: 38/40 stale reads without the
  dependency, 0/40 with `pack_stream.wait_event`.
- Model-level host-tier reads: 41/41 correct with the fence.
- Adjacent candidate/control/candidate gate: corrupted host-tier reads
  0/4 -> 4/4 -> 0/4.
- Integrated coverage included single TP4, two routed TP4 engines, concurrent
  and reused prefixes, and forced GPU-prefix eviction followed by host recall.

Checks run on this branch:

- `ruff check` on the three changed files: pass.
- `ruff format --check` on the three changed files: pass.
- Python bytecode compilation of the three changed files: pass.

The submission environment does not contain PyTorch/pytest, so the added unit
tests are left for ATOM CI to execute.

## Test Plan

- Verify one producer event is recorded per save step and shared by multiple
  save requests.
- Verify each save waits before calling `store()`.
- Verify the GPU connector enqueues `wait_event` on the pack stream and does
  not host-synchronize in the normal GPU path.
- Run the existing non-GPU unit-test suite in CI.

## Submission Checklist

- [x] The change is limited to dense LMCache save correctness and tests.
- [x] No recipe default or performance-policy change is included.
- [x] Static style and formatting checks pass.
- [ ] ATOM CI unit tests pass.


NidhoggD1:
Ready for review. The validation section now includes the Linux targeted unit
suite and a real-GPU LMCache save/retrieve smoke on the current PR head.

The GitHub Pre Checkin run is currently `action_required` because this PR comes
from a fork. Could a maintainer please approve the workflow run and review the
change when convenient? Thank you.

atom/kv_transfer/offload/dense/connector.py
```diff
@@ -179,6 +180,15 @@ def start_load_kv(self, metadata) -> None:
...
                 self._lookup_unpin(lookup_id)
+        save_ready_event = None
+        if self._do_save and any(
+            req.save_spec is not None for req in metadata.requests
+        ):
+            # Save metadata is dispatched after the producing forward. Record
+            # that stream here so the background pack stream cannot read KV
+            # blocks before their writes are complete.
+            save_ready_event = torch.cuda.Event()
+            save_ready_event.record(torch.cuda.current_stream())
```
IvanShan177: **The event record sits outside the dispatch loop, so a failure here drops the step's _loads_ as well as its saves, and the requests hang.**

Lines 183-191 run ahead of the `for req in metadata.requests` loop on line 192, which submits **both** load and save jobs. The raising call is really line 191 — `torch.cuda.Event()` is lazy in PyTorch (the `hipEvent` is created on first record), while `current_stream()` + `.record()` is a real runtime call, and on this path it is the *first* runtime call of the step. That makes it exactly where a sticky async HIP error from a previous step's kernel surfaces (async faults are reported at the next runtime call, not at the offending kernel).

The interleaving: `.record()` raises on the worker RPC thread -> the exception escapes `start_load_kv` -> `ModelRunner.process_kvconnector_output` (`model_runner.py:3333-3338`) calls it bare, with no handler -> the loop on 192 never runs, so **no load future is ever submitted**. Meanwhile the scheduler has already pinned those loads and is waiting on `finished_loading` / `failed_loading` from `get_finished()`.

Neither ever arrives, because the only thing that reports a load failure is `_guard` (`_offload_common.py:361-376`), and `_guard` *is* the submitted job — if submission never happens, `self._failed_load` is never filled. The `_lookup_unpin` loop on 180-182 has already run at that point, so the step's pins are gone too.

This is worse under `KimiK3OffloadConnector.start_load_kv` (`hybrid/kimi_k3/connector.py:259-285`), which calls `super().start_load_kv(metadata)` from a `finally` after `_arm_joint_loads`. Its own comment states the invariant this breaks:

> `ThreadPoolExecutor.submit` raises RuntimeError if the executor was shut down by a racing close() [...] and `ModelRunner.process_kvconnector_output` has no handler -- so each submit is isolated inside its helper [...] and super() runs in a finally so a KV load or save is never dropped.

So every raise site was deliberately isolated per item precisely because there is no handler upstream. A raise at 191 leaves the park armed with the KV leg never reporting (permanently parked), and the exception from the `finally` also masks whatever the `try` raised.

The sibling `DSV4OffloadConnector` already does this correctly for the same fence (`hybrid/dsv4/connector.py:745-762`): the create+record is inside the per-request loop, wrapped in `try/except`, and on failure it rejects just that one save and `continue`s. Suggest matching that shape here — or at minimum wrapping 190-191 and leaving `save_ready_event = None` so the saves degrade instead of taking the loads down with them.

atom/kv_transfer/offload/_block_gpu_connector.py
```diff
@@ -434,6 +434,22 @@ def _thread_state(self) -> _ThreadTransferState:
...
+        the thread-local pack stream preserves that ordering without blocking
+        the CPU save worker. The following ``batched_from_gpu`` call runs on
+        the same thread and therefore reuses this stream state.
+        """
+
+        state = self._thread_state()
+        if state.pack_stream is None:
+            producer_event.synchronize()
+            return
+        state.pack_stream.wait_event(producer_event)
```
IvanShan177: **This fence is thread-affine, and it silently degrades to a no-op if the pack ever runs off the calling thread.**

`wait_for_save_source` enqueues the dependency on the *calling* thread's `_ThreadTransferState.pack_stream`, and the docstring makes the assumption explicit ("The following `batched_from_gpu` call runs on the same thread").

The interleaving that breaks it: save worker T1 enqueues `wait_event(E)` on T1's pack stream, then calls `LMCacheEngine.store()`. If `store()` dispatches the GPU pack on any other thread T2 — an LMCache storage-manager worker, an async-store configuration, the `save_unfull_chunk` path — then `_thread_state()` on T2 lazily constructs a **fresh** `_ThreadTransferState` (`atom_lmcache_staging.py:98-111`) with brand-new pack/copy streams that carry no dependency on `E`. `batched_from_gpu` then reads KV concurrently with the model stream's still-pending writes: precisely the stale-KV persistence this PR exists to prevent, with no exception and no log line.

DSV4's equivalent uses a host-side `producer_event.synchronize()` (`hybrid/dsv4/connector.py:1523`), which is thread-independent. The existing `track_save_source` leans on the same same-thread assumption, but its failure mode is a benign degradation (no source-safe callback), not silent data corruption — so I don't think the precedent carries over.

Worth either recording the owning thread on `_ThreadTransferState` and asserting it here, or falling back to `producer_event.synchronize()` when the state was not created by the current thread.

atom/kv_transfer/offload/dense/connector.py
```diff
@@ -335,6 +351,8 @@ def _do_save_req(self, req: LMCacheReqMeta) -> None:
         t_store0 = time.perf_counter()
         self._reset_gpu_connector_transfer_stats()
         gpu_connector = self._engine.gpu_connector
+        if producer_event is not None:
+            gpu_connector.wait_for_save_source(producer_event)
```
IvanShan177: **This is now the only unguarded `gpu_connector` dereference in `_do_save_req`.**

Every neighbouring access is defensive: `_reset_gpu_connector_transfer_stats` and `_last_gpu_connector_transfer_stats` both early-return when `gpu_connector is None` (`_offload_common.py:336-360`), and the very next line uses `getattr(gpu_connector, "track_save_source", None)` plus a `callable(...)` check. `tests/test_dense_offload_connector.py:307-310` constructs `SimpleNamespace(gpu_connector=None, store=...)`, so `None` is a contemplated state rather than a theoretical one.

Because production now always passes a non-`None` `producer_event`, any engine whose `gpu_connector` is `None` — or which predates `wait_for_save_source` — turns **every** save into an `AttributeError`. That gets swallowed by `_guard` into `_record_save_failure`, i.e. 100% silent save failure (offload effectively off, visible only as an exception log), rather than a degraded-but-working save.

`getattr(gpu_connector, "wait_for_save_source", None)` with a `producer_event.synchronize()` fallback would keep both cases correct, and would match the style of the line directly below.

IvanShan177 (commented):
Reviewed at `54749d5`.

The core premise holds, and I checked it before looking for problems: `start_load_kv` is dispatched before `forward` (`process_kvconnector_output`, `atom/model_engine/model_runner.py:3333`), and the dense scheduler only emits saves for tokens whose forward has already completed, so recording on the current stream at dispatch time really does fence the producing writes. `pack_stream` is also the right stream to fence, since `batched_from_gpu` stage A is the only leg that touches KV. The `partial(...)` change is compatible with `OffloadWorkerMixin._guard(kind, fn, req)`, and DSV4's 3-arg `_do_save_req` is a separate class rather than a subclass.

Three comments below, all on error / `None` paths rather than on the fence itself. Concurrency first.

atom/kv_transfer/offload/dense/connector.py
NidhoggD1: Fixed in c6af1092. Fence creation is now lazy inside the per-save dispatch branch and wrapped in try/except. A record failure calls _record_save_failure(req) and continues the loop, so loads from the same metadata step are still submitted while no unfenced save is launched. Added a mixed load/save regression test that injects a record failure and verifies the load reaches finished_loading.

atom/kv_transfer/offload/_block_gpu_connector.py
NidhoggD1: Fixed in c6af1092 by removing the same-thread assumption. DenseOffloadConnector now forwards producer_event through LMCacheEngine.store(...); LMCache forwards the extra kwarg to BlockGPUConnector.batched_from_gpu(), which waits only after obtaining the actual calling thread state and pack stream. The GPU smoke confirms the event is consumed once on the real save-worker pack path without host synchronization.

atom/kv_transfer/offload/dense/connector.py
NidhoggD1: Fixed in c6af1092 as part of the event handoff change. _do_save_req no longer dereferences gpu_connector.wait_for_save_source; it passes producer_event through engine.store, while the adjacent track_save_source access remains guarded with getattr/callable. The existing gpu_connector=None-compatible unit path remains intact. The targeted Linux suite is now 44 passed.

yhl-amd (dismissed):

valarLip:
## Review — fence dense LMCache saves

Reviewed at the PR head (3 commits, +180/−8) in a detached worktree; the main checkouts were never touched.

**The premise checks out, so this is not a no-op.** I verified independently that the save frontier only covers already-forwarded chunks on both dispatch paths, that disagg-side streams rejoin the default stream, and that LMCache does forward `store(**kwargs)` through to `batched_from_gpu` in both shipped versions (the `0.5.5rc3` wheel and the `v0.4.5` source build pinned in `docker/`). The race is real and this closes it.

What follows is about the invariants it leaves unasserted, the legs it did not sweep, and the failure path. Two statements:

> **The fence is carried on a channel that can disappear silently — `producer_event=None` means "no fence" with no log, counter or assertion, and `LMCacheEngine.store()` has four early returns that never reach the connector at all. The comment justifying that channel is contradicted by the mechanism twelve lines above it.**

> **The failure path converts "one missing fence" into a permanent silent cache hole plus an unbounded traceback flood on the RPC thread — because it reports a never-attempted save as finished after the scheduler watermark has already advanced.**

And the channel this wants already exists in the tree, owned by the producer: `record_kv_cache_ready`.

**Provenance:** **[verified]** means the deciding code was read at this head or the behavior was confirmed against LMCache/torch on this box. **[reported]** means the shape matches the code but was not traced end to end.

---

# 1. The fence's arrival is unverifiable **[verified]**

`dense/connector.py:373`. `batched_from_gpu(..., producer_event=None)` skips `_wait_for_save_source` with **zero signal** — no log, no counter, no assertion. And `LMCacheEngine.store()` returns before `gpu_connector.batched_from_gpu` on four paths: `not is_healthy()`, `_is_passive()`, `is_frozen()`, and empty `memory_objs`.

So any LMCache bump that builds an explicit connector-kwargs dict, or a layerwise/remote store route, drops the kwarg and the race returns — observable only as an intermittent accuracy drop under offload hits, which this repo's own noise floor (~9pp at 100 GSM8K questions) makes effectively invisible.

The connector already carries the exact idiom for this class of question — `stats['batch_block_ids_enabled'] = 1` (`_block_gpu_connector.py:666`). Record a `producer_fenced` stat in `_capture_transfer_stats` and surface it in `[OFFLOAD-SAVE-PROF]`.

**The comment defending the kwarg route is contradicted in the same `with` block** (`:374`). It reads "even if LMCache changes its dispatch model" — but twelve lines above, `track_save_source` (`_block_gpu_connector.py:792`) carries save identity through a thread-local, and its docstring says it does so "without changing LMCache's public API". The same hypothetical dispatch change silently breaks that one: if LMCache ever runs `batched_from_gpu` on its own thread, the fence survives while early block release stops — `_source_safe_identities` (`:808-821`) returns `()` for a non-`SaveOperationId`, `_queue_source_safe_group` returns without publishing, no `DENSE_PAGE_SOURCE_SAFE_CHANNEL` completion is ever emitted, and every save's source blocks stay leased until `reclaim_stale_leases`. Either delete the overclaim, or put the event on the same thread-local channel — `self._engine.gpu_connector` is already reachable from `_do_save_req` (see `_last_gpu_connector_transfer_stats`, `_offload_common.py:337`).

# 2. The failure path is worse than having no fence **[verified]**

`dense/connector.py:204`. `_record_save_failure(req)` → `_record_store_terminal(req, False)` adds the operation to `_done_save` (i.e. `finished_saving`) **in both branches**. Four consequences compound:

1. A save that was never attempted is reported as finished.
2. `chunked_scheduler.py:435` already did `entry[1] = aligned` at emit time and nothing resets it, so **those chunks are never retried** — a permanent silent cache hole.
3. `_store_finished(succeeded=False)` (`chunked_scheduler.py:751-753`) deliberately does **not** call `_release_operation_lease` — "Failed stores keep unsafe ranges leased until abandon timeout" — so the block lease is held to the timeout.
4. Because `save_ready_event` stays `None`, a sticky HIP error — *the stated motivation for this change* — retries `torch.cuda.Event()` and emits a full `logger.exception` traceback for **every** save-bearing request, **every** step, forever, on the pre-forward RPC thread.

The comment "rejects only this save" is inaccurate: it rejects every save for the life of the error. Caching a per-step sentinel instead of `None` breaks item 4.

**No CUDA precondition on the producer side** (`:200`) **[verified]**. The consumer side explicitly contemplates non-CUDA (`state.pack_stream is None`), but `start_load_kv` calls `torch.cuda.Event()` / `torch.cuda.current_stream()` unconditionally for every step that has a save, with no `_use_cuda()`-style predicate and no once-only degradation. On any worker whose torch has no usable CUDA context, every save in every step takes the `except` branch: a stack-trace flood on the RPC thread — the thread this whole connector design exists to keep free — plus a terminal "finished" report per save, with the watermark already advanced so nothing is ever retried. Offload is silently dead rather than loudly rejected. One check at `register_kv_caches` is louder and cheaper.

# 3. Three legs not swept, one of them a subclass of the class being fixed

**KimiK3's state-tier stores have the identical hazard, by their own docstring** (`hybrid/kimi_k3/connector.py:399`) **[verified]**. They are dispatched from the same `start_load_kv`, **before** `super()`, and explicitly carry no producer fence:

> *"No producer fence and no staging copy: the source is the checkpoint's PAGE units, reserved out of the KV pool and pinned by the engine."*

Pinning answers "can these blocks be reallocated", not "have the producing writes retired". `start_load_kv` (`:283`) submits these to `self._state_tier`'s own executor threads before `super().start_load_kv(metadata)` (`:285`) creates the new fence. Under deferred output the host has not synced on forward N−1, so the tier packer gathers a torn mix of pre/post-write bytes into a recurrent-state checkpoint that restores cleanly and corrupts output on reuse. The sibling docstring at `:355` (`_start_state_loads`: "No producer fence, unlike the save path") is now self-contradictory.

**DSV4 is the other `_engine.store(` call site and was not swept** (`hybrid/dsv4/connector.py:1603`) **[verified]**. `grep -n '_engine.store('` returns exactly two hits; one was fixed. dsv4 does `producer_event.synchronize()` at `:1525` (and `:945`) then calls `store(tokens, mask=, block_ids=, req_id=)` with no kwarg — so the new `_wait_for_save_source` is **dead code for the entire DSV4 PAGE path**, which keeps the host stall this PR exists to remove, while the PR's own test is named `..._without_host_sync`. Note it cannot be converted mechanically: dsv4's `synchronize()` also gates the SLOT D2H, and the `_SlotStagingSyncError` quarantine depends on it raising. That makes it a design decision the PR should state, not omit.

**Loads get nothing, and the in-repo precedent fences both directions** (`dense/connector.py:188`) **[verified]**. `mp/backend.py:723-729` records one `torch.cuda.Event(interprocess=True)` and hands it to `_submit_load` **and** `_submit_save`. Here the load submit at `:188-193` runs before the event is even created, and `batched_to_gpu` has no parameter for it. A block freed by a request that finished or was preempted in step N's `schedule()` and reallocated to a loading request can take the H2D restore while forward N−1 (deferred output, not host-synced) is still reading or writing it; `wait_for_requests` fences at request granularity, not against the producing forward. Either the load direction genuinely does not need it — then say so, because the mp backend's symmetry now reads as an inconsistency someone will "fix" — or it is the same bug one line up.

**Under PP the invariant does not hold** (`dense/connector.py:201`) **[verified, latent]**. `Scheduler.advance_on_schedule` is `pipeline_parallel_size > 1` (`scheduler.py:598`), and `_advance_prefill_on_schedule` (`:2567`) does `seq.num_cached_tokens += num_scheduled_tokens[i]` inside `schedule()`, before the forward. `chunked_scheduler.build_connector_meta` reads that same field (`:397-408`; `_offload_finished_cached_tokens` is only set for retiring requests) and emits a `SaveSpec` covering the in-flight chunk. So `candidate_event.record(...)` runs before that forward is enqueued and `pack_stream.wait_event` orders the pack against nothing: the save packs **pre-forward** KV and stores a corrupt chunk under a valid content key, restored silently on a later prefix hit. Latent only because PP is currently broken on this box. Needs an assertion (save frontier ≤ previously-forwarded frontier) or a comment naming the `chunked_scheduler` invariant it rests on.

# 4. This is the third per-step producer event in the package, and the producer-owned channel already exists **[verified]**

`dense/connector.py:195`. `ModelRunner._record_kv_cache_ready` (`model_runner.py:3199`, fired at `:3290`/`:3312`/`:4659` at the **end** of forward, per final prefill chunk) already calls `connector.record_kv_cache_ready(req_ids)`, and `offload/connector.py:133` already forwards it to the impl. That docstring still says "no offload impl needs it now: the event exists only for the Mooncake producer" — which this PR makes false without updating it. Mooncake is the reference consumer (`mooncake_connector.py:1059-1064`: a `torch.cuda.Event()` on the post-forward stream, keyed per `req_id`, with explicit `pop` cleanup at three terminal points). The other two copies are `mp/backend.py:723` and `dsv4/connector.py:743-762`.

The right-depth fix is to implement `record_kv_cache_ready` on `OffloadWorkerMixin` and wait it in `_do_save_req`: one fence, owned by the producer, on a channel LMCache cannot drop, shared by dense/dsv4/kimi_k3/m3/mp, and symmetric for loads. That collapses §1, §3 and this section together.

# 5. Four mechanical issues **[verified]**

- **The wait is enqueued after work is already on the same stream** (`_block_gpu_connector.py:915`). `:910-914` runs `_prepare_block_id_stage`, whose prepared path calls `prepare([...], device=self.device, stream=state.pack_stream)` (`:659-663`) — real stream work — and only then does `:915-916` enqueue the wait. So "nothing this call submits runs before the producer event" is already false. Benign today because that work is a host-side block-id H2D, but the fence's validity is now an implicit, unstated property of every codec's `prepare_block_id_groups`. Swap the two lines (cost: nothing) and state the invariant in `_wait_for_save_source`'s docstring.
- **`producer_event` is a positional-or-keyword 4th parameter on a method a third-party library calls** (`:897`). The abstract contract is three positionals plus `**kwargs`, so any wrapper drift binds a memory-obj-ish value to `producer_event` and `_wait_for_save_source` calls `.wait_event(<not an event>)` inside the save thread — where `_guard` turns it into a save reported as finished (§2). Make it keyword-only. Separately, `batched_to_gpu` (`:934`) and `to_gpu` (`:889`) have no such parameter, so a `producer_event=` added to a retrieve "by symmetry" later falls into `**kwargs` → `_prepare_transfer` → `_ranges_to_block_ids`, which reads only `block_ids` and drops the rest: it would compile, pass, and fence nothing.
- **`functools.partial` silently degrades the failure log** (`dense/connector.py:216`). `_offload_common.py:368` logs `getattr(fn, "__name__", kind)`, and a partial has no `__name__` — so every dense save failure now logs `offload save failed for <req>` instead of `offload _do_save_req failed for <req>`, losing the one field that distinguishes a save-path failure from a submit-path one, on exactly the path where a new failure class was just introduced. Latent trap: `dsv4/connector.py:1078-1084` has its own `_guard` whose handler does a bare `fn.__name__`, so copying this idiom there raises `AttributeError` **inside** the except block. Use `functools.update_wrapper`, or widen `_guard` to `(kind, fn, req, **fn_kwargs)` — dsv4 already passes `producer_event` as a plain argument rather than binding it into a partial.
- **The `pack_stream is None` → `producer_event.synchronize()` branch is unreachable, untested, and would busy-spin** (`_block_gpu_connector.py:441`). `pack_stream` is None only when `_ThreadTransferState` was built with `use_cuda=False` (`atom_lmcache_staging.py:105-110`), but the only source of a `producer_event` is `dense/connector.py:201`, which requires a live CUDA context — and `_run_staged_pipeline` → `_assert_fused_chunk_major_available()` raises three lines later on a non-CUDA device anyway. Meanwhile this `torch.cuda.Event()` takes default flags, unlike every sibling in the package which writes `blocking=False` explicitly (`:839`), and this is the **first** event in the package ever host-`synchronize()`d — so the wait would spin a full core on the save daemon against a co-resident model. Delete the branch (let `_assert_fused_chunk_major_available` own the non-CUDA case) or raise; if it is kept, the event needs `blocking=True`.

# 6. All three new tests pass for weaker claims than their names assert **[verified]**

`tests/test_dense_offload_connector.py:333`.

**(a) The one-event-per-step property is unfalsifiable.** `torch.cuda.Event` is patched to `lambda: event` — a shared singleton — and `current_stream` globally to `rpc_stream`. An implementation that created a fresh event per request, or created and recorded it inside `_do_save_req` on the save thread (*the exact anti-design this shape exists to prevent*), yields an identical trace and stays green.

**(b) The failure test cannot distinguish failure from success.** `assert output.finished_saving == {save_operation}` (`:427`) is identical either way — dense sets `_early_release=True`, so the real verdict lives in `ConnectorCompletion(DENSE_PAGE_STORE_CHANNEL, op, succeeded)`, which the test never inspects. Changing `:209` to `_record_store_terminal(req, True)` keeps it green. Its load is also degenerate: `hbm=lmc=0` early-returns before `retrieve` is ever called.

**(c) The ordering test asserts only its own wiring.** It stubs five private methods, and `_prepare_block_id_stage`'s stub appends nothing to `trace` — so `assert trace == [("wait", ...), ("pipeline", ...)]` is true by construction and blind to §5's first item.

**(d)** The `pack_stream is None` branch has zero coverage; the only test that touches it wires `synchronize` to `pytest.fail`.

**(e)** The zero-arg stubs pin the exact signatures under review: adding `blocking=True` or a device argument makes both tests `TypeError`, so the fixes above will read as "the fix broke the tests".

# 7. Docs-match-code: every document describing this call chain is stale, including the comment that justifies the change **[verified]**

- `README.md:352` — the save-flow diagram is still `engine.store(tokens, mask, block_ids)` → `batched_from_gpu` with no fence.
- `README.md:273` — frames the pre-return current-stream touch as DSV4-only.
- `README.md:616-626` — the "Invariants enforced in code" table gains no row for the invariant the PR's own test treats as hard.
- `README.md:808-823` — documents `batched_from_gpu`/`batched_to_gpu` with no mention of the new parameter or the LMCache-kwargs dependency. That is **the single most fragile cross-repo assumption in the subsystem, documented only in a three-line inline comment**.
- `batched_from_gpu`'s own docstring (`:900`) is unchanged.
- `dense/connector.py:16` still claims `start_load_kv` "**only** submits ... so the worker RPC thread is free for forward", which the new CUDA-event record and `_record_save_failure` call break.
- `offload/connector.py:136` still says no offload impl needs `record_kv_cache_ready` (§4).

Finally, the **new** comment at `dense/connector.py:196` ("Save metadata is dispatched after the producing forward") reads as false against `engine_core.py:418` and contradicts dsv4's own comment at `:722` ("Metadata is dispatched before this batch's forward"). It should name the `chunked_scheduler` frontier invariant it actually depends on — which is also what §3's PP item needs.

---

Of the above, §2 is the one I would hold on: it is the only item that turns a transient failure into permanent silent data loss, and it does so on the exact error class named as the motivation. §1 is the one worth fixing even though the fence works today — an invariant whose presence cannot be observed will be removed by accident. §3's dsv4 and kimi_k3 legs need at minimum a sentence each saying whether they were considered; right now the PR reads as if dense were the only producer.

kvnloo:
Thanks for documenting the real-GPU smoke and the follow-up work on the save path. I'm working through the ATOM–LMCache integration downstream and don't have the target AMD hardware locally.

Would it be useful if I turned your existing `LocalCPUBackend` smoke into a small reproducible validation recipe, using whichever script and revision you think are appropriate?

I can handle the source/CPU checks and result write-up downstream. I'm mainly trying to preserve one narrow claim: actual production-path storage and reload with K/V + scale equality, separately from model accuracy or performance. I don't want to duplicate the fence work you're already doing here.
I also found the existing GPU/disk round-trip test useful as an earlier check, although because it explicitly synchronizes the producer I'd treat it as narrower evidence rather than validation of this async path.

Does that decomposition make sense, or is there an existing test/script you'd prefer we build the qualification around?

_AI-assisted source review and drafting; I don't have GPU results of my own to report yet._

NidhoggD1:
@valarLip thanks for the thorough pass. Addressed in `6aefe2ee` (scheduler retry) and `4a8d76be` (fence observability, ordering, tests, docs). Point by point:

**§1 fence arrival unverifiable** — done. `batched_from_gpu` now records `producer_fenced` in the transfer stats (default 0, set to 1 when the wait is enqueued) and `[OFFLOAD-SAVE-PROF]` prints it, so a store that reaches the connector without the event is visible per save. The "even if LMCache changes its dispatch model" comment is gone; the replacement names the `store(**kwargs)` forwarding as the one cross-repo assumption, and the README's `_block_gpu_connector.py` section says the same. I kept the kwargs route rather than the thread-local: if LMCache ever ran the pack on its own thread, the thread-local would lose the event silently, whereas kwargs survive that, and the stat now makes loss visible either way.

**§2 failure path** — done in `6aefe2ee`. A rejected or failed save no longer advances the watermark: the scheduler keeps the pre-emit frontier per operation, restores it on failure and re-emits the save once every TP rank is quiescent. A pre-submit rejection (fence record raised) reports quiescence directly; a post-submit failure becomes retryable when all of its source groups have been reported source-safe, otherwise the lease is reclaimed on the abandon timeout. A save that fails after its request was retired is dropped rather than retried against recycled blocks. Fence failure is logged once per failure streak, and one sentinel per step rejects the remaining saves of that step without re-recording. On the CUDA precondition: `_assert_fused_chunk_major_available` already refuses a non-CUDA device on the first pipeline run, so offload is rejected loudly there; I did not add a second check at registration.

**§3 legs not swept** —
- DSV4: deliberate, not converted. Its `producer_event.synchronize()` also gates the SLOT D2H and the `_SlotStagingSyncError` quarantine depends on it raising, so replacing it is a separate change with its own test plan. The PR description now says the PAGE fence is dense-only.
- KimiK3 state stores: not changed here. Whether the tier packer needs the same fence depends on whether the step that marks a checkpoint ready has host-observed the producing forward; I have not traced that and would rather fix it in a follow-up than assert it is safe.
- Loads: the hazard you describe (a partial-prefill request preempted in `schedule()`, its blocks reallocated to a load in the same step) is a block-reuse ordering question that exists independently of the save fence, and `batched_to_gpu` has no parameter for it today. Out of scope for this PR; worth its own issue.
- PP: the chunked scheduler now warns at init when `pipeline_parallel_size > 1` and saves are enabled, with a comment naming `advance_on_schedule` as the reason the frontier invariant does not hold there.

**§4 `record_kv_cache_ready`** — not adopted. `ModelRunner._record_kv_cache_ready` fires only for requests whose `is_final_chunk` is true, but the racing saves are exactly the intermediate prefill chunks (no token is produced, so nothing host-synchronizes before the next dispatch). Using that channel would require it to fire per chunk, which is a model-runner change I would keep separate. The `offload/connector.py` docstring stays accurate because dense still does not consume the callback.

**§5 mechanical** — all four done: the wait is enqueued before `_prepare_block_id_stage` and the docstring states that nothing else may precede it; `producer_event` is keyword-only; `_guard` takes `**fn_kwargs` so the `partial` (and the lost `fn.__name__`) is gone; the `pack_stream is None` branch raises instead of host-synchronizing.

**§6 tests** — (a) `torch.cuda.Event` is now a counting class: exactly one instance, recorded on the RPC thread, shared by both stores. (b) The failure test asserts `ConnectorCompletion(DENSE_PAGE_STORE_CHANNEL, op, False)` and the quiescent report. (c) The ordering test traces `_prepare_block_id_stage` and asserts `wait < block_ids < pipeline`, plus `producer_fenced == 1`. (d) New tests cover the keyword-only signature and the no-pack-stream raise. (e) The stubs were rewritten with the new signatures.

**§7 docs** — README save-flow diagram, `start_load_kv` bullet, invariants table and `_block_gpu_connector.py` section updated; the dense module docstring no longer claims `start_load_kv` only submits; the fence comment now states the frontier invariant (metadata dispatched before the forward; the save frontier covers only already-forwarded chunks).

The real-GPU smoke in the description was run on `c6af1092`; the GPU-facing change since then is the wait moving ahead of the block-ID upload. I will re-run the smoke on the current head and update the description.

NidhoggD1:
Re-ran the real-GPU smoke on the current head `4a8d76be` (MI350X, ROCm 7.2.53211, torch 2.10.0+rocm7.2.4, LMCache 0.4.5, single GPU, LocalCPUBackend, no model). Result: passed. Exactly one producer-event wait, executed on the save-worker pack thread against a live pack stream; the connector reported `producer_fenced=1`; two source-safe groups, one store terminal (succeeded) and one source-quiescent completion for the save; 16/16 tokens hit on the host tier; after zeroing the GPU KV, retrieval restored every K/V and scale tensor with exact equality. The PR description now reflects this run.

NidhoggD1:
@kvnloo yes, that decomposition makes sense, and building the qualification around the `LocalCPUBackend` smoke is the right call. Two pointers:

- Script and revision: use the version of the smoke that ran on `4a8d76be` (results and description above), against the current head `e9be221f`. It is the earlier script plus three checks the head now supports: the producer wait must land on the pack thread with a live pack stream, `last_transfer_stats()["producer_fenced"]` must be 1 for the store, and both `dense.page.store` (succeeded) and `dense.page.source_quiescent` completions must be emitted for the save operation. It needs one GPU, LMCache 0.4.5 and the PR checkout on `PYTHONPATH`; no model. I will attach it to this PR so you have a fixed artifact rather than a description.
- Agreed on `test_lmcache_offload_gpu_disk_e2e.py`: it records and forwards a producer event but the test itself is not a race, so treat it as coverage of the plumbing, not evidence for the async ordering.

One caveat for the write-up: the smoke proves "the shipped store path waits on the producer and reloads K/V + scales exactly". It does not exercise TP quorum or the failed-save retry path; those are unit-tested only.

kvnloo:
Thanks, this gives me a concrete starting point. I’ll build the downstream recipe around the linked smoke, using e9be221f as the target and recording the reported 4a8d76be run separately.

I’ll preserve the pack-thread/live-stream check, producer_fenced=1, both save completions, and exact K/V + scale restoration. The write-up will explicitly exclude TP quorum and failed-save retry coverage, and describe the disk test as plumbing coverage rather than an ordering-race test.

I’ll review the script and prepare the recipe before requesting GPU time. No additional testing needed from you at this stage, thanks for sharing the runner.

kvnloo:
Following up with the [pinned downstream recipe](https://github.com/kvnloo/ATOM/blob/4feea25b027927a2cff1648d83f4931672f943a5/docs/downstream/lmcache/DENSE_HOST_SMOKE_RECIPE.md). It reuses `pr2339_gpu_smoke.py` unchanged from Gist snapshot `c881bce7`, targets `e9be221f`, and records your reported `4a8d76be` run separately. I haven't run the smoke on a GPU.

Could you sanity-check just the acceptance table and environment assumptions before I seek a one-GPU reproduction? In particular, would running the unchanged smoke in an already-provisioned, compatible LMCache `0.4.5`/ROCm environment preserve its intended coverage without the machine-specific outer runner?

The recipe retains the pack-thread/live-stream wait, `producer_fenced=1`, both save completions, and exact host-backed K/V + scale restoration. It explicitly excludes normal registration, scheduler-driven eviction, TP quorum, failed-save retries, model accuracy, and ordering-race or performance claims.

This is a source-review request only—no additional run requested from you and no changes to the fence implementation. Thanks for providing the concrete starting point.

_AI-assisted source review and drafting; no GPU results of my own claimed._

